PR #2, opened 2017-09-21
T393: remove useless len() and set_level() functions

Taniadz:


dmbaturin:
Thanks, but I've already fixed that myself so that I could proceed with my own work. len() nonsense and excessive set_level()'s were the least of its problems. Please refrain from doing anything with this script until I push my changes.

PR #7, opened 2017-10-26
T393: write interface definition for lldp op mode

Taniadz:


hagbard-01:
I think that PR should be closed, since it has a conflict, is very old and seems not relevant anymore.

PR #8, opened 2018-03-03
T574: Add new style scripts and definitions for system host-name

jbrodriguez:
A first attempt at an interface definition.

A couple of questions.

I guess the node name must be 'host-name' (instead of hostname) for backwards compatibility, right ?

What is the convention for the script filenames:

- vyos-config-host-name.py
- vyos-config-hostname.py
- vyos-hostname.py

?

Who is writing the python script ?

### DIFF
--- a/interface-definitions/host-name.xml
+++ b/interface-definitions/host-name.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0"?>
+
+<!-- host-name configuration -->
+
+<interfaceDefinition>
+  <node name="system">
+    <children>
+      <node name="host-name" owner="${vyos_sbindir}/vyos-config-host-name.py">
+        <properties>
+          <help>Hostname settings</help>
+        </properties>
+        <children>
+          <leafNode name="hostname">
+            <properties>
+              <help>server's hostname </help>
+            </properties>
+          </leafNode>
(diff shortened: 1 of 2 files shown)


dmbaturin:
Hi Juan, thanks for your interest in the new style rewrites! Generally, the person who writes the definitions also writes the scripts.
Our development process is to first create a task in phabricator. Else, with this number of submodules we have, we'll be having very hard time producing release notes. I've created the task for you since you are new here, let's continue the discussion there: https://phabricator.vyos.net/T574
You can either register an account or login with github oauth.

dmbaturin:
I'm going to change some things, but the foundation is good I think, let's merge this and work from there.

dmbaturin:
Ok, I've paid my price for the sloppy code review, but next time please try to actually run your code before submitting pull requests. ;)

jbrodriguez:
Yeah, guilty as charged.

I didn't test it, since I didn't figure out how.

But I now see some tests were added, that definitely helps.

dmbaturin:
You can build the package (debuild -us -uc), install it on a live system with dpkg -i and try it out. Before the recent development that allowed us to start adding tests, it was the only way to test anything. :)

PR #10, opened 2018-05-08
T626 Add tests for vyos-update-crontab and initial setup for sonarcloud

higebu:
Add unit tests for vyos-update-crontab.py.


dmbaturin:
Good idea! One thing though: did you not use the testsuite and test runner APIs from the unitttest module on purpose?

higebu:
@dmbaturin I added it, but it needs renaming the script.

BTW, why does `generate` function in vyos-update-crontab.py write actual crontab file rather than `apply` function?

higebu:
@dmbaturin I understand about the `generate` function by reading the article you wrote. https://blog.vyos.net/vyos-development-digest-number-10

higebu:
@dmbaturin I marged `conf_mode` and `validators` packages into `vyos` package for support sonar-scanner. What do you think about this? We can separate these packages to other directory, but directory path needs to be same as package path for mapping the coverage report to SonarCloud.

higebu:
@dmbaturin Finally, I can support both testsuite ant unitttest.main. We can see the coverage on https://sonarcloud.io/component_measures?id=vyos%3Avyos-1x&metric=coverage

PR #14, opened 2018-06-08
T689: Converting simple op-mode commands from vyatta-op to new syntax

runborg:
After reading the blog post about new op-mode definitions i've taken upon the task to convert some simple commands to new syntax.
I've started with some simple commands in "vyatta-op". 

Finished so far:

* reboot
* poweroff
* show arp
* show bridge
* show date
* show disk
* show configuration
* show hardware
* show raid
* show users


runborg:
Hi! @c-po!

Yes, i'm able to generate new helper scripts for some of the commands. (i have'nt seen the complexity of all of the scripts)
Do you want everything in this pull-request, or should i drop this pr and divide it up into smaller pieces ?

do you want a rewrite to python only on the perl scripts. 
or should i also pythonize the short shell-scripts also?

c-po:
Just keep on extending this PR by pushing new commits to this branch. That's fine.

op-mode-definitions/show-hardware.xml
```diff
@@ -0,0 +1,95 @@
...
+                <properties>
+                  <help> Show system CPU details</help>
+                </properties>
+                <command>less $_vyatta_less_options --prompt='.cpu info' -- /proc/cpuinfo</command>
+              </node>
+              <node name="summary">
+                <properties>
+                  <help>Show CPU's on system</help>
+                </properties>
+                <command>${vyatta_bindir}/vyatta-cpu-summary.pl</command>
```
dmbaturin: I've made a replacement for the old CPU summary script, it will be in ${vyos_op_scripts_dir}/cpu_summary.py
Since the format is common in /proc, I've made vyos.util.colon_separated_to_dict function for that.

op-mode-definitions/show-hardware.xml
```diff
@@ -0,0 +1,95 @@
...
+              </node>
+            </children>
+          </node>
+
+          <node name="dmi">
+            <properties>
+              <help>Show system DMI details</help>
+            </properties>
+            <command>${vyatta_bindir}/vyatta-show-dmi</command>
```
dmbaturin: Any reason we are not calling dmidecode?
Anyway, the script is small, I suppose it can be renamed to dmi.sh and moved to the op mode scripts in vyos-1x

op-mode-definitions/show-configuration.xml
```diff
@@ -0,0 +1,39 @@
...
+          </properties>
+          <!-- no admin check -->
+          <command>cli-shell-api showCfg --show-show-defaults --show-active-only --show-hide-secrets</command>
+        </node>
+        <node name="commands">
+          <properties>
+            <help> Show running configuration as set commands </help>
+          </properties>
+          <!-- no admin check -->
+          <command>${vyatta_sbindir}/vyatta-config-gen-sets.pl</command>
```
dmbaturin: vyatta-config-gen-sets.pl is no more. I've replaced it with a better implementation, see https://phabricator.vyos.net/T684

op-mode-definitions/show-configuration.xml
```diff
@@ -0,0 +1,39 @@
...
+        </node>
+        <node name="files">
+          <properties>
+            <help> Show available saved configurations </help>
+          </properties>
+          <!-- no admin check -->
+          <command>${vyos_op_scripts_dir}/show-configuration-files.sh</command>
```
dmbaturin: I don't see this scripts. I'm missing something, or you forgot to commit it?

op-mode-definitions/show-arp.xml
```diff
@@ -0,0 +1,26 @@
...
+<interfaceDefinition>
+  <node name="show">
+    <children>
+    <node name="arp">
+      <properties>
+        <help>Show Address Resolution Protocol (ARP) information</help>
+      </properties>
+      <command>/usr/sbin/arp -e -n</command>
+    </node>
+    <tagNode name="arp">
```
dmbaturin: This will not work in the new system I think. The commands should be "run show arp" and "run show arp interface", where interface is a tag node.

op-mode-definitions/show-hardware.xml
```diff
@@ -0,0 +1,95 @@
...
+            <properties>
+              <help>Show CPU info</help>
+            </properties>
+            <command>lscpu</command>
+            <children>
+              <node name="detail">
+                <properties>
+                  <help> Show system CPU details</help>
+                </properties>
+                <command>less $_vyatta_less_options --prompt='.cpu info' -- /proc/cpuinfo</command>
```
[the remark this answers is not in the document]
runborg: agreed, i could change out all less'es with simple cat in all scripts.

dmbaturin:
@runborg Generally, the question we should ask is if it will create a maintenance burden or deter contributors. Shell scripts that are big and hairy or hard to follow definitely should be rewritten. Small and straightforward ones can be kept, nothing wrong with them.

dmbaturin:
Also, we need to remove the files from vyatta-op as we add them to vyos-1x, otherwise APT will not install them due to conflicting files.

op-mode-definitions/show-arp.xml
runborg: you mean, "show arp" as a node, and move the tag named "show arp" to "show arp interface" ?  
I did this "trick" to keep the original syntax, but if syntax change is accepted its easy to fix.

this trick is done in more places, so fixing this means changeing syntax on every command that does this.

op-mode-definitions/show-configuration.xml
runborg: File commited.. i forgot to commit it...

op-mode-definitions/show-date.xml
```diff
@@ -0,0 +1,30 @@
...
+        <properties>
+          <help>Show system date and time as Coordinated Universal Time</help>
+        </properties>
+        <command>/bin/date -u</command>
+        <children>
+          <node name="maya">
+            <properties>
+              <help>Show UTC date in Maya calendar format</help>
+            </properties>
+            <command>${vyatta_bindir}/maya-date.py $(date +%s)</command>
```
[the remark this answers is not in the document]
runborg: True!    another thing is if this "easter egg" is needed. as far as i know it can be removed from code without impact.. :)

op-mode-definitions/show-arp.xml
dmbaturin: There aren't _too_ many commands that use this arrangement I think. I also think this arrangement was always a bad idea. :)

op-mode-definitions/show-hardware.xml
runborg: as far as i see dmidecode is a more extensive output than the data fetched from the original vyatta-show-dmi script. 

Another question is if this information should be placed anywhere else instead of show hardware dmi ?

op-mode-definitions/show-users.xml
```diff
@@ -0,0 +1,30 @@
...
+<interfaceDefinition>
+  <node name="show">
+    <children>
+    <node name="users">
+      <properties>
+        <help>Show user information</help>
+      </properties>
+      <command>who -H</command>
+      <children>
+        <node name="recent">
```
runborg: is this command needed? wouldn't it be wiser to add " | tail "   or something like that to only list the latest

src/completion/list_raidset.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+echo -n `cat /proc/partitions | grep md | awk '{ print $4 }'`
```
runborg: I've added this as a simple wrapper script, but because its a simple oneliner it would fit inside the XML . is it wise to make it a wrapper script or should i move it to a inline part of the XML?

op-mode-definitions/show-hardware.xml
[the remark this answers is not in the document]
runborg: `less` is now removed and replaced with a simple `cat`

op-mode-definitions/show-arp.xml
runborg: 👍 I'll update syntax on all commands that does this :)

src/completion/list_raidset.sh
dmbaturin: Maybe not, I don't think it has a potential to grow, and even if it does, it's easy to move to a separate script.

op-mode-definitions/show-users.xml
dmbaturin: You mean "show users" or "show users recent"? I'd keep them both I think, and don't underestimate the number of people coming from platforms that lack |tail :)

op-mode-definitions/show-arp.xml
runborg: syntax updated

op-mode-definitions/show-arp.xml
```diff
@@ -0,0 +1,27 @@
...
+      <properties>
+        <help>Show Address Resolution Protocol (ARP) information</help>
+      </properties>
+      <command>/usr/sbin/arp -e -n</command>
+      <children>
+        <tagNode name="interface">
+          <properties>
+            <help>Show Address Resolution Protocol (ARP) cache for specified interface</help>
+            <completionHelp>
+              <script>${vyatta_sbindir}/vyatta-interfaces.pl --show ethernet</script>
```
runborg: NOTE: Move away from `vyatta-interfaces.pl` script

dmbaturin:
The list_interfaces.py completion script is now capable of listing interfaces by type and listing all broadcast interfaces (those sure have ARP), at the time, ethernet + bonding + bridge. Now we can use it in place of the old vyatta-interfaces.pl in this context.

``vyos@vyos-test# /usr/libexec/vyos/completion/list_interfaces.py --type tunnel
tun1 tun0
[edit]
vyos@vyos-test# /usr/libexec/vyos/completion/list_interfaces.py --broadcast
eth0 eth1 eth0.300 eth0.60 eth0.50 eth0.40 eth0.30

``

op-mode-definitions/show-users.xml
runborg: 👍  i ment `show users recent`. but i'll let them be there.

dmbaturin:
Kim also reviewed it, so now that every maintainers has taken a look, I'm merging this. We'll see what to do about the reboot and poweroff scripts, maybe just clean them up and move to vyos-1x, maybe rewrite.

PR #15, opened 2018-06-18
T201 - Beep When Fully Booted

hagbard-01:
Please review and merge. It depends on the pull request in vyos-build (beep packages installation and unit file which can be separately installed)


src/conf_mode/beep_if_fully_booted.py
```diff
@@ -0,0 +1,54 @@
...
+#
+
+import sys
+import subprocess 
+
+from vyos.config import Config
+from vyos import ConfigError
+
+def get_config():
+ conf = Config()
```
c-po: Please use an indention level of two

PR #16, opened 2018-06-21
T201 implmntation, it depends on the beep pacakge added in vyos-build

hagbard-01:
rebase didn't work well and made the things worse, I did a new fork and added my 2 files. Will hopefully make it easy for you to merge. Du you want me to remove the enable/disable logic and just have existent node or not existent node handled? If so let, me know, not a big deal to implement.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ interface_definitions:
 
 	# XXX: delete top level node.def's that now live in other packages
 	rm -f $(TMPL_DIR)/system/node.def
+	rm -f $(TMPL_DIR)/system/options/node.def
 	rm -f $(TMPL_DIR)/service/node.def
 	rm -f $(TMPL_DIR)/service/dns/node.def
 	rm -f $(TMPL_DIR)/protocols/node.def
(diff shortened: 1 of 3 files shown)


hagbard-01:
Made your suggested changes, it's now being enabled if node exists and disabled if the node is being removed.

c-po:
During ISO generation I see 
```
dpkg: error processing archive /root/packages/./vyos-1x_1.0.5_all.deb (--unpack):
 trying to overwrite '/opt/vyatta/share/vyatta-cfg/templates/system/options/node.def', which is also in package vyatta-cfg-system 0.20.44+vyos2+current11
```

Have you tried testing this on the target?

c-po:
Just build a fresh ISO with our modifications.

```
cpo@LR1# set system options beep-if-fully-booted

  Configuration path: [system options beep-if-fully-booted] requires a value
  Set failed
```

Looks non-tested to me. Property `<valueless/>` is missing

c-po:
This PR has been merged manually on the commandline and extended by removing the systemd. beep is now called directly. Thanks for the contribution.

PR #19, opened 2018-06-27
T253 limited all usernames to alphanumerical only

hagbard-01:
Limited input to avoid issues with white-spaces as well as the parser has issues if the the string has more than 246 characters.

### DIFF
--- a/interface-definitions/snmp.xml
+++ b/interface-definitions/snmp.xml
@@ -13,7 +13,7 @@
             <properties>
               <help>Community name [REQUIRED]</help>
               <constraint>
-                <regex>^[a-zA-Z0-9\-_]{1,100}</regex>
+                <regex>^[a-zA-Z0-9]{1,100}</regex>
               </constraint>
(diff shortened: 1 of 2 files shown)


dmbaturin:
Hi @hagbard-01!
Thanks for the contribution! Two things though. First, Linux doesn't limit usernames to only alphanumeric characters, at least underscore and hyphen are allowed. Dots are allowed too. We should not introduce arbitrary restrictions without a very good motivation.
Also, this restriction will break existing configs because in 1.1.8, this does work:

dmbaturin@reki# set system login user dmbaturin_1988-test authentication plaintext-password qwerty
[edit]
dmbaturin@reki# commit
[edit]

Second, regarding the 246 characters limit: perhaps we should fix the parser? Have you already found out where exactly the limitation is?

hagbard-01:
What are the plans,using the same pattern as in useradd?

dmbaturin:
The most important part is to ensure that no existing configs break, so yes, useradd regex sounds sensible.

PR #20, opened 2018-07-05
Fix for Phabricator T733 - a bug in the snmp config parser that means that ipv6 community strings are never set

jules-vyos:


jules-vyos:
Blimey Christian - that was quick ;)

Thank you.


-- Jules

PR #21, opened 2018-07-05
T720: filter is now the same like in useradd(8) including the debian …

hagbard-01:
I'm going to focus on user name only, for the snmp community the restriction is 100 chars but I have a look into the parser and try to fix it there. For any group or user account, I'm going to implement the useradd(8) restrictions and the debian restrictions, unless you guys have a problem with that.

### DIFF
--- a/interface-definitions/ssh.xml
+++ b/interface-definitions/ssh.xml
@@ -21,12 +21,20 @@
                   <leafNode name="group">
                     <properties>
                       <help>Allow members of a group to login</help>
+                        <constraint>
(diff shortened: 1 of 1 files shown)


dmbaturin:
Looks good now. What on earth are user names ending with $ are used for though? I've had no idea useradd allows it!

hagbard-01:
No idea, the standard is very old from what I have read and everyone puts their own restrictions in, so I used the debian useradd as example, that way we can avoid issues in general. I'm going ti fix the other ones later, just want to make sure everything is still working for everybody before I proceed. I'm watching the forum too and slack, if any issue arises with that. The restriction is also the same for local groups.

PR #22, opened 2018-07-20
Only look for xml files in interface and op template definitions

hagbard-01:


dmbaturin:
I'm not sure if we are realistically going to have non-XML files there, but I think it still may be a reasonable safeguard.

PR #23, opened 2018-07-24
T736 - Rewrite remote logging (syslog) to XML/Python

hagbard-01:
vyatta-cfg-system needs to be done first, before you merge this one int, otherwise you will have a conflicts for files referenced in both pkgs.


interface-definitions/syslog.xml
```diff
@@ -0,0 +1,859 @@
...
+            </tagNode>
+          </children>
+        </tagNode>
+        <tagNode name="host">
+          <properties>
+            <help>Logging to a remote host</help>
+            <valueHelp>
+              <format>x.x.x.x or host.domain.tld</format>
+              <description>Remote host name or IP address</description>
+              <!-- regex host or IP check? -->
```
dmbaturin: Perhaps no check at all — FQDNs are notoriously hard to check...

src/conf_mode/syslog.py
```diff
@@ -0,0 +1,267 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+
+import sys
+import os
+import jinja2
```
dmbaturin: A small note: the common approach to imports order which I follow is: standard library modules, third-party modules, local modules, single function imports.
Like:
import sys
import os

import jinja2

import vyos.config

from future import braces

src/conf_mode/syslog.py
```diff
@@ -0,0 +1,267 @@
...
+        if node == 'security':
+          node = 'auth'
+        selectors += node + "." + lvl
+    return selectors
+
+def generate(c):
+  tmpl = jinja2.Template(configs, trim_blocks=True)
+  config_text = tmpl.render(c)
+  #print (config_text)
+  with open('/etc/rsyslog.d/vyos-rsyslog.conf', 'w') as f:
```
dmbaturin, line 221: Perhaps move it to a constant at the top of the file?

src/conf_mode/syslog.py
```diff
@@ -0,0 +1,267 @@
...
+            raise ConfigError('Invalid facility ' + s + ' set in '+ conf + ' ' + item)
+          l = re.sub("^.+\.","",s)
+          if l not in lvl:
+            raise ConfigError('Invalid logging level ' + s + ' set in '+ conf + ' ' + item)
+
+def apply(c):
+  ### vyatta-log.conf is being generated somewhere
+  ### this is just a quick hack to remove the old configfile
+  
+  if os.path.exists('/etc/rsyslog.d/vyatta-log.conf'):
```
dmbaturin, line 254: Are there really situations when it would exist? After image upgrade, /etc/ would be clean.

src/conf_mode/syslog.py
```diff
@@ -0,0 +1,267 @@
...
+    os.remove('/etc/rsyslog.d/vyatta-log.conf')
+  os.system("sudo invoke-rc.d rsyslog restart >/dev/null")
```
dmbaturin: Syslog is a systemd service (rsyslog.service). I see no reason to use the SysV init compatibility interface for it.

dmbaturin:
Looks good, save for small things I've mentioned in the review.
You've tested it and it works for you?

src/conf_mode/syslog.py
hagbard-01: It's only used once in that line. Still want to have a constant?

src/conf_mode/syslog.py
hagbard-01: Yes. I did a fresh install, then installed my modified packages and that file did exist. So, the old vyatta-system-cfg generates it somewhere (haven't found yet, but I also didn't look too hard). So updates would be affected. I can also put instead of global a global1 or so as channel name, but I thought I stick with global since it would refer directly to the config section. If both files exist, the one which is loaded first is being the global, which could lead to very confusing runtime behavior.

src/conf_mode/syslog.py
hagbard-01: invoke-rc.d is a script testing is at the first place (it is used for packages), I'll change it to systemctl then.

src/conf_mode/syslog.py
hagbard-01: in progress

interface-definitions/syslog.xml
hagbard-01: If unchecked a 'set system syslog host aaa bbb ...' inserts you a space, that leads to a lot of fun in the parser. We definitely need to check it, or I need to extend syslog.py and check it there for whites spaces, backslashes etc. Just wanted to know your opinion and how you guys usually deal with that.

hagbard-01:
Yes, everything is working, I also implemented user which was not in the original perl script. The user option did just nothing.

interface-definitions/syslog.xml
hagbard-01: set system syslog host "host name" facility all

+        host host name {
+            facility all {
+            }
+        }


would be the result, I can check in verify() but it will be already in the config then, that's why I think checking it with a regex is better.

dmbaturin:
Thanks, I'll review the changes.

hagbard-01:
Will this be merged?

dmbaturin:
Sorry for late reply, been busy with the VRRP stuff.
Looks good now, I may clean up some minor issues later but otherwise it looks fine.

PR #24, opened 2018-08-02
T754: add DNSSEC to DNS forwarder

mb300sd:
Add dnssec support to forwarder. Keeps existing default from powerdns

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -115,3 +115,4 @@ debian/vyos-1x.*
 
 # Sonar Cloud
 .scannerwork
+/.vs
--- a/interface-definitions/dns-forwarding.xml
+++ b/interface-definitions/dns-forwarding.xml
@@ -32,6 +32,34 @@
                   <multi/>
                 </properties>
               </leafNode>
+              <leafNode name="dnssec">
+                <properties>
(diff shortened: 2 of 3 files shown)


dmbaturin:
Looks good? Do we have a task about it in https://phabricator.vyos.net ?
Our standard procedure it to create a task and reference it in commits (e.g. "T42: Add DNSSEC support to DNS forwarder"), so that they are automatically linked with commits, and it's easy to build a changelog from all VyOS packages.

c-po:
Phybricator task is at https://phabricator.vyos.net/T754:

@mb300sd please adjust your commit headline (prefix with task id) to: `T754: add DNSSEC to DNS forwarder` thus the commit will be automatically referenced in the phabricator task. 

Thanks for the contribution, much appreciated!

dmbaturin:
Yes, it seems like a formality, but if you have more than one package, tracking changes and making changelog can become a real nightmare if conventions are not followed!

mb300sd:
All done. Sorry, new to this and the phabricator is kinda confusing.

dmbaturin:
We usually prepend tasks to all commits, but this is sufficient I believe.

PR #27, opened 2018-08-11
T427, T773 adding wirguard support

hagbard-01:
Since I had the routing and keying part added to the conf mode script. I've tested the package via manual install, when I build the iso it downloads from dev.packages.vyos.net, so I just did an install image and a dpkg -i to check that there are no issues. Once I have the conf mode script cleaned up, I do another pull request for it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ clean:
 
 .PHONY: test
 test:
-#	PYTHONPATH=python/ python3 -m "nose" --with-xunit src --with-coverage --cover-erase --cover-xml --cover-package src/conf_mode,src/op_mode,src/completion,src/helpers,src/validators --verbose
+	PYTHONPATH=python/ python3 -m "nose" --with-xunit src --with-coverage --cover-erase --cover-xml --cover-package src/conf_mode,src/op_mode,src/completion,src/helpers,src/validators --verbose
 
 .PHONY: sonar
 sonar:
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ clean:
 
 .PHONY: test
 test:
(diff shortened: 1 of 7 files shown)


dmbaturin:
Wait, but where is the conf mode definitions and scripts for it now?

hagbard-01:
conf mode and the conf script comes shortly. I have to remove the key generation and routing options like you suggested and do a few tests with it. I just pushed the stuff which can be released already. I'm currently finishing up the vyos-build (automatic module loading), after that I can remove the loader from the scripts (they don't hurt, but it's the later on cleaner). The conf script and interface is then just the last piece, I have it already in my own branch, but as mentioned needs some cleanup, since I had everything in one script.

hagbard-01:
Is still anything holding a merge up?

dmbaturin:
I see you've added the conf mode. Is it complete now?

hagbard-01:
Yes it is. It is a working version, but I plan to add a few things in the next couple of weeks, like show status and the such. I was 't able to reproduce Kim's comment, I downloaded the latest rolling, installed the wireguard packages and my version of vyos-1x, rebooted and didn't see that kernel module was loaded. I have a few code lines in both scripts which check if the module is loaded and if not it does it automatically and drops a message to syslog. I was playing around with it, it works pretty stable, I didn't come over any issues so far, but I also tested everything in a local network (checked with tcpdump the traffic). For the future, we need some type of standardized procedure to have kmods automatically loaded, in my opinion they should be in the package if the pkg is provided by vyos, if it comes via debian it should be in vyos-build.

dmbaturin:
Ok, let's merge it then. Please make new PRs when you add more features!

PR #28, opened 2018-08-15
T689: Added show ntp op mode definition

alkersan:
The corresponding old-style def can be found [here](https://github.com/vyos/vyatta-op/pull/14/files).
While this is a tiny change, there are few things I'd like to clarify.

1. Is GitHub appropriate for discussions or everything should go through Phabricator?
2. The old style node was simultaneously a tag and a command, i.e. both operations were possible `show ntp` and `show ntp 0.pool.ntp.org`. ~~Unfortunately I didn't find a way to express this with new xml schema, thus the ported syntax has intermediate server tag `show ntp server 0.pool.ntp.org`~~.
3. The `list_ntp_servers.sh` completion script seems redundant. My intuition is that path-based completion helper `<path>system ntp server</path>` should produce the same list. But currently the op template generator [uses](https://github.com/vyos/vyos-1x/blob/19a1576263cf72de200094de79095b1a0ae81ec8/scripts/build-command-op-templates#L114) `cli-shell-api listNodes`, which [segfaults](https://phabricator.vyos.net/T734#16682) when used not in conf mode. Would it appropriate to generate something like [this](https://github.com/alkersan/vyos-1x/compare/ntp_show...alkersan:use_list_effective_nodes?expand=1) for the op mode path completions?
```
local -a ARR=$(cli-shell-api -- listEffectiveNodes system ntp server) && echo ${ARR[@]//\'/}"
```


c-po:
Hi,

I just did a quick build/test and see the following:

**VyOS 1.2.0-rolling+201808151055**
```
cpo@CR1:~$ show ntp
Possible completions:
  <Enter>       Execute the current command
  0.pool.ntp.org
                Show date and time of specified NTP server
  1.pool.ntp.org
  2.pool.ntp.org


cpo@CR1:~$ show ntp
     remote           local      st poll reach  delay   offset    disp
=======================================================================
=199.182.221.110 172.16.37.251   16 1024    0 0.00000  0.000000 3.99217
=5.189.146.13    172.16.37.251   16 1024    0 0.00000  0.000000 3.99217
=80.151.151.109  172.16.37.251   16 1024    0 0.00000  0.000000 3.99217
```

**installing your PR on top of VyOS 1.2.0-rolling+201808151055**

```
cpo@LR1:~$ show ntp
Possible completions:
  <Enter>       Execute the current command
  <text>        Show date and time of specified NTP server


cpo@LR1:~$ show ntp
/usr/bin/ntpdc: read: Connection refused
```

For question 3 please see https://github.com/vyos/vyos-1x/blob/current/op-mode-definitions/show-configuration.xml#L10, you should use something like: `<command>cli-shell-api -- listEffectiveNodes system ntp server</command>`

alkersan:
@c-po thank you for review. 
Regarding the `ntpdc` error - I've missed to check if the daemon is actually running, as it was [in the original](https://github.com/vyos/vyatta-op/pull/14/files#diff-b965fc02367556d897512d9108ba4a26L1) vyatta-op script. Now it's [inlined](https://github.com/vyos/vyos-1x/pull/28/files#diff-fcb80d1eaaee16a7c7868ee40837929bR9) into the xml command.

Regarding the completion helper from question 3. I'll try to rephrase what I mean.
I think [there is](https://github.com/vyos/vyos-1x/blob/current/scripts/build-command-op-templates#L113) is a bug in the code generator for the `path-based completion helper` expressions of the op-mode definitions. For example, given this definition:
```
<completionHelp>
  <path>system ntp server</path>
</completionHelp>
```

the resulting node.def will contain:
```
allowed: /bin/cli-shell-api listNodes system ntp server
```
which in turn leads to a segfault, if the completion is called not from the conf-mode (as was described [here](https://phabricator.vyos.net/T734#16682)).
This means that all path completions in the [op-mode-definitions](https://github.com/vyos/vyos-1x/tree/current/op-mode-definitions) are broken at the moment. Good thing is that now there are literally zero of them, thus the error was probably not observed. `show ntp` operation, i guess, is the first one that may use this completion helper. That's why the proposed change [here](https://github.com/vyos/vyos-1x/pull/28/commits/c9fc3691a0400fed3f49ee413096e26834f16dc1#diff-4029b3c4f4693779fd64c7ddbd3888f9R113)

c-po:
Thanks for the detailes explanation, now I understand. The current behavior is correct as it only supports `listNodes` and not `listEffectiveNodes ` (some background: https://wiki.vyos.net/wiki/Perl_API#Effective_config) but this is confusing. A developer would expect this to work. @dmbaturin is the master of vyos-1x and maybe he can say some words about this.

For the time beeing I propose to use the same implementation as everybody else does for op-mode commands by using: `<command>cli-shell-api -- listEffectiveNodes system ntp server</command>` instead of 

```
<completionHelp>
  <path>system ntp server</path>
</completionHelp>
```

until this can be refactored at once. Your change in `scripts/build-command-op-templates` looks promising but IMHO this should be done in a separate commit refactoring all current op-mode commands to use this better implementation.

alkersan:
But the `<command>` can't be used for completionHelp, right? I mean, in this specific case, 
`show ntp <TAB>` shoud propose a list of currently configured NTP peers, so that it can be used in the actual command as: `<command>/usr/sbin/ntpdate -q "$3"</command>`
I can revert it to the initial workaround:
```
<completionHelp>
  <script>${vyos_completion_dir}/list_ntp_servers.sh</script>
</completionHelp>
```
where the `list_ntp_servers.sh` should be something like:
```
local -a ARR=$(cli-shell-api -- listEffectiveNodes system ntp server)
echo "${ARR[@]//\'/}"
```
As described in the [cli shell wiki](https://wiki.vyos.net/wiki/Cli-shell-api#cite_ref-2), the results of the `list*` commands should be evaluated. Also the quotes should be stripped, otherwise the completions will be quoted:
```
$> show ntp
Possible completions:
  <Enter>       Execute the current command
  '0.pool.ntp.org'
                Show date and time of specified NTP server
  '1.pool.ntp.org'
  '2.pool.ntp.org'
```

c-po:
I'm totally ok with the helper for the time beeing.

alkersan:
Sorry if I made a lot of noise about nothing.

Oh, I've also did one change I've forgot to mention. [Currently](https://github.com/vyos/vyatta-op/blob/current/templates/set/date/ntp/node.def#L1) the `set date ntp` uses the `0.vyatta.pool.ntp.org` server as a source. The pool [seems alive](https://stat.ripe.net/0.vyatta.pool.ntp.org#tabId=dns), but is it safe to use, given that other Vyatta web resources are effectively dead by now? That was the reasoning behind [the change](https://github.com/vyos/vyos-1x/pull/28/files#diff-9757708e44286eafcaf32842d9f31bcfR47) for `0.pool.ntp.org`. Is this acceptable?

c-po:
Definately! Good catch!

captify-daleksandrov:
Ok, I think that's all for such a tiny contrib. At least now I have a bit more clear picture of what is going on in VyOS development. These tests were run manually on top of the version `1.2.0-rolling+201808171128`:
```
show ntp
show ntp 169.254.169.123
set date 08171200
set date ntp
show date utc
set date ntp 169.254.169.123
show date utc
configure
delete system ntp
commit
run show ntp
```

alkersan:
Ok, I think that's all for such a tiny contrib. At least now I have a bit more clear picture of what is going on in VyOS development. These tests were run manually on top of the version `1.2.0-rolling+201808171128`:
```
show ntp
show ntp 169.254.169.123
set date 08171200
set date ntp
show date utc
set date ntp 169.254.169.123
show date utc
configure
delete system ntp
commit
run show ntp
```

PR #29, opened 2018-08-16
T689: Rewritten poweroff/reboot script to use systemd-shutdownd service

runborg:
After rewriting xml of poweroff/reboot script the perl script was not rewritten. 
I've rewritten it in python to make use of the systemd-shutdownd service instead of scheduling with AT.

### DIFF
--- a/op-mode-definitions/poweroff.xml
+++ b/op-mode-definitions/poweroff.xml
@@ -16,31 +16,41 @@
 
       <leafNode name="cancel">
         <properties>
-          <help>Cancel a pending reboot</help>
+          <help>Cancel a pending poweroff</help>
         </properties>
         <command>sudo ${vyos_op_scripts_dir}/powerctrl.py --cancel</command>
       </leafNode>
 
     <tagNode name="in">
       <properties>
-        <help>Reboot in X minutes</help>
+        <help>Poweroff in X minutes</help>
         <completionHelp>
-          <list>Minutes</list>
+          <list>&lt;Minutes&gt;</list>
         </completionHelp>
       </properties>
-      <command>sudo ${vyos_op_scripts_dir}/powerctrl.py --yes --poweroff '$3'</command>
+      <command>sudo ${vyos_op_scripts_dir}/powerctrl.py --yes --poweroff $3 $4</command>
 
     </tagNode>
 
     <tagNode name="at">
       <properties>
-        <help>Reboot at a specific time</help>
+        <help>Poweroff at a specific time</help>
         <completionHelp>
           <list>HH:MM</list>
         </completionHelp>
       </properties>
-      <command>sudo ${vyos_op_scripts_dir}/powerctrl.py --yes --poweroff '$3'</command>
-
+      <command>sudo ${vyos_op_scripts_dir}/powerctrl.py --yes --poweroff $3</command>
+      <children>
+        <tagNode name="date">
+          <properties>
+            <help>Poweroff at a specific date</help>
+            <completionHelp>
+              <list>&lt;DDMMYYYY&gt; &lt;DD/MM/YYYY&gt; &lt;DD.MM.YYYY&gt; &lt;DD:MM:YYYY&gt;</list>
+            </completionHelp>
+          </properties>
+          <command>sudo ${vyos_op_scripts_dir}/powerctrl.py --yes --poweroff $3 $5</command>
+        </tagNode>
+      </children>
     </tagNode>
 
     </children>
--- a/op-mode-definitions/poweroff.xml
+++ b/op-mode-definitions/poweroff.xml
@@ -36,7 +36,7 @@
       <properties>
         <help>Poweroff at a specific time</help>
         <completionHelp>
-          <list>HH:MM</list>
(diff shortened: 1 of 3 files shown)


op-mode-definitions/poweroff.xml
```diff
@@ -2,37 +2,45 @@
 <interfaceDefinition>
   <node name="poweroff">
     <properties>
-      <help>Poweroff the system</help>
+      <help>Shutdown the system</help>
```
c-po: In the UNIX world there is a difference between shutdown and poweroff. Shutdown will only shutdown the Kernel but the system will continue to run. With poweroff we really shutdown the Kernel and the power-off the system. Please keep the old help text.

op-mode-definitions/poweroff.xml
```diff
@@ -2,37 +2,45 @@
...
     </properties>
-    <command>/opt/vyatta/bin/sudo-users/vyatta-poweroff.pl --action poweroff</command>
+    <command>sudo ${vyos_op_scripts_dir}/powerctrl.py --poweroff now</command>
+
     <children>
       <leafNode name="now">
         <properties>
-          <help>Poweroff the system without confirmation</help>
+          <help>Shutdown the system imidiently</help>
```
c-po: Typo `immediately` instead of `imidiently`. Also see comment above, I prefer the explicit wording of `without confirmation`, instead of the vague term `immediately`.

op-mode-definitions/poweroff.xml
```diff
@@ -2,37 +2,45 @@
...
+
+    </tagNode>
 
     <tagNode name="at">
       <properties>
-        <help>Poweroff at a specific time</help>
+        <help>Reboot at a specific time</help>
         <completionHelp>
           <list>HH:MM</list>
-          <list>MMDDYY</list>
```
c-po: Please keep this completion as it was present in VyOS 1.1.x.

op-mode-definitions/reboot.xml
```diff
@@ -4,35 +4,43 @@
...
       <help>Reboot the system</help>
     </properties>
-    <command>/opt/vyatta/bin/sudo-users/vyatta-reboot.pl --action reboot</command>
+    <command>sudo ${vyos_op_scripts_dir}/powerctrl.py --reboot now</command>
+
     <children>
       <leafNode name="now">
         <properties>
-          <help>Reboot the system without confirmation</help>
+          <help>Reboot the system imidiently</help>
```
c-po: See comments in poweroff command

op-mode-definitions/reboot.xml
```diff
@@ -4,35 +4,43 @@
...
+      <command>sudo ${vyos_op_scripts_dir}/powerctrl.py --yes --reboot $3</command>
+
+    </tagNode>
 
     <tagNode name="at">
       <properties>
         <help>Reboot at a specific time</help>
         <completionHelp>
           <list>HH:MM</list>
-          <list>MMDDYY</list>
```
c-po: See comments in poweroff command

op-mode-definitions/poweroff.xml
runborg: undoing change :)

op-mode-definitions/poweroff.xml
runborg: undoing

op-mode-definitions/reboot.xml
runborg: undoing

op-mode-definitions/poweroff.xml
runborg: The problem with this is that systemd-shutdownd doesent allow us to specify dates directly, an implementation of this could be added but it need to be converted into minutes until this date.   i'll try to implement it.

op-mode-definitions/reboot.xml
runborg: see comment on poweroff.xml

c-po:
Maybe something like https://stackoverflow.com/questions/28828611/date-time-format-conversion-in-python would be applicable?

op-mode-definitions/reboot.xml
runborg: implemented as reboot/poweroff at HH:MM date DD.MM.YYYY. the difference from 1.1.8 is that a time needs to be configured. but can still be sett to 00:00 to give same action as in 1.1.8.. The current implementation uses systemd-shutdownd's "add by minutes" functionality, and so the second counter is ticking and i cant reset it to :00 
Adding only time with the format HH:MM uses "add by timespec" inside systemd-shutdownd and automatically resets seconds counter to :00

runborg:
I've updated code to address the issue with not allowing input of dates, input of a date is now a argument after a time is entered. f.ex:   "reboot at 00:00 date 23.10.2018". date field tries to detect what format it is in, using . / or : as separator.     an still unresolved cosmetic issue is that shutdownd returns the date in YYYY-MM-DD format to the user as a confirmation, but i think i'll leave it that way

dmbaturin:
Looks good I guess. @c-po shall we merge it now?

PR #30, opened 2018-08-17
T427, T783

hagbard-01:
T427: make endpoint an optional setting (only client config needs it, if you run as server then it pick it from the authenticated connection automatically)
T783: feature request to implement persistent-keepalive as an optional parameter


hagbard-01:
I found a bug when you delete the variable. Will request a new one, once I fixed it.

PR #34, opened 2018-08-18
T789: Quick hack to get reboot working again

hagbard-01:
Maybe it would be better to use a dict rather than an array and test it's length. This is just a quick hack to get reboots working again.


runborg (commented):
I think this is the wrong place to patch this fault. 

try this instead:  (in the end of main():

        if args.reboot is not None:
            execute_shutdown(args.reboot, reboot=True, ask=args.yes)
        if args.poweroff is not None:
            execute_shutdown(args.poweroff, reboot=False,ask=args.yes)

hagbard-01:
It was just a quick hack after I noticed in my test/dev environment that I was unable to reboot via vbash. Would you mind grabbing https://phabricator.vyos.net/T789?
I closed this PR now.

PR #35, opened 2018-08-19
T688: parsing component version string

ovsiannikov:
reference: https://phabricator.vyos.net/T688
Implemented the extraction of the components versions dictionary.


dmbaturin:
Since current is now separate from crux, I suppose it's time to start merging patches meant for 1.3.0

I would argue the get from file function is redundant, but I'll fix it myself. Thanks!

PR #37, opened 2018-08-20
T783 Bugfix for persistent-keepalive

hagbard-01:
fixed the regex and removed the functional if config value equals effective value


hagbard-01:
Oops, gotta rebase to remove the quick hack for reboot, which is by the way still broken.

PR #38, opened 2018-08-20
T427, T783 minor bugfixes

hagbard-01:
- fixes the regex, now at least 1 digit is required to set (1 sec for instance)
- fixes an issue if persistent-keepalive isn't set and an update on the config is done, it would show a keyerror for it since it's not set in the config dict.


c-po:
For commit 2a4b380ecdf5c51a41dfee64652cd77789a8204c a prefix would be nice indicating what part is changed (`wireguard.py:`) for instance. THX for the continuing support.

PR #39, opened 2018-08-20
T790: wireguard: add status commands

hagbard-01:
more commands can be implemented upon request, wouldn't make much sense to display the same information under different trees.


hagbard-01:
Just read your comment on #38, change log:
- Makefile: added the removal of op-mode show/interfaces/node.def when pkg is being generated
- added status commands to op-mode-definitions/wireguard.xml (show interfaces wireguard ...)

PR #40, opened 2018-08-21
T689: Moved clear console and reset terminal ops from vyatta-op

alkersan:
vyatta-op cleanup PR is [here](https://github.com/vyos/vyatta-op/pull/16)


alkersan:
@dmbaturin seems that cleaning [PR](https://github.com/vyos/vyatta-op/pull/16) was merged a bit early

PR #47, opened 2018-08-30
Bring VRRP configuration in line with keepalived config documentation.

Watcher7:
Minor changes. Related to T814

### DIFF
--- a/interface-definitions/vrrp.xml
+++ b/interface-definitions/vrrp.xml
@@ -146,7 +146,7 @@
                 <properties>
                   <help>Preempt delay (in seconds)</help>
                   <constraint>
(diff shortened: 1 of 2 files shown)


c-po:
[Manpage](http://www.keepalived.org/manpage.html) states:
```
# Seconds after startup or seeing a lower priority master until preemption
# (if not disabled by "nopreempt").
# Range: 0 (default) to 1000 (e.g. 4.12)
# NOTE: For this to work, the initial state of this
# entry must be BACKUP.
preempt_delay 300    # waits 5 minutes
```

We now make use of the default settings of Keepalived which is preferrable.

PR #48, opened 2018-09-03
T71: Add more strongSwan Charon settings

unixninja92:
Added more strongswan charon settings for T71

### DIFF
--- a/interface-definitions/ipsec-settings.xml
+++ b/interface-definitions/ipsec-settings.xml
@@ -9,10 +9,52 @@
               <help>Global IPsec settings</help>
             </properties>
             <children>
+              <leafNode name="unencrypted-mainmode">
+                <properties>
+                  <help>Accept unencrypted ID and Hash messages from legacy devices</help>
+                  <valueless/>
+                </properties>
+              </leafNode>
+              <leafNode name="ikesa-sa-segments">
+                <properties>
+                  <help>The number of segments to breakup the IKE_SA hash table into</help>
+                  <constraint>
+                    <validator name="numeric" argument="--positive"/>
+                  </constraint>
+                  <constraintErrorMessage>must be positive integer</constraintErrorMessage>
+                </properties>
+              </leafNode>
+              <leafNode name="ikesa-sa-table-size">
+                <properties>
+                  <help>Size of the IKE_SA hash table</help>
+                  <constraint>
+                    <validator name="numeric" argument="--positive"/>
+                  </constraint>
+                  <constraintErrorMessage>table size must be positive integer</constraintErrorMessage>
+                </properties>
+              </leafNode>
               <leafNode name="disable-route-autoinstall">
                 <properties>
                   <valueless/>
                   <help>Do not automatically install routes to remote networks</help>
+                </properties>
+              </leafNode>
+              <leafNode name="disable-virtual-ip-autoinstall">
+                <properties>
+                  <valueless/>
+                  <help>Do not automatically install virtual ip addresses to outbound interface</help>
+                </properties>
+              </leafNode>
+              <leafNode name="cisco-unity">
+                <properties>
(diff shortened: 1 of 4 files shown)


interface-definitions/ipsec-settings.xml
```diff
@@ -9,12 +9,54 @@
               <help>Global IPsec settings</help>
             </properties>
             <children>
+              <leafNode name="enable-unencrypted-mainmode-messages">
```
c-po: Can we probably rename this leafNode to: `unencrypted-mainmode`? CLI should be not to long. B/c everything else is stated in the help message.

interface-definitions/ipsec-settings.xml
```diff
@@ -9,12 +9,54 @@
...
+                <properties>
+                  <help>Accept unencrypted ID and Hash messages from legacy devices</help>
+                  <valueless/>
+                </properties>
+              </leafNode>
+              <leafNode name="ikesa-table-segments">
```
c-po: Maybe `ike-sa-segments` is more human readable?

interface-definitions/ipsec-settings.xml
```diff
@@ -9,12 +9,54 @@
...
+                <properties>
+                  <help>The number of segments to breakup the IKE_SA hash table into</help>
+                  <constraint>
+                    <validator name="numeric" argument="--positive"/>
+                  </constraint>
+                  <constraintErrorMessage>must be positive integer</constraintErrorMessage>
+                </properties>
+              </leafNode>
+              <leafNode name="ikesa-table-size">
```
c-po: Can we use `ike-sa-table-size` to make it more human readable?

interface-definitions/ipsec-settings.xml
```diff
@@ -9,12 +9,54 @@
...
+                  <constraintErrorMessage>table size must be positive integer</constraintErrorMessage>
+                </properties>
+              </leafNode>
               <leafNode name="disable-route-autoinstall">
                 <properties>
                   <valueless/>
                   <help>Do not automatically install routes to remote networks</help>
                 </properties>
               </leafNode>
+              <leafNode name="disable-virtual-ip-autoinstall">
```
c-po, line 42: Could this be shortened to `disable-address-autoinstall` then we can use it for IPv4 and IPv6.

interface-definitions/ipsec-settings.xml
```diff
@@ -9,12 +9,54 @@
...
+                <properties>
+                  <valueless/>
+                  <help>Do not automatically install virtual ip addresses to outbound interface</help>
+                </properties>
+              </leafNode>
+              <leafNode name="send-cisco-unity-id">
```
c-po: I would name it only `cisco-unity` b/c everything else is explained in the help below.

interface-definitions/ipsec-settings.xml
```diff
@@ -9,12 +9,54 @@
...
+                <properties>
+                  <valueless/>
+                  <help>Send Cisco Unity vendor ID payload (IKEv1 only)</help>
+                </properties>
+              </leafNode>
+              <leafNode name="send-strongSwan-id">
```
c-po: I think this is the first node having camelCase. Can we name it `strangswan-id` instead?

interface-definitions/ipsec-settings.xml
unixninja92: good idea

unixninja92:
I can squash down to one or two commits if needed.

c-po:
@unixninja92 plesae do if you don't mind.

unixninja92:
squashed

c-po (commented):
WHat about the possible shorting of other nodes in this commit? Makes everyones life a bit easier?. Later API changes are always painfull.

PR #55, opened 2018-10-15
Bugfixes: T870,T886

hagbard-01:
- T870: Commit-confirm restarts the server even after commit
- T886: validation logic in `interfaces wireguard wgX address x.x.x.x broken

### DIFF
--- a/interface-definitions/wireguard.xml
+++ b/interface-definitions/wireguard.xml
@@ -29,7 +29,7 @@
               </valueHelp>
               <multi/>
               <constraint>
-                <validator name="interface-address"/>
+                <validator name="ip-prefix"/>
               </constraint>
             </properties>
           </leafNode>
--- a/src/op_mode/powerctrl.py
+++ b/src/op_mode/powerctrl.py
@@ -2,141 +2,158 @@
 import os
 import sys
 import argparse
+import subprocess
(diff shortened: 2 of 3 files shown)


dmbaturin:
@hagbard-01 If you make pull requests despite being a committer now, please make them one feature per pull request still. ;)

Also, this should go to current first and then get cherry-picked/backported to long term support branches.

hagbard-01:
Gotcha. I was under the assumption that I always have to do PRs and the merge them in.

dmbaturin:
@hagbard-01 My objection is not against pull requests, but against merging your own pull requests, which defeats the purpose and only creates noise in the commit history.

Ideally we should strive to get all code reviewed by one another before it gets into mainline. In practice, I believe uncontroversial fixes are impractical to handle this way; while syntax and behaviour changes should always be discussed first.

hagbard-01:
@dmbaturin I totally agree, but that will delay development quite a bit. However, for all rc merges it would be a good thing in my opinion. I can also do PRs and wait until it's merged, but as mentioned already it slows down everything a bit, since I usually test everything fir a new generate rolling image, to see that all my changes work as expected.
I think sooner or later we should structure ourselves into some type of teams who review each others work.
Let me know how you want to proceed, I'm open for everything.

PR #56, opened 2018-11-08
T979: Allow spaces in wireguard interface descriptions

bswinnerton:
Previous to this pull request, setting a Wireguard interface description would result in a validation error similar to the following:

```
brooks@border# set interfaces wireguard wg0 description "Tunnel"
[edit]
brooks@border# set interfaces wireguard wg0 description "Tunnel tunnel
tunnel"

  interface description is too long (limit 100 characters)
  Value validation failed
  Set failed

[edit]
```

This commit makes the regex less restrictive up to 100 characters.

### DIFF
--- a/interface-definitions/wireguard.xml
+++ b/interface-definitions/wireguard.xml
@@ -18,7 +18,7 @@
         <children>
           <leafNode name="address">
             <properties>
-              <help>IP address</help> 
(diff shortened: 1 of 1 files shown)


hagbard-01:
Thank you for your contribution.

PR #58, opened 2018-11-08
T978: Support PowerDNS Recursor outbound queries over IPv6.

gsadams:
This requires adding a query-local-address6 setting to enable outbound
IPv6 queries in general, and also formatting upstream nameserver IPv6
addresses in such a way that Recursor can parse them.

### DIFF
--- a/src/conf_mode/dns_forwarding.py
+++ b/src/conf_mode/dns_forwarding.py
@@ -39,6 +39,8 @@ threads=1
 allow-from=0.0.0.0/0, ::/0
 log-common-errors=yes
 non-local-bind=yes
+query-local-address=0.0.0.0
+query-local-address6=::
 
 # cache-size
 max-cache-entries={{ cache_size }}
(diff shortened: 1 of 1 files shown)


c-po:
Thx for the contribution. Usually we prefix our commit headline with the phabricator task id, thus commits can be automassigned to tasks. Can you please reword your commit?

src/conf_mode/dns_forwarding.py
```diff
@@ -39,6 +39,7 @@
 allow-from=0.0.0.0/0, ::/0
 log-common-errors=yes
 non-local-bind=yes
+query-local-address6=::
```
c-po, line 43: Could you please add `query-local-address=0.0.0.0/0` for completion, too? 
Then we have both IPv4 and IPv6 statements.

gsadams:
Done.

c-po (changes requested):
Could you please add query-local-address=0.0.0.0/0 for completion, too?
Then we have both IPv4 and IPv6 statements.

src/conf_mode/dns_forwarding.py
gsadams: Done (but without the /0; it doesn't like that). This is the default value, so I didn't include it, but it does seem valuable to be explicit.

c-po (approved):

PR #59, opened 2018-11-26
add tests & install pipenv

kamijin-fanta:
- add tests for ntp configure
  - there was a bug that does not deepCopy the default_config_data in `conf_mode/ntp.py`, so I fixed it.
- install pipenv
  - standard version control system in Python
  - Pipfile, Pipfile.lock was added
- fix makefile
  - the test with main is not executed by default, I added `--exe` flag


c-po:
Thanks for the contribution. Our `Pipfile.lock` contains a Cygwin/LXSS path for a Windows machine, can this be?

kamijin-fanta:
@c-po I fixed it! Thank you!

PR #62, opened 2018-12-16
T1104 : "show vpn ipsec sa" crashes

daniel-pro:
Updated show_ipsec_sa.py, added int() fuction to convert bytes_in and bytes_out from string to int.


dmbaturin:
Thanks!

PR #63, opened 2018-12-18
T1077: Update show_ipsec_sa.py

daniel-pro:
Updated show_ipsec_sa.py, it now considers the tunnel down when the ipsec statusall command shows the tunnel in CONNECTING state so that it doesn't parse the string looking for ESTABLISHED and doesn't raise the error :
File "/usr/libexec/vyos/op_mode/show_ipsec_sa.py", line 51, in
raise e
File "/usr/libexec/vyos/op_mode/show_ipsec_sa.py", line 39, in
time, _, _, ip, id = parse_conn_spec(status)
File "/usr/libexec/vyos/op_mode/show_ipsec_sa.py", line 11, in parse_conn_spec

### DIFF
--- a/src/op_mode/show_ipsec_sa.py
+++ b/src/op_mode/show_ipsec_sa.py
@@ -32,7 +32,7 @@ status_data = []
 
 for conn in connections:
     status = subprocess.check_output("ipsec statusall {0}".format(conn), shell=True).decode()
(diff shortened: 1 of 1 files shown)


daniel-pro:
I might have referred the wrong task number. If that's the case please let me know.

PR #65, opened 2019-02-21
T1256: Execute "show ipsec vpn ipsec sa" returns incorrect results

daniel-pro:
Not sure it's a normal case scenario, the one highlighted in T1256.
To manage it I changed the "if" logic.


dmbaturin:
Sounds reasonable I think.

By the way, @UnicronNL and I got StrongSWAN's vici to work, so we'll soon be replacing that script and others with much simpler, vici-based versions.

PR #66, opened 2019-04-18
router bytes zeros

tmarlok88:
In the push classless route part the script removes zeros from the router bytes. This is not the desired behavior.
For example this

```
static-route {
    destination-subnet "172.22.128.0/17"
    router "172.22.0.12"
}
```
will create this in the `dhcpd.conf`:
```
option rfc3442-static-route 17,172,22,128,172,22,12;
option windows-static-route 17,172,22,128,172,22,12;
```
(The zero disappeared in the router address)

### DIFF
--- a/src/conf_mode/dhcp_server.py
+++ b/src/conf_mode/dhcp_server.py
@@ -572,10 +572,9 @@ def get_config():
                         # add router bytes
                         bytes = subnet['static_router'].split('.')
                         for b in bytes:
-                            if b != '0':
-                                string += b
-                                if b is not bytes[-1]:
-                                    string += ','
+                            string += b
+                            if b is not bytes[-1]:
+                                string += ','
 
                         subnet['static_route'] = string
 


dmbaturin:
Hi @tmarlok88,
Thanks for your contribution. Please note that our usual procedure is to create a task in https://phabricator.vyos.net and reference the task number in the commits (e.g. "[dhcp-server] T9999: do not remove trailing zeroes from subnets in DHCP static route config"). Else tracking changes across multiple submodules and making release notes becomes nearly impossible.

Could you make a task with an explanation of why the old behaviour is wrong?

tmarlok88:
Hi @dmbaturin,
Sorry about that. I created a task: https://phabricator.vyos.net/T1343
Please feel free to correct me if I missed something, this is my first PR.
Should i make a new commit referencing the task?

dmbaturin:
Thanks! You can edit existing commit and force push perhaps. Or make a new pull request if nothing else.

PR #72, opened 2019-06-16
T1439 dhcp server static mapping fix

jjakob:
See https://phabricator.vyos.net/T1439
Reopened from new branch on fork.


interface-definitions/dhcpv6-server.xml
```diff
@@ -278,25 +278,36 @@
...
                       <leafNode name="identifier">
                         <properties>
-                          <help>Client identifier for this static mapping</help>
+                          <help>Client identifier (DUID) for this static mapping [REQUIRED]</help>
+                          <valueHelp>
+                            <format>h[[:h]...]</format>
+                            <description>DUID: colon-separated hex list (as used by isc-dhcp option dhcpv6.client-id)</description>
+                          </valueHelp>
+                          <constraint>
+                            <validator name="dhcpv6-client-id"/>
```
c-po: Why can't you use an inlined regex here?

```
          <constraint>
            <regex>([0-9A-Fa-f]{1,2}[:])*([0-9A-Fa-f]{1,2})$</regex>
          </constraint>
```

Then there is no need for the file `src/validators/dhcpv6-client-id`

c-po (approved):

interface-definitions/dhcpv6-server.xml
```diff
@@ -278,25 +278,36 @@
                   <tagNode name="static-mapping">
                     <properties>
                       <help>Name of static mapping</help>
+                      <constraint>
+                        <regex>^[-_a-zA-Z0-9.]+$</regex>
```
c-po: There is no need to preceed a regex with ^ as this is alsays done „under the hood“

jjakob: There are 5 other places in the file (and more in others) that also begin with ^, shouldn't they all be the same?

interface-definitions/dhcpv6-server.xml
jjakob: Already done in https://github.com/vyos/vyos-1x/pull/72/commits/f47d1d0fb50784e940a1bedba2806e236b6f8984

interface-definitions/dhcpv6-server.xml
jjakob: I see, validate-value.py uses re.fullmatch so the ^ is implied. IMO it should be removed everywhere then and not just here, to be consistent.

c-po: True, but that should be a separate commit then.

jjakob: removed https://github.com/vyos/vyos-1x/pull/72/commits/6b86cd54c12d34037d00667eaeaf2dc68d0a41a1

PR #77, opened 2019-06-24
 T1470: improve output of "show dhcpv6 server leases"

jjakob:
- change DUID to IAID_DUID
- format IAID_DUID as colon-separated hex list
- implement functions: pool, sort, state
- add op-mode definitions for pool, sort, state
- add columns: State, Type, Last communication, Pool
- implement json output


jjakob:
Tested working on current rolling

PR #78, opened 2019-06-25
[pdns-recursor] T1469 - specified dns forwarding not work

qiuchengxuan:
when conflict exists between forward-zone-recurse entry,
the lower one hides the upper one, which leads to inactive dns forwarding configuration


qiuchengxuan:
@dmbaturin would like to merge this pull request?

PR #85, opened 2019-07-04
T1435 plus other dhcp/dhcpv6-server enhancements

jjakob:
- T1435: dhcp-server: make ip-address optional in static-mapping
- remove [REQUIRED] from dhcpv6-server static-mapping identifier and ipv6-address
- verify if static-mapping ipv6-address is in subnet
- make help and error messages in conf-mode more descriptive
- remove regex ^$ anchors (implied in re.fullmatch)


jjakob:
tested all on current/rolling

PR #89, opened 2019-07-13
T1523  Fix bind param if outside-address not present

DmitriyEshenko:
If in config exist `bind=` without value, accel-ppp listen wrong ip address 255.255.255.255:1723. If need default behavior with listening on 0.0.0.0:1723 we don't set empty bind option.

### DIFF
--- a/src/conf_mode/accel_pptp.py
+++ b/src/conf_mode/accel_pptp.py
@@ -84,7 +84,9 @@ wins2={{wins[1]}}
 {% endif %}
 
 [pptp]
+{% if outside_addr %}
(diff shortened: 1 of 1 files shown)


dmbaturin:
Do we have a task in phabricator about it?
This is a non-trivial change and a fix for a bug, so we should have it properly recorded.

The usual procedure is to first create a task, then reference it in the commit (T42: fix this and that).

DmitriyEshenko:
@dmbaturin, thank you for your reply. I create task in phabricator and change PR name. T1523

PR #92, opened 2019-07-21
Equuleus rebase

runborg:
rebase of vyos-1x:current into vyos-1x:equuleus

No issues found under rebase


dmbaturin:
I've just pushed a merge so that we can avoid duplicate commits.

PR #93, opened 2019-07-25
T1541 Fix: adding additional check

DmitriyEshenko:


c-po:
Thanks for the contribution. A more descriptive headline would be appreciated next time - e.g. `[dhcpv6-relay] T1541: fix missing if statement`

PR #103, opened 2019-08-13
[service https] T1443: add self-signed TLS certificate

jestabro:


jestabro:
Added config setting for lifetime of certificate; refactored.

PR #108, opened 2019-08-19
[op-mode] T1596 rewrite 'telnet' and 'traceroute' operations to xml style

alkersan:
Please review these op-mode rewrites. See the corresponding PR in vyatta-op [here](https://github.com/vyos/vyatta-op/pull/25).

**Notes and questions**
1. Are [these](https://github.com/vyos/vyos-1x/pull/108/files#diff-611eddb10fca5cae1ad6194da734d881R8) [usages](https://github.com/vyos/vyos-1x/pull/108/files#diff-611eddb10fca5cae1ad6194da734d881R17) of 'nameless' `tagNode` valid? Didn't find other way to reproduce the original [node structure](https://github.com/vyos/vyatta-op/tree/437867584488ec3bf1bdd28ddfdd45410e9a5bf9/templates/telnet) as a series of `node.tag`s. Any named tagNode in xml creates two levels of nesting. Also the root node can't be tagged.

2. What was the point in doing hostname resolution manually in the [original traceroute scripts](https://github.com/vyos/vyatta-op/blob/437867584488ec3bf1bdd28ddfdd45410e9a5bf9/templates/traceroute/ipv4/node.tag/node.def#L14)? It seems that traceroute does that itself when a hostname is passed.

3. Should the required packages be specified as deb dependencies in the [control script](https://github.com/vyos/vyos-1x/pull/108/files#diff-d837a1c17de0268bcea321239ddbed47R64)? It seems like yes, but [vyatta-op listed](https://github.com/vyos/vyatta-op/pull/25/files#diff-d837a1c17de0268bcea321239ddbed47L14) only one tool

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -61,6 +61,9 @@ Depends: python3,
   openvpn,
   openvpn-auth-ldap,
   openvpn-auth-radius,
+  mtr-tiny,
+  telnet,
+  traceroute,
   ${shlibs:Depends},
   ${misc:Depends}
 Description: VyOS configuration scripts and data
--- a/op-mode-definitions/telnet.xml
+++ b/op-mode-definitions/telnet.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0"?>
+<interfaceDefinition>
+  <node name="telnet">
+    <properties>
+      <help>Telnet to a node</help>
+    </properties>
+    <children>
+      <tagNode name="">
+        <properties>
+          <help>Telnet to a host</help>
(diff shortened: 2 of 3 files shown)


c-po:
You are absolutely right baout adding the tools to the dependenvy section of vyos-1x.

About manual hostname resolution - I have no idea but I would totally let traceroute take care about it.

Regarding empty `tagNodes` I would have done it using a regular `node` with children (but I am more fit using cfg mode instead of op mode)

PR #114, opened 2019-08-23
[dummy] T1609 Fixing dummy interface state

DmitriyEshenko:


c-po:
Now we mix both `vyos.configinterface` and `vyos.interfaceconfig` where the former was an abstraction before `vyos.interfaceconfig` was available.

Please take the chance and migrate to `vyos.interfaceconfig`.

PR #122, opened 2019-08-31
[dummy] Deleting default description/alias if it not set by cli

DmitriyEshenko:


c-po:
Thanks for the fix.

Its already in aa5054e524a4996ba9cc11deec25fdf4279b29cf and also for loopback in c306c248ff2ebdbd2f669e8c052a6ecd85a50fad so I close this.

PR #130, opened 2019-09-19
[OpenVPN] T1670: Added OpenVPN options: tls-auth, tun-mtu, ncp-ciphers, minimum tls ve…

vindenesen:
…rsion and added keepalive option to client section.

Also added checks to see if certificate files exists before trying to check the content of them.

Btw, this is my first ever pull request on Github, so please bear with me if I'm not following proper procedure.

### DIFF
--- a/interface-definitions/interfaces-openvpn.xml
+++ b/interface-definitions/interfaces-openvpn.xml
@@ -543,6 +543,34 @@
                   <help>File containing this host's private key</help>
                 </properties>
               </leafNode>
+              <leafNode name="auth-file">
+                <properties>
+                  <help>File containing tls static key for tls-auth</help>
+                </properties>
+              </leafNode>
+              <leafNode name="minimum-tls-version">
+                <properties>
+                  <help>Specify the minimum required TLS version</help>
+                  <completionHelp>
+                    <list>1.0 1.1 1.2</list>
+                  </completionHelp>
+                  <valueHelp>
+                    <format>1.0</format>
+                    <description>TLS v1.0</description>
+                  </valueHelp>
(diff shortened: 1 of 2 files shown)


c-po:
Thanks for the contribution. I have some question abot commit fdf5479128087902ef93a958eb6a0192db2f495f b/c this one should be split into 4 individual commits each adding one enhancement.

In addition your MTU values confuse me. Why is the lower end 1000 bytes? Is it a limit of OpenVPN? Why is the upper limit > 1500 bytes? Using a tunnel MTU >= 1500 bytes on a UDP connection will render the tunnel useless is 90% of the cases b/c encapsulating traffic (1500 bytes) and adding the overhead for IP, UDP and ethernet will easily exceed the 1500 bytes MTU in most of the provider networks.

vindenesen:
Thanks for the feedback! I will try to create four different commits/pull requests for this instead.

Regarding MTU, I have no idea what the lower and upper limits should be, so I just used 1000 - 2000. But from what I understand from your comment, the upper limit should probably be no more than 1500. I can't find any lower limit in the OpenVPN documentation, but an MTU of 1 would probably not be very useful either. I see that the MTU lower end on ethernet interface is set to 68, so I will use that instead, unless you have any other suggestions.

c-po:
There was no need to close this pull request. YOu could have splitted the commit and then force push to your fork.

c-po:
MTU has a lower limit of 68 bytes. For further information conduct RFC791.

vindenesen:
I'm a complete noob with Github. Nice to know when it becomes necessary
again.
Thanks, will read up on it.

PR #133, opened 2019-09-20
[OpenVPN] T1675: Added setting for minimum tls version

vindenesen:
```
vyos@vyos# set int open vtun0 tls tls-version-min
Possible completions:
   1.0          TLS v1.0
   1.1          TLS v1.1
   1.2          TLS v1.2


```


c-po:
Thanks for the contribution. It would be nice if the individual commits could be prefixes with the task-id, too. That makes browsing the history easier :)

PR #141, opened 2019-09-27
[NTP]: T1696: Updated ntp-test after interface ignore wildcard was moved

vindenesen:


vindenesen:
T1694 deleted this test, so this pull request isn't required anymore.

PR #146, opened 2019-10-13
[OpenVPN]: T1704: Added function for ncp-ciphers, and ability to disable it.

vindenesen:
[OpenVPN]: T1704: Changed config structure for OpenVPN encryption to support ncp-ciphers.
[OpenVPN]: T1704: Added migration scripts for interface 2-to-3

Be aware of that this changes the version of interface to 3. It seemed logical to add ncp-ciphers under encryption settings like this:
```
vyos@vyos# set int open vtun0 encryption
Possible completions:
   cipher       Standard Data Encryption Algorithm
   disable-ncp  Disable support for ncp-ciphers
+  ncp-ciphers  Data Encryption Algorithm list for use in server or client mode

```
Migration script is included. Note, this PR depends on https://github.com/vyos/vyatta-cfg-system/pull/108


c-po:
Please don't forget to also update the documentation (`equuleus`branch)

c-po:
Looks good to me. So does this NCP thing also work with old VyOS versions (1.2 `crux`) which does not have this feature? How are the cipers set here?

vindenesen:
NCP does not work with crux, OpenVPN will fall back to the cipher setting (or the default cipher if not specified). OpenVPN 2.4.0 or newer is needed for NCP to work. Clients that have NCP disabled or is older than 2.4.0 will use the cipher setting instead of ncp-ciphers. A good explanation here: https://sourceforge.net/p/openvpn/mailman/message/35669655/

I will update the documentation and set this PR ready for review when that's done.

c-po:
Looks good - you can remove the draft status for merging

vindenesen:
Finally got around to creating a PR to the documentation: https://github.com/vyos/vyos-documentation/pull/143

PR #147, opened 2019-10-21
T1755: Python KeyError exceptions raised with 'show vpn ipsec sa' command under use of certain IPSEC cipher suites.

jdmac87:
Related to [T1755](https://phabricator.vyos.net/T1755)

- Commit c761e94 resolves the issue where integ-alg is complained about as a python KeyError exception when using proposals that don't use HMAC specifically (GCM, CHACHA20_POLY1305).
- Commit 915116e resolves the issue where encr-keysize is complained about as a python KeyError exception when using proposals that don't yield a keysize specifically (CHACHA20_POLY1305 for instance).

### DIFF
--- a/src/op_mode/show_ipsec_sa.py
+++ b/src/op_mode/show_ipsec_sa.py
@@ -82,7 +82,10 @@ for sa in sas:
                 pkts_str = re.sub(r'B', r'', pkts_str)
 
                 enc = isa["encr-alg"].decode()
-                key_size = isa["encr-keysize"].decode()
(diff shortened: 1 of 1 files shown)


c-po:
@jdmac87 thanks for the contribution. Can you please put the Phabricator Task ID in front of the commit message so our Phabricator instance will do automatic tracking of this commit(s).

jdmac87:
@c-po updated and forced pushed the commit amendments. Does that look correct for Phab to pickup?

PR #151, opened 2019-10-22
T1759: Bugfix for missing interface, cleanup

kroy-the-rabbit:
There was leftover code that should have been deleted. 

Some of the display code crashed when passing in an invalid interface.


dmbaturin:
Is there a reason you are using the narrow `is not None`? The next version of the `vyos.config` library returns an empty list when there are no values, and `if ips:` would easily cover both old and new versions.

kroy-the-rabbit:
@dmbaturin Mostly because of my inexperience with python.  I'll adjust.

Closing this pull request and starting a new one because I got into some trouble with git (also due to inexperience)

PR #161, opened 2019-11-12
T1787: escape isolated single backslashes in output of showConfig

jestabro:
Escape isolated single backslashes in output of showConfig before passing to ConfigTree().


jestabro:
This should be moved to a bug task, as it affects several tasks, rather than only referring to T1787.

PR #162, opened 2019-11-12
T1801: escape isolated backslashes before passing to ConfigTree()

jestabro:


jestabro:
This requires adding python3-psutil to the Dockerfile, as the nosetests now pull in util.py from config.py, which requires psutil; do not merge until added ...

PR #170, opened 2019-12-03
T1801: move escaping of backslashes into configtree

jestabro:


jestabro:
Reviewed by dmbaturin.

PR #171, opened 2019-12-03
T1844 Fix interface configuration

varesa:
Revert an extraneous change from T1762 which changed the
parameter to conf.exists() from a string to a list

https://phabricator.vyos.net/T1844
https://phabricator.vyos.net/T1762


hagbard-01:
https://github.com/vyos/vyos-1x/pull/172 should fix that behavior. It only raised an exception if the node didn't exist.

varesa:
The issue actually seems to be a side effect of something much bigger. See https://phabricator.vyos.net/T1844 for more details

PR #174, opened 2019-12-04
T1846 Fix Config-object being affected by the shell env

varesa:
The Config-object and the contained ConfigTree seems to be expected
to be absolute paths from the root of the tree. However the way it
uses cli-shell-api with an inherited environment means that the
session_config only contains configuration nodes under the current
[edit ...] level. Override the edit-level environment variables.

https://phabricator.vyos.net/T1846


jestabro:
This has been resolved by commit 12a21a4bba6dfeb3ebb005b62748a2dbaabd84df; PR will be closed.

varesa:
Ah, didn't notice that was an option. Great!

PR #178, opened 2019-12-06
Add GCC preprocessor support for XML files

c-po:
A lot of XML code is duplicated (VLAN, interface address) for instance. Such XML definitions should be moved to `feature.xml.i` files and then just pulled in via GCC preprocessor `#include <feature.xml.i>` in the bond or ethernet xml definition. Thus we gain a single source and modularity.

When the XML files are rendered, the GCC preprocessor should be called first, inserting the xml.inc statements into an intermediate xml file which is the ised to generate the node defs.

Advantages:
* Pro: Single source
* Pro: Maintainability
* Pro: Modularity
* Con: It will look strange in the beginning

An example of an `address-ipv4-ipv6-dhcp.xml.i` file which sets an interface IPv4/IPv6 or DHCP address can look like:
```
          <leafNode name="address">
            <properties>
              <help>IP address</help>
              <completionHelp>
                <list>dhcp dhcpv6</list>
              </completionHelp>
              <valueHelp>
                <format>ipv4net</format>
                <description>IPv4 address and prefix length</description>
              </valueHelp>
              <valueHelp>
                <format>ipv6net</format>
                <description>IPv6 address and prefix length</description>
              </valueHelp>
              <valueHelp>
                <format>dhcp</format>
                <description>Dynamic Host Configuration Protocol</description>
              </valueHelp>
              <valueHelp>
                <format>dhcpv6</format>
                <description>Dynamic Host Configuration Protocol for IPv6</description>
              </valueHelp>
              <constraint>
                <validator name="ip-cidr"/>
                <regex>(dhcp|dhcpv6)</regex>
              </constraint>
              <multi/>
            </properties>
          </leafNode>
```

Further more it should be possible that an include file includes statements of other code to get real single-source definitions. This is done for example in `vif.xml.i` which looks like:

```
<tagNode name="vif">
  <properties>
    <help>Virtual Local Area Network (VLAN) ID</help>
    <valueHelp>
      <format>0-4094</format>
      <description>Virtual Local Area Network (VLAN) ID</description>
    </valueHelp>
    <constraint>
      <validator name="numeric" argument="--range 0-4094"/>
    </constraint>
    <constraintErrorMessage>VLAN ID must be between 0 and 4094</constraintErrorMessage>
  </properties>
  <children>
    #include <include/address-ipv4-ipv6-dhcp.xml.i>
    #include <include/interface-description.xml.i>
    #include <include/dhcp-dhcpv6-options.xml.i>
    #include <include/interface-disable-link-detect.xml.i>
    #include <include/interface-disable.xml.i>
    <leafNode name="egress-qos">
      <properties>
        <help>VLAN egress QoS</help>
        <completionHelp>
          <script>echo Format for qos mapping, e.g.: '0:1 1:6 7:6'</script>
        </completionHelp>
        <constraint>
          <regex>[:0-7 ]+$</regex>
        </constraint>
        <constraintErrorMessage>QoS mapping should be in the format of '0:7 2:3' with numbers 0-9</constraintErrorMessage>
      </properties>
    </leafNode>
    <leafNode name="ingress-qos">
      <properties>
        <help>VLAN ingress QoS</help>
        <completionHelp>
          <script>echo Format for qos mapping '0:1 1:6 7:6'</script>
        </completionHelp>
        <constraint>
          <regex>[:0-7 ]+$</regex>
        </constraint>
        <constraintErrorMessage>QoS mapping should be in the format of '0:7 2:3' with numbers 0-9</constraintErrorMessage>
      </properties>
    </leafNode>
    <node name="ip">
      <children>
        #include <include/interface-arp-cache-timeout.xml.i>
        #include <include/interface-enable-proxy-arp.xml.i>
      </children>
    </node>
    #include <include/interface-mac.xml.i>
    #include <include/interface-mtu-68-9000.xml.i>
  </children>
</tagNode>
```


c-po:
The code of the dummy interface is thus shrinked down to:

```
<?xml version="1.0"?>
<interfaceDefinition>
  <node name="interfaces">
    <children>
      <tagNode name="dummy" owner="${vyos_conf_scripts_dir}/interfaces-dummy.py">
        <properties>
          <help>Dummy interface name</help>
          <priority>300</priority>
          <constraint>
            <regex>dum[0-9]+$</regex>
          </constraint>
          <constraintErrorMessage>Dummy interface must be named dumN</constraintErrorMessage>
          <valueHelp>
            <format>dumN</format>
            <description>Dummy interface name</description>
          </valueHelp>
        </properties>
        <children>
          #include <include/address-ipv4-ipv6.xml.i>
          #include <include/interface-description.xml.i>
          #include <include/interface-disable.xml.i>
        </children>
      </tagNode>
    </children>
  </node>
</interfaceDefinition>
```

dmbaturin:
I still think in most cases the right solution is to restructure the CLI so that everything that maps to a single underlying component is in the same place, but address options is clearly a case where it's not possible.

PR #180, opened 2019-12-07
 scripts: import_node_def: T1836: scripts fails to create xml

elbuit:
- solved xml export
  - Add regexp field for multiple options


c-po:
Thanks for the contribution. Could you please split your commit into two - as you added two changes which are unrelated according to the commit message.

* `node.def import: T1836: fix XML export`
* `node.def import: T1836: add regex filter for condition import`

PR #199, opened 2020-01-07
bfd: T1949: fix verification logic for IPv6 BFD peers

bmhughes:
As described in T1949 IPv6 BFD peers only require a source address to be configured.

[FRR Docs](http://docs.frrouting.org/en/latest/bfd.html#clicmd-peer%3CA.B.C.D|X:X::X:X%3E[{multihop|local-address%3CA.B.C.D|X:X::X:X%3E|interfaceIFNAME|vrfNAME}])

### DIFF
--- a/python/vyos/validate.py
+++ b/python/vyos/validate.py
@@ -52,6 +52,17 @@ def is_ipv6(addr):
 
     return False
 
+def is_ipv6_link_local(addr):
+    """
+    Check addr if it is an IPv6 link-local address/network. Returns True/False
+    """
+
+    if is_ipv6(addr):
+        if ipaddress.IPv6Address(addr).is_link_local:
+            return True
+
+    return False
+
 def is_intf_addr_assigned(intf, addr):
     """
     Verify if the given IPv4/IPv6 address is assigned to specific interface.
--- a/src/conf_mode/protocols_bfd.py
+++ b/src/conf_mode/protocols_bfd.py
@@ -163,10 +163,15 @@ def verify(bfd):
     conf = Config()
 
     for peer in bfd['new_peers']:
-        # IPv6 peers require an explicit local address/interface combination
-        if vyos.validate.is_ipv6(peer['remote']):
+        # IPv6 link local peers require an explicit local address/interface
+        if vyos.validate.is_ipv6_link_local(peer['remote']):
             if not (peer['src_if'] and peer['src_addr']):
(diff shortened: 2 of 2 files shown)


zdc:
Hello, @bmhughes!
Can you confirm that a configuration without interface works well? Because, at the moment of BFD implementation, it was necessary to configure the interface also for proper work, regardless of what was noted in FRR documentation.

bmhughes:
Hi @zdc!

Yes I have a BFD peer configured like so with my patched version:
```
peer 2001:8b8:6::31 {
    multihop
    source {
        address 2001:8b8:6::30
    }
}
```

The BFD peering is up and I'm running a IPv6 BGP session with it without issue. From my take of the FRR docs you only need to specify the address, the interface is optional.

zdc:
@bmhughes, looks like you are right: now FRR accepts IPv6 neighbor without local interface with one exception - if local or peer IP address is link-local, it leads to configuration failure or bfdd crash.
Can you extend your PR a bit by adding a check for link-local addresses and raise error only when it is used without interface name? This should make check more accurate without risks for reliability.

bmhughes:
> 
> 
> @bmhughes, looks like you are right: now FRR accepts IPv6 neighbor without local interface with one exception - if local or peer IP address is link-local, it leads to configuration failure or bfdd crash.
> Can you extend your PR a bit by adding a check for link-local addresses and raise error only when it is used without interface name? This should make check more accurate without risks for reliability.

Sure no problem, I'll get that added.

zdc:
@bmhughes do not hurry so much with this change. You got confused in variable names a bit (peer['local'). :)
Try to make testing with different configuration combinations. Even FRR CLI for BFD is still buggy by itself, so you will never know what to expect. :)

bmhughes:
> 
> 
> @bmhughes do not hurry so much with this change. You got confused in variable names a bit (peer['local'). :)
> Try to make testing with different configuration combinations. Even FRR CLI for BFD is still buggy by itself, so you will never know what to expect. :)

Yes you're right I rushed that. Is there an easy way I can build an ISO off of this to test with? There doesn't seem to be an obvious way. This get build into deb packages which vyos-build picks up I'm presuming.

**edit:** Scratch that I've found it.

bmhughes:
I've tested with multiple different IPv6 configurations including using link-local address and all seems to be working as expected.

PR #210, opened 2020-01-29
T1994 - Fix lldpd jinja template

bbs2web:
Undo removal of line break and fix match expression.


c-po:
manually picked patch

PR #214, opened 2020-02-09
Correcting typo in validation of SSH key-type

dchewy:
Additional “s” in regex constraint prevents configuration of “ssh-ed25519” keys


c-po:
Thanks for the contribution. We have also encountered the issue https://phabricator.vyos.net/T2020 and I pushed the exact same fix. Sorry!

PR #216, opened 2020-02-10
get_config_dict failed when configuration was empty/missing

thomas-mangin:
using get_config_dict failed when the configuration section was missing did return a string which could not be parsed to the JSON parser for deserialisation.

### DIFF
--- a/python/vyos/config.py
+++ b/python/vyos/config.py
@@ -253,6 +253,9 @@ class Config(object):
             path = " ".join(path)
         try:
             out = self._run(self._make_command('showConfig', path))
+            # ugly but can not see a better way to handle this
+            if out == 'Configuration under specified path is empty\n':
+                return ''
             return out
         except VyOSError:
             return(default)


dmbaturin (changes requested):
Our standard procedure is to create a bug in https://phabricator.vyos.net and reference the task number in the commit message (`Txxxx: Fix incorrect behaviour of get_config_dict with empty configuration` or similar).
That way commits get automatically linked to tasks. For small additions like you did for the deb target, we sometimes skip that part, but with bugs it shouldn't be skipped because we need it in release notes.

python/vyos/config.py
```diff
@@ -253,6 +253,9 @@ def show_config(self, path=[], default=None, effective=False):
             path = " ".join(path)
         try:
             out = self._run(self._make_command('showConfig', path))
+            # ugly but can not see a better way to handle this
+            if out == 'Configuration under specified path is empty\n':
```
dmbaturin, line 257: This is fragile. I see your point, but let me check if I can make cli-shell-api exit with 1 in such cases.

thomas-mangin:
@dmbaturin ack.

I was explained by @c-po that I should use config.exists and config.exists_effective to prevent this.
no need for this patch. closing.

dmbaturin:
No need to close, you can always force push to the branch. Use 'git push --force' to override existing commits.
@c-po is right that `config.exists` can be used to prevent thig, but I think I'm close to making cli-shell-api correctly fail, then an updated version of this patch can be a better way to handle empty config situation.

dmbaturin:
https://github.com/vyos/vyatta-cfg/commit/7f6eb1ed2c2a9dc76227ac355446b6d4b6c27733 After this, we can check if `showConfig` succeeded, rather than check if a path exists. It also exits with different code for invalid path and empty config now, which we can translate to Python exceptions.

thomas-mangin:
implementation of a better solution in #217 / T2027 taking advantage of T2026

python/vyos/config.py
thomas-mangin: T2027 is now handling this in a cleaner way

PR #217, opened 2020-02-10
T2027: ignore harmless showConfig error

thomas-mangin:
Building on T2026, this will ignore empty configuration sections
Removing the need to always use exists or exists_effective before
calling get_config_dict.

### DIFF
--- a/python/vyos/config.py
+++ b/python/vyos/config.py
@@ -133,6 +133,16 @@ class Config(object):
             raise TypeError("Path must be a whitespace-separated string or a list")
         return (self._level + path)
 
+    def _soft_fail(self, command, code, out):
+        # _soft_fail allows to intercept known program failures
+        # and handle them gracefully, by ignoring them and
+        # returning the inital data, or something else.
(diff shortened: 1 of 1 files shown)


thomas-mangin:
another way to fix the issue could be :
`res = self.show_config(self._make_path(path), default='', effective=effective)`
but it would swallow the error code 2.

dmbaturin:
I believe we should throw different exceptions for different error codes, e.g.

```
class EmptyConfigError(VyOSError):
  pass

class InvalidConfigPathError(VyOSError):
  pass
```

Then in `get_config_dict` we can return an empty dict if a config is empty, or let the exception bubble up to the caller.
If we make all new exceptions subclasses of existing VyOSError, then old code won't break.

thomas-mangin:
@dmbaturin  `self._run` does not return the error code 
```out = self._run(self._make_command('showConfig', path))```
so it would require to change the API to return the error code on top of the string, a much larger patch.

PR #221, opened 2020-02-14
T2037: use a systemd file for vyos path configuration

thomas-mangin:
Change the code to use a environment file in /usr/share/vyos to set the folder locations.


c-po:
@dmbaturin I build myself a custom ISO image with this change on top of vyos-1x and all smoketests `/usr/bin/vyos-smoketest` succeed. It looks cleaner to me so we probably should merge it. Maybe it also fixes https://phabricator.vyos.net/T685.

thomas-mangin:
This code now conflicts with other change. Please keep it open until I can rework it (and hopefully be better).

thomas-mangin:
lack of love for this idea it seems, sending the patch to the graveyard.

PR #222, opened 2020-02-14
tunnel: T2028: move interface tunnel to XML/Python

thomas-mangin:
This patch migrates the "interface tunnel" section to xml/python

### DIFF
--- a/interface-definitions/include/interface-mtu-64-8024.xml.i
+++ b/interface-definitions/include/interface-mtu-64-8024.xml.i
@@ -1,0 +1,13 @@
+<leafNode name="mtu">
+  <properties>
+    <help>Maximum Transmission Unit (MTU)</help>
+    <valueHelp>
+      <format>64-8024</format>
+      <description>Maximum Transmission Unit</description>
+    </valueHelp>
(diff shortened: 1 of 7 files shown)


c-po:
Thanks for this huge PR. Infact I do not like the part where you alter existing behavior in `vyos.ifconfig` and just place it in the very same commit. Usually such changes - even when they are required for future changes - should be placed in an individual commit. This makes tracking changes and the evolution of code much easier.

Can I ask you to place that modification in an individual commit?

thomas-mangin:
@c-po no issues at all. I perfectly agree that this should be moved out. I was just unsure if you would have wanted to see some of the other creation logic moved into "Interface". I have created https://github.com/vyos/vyos-1x/pull/223 as requested.

thomas-mangin:
for clarity, I am not asking for this PR to be merged at this point. I just wanted to check if there was an issue with the approach if the use of `ConfigurationState` as it is was fine. I am still to write the unit testing and related show commands.

c-po:
Please follow the commit rules where each commit must be prefixed with a topic and a task id.

PR #223, opened 2020-02-17
ifconfig: T2046: _delete/_create for Interface

thomas-mangin:
Not all interfaces use ip link add dev for creation.

For example, tunnel interfaces may use 'ip tunnel add' or 'ip -6 tunnel add'. The interface creation command could/should allow sub-class to modify the command used for creation and deletion.


thomas-mangin:
overtaken by T2057

PR #227, opened 2020-02-20
ifconfig: T2057: generalised Interface configuration

thomas-mangin:
Provides a consistent way to pass options to sub-classes of Interface. Using python **kwargs, parameters, options can be passed one by one or using a dict.

MACVLANIf("name0", option1=value1, option2=value2)

or 

options = { 'option1': value1, 'option2': value2, }
MAVLANIf("name0", **options)

The interface creation/deletion code was moved into a sub-function which was redefined by sub-classes, tidying the code.

### DIFF
--- a/python/vyos/ifconfig.py
+++ b/python/vyos/ifconfig.py
@@ -19,6 +19,7 @@ import jinja2
 import json
 import glob
 import time
+from copy import deepcopy
 
 import vyos.interfaces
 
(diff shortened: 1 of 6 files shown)


dmbaturin:
Are there places where this can cause incompatibilities?

thomas-mangin:
The ** is a python feature. 

You can not have something like Interface('name',option-with-minus=10) but you can have Interface('name',**{'option-with-minus:10}) is it what you are asking ?

c-po:
going to validate this the next couple hours

thomas-mangin:
@dmbaturin no it does not cause any incompatibilities.

c-po:
I took the latest commits here and build myself an ISO. It turned out to be not bootable as in the configuration was not properly applied leading to no network connectivity.

Please test by building yourself a dedicated ISO image.

thomas-mangin:
Thank you @c-po sorry for this. I will do this and look at why it works when updating the dpkg but not building the iso.

PR #230, opened 2020-02-26
ipoe-server: T2073: reset op-mode command for sessions

DaniilHarun:


op-mode-definitions/ipoe-server.xml
```diff
@@ -1,5 +1,41 @@
...
+        <properties>
+          <help>Clear ipoe-server sessions or process</help>
+        </properties>
+        <children>
+          <node name="session">
+            <properties>
+              <help>Clear ipoe-server session</help>
+            </properties>
+            <children>
+              <tagNode name="username">
```
hagbard-01: Is there a reason for using tagNode instead of leafnode?

op-mode-definitions/ipoe-server.xml
```diff
@@ -1,5 +1,41 @@
...
+                <properties>
+                  <help>Clear ipoe-server session by username</help>
+                  <completionHelp>
+                    <script>/usr/bin/accel-cmd -p 2002 show sessions username | sed -e 's/ \r//g' | tail -n +3</script>
```
hagbard-01: Are you sure that command clears the users sessions?

hagbard-01 (changes requested):
Hi DaniilHarun,

thank you for your contribution. Please review my comments below.

op-mode-definitions/ipoe-server.xml
DaniilHarun: Hello! no reason...

hagbard-01: Can you use then leafNode please?
Also, since it appears that you are using the IPoE service within VyOS, I would be very interested on feedback and user scenarios.

hagbard-01 (commented):
Thank you very much.

PR #232, opened 2020-02-29
dhcp-server: Allow configure bootstrap root path

elbandi:
You can set the root path for tftpboot.

http://www.networksorcery.com/enp/protocol/bootp/option017.htm


c-po:
Thanks for the contribution. Our process requires that a corresponding task in VyOS Phabricator (https://phabricator.vyos.net) is generated and the task number is referenced in the commit headline.

A detailed documentation can be found here: https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch

c-po:
No activity and conflicting files in this PR - closing.

PR #233, opened 2020-03-02
macvlan: T2057: fix creation of interface

thomas-mangin:
fix a typo in the create command for the macvlan interface introduced with T2057


thomas-mangin:
a mistake on my side. The code uses `intf`.

PR #234, opened 2020-03-03
dhcp-server: T2062: Fix static route bytes

elbandi:
Network bytes calculated wrong here:
https://github.com/vyos/vyos-1x/blob/current/src/conf_mode/dhcp_server.py#L596-L600

if the network_address (with /24 netmask) is 10.10.0.X or 10.0.10.X, the resulted string will be same: "10, 10,"

https://phabricator.vyos.net/T2062


dmbaturin (approved):
Looks like a sensible fix.

PR #241, opened 2020-03-05
ifconfig: T2101: Fix VXLAN config option parsing

tjjh89017:
Need to fix undefined reference, wrong variable from config['remote'] to self.config['remote']
Need to fix VXLAN parser to self.config, need fill self.options in VXLANIf because of

```
for k in self.options:
    if k in kargs:
        self.config[k] = kargs[k]
```


tjjh89017:
related to https://phabricator.vyos.net/T2101

PR #242, opened 2020-03-06
ifconfig: T2104: splt ifconfig.py into multiple files

thomas-mangin:
Split the ifconfig.py file into its components.


dmbaturin (approved):
Sensible, I think. I wonder if the amount of per-interface code can be further reduced through generalization though.

PR #245, opened 2020-03-15
Fix `NameError: name 'config' is not defined` on commit

petoem:
Fixes `commit failed` on vyos-1.3-rolling-202003150217 for vxlan interface.

### DIFF
--- a/python/vyos/ifconfig/vxlan.py
+++ b/python/vyos/ifconfig/vxlan.py
@@ -13,7 +13,7 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 
-
+from vyos import ConfigError
 from vyos.ifconfig.interface import Interface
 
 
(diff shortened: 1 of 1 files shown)


dmbaturin (changes requested):
Is there a compelling reason why copying the config hash and reformatting the values to include iproute2 keywords is better than just using different iproute2 command templates for the unicast and multicast cases?

I don't think that would be a bad kind of code duplication since it makes the code easier to read, and it's also faster than copying (even though for a small dict copy isn't expensive).

```
if unicast:
  cmd = 'ip link add ... remote {remote}'
else:
  cmd = 'ip link add ... group {group}'
```

c-po:
@thomas-mangin

thomas-mangin:
@dmbaturin your suggestion makes sense and is clearer: reading the code I had not understood it was what was happening.

petoem:
@dmbaturin I am not a python dev and I made this on a vyos installation I use. When dealing with an unknown code base, I usually try to keep the changes minimal and try not to change values that could be used elsewhere. That's why I copied the dict. 

Anyway, your suggestion does make sense and I'll try to fix it.

c-po:
@petoem thnks for beeing so kind and just sending a PR to use. I have merged it and adopted to our development guidelines @ https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch

PR #248, opened 2020-03-19
T2139: openvpn: allow unset dh-file when using EC keys

jjakob:


dmbaturin (changes requested):
Why use a special "none" value? That option becomes a syntactic noise in the config.
I think a better way to go about it is to remove the `file-exists` validator from the command definition, and allow the user to omit the `dh-file` option if (and only if) the private key header contains `EC PRIVATE KEY` rather than `RSA PRIVATE KEY`.
Appropriate commit message like "Diffie-Hellman prime file is unspecified, assuming ECDH" would be nice too.

jjakob:
I think it's possible to set "dh none" without ECDH keys too, but it's probably not wanted to do so. Your suggestion does make sense so I'll rework the code. This was the easiest to implement and also matched the syntax of openvpn's "dh" option, that's why I decided on this way.

Edit: I think omitting the validator isn't necessary as the special case is done by not setting dh-file at all. The logic is done in the conf_mode script itself.

jjakob:
Tested working with:
-RSA keys, dh-file: works as befire
-EC keys, no dh-file: prints "Diffie-Hellman prime file is unspecified, assuming ECDH", adds "dh none" to openvpn config
-EC keys, dh-file: prints "Warning: using dh-file and EC keys simultaneously will lead to DH ciphers being used instead of ECDH"

Due to a existing bug it requires PR #249 to work correctly.

c-po:
You now re-assign a value generated in `generate()` inside `verify()`: `openvpn['tls_dh'] = 'none'`.
All value assignments should go into `generate()`, please change your script syntax accordingly.

jjakob:
Where should it print the warnings and notices, in get_config or verify?

dmbaturin:
Warnings are generally printed at the verify stage.

jjakob:
It currently prints both the warning ant notice in the generate stage, should both be printed in verify?

jjakob:
Both notice and warning moved to verify, modifying variable done in generate.
Tested all 3 cases as above, works

c-po:
Sorry, I made a mistake. It should have been:

All value assignments should go into `get_config()`, please change your script syntax accordingly.

can you please adjust once more?

jjakob:
Moved value assignment from generate() to get_config()

jjakob:
Moved value assignments from generate() to get_config(), tested working same as above.

PR #254, opened 2020-03-22
WIP: Make config scripts more testable

rhn:
This series of patches comes from me not having a sacrificial VyOS setup for testing DHCP.

Seeing how configuring DHCP in VyOS is a matter of turning a text file into some text files, I was rather surprised that the scripts can't run in a standalone fashion. After applying this, I managed to actually achieve this.

Applying this wider opens up the possibility to do some integration testing for individual config scripts, e.g. in a CI system.

To test, use the commit https://github.com/rhn/vyos-1x/commit/f3661b4ef97cecf7e9c114a9ac2a8bcc773a689a and follow the instructions in the commit message.

Completely untested on anything live.

### DIFF
--- a/python/vyos/config.py
+++ b/python/vyos/config.py
@@ -88,7 +88,7 @@ class Config(object):
     the only state it keeps is relative *config path* for convenient access to config
     subtrees.
     """
-    def __init__(self, session_env=None):
+    def __init__(self, session_env=None, config_file='/opt/vyatta/etc/config/config.boot'):
         self._cli_shell_api = "/bin/cli-shell-api"
         self._level = []
         if session_env:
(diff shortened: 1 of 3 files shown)


c-po:
Thanks for this contribution. Please follow the VyOS dvelopment guidelines described here: https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch

rhn:
Updated for the guidelines with https://phabricator.vyos.net/T3754 .

Commit for testing is 77f1f095f4fbb9164751c6678e213bab2ab17cac , which is not based on those because I had to change commit message of parents before submitting.

It would be nice if someone could reopen.

PR #263, opened 2020-03-24
T2133: add ipv6 disable to grub menuentry command

elbandi:
IPV6 is build directly into the kernel. So, it's not loaded by modprobe, so the modprobe configuration is not involved.

Builtin module parameters can be configured through kernel command line.

### DIFF
--- a/src/conf_mode/system-ipv6.py
+++ b/src/conf_mode/system-ipv6.py
@@ -87,24 +87,25 @@ def apply(ip_opt):
         print('Changing IPv6 disable parameter will only take affect\n' \
               'when the system is rebooted.')
 
-    # configure multipath
-    sysctl('net.ipv6.fib_multipath_hash_policy', ip_opt['mp_layer4_hashing'])
+    if os.path.exists("/proc/sys/net/ipv6"):
+        # configure multipath
+        sysctl('net.ipv6.fib_multipath_hash_policy', ip_opt['mp_layer4_hashing'])
 
-    # apply neighbor table threshold values
-    sysctl('net.ipv6.neigh.default.gc_thresh3', ip_opt['neighbor_cache'])
-    sysctl('net.ipv6.neigh.default.gc_thresh2', ip_opt['neighbor_cache'] // 2)
-    sysctl('net.ipv6.neigh.default.gc_thresh1', ip_opt['neighbor_cache'] // 8)
+        # apply neighbor table threshold values
+        sysctl('net.ipv6.neigh.default.gc_thresh3', ip_opt['neighbor_cache'])
+        sysctl('net.ipv6.neigh.default.gc_thresh2', ip_opt['neighbor_cache'] // 2)
+        sysctl('net.ipv6.neigh.default.gc_thresh1', ip_opt['neighbor_cache'] // 8)
 
-    # enable/disable IPv6 forwarding
-    with open('/proc/sys/net/ipv6/conf/all/forwarding', 'w') as f:
-        f.write(ip_opt['ipv6_forward'])
+        # enable/disable IPv6 forwarding
+        with open('/proc/sys/net/ipv6/conf/all/forwarding', 'w') as f:
+            f.write(ip_opt['ipv6_forward'])
 
-    # configure IPv6 strict-dad
-    for root, dirs, files in os.walk('/proc/sys/net/ipv6/conf'):
-        for name in files:
-            if name == "accept_dad":
-                with open(os.path.join(root, name), 'w') as f:
-                    f.write(str(ip_opt['strict_dad']))
+        # configure IPv6 strict-dad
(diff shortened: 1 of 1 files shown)


src/conf_mode/system-ipv6.py
```diff
@@ -69,12 +70,19 @@ def generate(ip_opt):
 def apply(ip_opt):
     # disable IPv6 address assignment
     if ip_opt['disable_addr_assignment']:
+        # this file is useless, but it's an indicator for grub setup scripts
```
elbandi, line 73: For vyatta-grub-setup script, if user set this before install image
and for install-image-existing, if user add a new image

src/conf_mode/system-ipv6.py
```diff
@@ -69,12 +70,19 @@ def generate(ip_opt):
...
         with open(ipv6_disable_file, 'w') as f:
             f.write('options ipv6 disable_ipv6=1')
+
+        if os.path.exists(grub_cfg_file):
+            os.system("sed -i 's/noautologin \+net.ifnames/noautologin ipv6.disable=1 net.ifnames/' "+grub_cfg_file)
```
jjakob: I think this should be more robust to changes in the cmdline, if the cmdline parameters are changed in the future, this will stop working. Perhaps append it to the end of the line and search for some other string that's more permanent.

jjakob: For example 
```
sed -i '/[[:space:]]*linux / s/$/ ipv6.disable=1/'
```
and replace the other places with matching code

Edit: this doesn't handle the ipv6.disable parameter already existing, I'd extend the code to first 's/ipv6.disable=?/ipv6_disable=1/' and if that fails, do the sed append as above.

jjakob:
```
if ! sed -i 's/ipv6.disable=?/ipv6.disable=1/' $file; then
    sed -i '/[[:space:]]*linux / s/$/ ipv6.disable=1/' $file
fi
```

src/conf_mode/system-ipv6.py
```diff
@@ -69,34 +70,42 @@ def generate(ip_opt):
 def apply(ip_opt):
     # disable IPv6 address assignment
     if ip_opt['disable_addr_assignment']:
+        # this file is useless, but it's an indicator for grub setup scripts
         with open(ipv6_disable_file, 'w') as f:
             f.write('options ipv6 disable_ipv6=1')
+
+        if os.path.exists(grub_cfg_file):
+            os.system("sed -i 's/ipv6.disable=[^[:space:]]*/ipv6.disable=1/;/^[[:space:]]*linux / {/ipv6.disable=/! s/$/ ipv6.disable=1/}' "+grub_cfg_file)
```
c-po, line 78: I do not know if it is the right way to offload this to `os.system()` or rather parse it via means of Python. @dmbaturin ?

c-po: I like the patch in general and the idea - this is a minor detail we can improve later on.

c-po (approved):

c-po:
When looking at the code we utilize for GRUB vyatta-op: /opt/vyatta/bin/vyatta-boot-image.pl I get a strong feeling the IPv6 disable feature should be moved over there.

What happens if you disable IPv6 with your commit, save the config and then add a new system image by `add system image`? Is the change of disabling IPv6 transferred or not?

src/conf_mode/system-ipv6.py
elbandi: i'm not a python coder. i found this simple. if there is a robust/community standard solution for replace sed in python, good for me.

elbandi:
@c-po yes, add system image sets ipv6 too. Moveover, if someone disable the ipv6 when running from live iso, and install vyos to disk, ipv6 will be disabled too. check this pr: https://github.com/vyos/vyatta-cfg-system/pull/121

c-po (approved):

c-po:
Actually why do you wan't to disbale IPv6 on the system? I think this is a huge workpackage.

What should happen to all the ipv6 nodes in the system? Should they "vanish" or should they stay as they are?

One example would be disabling IPv6 but enabling either DHCPv6 or configuring an IPv6 address on an interface - thus will result in commit errors as addresses would likely be rejected by the system.

elbandi:
Our switches doesnt support ipv6, and we had issues when other system has enabled (some apps trying to use AAAA addresses, etc). so we disable it everywhere

c-po:
@Elbandi maybe you can adjust this PR to the requirements in https://phabricator.vyos.net/T2447 - thus you can set it if you need to and the VyOS base stays like it is?

c-po:
If you want to disable IPv6 you can now also use the `set interfaces ethernet eth1 ipv6 address no-default-link-local` which will remove even the link-local address from the interface.

PR #265, opened 2020-03-24
ifconfig: T2057: break down DHCP, add register, STP and VLAN as adapters

thomas-mangin:


c-po:
Why do we need a PPPoE interface here? PPPoE is solely created by PPP and its corresponding PPP configuration only

thomas-mangin:
Completeness and consistency. The interface data is used to generate the list of possible interface section names for example. See register.py line 79

PR #267, opened 2020-03-24
openvpn: T2146: remove old files

jjakob:


dmbaturin (approved):

dmbaturin (approved):

src/conf_mode/interfaces-openvpn.py
```diff
@@ -899,6 +899,13 @@ def generate(openvpn):
     interface = openvpn['intf']
     directory = os.path.dirname(get_config_name(interface))
 
+    # we can't know which clients were deleted, remove all client configs
+    if os.path.isdir(directory + '/ccd/' + openvpn['intf']):
```
thomas-mangin: This works but a better way would be to `os.path.join(directory,'ccd',openvpn['intf'])`

c-po:
Looks good to me, too - is it still a WIP?

src/conf_mode/interfaces-openvpn.py
jjakob: Fixed

jjakob:
Ready to merge.

jjakob:
@Elbandi Indeed, I didn't think of that at the time. It would require more code to list all the files in the directory, find which ones aren't configured clients, and delete just those. The script recreates all the ccd files from scratch soon after so it's not a problem just to delete them all. The server won't read them until being restarted anyway.

PR #270, opened 2020-03-25
ifconfig: T2057: fix finding section name from interface

thomas-mangin:


thomas-mangin:
This is in conjunction with https://github.com/vyos/vyos-smoketest/pull/8 as both are related

PR #271, opened 2020-03-25
vrf: T31: prevent vrf with interface namesake

thomas-mangin:


c-po:
This is definately the wrong place for this check. As stated on slack VRF name validation must happen in the CLI frontend as this is the first point where we can drop rubbish. A proper

```
 <constraint>
   <regex> awesomeexpression for bond interfaces</regex>
   <regex> awesomeexpression for ethernet interfaces</regex>
   <regex> awesomeexpression for pppoe interfaces</regex>
</constraint>
```

thomas-mangin:
will could/should to be a dynamic script so new interface name are automatically added. but agree need redoing.

c-po:
new interface names are handled by the regex - we rarely add ned interface types.

PR #272, opened 2020-03-26
ifconfig: T2158: never fail when setting flowcontrol

thomas-mangin:


c-po:
This is a good example where you should describe the why in the commit message so we know the reason for this change in 6 months time when we ask us again why some lines have been added to the codebase, please see https://docs.vyos.io/en/latest/contributing/development.html#writing-good-commit-messages

PR #275, opened 2020-03-27
ifconfig: T2057: explicity name state functions

thomas-mangin:
The Interface get_state/set_state were not clear about
if they edited the admin or operational state.

functions are now using admin_state and oper_state
for clarity.

### DIFF
--- a/python/vyos/ifconfig/bond.py
+++ b/python/vyos/ifconfig/bond.py
@@ -101,7 +101,7 @@ class BondIf(Interface):
         for s in self.get_slaves():
             slave = {
                 'ifname': s,
-                'state': Interface(s).get_state()
+                'state': Interface(s).get_admin_state()
             }
             slave_list.append(slave)
 
@@ -112,7 +112,7 @@ class BondIf(Interface):
         # physical interface
         for slave in slave_list:
              i = Interface(slave['ifname'])
-             i.set_state(slave['state'])
+             i.set_admin_state(slave['state'])
 
     def set_hash_policy(self, mode):
         """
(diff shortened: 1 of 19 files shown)


dmbaturin (approved):
Sensible,  I think.

c-po:
Reading operstate is definately required when the `show interfaces` commands get migrated. But it makes no sense to "write" into the `operstate` as it can not be written at all (it would make no sense, too).


The values according to the Kernel ABI are:

c-po (changes requested):
Reading operstate is definately required when the `show interfaces` commands get migrated. But it makes no sense to "write" into the `operstate` as it can not be written at all (it would make no sense, too).


The values according to the Kernel ABI are:
```
What:           /sys/class/net/<iface>/operstate
Date:           March 2006
KernelVersion:  2.6.17
Contact:        netdev@vger.kernel.org
Description:
                Indicates the interface RFC2863 operational state as a string.
                Possible values are:
                "unknown", "notpresent", "down", "lowerlayerdown", "testing",
                "dormant", "up".
```

Also the sysfs node is clearly marked as RO in the Kernel Source code `static DEVICE_ATTR_RO(operstate);` (see commit 6be8aeef348a7a4a1 ("net: core: convert class code to use dev_groups")), thus `_sysfs_set['oper_state']` must be droped from the implementation.

thomas-mangin:
LOL - ok make sense - implementing get/set without thinking enough .. removing the useless code.

thomas-mangin:
set_oper_state removed.

PR #276, opened 2020-03-27
ifconfig: T2057: add get_alias function

thomas-mangin:
add get_alias to the Interface API as it is required to convert vyatta-show-interfaces.pl to python


thomas-mangin:
forced pushed to improve the docstring

PR #283, opened 2020-03-29
ifconfig: T2057: Move code from interface.py to wireguard.py 

thomas-mangin:


thomas-mangin:
This was the initial patch to remove the circular import with the move.

PR #295, opened 2020-04-03
T2205: Don't change offload sg if driver is 'hv_netvsc'

suuei:


c-po (changes requested):
Why is a modechange to 755 required?

suuei:
Sorry, the permissions have been changed due to an operational error.
Is this the right way to fix it?

jjakob:
I think the right way would be to not fail on setting these and silently continue, like https://github.com/vyos/vyos-1x/pull/272 . Not just scatter-gather but all properties set in this file. In addition I'd add a check in config mode, when a setting wants to be set on a interface that doesn't support it, print a warning to the console. (Exit status of ethtool --get-... will probably be 76, as in code from above PR, the validate function should check all of these and print warnings).

You can modify the PR by doing a rebase (`git rebase -i HEAD~`), picking the commit you want to edit, chmod the file back to the old mode,  make any changes you want to the commit, then `git commit --amend -a`, `git rebase --continue` and `git push --force`.

c-po:
Please also check our development guidelines at https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch as there is no task id assigned to the commit.

suuei:
Thank you for your advice.

I tried to check an interface that supports offload sg using `ethtool -k eth0` .
However, regardless of whether the interface supports changing offload sg settings, the result contained the following lines:
```
scatter-gather: on
	tx-scatter-gather: on
	tx-scatter-gather-fraglist: off [fixed]
```
Do you have any good ideas?

jjakob:
Discussion continued at T2205

thomas-mangin:
Yes, the behaviour was changed/reverted with
https://github.com/vyos/vyos-1x/commit/dcbd5f430907747f8c29111fbe9f5737865a7ae8

However explicit messages for known drivers are good for the users.

PR #298, opened 2020-04-04
openvpn: T2222: set 'multihome' option when local-host is unset

jjakob:
If configured with protocol udp and no local-host, openvpn needs
the 'multihome' option to correctly work on all addresses.


c-po:
Damn! You have picked task 2222 😸 👍

PR #299, opened 2020-04-04
ifconfig: T2205: silence ethtool harmless failures

thomas-mangin:
Not all interfaces are capable of all features. Since commands are
now checked for valid completion, ethtool command failure must
be ignored.


c-po:
Damn! I like that possible feature flag 👍

PR #300, opened 2020-04-04
show_openvpn: T2184: skip unactive interface

thomas-mangin:
Should the openvpn status file not be present, present a blank
interface.


thomas-mangin:
This task was assigned to me so I am providing this patch, however I have never used OpenVPN in my life, so it is not tested as indicated on the phabricator entry.

PR #322, opened 2020-04-10
logging: T2245: Added basic common logging module

zdc:
This allows to quickly add everything that needs for logging into any Python script with one line:
```
from vyos.logger import logger
logger.level("Message")
```
The `level` should be one from: `exception`, `critical`, `error`, `warning`, `info`, `debug`.

Also, some basic configurations available:

- logs can be configured to be displayed in stderr (or any other stream) or saved to file;
- the logging level can be changed globally or for a single handler.

### DIFF
--- a/python/vyos/logger.py
+++ b/python/vyos/logger.py
@@ -1,0 +1,63 @@
+# Copyright 2020 VyOS maintainers and contributors <maintainers@vyos.io>
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


python/vyos/logger.py
```diff
@@ -0,0 +1,63 @@
...
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+# import logging and related modules
+import logging
```
c-po, line 17: Why import the complete class and one line lter only specific components?

c-po (changes requested):

thomas-mangin:
Hi @zdc, I do not understand why we need  class for this. could you please share use case ?

python/vyos/logger.py
zdc: Not all standard logging handlers are inside the `logging` module, it includes only `NullHandler`, `StreamHandler` and `FileHandler`. `SysLogHandler` and `RotatingFileHandler` need to be imported from the `logging.handlers` module additionally.

thomas-mangin:
@zdc I have like the feeling that this PR can be closed 😉

PR #326, opened 2020-04-10
dhcp: T2265: refactor DHCP class

thomas-mangin:
Breaks the code down to make it easier to understand, simplify API


python/vyos/ifconfig/dhcp.py
```diff
@@ -16,103 +16,55 @@
...
-        """
-        return self._dhcp[6]['options']
-
-    def set_dhcpv6_options(self, options):
-        """
-        Store new DHCP options used by next run of DHCP client.
-        """
-        self._dhcp[6]['options'] = options
+class _DHCPv4 (_DHCP):
+    template = """\
```
c-po: Please migrate the template to data/templates folder

c-po (changes requested):
Please migrate the templates to data/templates folder

python/vyos/ifconfig/dhcp.py
thomas-mangin: Hi Christian, I agree it should be done but could you please merge as the current code does not load them from there neither so it is not a step backward.

c-po: Can‘t as there are conflicting files.

thomas-mangin: argh ! ok - will re,re,...base :-) and add the template at the same time.

thomas-mangin:
migration to templated added to the patchset.

PR #332, opened 2020-04-11
template: T2230: use render to generate templates

thomas-mangin:
convert all call to jinja to use template.render


c-po:
This branch has conflicts that must be resolved

PR #334, opened 2020-04-12
ifconfig: T31: add skeleton VTI and input classes

thomas-mangin:
also, it adds a function to Section which provides a list of reserved names
useful to figure out the reserved name of interfaces which can not be used.


c-po (commented):
Why is vti prefixed with pppoe?

thomas-mangin:
I do not know which others of more are required and it allows to enumerate all the sections of the interface.. set interface <section>

thomas-mangin:
Let me check vti.

PR #336, opened 2020-04-12
Update Ethernet hw-id to be clear

tjharman:
hw-id currently has the same help text as the "mac" command.  Update the text to make it clear how the hw-id mac address is different

### DIFF
--- a/interface-definitions/interfaces-ethernet.xml.in
+++ b/interface-definitions/interfaces-ethernet.xml.in
@@ -58,10 +58,10 @@
           </leafNode>
           <leafNode name="hw-id">
             <properties>
-              <help>Media Access Control (MAC) address</help>
+              <help>Map to Media Access Control (MAC) address</help>
               <valueHelp>
                 <format>h:h:h:h:h:h</format>
-                <description>Hardware (MAC) address</description>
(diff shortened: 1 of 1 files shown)


c-po:
Hi @tjharman, thanks for the contribution.

The `hw-id` is actual a parameter you can not set and will be ignored if done so. There are some plans for removing this in the future if I recap this right. Currently it only holds the physical MAC address burned into the card.

To change the intrface MAC address use the `mac` node instead.

In addition there is a guideline on submitting and merging patches, please see https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch.

The help text could indeed be updated to what it really is - will do so.

Thanks,
Christian

tjharman:
Thanks c-po, appreciate the kind feedback.
My only comment is that when I set hw-id recently, I ended up with a system that couldn't find eth0 (the device I'd applied hw-id to), it was in a down/down state, and I ended up with an extra "eth2" interface (usually I'd have eth0 and eth1)

A number of reboots didn't fix it, it wasn't until I realised I wanted "mac" and not "hw-id" and swapped it, then on reboot everything came up as expected.

Anyway, that's a seperate issue, I'll investigate and log something on Phabricator.

PR #337, opened 2020-04-12
cmd: T2226: add the a full log of all commands

thomas-mangin:
when /tmp/vyos.developer.debug is enabled, send all command run (from boot onwards) to /tmp/full-log allowing to diagnose boot level issues.

### DIFF
--- a/python/vyos/util.py
+++ b/python/vyos/util.py
@@ -35,7 +35,10 @@ def debug(flag):
     # this is to force all new flags to be registered here to be documented:
     if flag not in ['developer', 'ifconfig']:
         return ''
-    return flag if os.path.isfile(f'/tmp/vyos.{flag}.debug') else ''
+    for folder in ('/tmp', '/config'):
+        if os.path.isfile(f'{folder}/vyos.{flag}.debug'):
+            return flag
+    return ''
 
 
 def debug_msg(message, flag=''):
@@ -45,6 +48,20 @@ def debug_msg(message, flag=''):
 
     if debug(flag):
         print(f'DEBUG/{flag:<6} {message}')
+
+    if not debug('developer'):
+        return
+
+    logfile = '/tmp/full-log'
+    existed = os.path.exists(logfile)
+
+    with open(logfile, 'a') as f:
+        f.write(f'DEBUG/{flag:<6} {message}\n')
+        if not existed:
+            # at boot the file is created as root:vyattacfg
+            # at runtime the file is created as user:vyattacfg
+            # do not use run/cmd to not have a recursive call to this code
+            os.system(f'chmod g+w {logfile}')
 
 
 # There is many (too many) ways to run command with python


c-po:
How to enable this for system bootup as I can not touch this file during boot?

thomas-mangin:
@cpo you are fast .. Yes, the patch was wrong and is fixed.

python/vyos/util.py
```diff
@@ -46,6 +49,20 @@ def debug_msg(message, flag=''):
...
+    logfile = '/tmp/full-log'
+    existed = os.path.exists(logfile)
+
+    with open(logfile, 'a') as f:
+        f.write(f'DEBUG/{flag:<6} {message}\n')
+        if not existed:
+            # at boot the file is created as root:vyattacfg
+            # at runtime the file is created as user:vyattacfg
+            # do not use run/cmd to not have a recursive call to this code
+            os.system(f'sudo chmod g+w {logfile}')
```
c-po: Why you use always sudo? It's not required

c-po (changes requested):

python/vyos/util.py
```diff
@@ -35,7 +35,10 @@ def debug(flag):
     # this is to force all new flags to be registered here to be documented:
     if flag not in ['developer', 'ifconfig']:
         return ''
-    return flag if os.path.isfile(f'/tmp/vyos.{flag}.debug') else ''
+    for folder in ('/tmp', '/config'):
```
c-po, line 38: Is there any documentation in readthedocs or the source about the new location?

c-po (changes requested):

python/vyos/util.py
thomas-mangin: How can there be documentation for a feature which does not exist yet? 😛 

Also, the only reason I know about this feature is as I read the code and "stubbled" upon it.

I was not aware it was supposed to be documented as searching for it does not return anything:

```
https://docs.vyos.io/en/latest/search.html?q=vyos.debug.ifconfig
<no result>
```

on the latest master of vyos-documentation:
```
# grep "vyos.ifconfig.debug" . -r
#
```

Is it supposed to be documented ?

python/vyos/util.py
thomas-mangin: ok will remove.

python/vyos/util.py
c-po: The old approach with the /tmp file was documented in the source, so should be the new one. As it is now persistend it should also be mentioned here: https://docs.vyos.io/en/latest/contributing/development.html#debugging but his can happen after this is merged.

thomas-mangin: Happy to update the documentation, but as I failed to notice it was public, should I also add a "command" debugging as we are at it.

c-po: Everything which is not yet documented should be added into the debugging section for all

PR #347, opened 2020-04-15
ipoe: T2294: Fix templates and migrate to systemd

DmitriyEshenko:


c-po:
That should habe been actually two commits as two independant changes. Could you split that please?

DmitriyEshenko:
Should I create subtask and cancel this PR?
Note: As for me, this commits related, because ipoe was broken

c-po:
Nope, just split the vommit and keep the Taskid thats fine. Those might be related in a task way but not in the git way of doing coherent commits which only fix one thing at a time. that would be the template rendering and on the other habd systemd - two non related changes which deal witbthe greater good for T2294

PR #360, opened 2020-04-20
airbag: T2186: generic syslog and better text

thomas-mangin:
This introduce a logger module to provide consistent logging through VyOS and convert airbag to it.

It also improves the information reported by airbag (including the content of show version), and remove the dependency on pystache for the rendering of show version using python3 format feature instead.

It also improves the util function read_file (to return a default string/object on failure instead of raising) and add a read_json helper which reads and parse json files.

### DIFF
--- a/python/vyos/airbag.py
+++ b/python/vyos/airbag.py
@@ -13,15 +13,14 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 
-import os
 import sys
 from datetime import datetime
 
 from vyos import debug
 from vyos.config import Config
-from vyos.version import get_version
-from vyos.util import run
 from vyos.logger import syslog
+from vyos.version import get_version
+from vyos.version import get_full_version_data
 
 # we allow to disable the extra logging
 DISABLE = False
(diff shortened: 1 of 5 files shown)


python/vyos/airbag.py
```diff
@@ -124,29 +118,45 @@ def InterceptingException(excepthook,_singleton=[False]):
...
     InterceptingException(intercepter)
 
 
 # Messages to print
+# if the key before the value has not time, syslog takes that as the source of the message
 
 FAULT = """\
-Date:       {date}
-VyOS image: {version}
+Report Time:      {date}
```
c-po, line 129: Why duplicate this code and not use the `show system version` output? That would give us single source

c-po (changes requested):

python/vyos/airbag.py
thomas-mangin: I have refactored the function which gets the dict but the output is slightly different.
If there is no space in the text, the the first word is taken by syslog to be the service name:

This is why the code has "Report Time" (and not Date) anymore and "Image Version" (and not Version).

```
Apr 20 12:07:38 vyos Date[25010]:       2020-04-20 12:07:38
Apr 20 12:07:38 vyos python3[25010]: VyOS image: 1.3-rolling-202004180117
Apr 20 12:07:38 vyos python3[25010]: Traceback (most recent call last):
Apr 20 12:07:38 vyos python3[25010]:   File "/usr/libexec/vyos/conf_mode/interfaces-dummy.py", line 138, in <module>
Apr 20 12:07:38 vyos python3[25010]:     apply(c)
Apr 20 12:07:38 vyos python3[25010]:   File "/usr/libexec/vyos/conf_mode/interfaces-dummy.py", line 103, in apply
Apr 20 12:07:38 vyos python3[25010]:     raise RuntimeError('test')
Apr 20 12:07:38 vyos RuntimeError[25010]: test
```

PR #361, opened 2020-04-20
openvpn: T2339: fix for IPv4 local-host addresses

jjakob:
Commit bb9f998 introduced a bug where openvpn fails to start if
'local-host' is an IPv4 address due to 'proto' wanting a IPv6 socket.
This adds a conditional check and uses normal proto if it's IPv4.

### DIFF
--- a/data/templates/openvpn/server.conf.tmpl
+++ b/data/templates/openvpn/server.conf.tmpl
@@ -18,7 +18,7 @@ dev {{ intf }}
 persist-key
 iproute /usr/libexec/vyos/system/unpriv-ip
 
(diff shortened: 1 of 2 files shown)


mrozentsvayg:
this PR is not breaking remote IPv4, but it's only fixing configurations with local-host allowed (not client mode)
proposed fix for client mode:
https://phabricator.vyos.net/T2550#66134

mrozentsvayg:
> Unfortunately this breaks OpenVPN with remote IPv4 hostnames.

this PR is not breaking remote IPv4, but it's only fixing configurations with local-host allowed (not client mode)
proposed fix for client mode: https://github.com/vyos/vyos-1x/pull/443 (https://phabricator.vyos.net/T2550#66134)

PR #367, opened 2020-04-21
l2tpv3: T2352: check that the local-ip exists

thomas-mangin:


dmbaturin (approved):

c-po:
Why not make use of an already existing function? `vyos.validate.is_addr_assigned()` ?

c-po (changes requested):
Why not make use of an already existing function? `vyos.validate.is_addr_assigned()` ? The wheel should not be invented multiple times.

thomas-mangin:
Agree this is duplication. I thought that validation was performing data validation (regex / checks) this is a query of the operation of the router, hence why I placed it under operational but I think you are right. opening a task to discuss.

c-po:
No Task needed - please use it. The file has been called validate b/c it is called by most verify() functions which is true here, too.

thomas-mangin:
kept the function as a separate commit. using the validate function in the code.

c-po:
This means we have unused code now? Not a fan of it.

thomas-mangin:
Just merge half the patch.

src/conf_mode/interfaces-l2tpv3.py
```diff
@@ -170,6 +175,9 @@ def verify(l2tpv3):
     if not l2tpv3['local_address']:
         raise ConfigError(f'Must configure the l2tpv3 local-ip for {interface}')
 
+    if not l2tpv3['assigned']:
```
c-po: Why not call `is_addr_assigned(l2tpv3['local_address'])` here directly?

src/conf_mode/interfaces-l2tpv3.py
```diff
@@ -198,7 +206,7 @@ def apply(l2tpv3):
     conf = deepcopy(L2TPv3If.get_config())
 
     # Check if L2TPv3 interface already exists
-    if l2tpv3['intf'] in interfaces():
+    if l2tpv3['intf'] in l2tpv3['interfaces']:
```
c-po: Why use a cached version of `interfaces()``

c-po (changes requested):

PR #371, opened 2020-04-21
interfaces: T2362: IPv6 link-local and EUI64 address fixes

jjakob:


jjakob:
This is the initial fix to make IPv6 work at all, later a fix for no-default-link-local is needed (set sysctl addr_gen_mode=1 to tell the kernel to not generate a link-local, otherwise the intf's MAC changes after a member is added so we can't delete the kernel's link-local any more), and adding the code to not set/remove link-locals when interfaces are added to a bridge

c-po:
@jjakob I really like this contribution - can you please split this into two individual commits? (use `git reset HEAD~1` && `git cehckout -- .`) to move all files back into the index and then back into your working copy? You then can perform `git add` again for split of the commits, or `git add --patch` to select individual hunks to add the the Git index.

By splitting the change It becomes more maintainable in the future (cherry-pick/bisect). Also it looks like it could fix one of the OSPFv3 problems I had in the past where a link-local address was missing, too.

Why can't we always ensure an interface has a link-local address? Is there any reason to not always set this?

jjakob:
Which two commits do you mean? One to split the functions and add the default link local, the other one to add no-default-link-local and multi on eui64? I can add the other two things still left to do while I'm at it.

Link-local addresses are like normal IPv6 addresses, just that they can only be used on the L2 segment and can't be routed. Several protocols require a link-local address to work, including RA, NDP, SLAAC, DHCPv6, and OSPF (so yeah, your problem was definitely due to this).
Services configured to listen on all addresses will listen on link-locals too, even on interfaces you might not have configured any address on at all. Obviously this is a security risk, since the user doesn't know the interface really has an address, and doesn't think it needs a firewall or anything.

Now that I think about it, this should probably be changed to not add a link-local by default, and only add one it's configured to do so. But if a user wanted to use IPv6 on an interface, he'd have to do it by hand, which would need documenting and would need a migrator (to turn on link-locals in the config on all interfaces for old users) else we'd have to have some sort of conditional logic - add a link-local if either: the interface has a IPv6 address, is configured to do RA, other cases I might not be thinking of right now,...

So for now I'd keep the logic of adding one by default and having an option to disable it, as the opposite would require more work, but would be better from a security standpoint in the long run.

As regards the bridge members - adding a link-local to a bridge slave will make the kernel send out packets on the interface, but all packets conimg back will go up to the bridge. I don't think that's wanted behavior.

c-po:
Actually by reading it again I would split it by 4:

- break apart set_ipv6_eui64_address into add_ and del_ functions
- add the ability to delete the default link-local address via
'ipv6 address no-default-link-local'
- set a default 'fe80::/64' link-local address on all relevant interfaces
- expand the eui64 config node and code to support more than 1 prefix

jjakob:
Right. I'm working on it, and adding the two things left over from before.

jjakob:
This is ready to be merged. I tested it on 2 routers - one with bonds, bond vifs and eth's, and one with eths, eth vifs, bridges and vtuns. No errors on any of them, interfaces have link-locals if their interface conf mode scripts add it.

Currently yes: bonding, bridge, ethernet, l2tpv3, openvpn in TAP mode, pseudo-ethernet, vxlan, wireless, vlan
no: dummy, geneve, loopback, pppoe, tunnel, wireguard, wirelessmodem

If a interface works only on L3, there's no point in adding a link-local to it.

Auto-generation of link-local addresses in the kernel is disabled via sysctl.

I had to add the EUI64 address code to the vlan scripts, as they currently had none, even though their interface definitions included it (so you could set a eui64 on a vlan, but it was a no-op).

The ordering for adding and deleting the EUI64 addresses was set so that the interface has the new mac address before the addresses are added, so bridge members are added before that. Removing of EUI64 addresses is done before the MAC is changed and there is code to detect if the interface MAC is going to chenge, in which case all the addresses are removed and re-added to update their EUI64.

The only thing left to do is deleting and re-adding addresses when interfaces are added and taken out of a bridge, since this is a larger change that isn't really connected with this issue, I won't include it here.

c-po (approved):

jjakob:
Added a bugfix on the 'delete and re-add all EUI64 addresses if MAC has changed' commit.

jjakob:
There is one more bug in the mac-checking code that I need to fix before this gets merged.

jjakob:
The bugs are fixed so it can be merged.

I'm not sure how to check the existence of a VLAN interface or get its MAC (it's not in the vyos.ifconfig class so get_mac method can't be used on it) so it just always deletes and re-adds all EUI64 addresses from it if a custom mac is set. Should be fine for now.

PR #376, opened 2020-04-24
log: T2377: do not modify log file if not 666

thomas-mangin:


c-po:
For future commits please use the commit body to explain why this change is needed. That's easier then digging up Phabricator :)

PR #379, opened 2020-04-26
util: T2226: multiple improvements

thomas-mangin:
patch 1:

    Do not agreggate stderr with stdout. So if a command reports
    a message on stderr but does not report an error, it will
    not be send to the user to confuse him.

    Explicitely set encoding to utf-8, which does not change
    the code behaviour but simplify the code.

patch 2:

    add an option to cmd() allowing to define a list of error codes
    which are expected when runnin the command. the default is only
    to expect zero (no error).

patch 3:

    debug.noteworthy can be used to record noteworhy event during
    the lifetime of the program. Should anything then cause the
    program to fail and cause an airbag report to the user,
    then this information will also be included.

### DIFF
--- a/python/vyos/airbag.py
+++ b/python/vyos/airbag.py
@@ -26,6 +26,17 @@ from vyos.version import get_full_version_data
 DISABLE = False
 
 
+_noteworthy = []
+
+def noteworthy(msg):
+    """
+    noteworthy can be use to take note things which we may not want to
+    report to the user may but be worth including in bug report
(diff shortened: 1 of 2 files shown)


thomas-mangin:
sorry, not tested with make iso / make test .. will do and update when it was done.

thomas-mangin:
no issues with `make iso; make test`

c-po:
Why not add those nice explanation to the commit itself? Then its archived forever.

PR #383, opened 2020-04-28
disable :T2372: disable sub-interface if parent is

thomas-mangin:
The patchset rewrites the vlan_to_dict code so that if an interface is disabled, all the VLAN of that interface are also disabled (admin down).

The code creates a new helper function to look for the disable option in parent interfaces.

It also refactor much of the dictionary generation for interfaces and VLANs (as they are indicatical,  and create a helper function to parse vif, vif-s, vif-c.

### DIFF
--- a/python/vyos/configdict.py
+++ b/python/vyos/configdict.py
@@ -18,7 +18,12 @@ A library for retrieving value dicts from VyOS configs in a declarative fashion.
 
 """
 
+from enum import Enum
+from copy import deepcopy
+
 from vyos import ConfigError
+from vyos.ifconfig import Interface
+
 
 def retrieve_config(path_hash, base_path, config):
     """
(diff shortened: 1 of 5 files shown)


python/vyos/configdict.py
```diff
@@ -98,192 +103,339 @@ def get_ethertype(ethertype_val):
         raise ConfigError('invalid ethertype "{}"'.format(ethertype_val))
 
 
-def vlan_to_dict(conf):
+vlan_default = {
+    # get the '100' in 'interfaces bonding bond0 vif-s 100'
```
c-po: What do you mean with this comment?

python/vyos/configdict.py
```diff
@@ -98,192 +103,339 @@ def get_ethertype(ethertype_val):
...
+    # Get the interface IPs
+    eff_addr = conf.return_effective_values('address')
+    act_addr = conf.return_values('address')
+
+    # Get prefixes for IPv6 addressing based on MAC address (EUI-64)
+    eff_eui = conf.return_effective_values('ipv6 address eui64')
+    act_eui = conf.return_values('ipv6 address eui64')
+
+    # Determine interface addresses (currently effective) - to determine which
+    # address is no longer valid and needs to be removed from the bond
```
c-po: I guess bond here is just a leftover.

c-po (changes requested):

c-po:
I really like the idea and besides those two small comments all looks good to me. Less duplicated code!

python/vyos/configdict.py
thomas-mangin: leftover from line 109 ..

python/vyos/configdict.py
```diff
@@ -98,192 +103,339 @@ def get_ethertype(ethertype_val):
...
+
+def intf_to_dict(conf, default):
     """
     Common used function which will extract VLAN related information from config
     and represent the result as Python dictionary.
 
     Function call's itself recursively if a vif-s/vif-c pair is detected.
     """
-    vlan = {
-        'id': conf.get_level()[-1], # get the '100' in 'interfaces bonding bond0 vif-s 100'
```
thomas-mangin, line 109: @c-po this left over.

python/vyos/configdict.py
thomas-mangin: Yes, this comment is also a left-over !

jjakob:
@thomas-mangin one of us will need to rebase our changes as we were working on the same code. Look at my PR.

Does this work fine regarding the link-local address code?

c-po:
@thomas-mangin can you please fox the leftovers?

thomas-mangin:
I fixed the two comments

PR #384, opened 2020-05-01
T2241: fix interfaces falling out of bridge

jjakob:
Untested. If there are too many commits I can join them.

Edit: I think me and @thomas-mangin were working on some of the same code. I need to look at his PR changes and see if there is any duplication.

### DIFF
--- a/python/vyos/configdict.py
+++ b/python/vyos/configdict.py
@@ -266,9 +266,6 @@ def intf_to_dict(conf, default):
     if conf.exists('ipv6 disable-forwarding'):
         intf['ipv6_forwarding'] = 0
 
-    # Media Access Control (MAC) address
-    if conf.exists('mac'):
-        intf['mac'] = conf.return_value('mac')
     # check if interface is member of a bridge
     intf['is_bridge_member'] = is_member(conf, intf['intf'], 'bridge')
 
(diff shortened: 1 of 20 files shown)


c-po:
I like small commits :)

jjakob:
Ok. Yeah, looking at #383 one of us will definitely need to rebase as the vlan code has substantial changes in both.

c-po:
This still a WIP?

jjakob:
Yes, still testing

jjakob:
I'll rebase after #388 is merged

python/vyos/configdict.py
```diff
@@ -119,6 +119,8 @@ def vlan_to_dict(conf):
         'disable_link_detect': 1,
         'egress_qos': '',
         'egress_qos_changed': False,
+        'id': '',
```
thomas-mangin, line 122: This is not shared by all the code, this is why when I moved this "vlan" (really I should have renamed this ! I did not add it. I believe that adding it, would set the expectation that intf_ code would set it.

python/vyos/ifconfig/section.py
```diff
@@ -54,7 +54,7 @@ def _basename (cls, name, vlan):
         name = name.rstrip('0123456789')
         name = name.rstrip('.')
         if vlan:
-            name = name.rstrip('0123456789')
+            name = name.rstrip('0123456789.')
```
thomas-mangin, line 57: how could we have `eth0.0.`?

jjakob:
vif-s 1 vif-c 2 -> eth0.1.2
Without the patch this would result in 'eth0.'

Edit: the comment I replied to was deleted.

jjakob:
Ready to merge - tested basic functionality for ethernet, bridge, bond, vif. Other interface types need testing.

There is a weird bug with the ipv6 link-local code but it is present without this PR too - it started after PR #383 - I'll investigate and create a task for it after this is merged

PR #386, opened 2020-05-02
validator: T2417: try to make the code clearer

thomas-mangin:
Make sure all the function have similar behavior:
 - sys.exit(string) to have exit(1)with message on stderr
 - use of __main__
 - if tests as the exception / error path


thomas-mangin:
I have removed my .git ... I copied a clean one copied it from another clean repo "current" branch, and then re-committed the change. Still, I have all the above ... any explanation of what is happening would be appreciated so I can fix it !

c-po:
Rebase onto origin/master

jjakob:
or `git rebase upstream/current` if your remote for vyos-1x is names upstream (I name my forks origin)
Depending on what you've done, just manually fixing each merge conflict and continuing the rebase may be enough, you can also drop unwanted commits with `rebase -i HEAD~n`. If it's too messed up, you may need to reset to upstream and force push.

src/validators/vrf-name
```diff
@@ -14,27 +14,30 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+
 import re
-from sys import exit, argv
+import sys
```
c-po, line 19: Why pull in everything when we can only pull in two functions?

src/validators/mac-address
```diff
@@ -15,12 +15,15 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import re
-from sys import exit, argv
+import sys
```
c-po, line 18: Why pull in everything when we can only pull in two functions?

c-po (changes requested):

src/validators/mac-address
jjakob: According to some python gurus, rhis shouldn't make any difference in the execution time or memory usage, the whole module is loaded in each case. It does have benefits, to prevent function name clashes and make the code terser.

c-po:
Closing - source branch vanished

thomas-mangin:
I will look for the code on my machine and re-submit, but as you commented here, I will reply :-)

For the question "why pull everything when you can pull two functions.

Unlike C where an import will only "copy" the code for the function via the pre-processor, and therefore the less is the better as it reduces compilation time and the number of "useless" object in the binary, for a module to work it must be fully initialised.

So the module is fully initialised (fully loaded and most often compiled to some bytecode). Then a reference to the object/function/... requested is inserted in the scope of the current code. So there is no benefit to "limit" the import.

The only reason why "from .. import *" is bad is that you do not know what you import and risk conflicts between the module code and yours where your code overwrites a module function/feature. 

And for why I use `import sys` and not using `from ... import ...` when using the standard library.

I have no issue with using "from" import, in most cases, they are perfectly fine. When the name is a mouthful `datetime.datetime` comes to mind, and when no other feature of the lib is used, it makes perfect sense.

`sys.exit` is a clear indicator that it is a standard library call an not another local function and therefore helps the code clarity.

There seems to be a belief that "from sys import exit" will yield better code performance due. This is simply not true in most cases.

However, Using "." and will cause object deference will indeed have a small impact. In a tight loop, we should indeed try to make sure objects in the `locals()` scope is used. I co-wrote an HTTP server in python and this kind of things become important when you start to serve multiple hundred of megabits per seconds (but is not where the bottleneck will often first appear).

Also adding "more" to the local scope (with many from) does not cause lookup slow-down as the hashing algorithm. https://wiki.python.org/moin/TimeComplexity. 

My only advice would be avoid `if value in list` for any form of sizeable data as the scanning of `list` is 0(n). The same can be done with a dict at a fraction of the CPU cost.

thomas-mangin:
resubmited as https://github.com/vyos/vyos-1x/pull/395

PR #388, opened 2020-05-02
T2367: flush addresses when adding bond/bridge members

jjakob:


python/vyos/ifconfig/interface.py
```diff
@@ -637,39 +635,51 @@ def add_addr(self, addr):
...
-            self.dhcp.v6.set()
+        # XXX: normalize/compress with ipaddress if calling functions don't?
+        # is subnet mask always passed, and in the same way?
+        ret = True
+
+        # we can't have both DHCP and static IPv4 addresses assigned
+        for a in self._addr:
+            if ( ( addr == 'dhcp' and a != 'dhcpv6' and is_ipv4(a) ) or
+                    ( a == 'dhcp' and addr != 'dhcpv6' and is_ipv4(addr) ) ):
+                raise ConfigError((
```
thomas-mangin, line 646: why is it not possible?

python/vyos/ifconfig/interface.py
```diff
@@ -637,39 +635,51 @@ def add_addr(self, addr):
...
+
+        # do not add same address twice
+        if addr not in self._addr:
+            # add to interface
+            if addr == 'dhcp':
+                self.dhcp.v4.set()
+            elif addr == 'dhcpv6':
+                self.dhcp.v6.set()
+            else:
+                if not is_intf_addr_assigned(self.config['ifname'], addr):
```
thomas-mangin: you could (if you want) use self.ifname here and use f-string.

python/vyos/ifconfig/interface.py
```diff
@@ -681,11 +691,37 @@ def del_addr(self, addr):
...
             self.dhcp.v6.delete()
         else:
             if is_intf_addr_assigned(self.config['ifname'], addr):
                 cmd = 'ip addr del "{}" dev "{}"'.format(addr, self.config['ifname'])
-                return self._cmd(cmd)
+                ret = self._cmd(cmd)
+            else:
+                return False
+
+        try:
```
thomas-mangin: `if addr in self._addr:` instead of using an exception ?

src/conf_mode/interfaces-bridge.py
```diff
@@ -346,6 +347,13 @@ def apply(bridge):
 
         # add interfaces to bridge
         for member in bridge['member']:
+            # flushes address of only children of Interfaces class
```
thomas-mangin, line 350: This code is used multiple time, should it be moved within ifconfig ?

thomas-mangin:
Sorry for the many comments - cool patch.

python/vyos/ifconfig/interface.py
jjakob: I don't know, that's what the function did previously. If there's no other limitation, this can be removed?

jjakob:
Thanks, I'll fix the things - they're all good comments.

src/conf_mode/interfaces-bridge.py
jjakob: Can you be more specific? We can't move the `if intf in Section.interfaces`, the other 2 klass lines maybe, but I don't know exactly where it would make sense.

python/vyos/ifconfig/interface.py
jjakob: I tried it quickly and an interface can have dhcp and static, but even if this can be removed it should be a separate PR

PR #393, opened 2020-05-05
T2427: fix interface addressing, bugfix config library, migrate VLAN config to dicts, move VLAN adding to common function

jjakob:


thomas-mangin:
Hi @jjakob - I like the patch but I see you removed some of the logic which deals with deciding what IP to add/remove when the interface is disabled or enabled. I am not clear on how you make sure to correctly handle it now.

There is a big difference between staying disabled, nothing to do, and going disabled, where all the IP must be removed even if some were added to the interface at the same time.

jjakob:
@thomas-mangin yes, that was intentional. There are other services that depend on having addresses configured and present on the system to work properly, dhcp-server is only one of them that has such a check in verify() - see 7a1c1fe commit message. I didn't check, but I assume other daemons won't like it if the addresses they're listening on just suddenly disappear from the system - they may crash and fail to start back up.

I don't see a harm in leaving the addresses on the interface - the interface gets put admin down so no communication can occur on it anyway. This was the old behavior too. I guess we have different concepts of what 'disabled' for an interface means.

jjakob:
Oh, I forgot to mention that the changes in config.py - using list.copy() - only work for Python > 3.3. If this is used on Crux too, the python version there needs to be checked - I can amend the commit to be Python 2 compatible.

thomas-mangin:
I believe we need a definition of disabling. For me, I should be able to re-use the IP on another interface after disabling one with it.

thomas-mangin:
On JunOS you have the disable (like we have set and delete) and it marks a part of the configuration as "here as you want to reuse it" but not here anymore. I saw the disable keyword the same way.

thomas-mangin:
@jjakob I did not realise copy() was added to list. you could replace the copy() with [:] as it does the same if it matters.

jjakob:
@thomas-mangin What you want would also require checking all services (dhcp, dns, ntp,...) and VPNs for listening addresses and refuse to disable if one of them uses an address on the interface, and vice-versa. You can't just remove addresses from the system without checking what depends on them and reconfiguring those services.
AFAIK 'disable' in VyOS has always meant to just bring the interface down. We'd need to check the old Vyatta code to be sure. In any case, it's needed to take out this functionality right now to fix bugs that make the system unusable, you can re-add it later with the appropriate checks as I described.

jjakob:
I checked, Crux has Python 3.7 so there should be no issue leaving copy() in there - Crux's config.py hasn't been updated in a long time, either, I doubt this will ever be backported :)

c-po:
That definately wont make it in to crux so python3 is safe

thomas-mangin:
@dmbaturin has confirmed that disable should admin down the interface not remove it. So indeed my changes were wrong.

PR #399, opened 2020-05-07
wireless: T2427: migrate to use common configdict and vlan functions, add common interface includes to template

jjakob:
Fixes commit error with vif on wireless interface


c-po:
Should we really add vif-s dunctionality on a wireless link? I even find vif doubtfull :)

PR #413, opened 2020-05-19
T2465: vyos-hostsd-client needs sudo

kroy-the-rabbit:
There have been a number of complaints about DHCP not getting inserted into the `/etc/hosts` file.   

This should correct that problem.

### DIFF
--- a/src/system/on-dhcp-event.sh
+++ b/src/system/on-dhcp-event.sh
@@ -43,13 +43,13 @@ case "$action" in
        exit 1
     fi
     # add host
-    /usr/bin/vyos-hostsd-client --add-hosts --tag "DHCP-$client_ip" --host "$client_fqdn_name,$client_ip"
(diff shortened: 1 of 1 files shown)


jacobweinstock:
I tested this out and don't believe this will work. Reason is, is that https://github.com/vyos/vyos-1x/blob/current/src/systemd/isc-dhcp-server.service runs dhcpd as user: `nobody` and `nobody` is not in the sudo group, as far as i can tell. This is the error message im seeing.
```bash
May 19 21:37:45 vyos sudo[6141]:   nobody : user NOT in sudoers ; TTY=unknown ; PWD=/ ; USER=root ; COMMAND=/usr/bin/vyos-hostsd-client
```

kroy-the-rabbit:
Perfect.  

That's actually the root of this problem then.  So this wasn't the fix but it helped track it down.  

Thanks!

PR #414, opened 2020-05-19
util: T2467: automatically add sudo to known commands

thomas-mangin:
This patch will automatically add "sudo" to commands which do need it, currently only systemctl.


dmbaturin (approved):
I like the idea and I think the implementation is sensible.

c-po:
Me too

PR #416, opened 2020-05-20
T2465: Permissions on vyos-hostsd socket incorrect

kroy-the-rabbit:
The DHCP server is unable to apply entries to the hosts file because the permissions on the socket are getting created wrong.

```
$ ls -al /run/vyos-hostsd.sock
srwxrwxrwx 1 root vyattacfg 0 May 20 01:38 /run/vyos-hostsd.sock

```

This gives it the correct permissions so that the nobody/nobody user/group can change it.

### DIFF
--- a/src/services/vyos-hostsd
+++ b/src/services/vyos-hostsd
@@ -278,7 +278,11 @@ if __name__ == '__main__':
 
     context = zmq.Context()
     socket = context.socket(zmq.REP)
+    
+    # Set the right permissions on the socket, then change it back
+    o_mask = os.umask(0)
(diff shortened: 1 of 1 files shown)


kroy-the-rabbit:
It's worth asking the question if setting the socket to 777 is the right step here.  

The other option would be to put the isc-dhcp server in the vyattacfg group

dmbaturin:
I think proper group settings is ultimately a better idea, but we need some plan to get rid of the hardcoded vyattacfg group. Until we agree on the plan to get rid of it, I think 666 is fine (exec bit doesn't harm, but why it's needed?)

PR #419, opened 2020-05-20
T2465/T2483 Fix DNS resolution from DHCP

kroy-the-rabbit:
This should migrate the pdns restart code from the DHCP update script to vyos-hostsd


c-po (approved):
Nice counting of work.

src/services/vyos-hostsd
```diff
@@ -255,6 +263,13 @@ def handle_message(msg_json):
     with open(STATE_FILE, 'w') as f:
         json.dump(STATE, f)
 
+    if changes > 0:
+        (ret,return_code) = popen("pidof pdns_recursor");
```
c-po: Why not use `vyos.util.process_named_running()` here?

kroy-the-rabbit: Sounds good to me.  Also found a bug (which python probably would ignored)

kroy-the-rabbit:
Royally messed up my commits here.  Resubmitting.

c-po:
you can always git push —force

PR #420, opened 2020-05-21
nat: T2460: add src/op_mode/show_nat_translations.py

thomas-mangin:
This commit adds the program but does not make use of it


src/op_mode/nat-translation.py
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env python3
```
dmbaturin: Could you add a GPLv2+ license header?

src/op_mode/nat-translation.py
```diff
@@ -0,0 +1,185 @@
...
+import ipaddress
+import argparse
+import xmltodict
+
+from vyos.util import popen
+from vyos.util import DEVNULL
+
+conntrack = '/usr/sbin/conntrack'
+
+verbose_format = "%-20s %-18s %-20s %-18s"
```
dmbaturin: There's a great library called tabulate. We use it in op_mode/show_dhcp.py for example.
Makes things so much simpler than manual formats, has built-in support for printing headers too.

src/op_mode/nat-translation.py
```diff
@@ -0,0 +1,185 @@
...
+        if ipaddr:
+            command += f' --orig-src {ipaddr}'
+    if srcdest == 'destination':
+        command += ' -g'
+
+    return command
+
+
+def run():
+    if not os.path.exists(conntrack):
```
dmbaturin: I don't think we should be testing this. If it's stated as a dependency in debian/control, then there's no chance it can be missing.

dmbaturin (changes requested):
Looks good overall, I left some comments about possible improvements.

src/op_mode/nat-translation.py
thomas-mangin: Yes, I have used it for other tools. In that case the output for "--verbose" cause every odd/even line to be formated differently so it is not really suitable.

src/op_mode/nat-translation.py
thomas-mangin: sure :-)

src/op_mode/nat-translation.py
thomas-mangin: I was just replicating "line for line" what the perl does - happy to drop :-)

PR #422, opened 2020-05-23
config: T2499: expose showConfig

thomas-mangin:
The patchset tidy a bit Config and expose the cli-shell-api showConfig functions which are used to set up the class at boot.

### DIFF
--- a/python/vyos/config.py
+++ b/python/vyos/config.py
@@ -91,10 +91,7 @@ class Config(object):
     def __init__(self, session_env=None):
         self._cli_shell_api = "/bin/cli-shell-api"
         self._level = []
-        if session_env:
-            self.__session_env = session_env
-        else:
-            self.__session_env = None
+        self.__session_env = session_env
 
         running_config_text = self.get_active()
         session_config_text = self.get_working() or running_config_text
--- a/python/vyos/config.py
+++ b/python/vyos/config.py
@@ -88,8 +88,11 @@ class Config(object):
     the only state it keeps is relative *config path* for convenient access to config
     subtrees.
     """
+    _cli_shell_api = "/bin/cli-shell-api"
+    _show_active =  '/bin/cli-shell-api --show-active-only --show-show-defaults --show-ignore-edit showConfig'
+    _show_working = '/bin/cli-shell-api --show-working-only --show-show-defaults --show-ignore-edit showConfig'
+
     def __init__(self, session_env=None):
-        self._cli_shell_api = "/bin/cli-shell-api"
         self._level = []
         self.__session_env = session_env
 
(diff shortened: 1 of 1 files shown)


thomas-mangin:
building installing and rebooting works, but it seems that debian changed something which broke make iso
```
06:41:10 4630 < Preparing to unpack .../acpid_1%3a2.0.31-1_amd64.deb ...
Unpacking acpid (1:2.0.31-1) ...
06:41:10 4631 < Setting up acpid (1:2.0.31-1) ...
06:41:10 4632 < Running in chroot, ignoring request.
All runlevel operations denied by policy
06:41:10 4633 < invoke-rc.d: policy-rc.d denied execution of start.
06:41:10 4634 < Processing triggers for systemd (241-7~deb10u4) ...
06:41:11 4635 < I: Create initramfs if it does not exist.
Extra argument '4.19.124-amd64-vyos'
06:41:11 4636 < Usage: update-initramfs {-c|-d|-u} [-k version] [-v] [-b directory]

Options:
 -k version	Specify kernel version or 'all'
 -c		Create a new initramfs
 -u		Update an existing initramfs
 -d		Remove an existing initramfs
 -b directory	Set alternate boot directory
 -v		Be verbose

See update-initramfs(8) for further details.

E: config/hooks/live/17-gen_initramfs.chroot failed (exit non-zero). You should check for errors.
P: Begin unmounting filesystems...
```

dmbaturin:
Why do you want to expose it?
We are trying to _abstract away_ the legacy, not increase our reliance on it.

Is there a specific short-term gain of sufficient magnitude that justifies it?

thomas-mangin:
I am working on the same 40% speedup (and then more) on all command using Configuration as @jestabro with https://github.com/vyos/vyos-1x/pull/424 - he beats me to it.

He achieved it by using some files on the file system as cache. As I explained Thursday (or at least I thought I did !) I am working toward achieving the same result by having the configuration objects cached in memory and making all the functions in the class Config use RPC using the vyos-hostd daemon to answer them (I see no reason to have a myriad of daemon when this is a perfectly good one which is already setup).

This is a break down of my uncompleted incoming patchset to do so. I just wanted to provide the work in easily digestable chunks, where the Config class is renamed DirectConfig, and used in vyos-hostsd, the Config class used by the rest of the code then just implement all the current function as RPC. 

In this patch PR, I remove all the calls made in __init__ to function to be able to send them via RPC later on.

Part of this work I also noticed that `systemctl restart vyos-hostsd` was having some lovely side effect - hence PR #423 too.

Once this is done, any commands which could gain from caching could be added to the list, such as ""render" as the loading of the jinja template is costly and would gain nicely from being cached by the daemon too futher improving commit time for example, but this is just the once example I can think of while writing this answer.

thomas-mangin:
For clarity, my solution only have works during boot. Until the /tmp/vyos-config-status is written, as until then we know that there is no write to the configuration and therefore caching the initial reads is possible. This is not true after boot. 
@jestabro solution does optimise the cost past boot, when I was mostly concerned with the boot time.

PR #423, opened 2020-05-23
systemd: T2494: allow to restart vyos-hostsd without consequence

thomas-mangin:
This patchset tidy up the systemd setup for vyos-hostsd and vyos-http-api


jjakob:
vyos-hostsd needs to be running before vyos-router starts, so it probably needs a Before=vyos-router.service

thomas-mangin:
I did not realise it did as it is mostly FRR and Firewall - thank you adding it to the list :-)

jjakob:
It starts the boot config load and commit, so it does need vyos-hostsd running as various scripts then depend on it running (e.g. host_name.py, dns_forwarding.py, dhcp-server, dhclient...)

dmbaturin:
Note that `Before=` by itself doesn't create a dependency. Does vyos.target include cloud-init now? If yes, it shouldn't do that in non-cloud-init'y flavors.

My `RequiredBy` dependency was tailor-made to have it start before cloud-init if cloud-init is enabled. The means don't matter, but that's how it must work.

thomas-mangin:
Correct `Before=` is just for the boot ordering the `WantedBy=` is what ensure the service is enabled and configured to run.

RequiredBy= and WantedBy= are similar. RequiredBy=  will cause the activation to fail if not met. But it also means (not seen in the doc) that if the service restart (daemon dying and restarted) any service listed will also restart, surely not what we want for cloud-init and vyos-router 😄

thomas-mangin:
@dmbaturin Do you think a "Before" for cloud-init-local.service would be enough ?

jjakob:
We then need to add Wants=vyos-hostsd.service to one of the services (probably vyos-router). If cloud-init needs vyos-hostsd, then add the Wants to cloud-init-local.service too.
I'd also use After= in vyos-router rather than Before= in vyos-hostsd.
The service requiring vyos-hostsd needs it in both Wants= and After=.
IMO that's better than WantedBy or RequiredBy, it's more clear (direct dependency rather than reverse dependency).
It would mean that if vyos-hostsd fails to start for any reason, the system hostname, /etc/hosts and resolv configuration will be broken. But at least the system will boot up so that the fault can be found, as opposed to stalling on bootup. The console will have a printed failure message, and we can restart vyos-hostsd without restarting vyos-router.

dmbaturin:
@thomas-mangin Last I tried, it wasn't, which is why I put both vyos-router and clout-init in RequiredBy.

thomas-mangin:
@dmbaturin it is easy to reproduce: send kill signal to the vyos-hostsd process and look at the printing on the console as vyos-router restarts.

thomas-mangin:
rebased the patch to current

PR #424, opened 2020-05-23
config: T2497: cache config strings during a commit

jestabro:
Simple caching; requires companion pr of vyatta-cfg. Its effect can be seen by comparing the time taken to add 100 dummy interfaces:

without caching:

real 0m37.466s
user 0m17.708s
sys 0m7.440s

with caching:

real 0m19.347s
user 0m12.524s
sys 0m2.708s

### DIFF
--- a/python/vyos/config.py
+++ b/python/vyos/config.py
@@ -70,6 +70,11 @@ import subprocess
 
 import vyos.configtree
 
+config_cache = True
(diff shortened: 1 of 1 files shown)


thomas-mangin:
Should the file not better saved in `/var/run`?

jestabro:
This was a simple workaround, which is no longer needed; a proper solution will follow.

PR #426, opened 2020-05-24
Corrected help text for pppoe statistics command

njh:


njh:
Thanks for merging so fast!

PR #427, opened 2020-05-24
airbag: T2088: do not call Config.in_session

thomas-mangin:
It was found that calling Config() was extremely slow and
therefore was affecting performance. It was called by airbag at
each import. The first band-aid is to only call the function if
the router if fully loaded and detect boot using vyos-config-status

For clarity, it would only run once even if imported multiple time as the module is only initialised once and then referenced when imported.

### DIFF
--- a/python/vyos/airbag.py
+++ b/python/vyos/airbag.py
@@ -13,6 +13,7 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
 import sys
 from datetime import datetime
 
@@ -24,6 +25,7 @@ from vyos.version import get_full_version_data
 
 # we allow to disable the extra logging
 DISABLE = False
+STATUS_FILE = '/tmp/vyos-config-status'
 
 
 _noteworthy = []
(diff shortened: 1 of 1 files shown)


jestabro:
The issue stills remains with this patch; timing is unaffected (adding 100 dummy interfaces):

with airbag (same as before):
vyos@vyos# time commit

real    0m37.030s
user    0m16.931s
sys     0m7.978s

without:
vyos@vyos# time commit

real    0m26.742s
user    0m13.513s
sys     0m5.136s

Also, although it is true that the module is referenced after the first import, one has this import in all _processes_ which have it in the path, through its inclusion via vrrp.py or util.py.

thomas-mangin:
patch changed to have False set instead of True .. disabling at boot (does not change the code path for when Config() is called) but it should not be called until the end of the boot.

Looking at the system trace we used / have to see two calls to cli-shell-api at boot when the `__init__` of Config is called. That init should not be called by airbag at boot by the patch now.

However, all the other commit command will still Config and will have the slow down unless https://github.com/vyos/vyos-1x/pull/424 was applied, then it is still called but not as slow.

c-po:
Is this different from Config.in_session?

https://github.com/vyos/vyos-1x/blob/current/python/vyos/config.py#L223

jestabro:
Not different; that is the call that is causing the slow down.

thomas-mangin:
Hum, ... I need to look at how to generate the systemd trace with this patch

c-po:
It is described here: https://docs.vyos.io/en/latest/contributing/development.html#boot-timing

thomas-mangin:
This is withdrawn and #439 is proposed instead.

PR #431, opened 2020-05-26
ethernet interface: T2514: add warning msg when changing mac for bond…

L6NqLW:
… member


c-po:
Probably this should be changed to an error?

PR #437, opened 2020-05-28
dhcpv6-pd: T421: Start dhcp6c in all cases when DHCPv6-PD is configured

gsadams:
This is actually two changes; the first cleans up a bit of config change that was left out of a recent change, and the second makes DHCPv6-PD work when the router isn't getting its ethernet IPv6 address via DHCPv6. This is probably the most common configuration, since in most cases, the router probably either gets its local network configuration from its upstream partner or is statically configured.

### DIFF
--- a/python/vyos/configdict.py
+++ b/python/vyos/configdict.py
@@ -234,9 +234,9 @@ def intf_to_dict(conf, default):
 
     # DHCPv6 prefix delegation (RFC3633)
     current_level = conf.get_level()
-    if conf.exists(['dhcpv6-options', 'delegate']):
-        for interface in conf.list_nodes(['dhcpv6-options', 'delegate']):
-            conf.set_level(current_level + ['dhcpv6-options', 'delegate', interface])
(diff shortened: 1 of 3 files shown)


c-po:
HI @gsadams,

thank you for your contribution. I actually have the same fixes on my side but did not yet pushed them. Sorry for the trouble.

In terms of `request_prefix_delegation()` please configure `address dhcpv6` on the interface until a "firendly" fix has been found how this should be handled at all.

gsadams:
I am happy to do the work to make this a complete fix. I specifically do not want `address dhcpv6`, but I know that this can specifically be worked around with `dhcpv6-options parameters-only`; I'll give it a try. But it feels dirty.

No changes to the configuration syntax should be required for a basic implementation, although I can think of a couple small changes that would make it a little cleaner and clearer to the user.

I am happy to do this work; I am an experienced software engineer. Shall I open a new task to propose the changes?

c-po:
Please do so, thanks.

gsadams:
Very good; will do.

PR #438, opened 2020-05-28
systemd: T421: Fix the patch for dhcp6c service boot failure.

jack9603301:
During the first load after boot, some interfaces may not be initialized, which will lead to the failure of dhcp6c startup. Using this setting, you can restart automatically after a period of time after startup failure to complete the fault recovery of dhcp6c.

Adding infinite failure restart recovery to dhcp6c ensures that dhcp6c can recover from failure after all interfaces are initialized.

### DIFF
--- a/src/systemd/dhcp6c@.service
+++ b/src/systemd/dhcp6c@.service
@@ -10,7 +10,6 @@ Type=forking
 PIDFile=/run/dhcp6c/dhcp6c.%i.pid
 ExecStart=/usr/sbin/dhcp6c -D -k /run/dhcp6c/dhcp6c.%i.sock -c /run/dhcp6c/dhcp6c.%i.conf -p /run/dhcp6c/dhcp6c.%i.pid %i
 
-
 Restart=on-failure
 
 StartLimitIntervalSec=0
--- a/src/systemd/dhcp6c@.service
+++ b/src/systemd/dhcp6c@.service
@@ -10,6 +10,7 @@ Type=forking
 PIDFile=/run/dhcp6c/dhcp6c.%i.pid
 ExecStart=/usr/sbin/dhcp6c -D -k /run/dhcp6c/dhcp6c.%i.sock -c /run/dhcp6c/dhcp6c.%i.conf -p /run/dhcp6c/dhcp6c.%i.pid %i
 
+
 Restart=on-failure
 
 StartLimitIntervalSec=0
--- a/src/systemd/dhcp6c@.service
+++ b/src/systemd/dhcp6c@.service
@@ -10,7 +10,11 @@ Type=forking
 PIDFile=/run/dhcp6c/dhcp6c.%i.pid
 ExecStart=/usr/sbin/dhcp6c -D -k /run/dhcp6c/dhcp6c.%i.sock -c /run/dhcp6c/dhcp6c.%i.conf -p /run/dhcp6c/dhcp6c.%i.pid %i
 
-Restart=always
+Restart=on-failure
+
+StartLimitIntervalSec=0
+
+RestartSec=10
 
 [Install]
 WantedBy=multi-user.target


jack9603301:
Specifically explained here, I set the restart time to 10s, originally I set it to 1s, but the reported restart interval is too short, so the automatic failure restart of sysemctl does not take effect.

c-po:
Thanks for the contribution. Please describe why this change is required. See https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch for more info

jack9603301:
During the first load after boot, some interfaces may not be initialized, which will lead to the failure of dhcp6c startup. Using this setting, you can restart automatically after a period of time after startup failure to complete the fault recovery of dhcp6c.

c-po:
Please read the link again, the commit message still does not reflect it.

jack9603301:
I've looked at the links and referenced other people's PR requests to describe how to write, and if you have any questions, please let me know.

c-po:
Well if you would have read the link on how to submit a patch and write a commit message read carefully you would notice that the commit message is incomplete and could be more verbose as requested above. But you just continue to prefer to just rant around and tell other people what to do, this does not help.

jack9603301:
How about that? Sorry, it's the first time I've actually submitted PR, so please let me know if it's still wrong. From the links and other people's PR, you can see the specific changes you want to make in the description

c-po:
### Writing good commit messages

The format should be and is inspired by: https://git-scm.com/book/ch5-2.html

* A single, short, summary of the commit (recommended 50 characters or less, not exceeding 80 characters) containing a prefix of the changed component and the corresponding Phabricator reference e.g. `snmp: T1111: or ethernet: T2222:` - multiple components could be concatenated as in `snmp: ethernet: T3333`

* In some contexts, the first line is treated as the subject of an email and the rest of the text as the body. The blank line separating the summary from the body is critical (unless you omit the body entirely); tools like rebase can get confused if you run the two together.

* Followed by a message which describes all the details like:
  * What/why/how something has been changed, makes everyone’s life easier when working with git bisect
  * All text of the commit message should be wrapped at 72 characters if possible which makes reading commit logs easier with `git log` on a standard terminal (which happens to be 80x25)
  * If applicable a reference to a previous commit should be made linking those commits nicely when browsing the history: `After commit abcd12ef ("snmp: this is a headline") a Python import statement is missing, throwing the following exception: ABCDEF`

* Always use the `-x` option to the `git cherry-pick` command when back or forward porting an individual commit. This automatically appends the line: (`cherry picked from commit <ID>`) to the original authors commit message making it easier when bisecting problems.

* Every change set must be consistent (self containing)! Do not fix multiple bugs in a single commit. If you already worked on multiple fixes in the same file use git add –patch to only add the parts related to the one issue into your upcoming commit.

A comment in a PR is not a commit message! But I'm also not convinced that this is the proper solution.

jack9603301:
I always thought that there was a problem with the description of PR request. Sorry, I didn't know that the current understanding was correct. I have created a new git commit to fix the errors described in the previous commit.

Zer0t3ch:
I don't currently know much about the VyOS codebase, and I assume y'all have already thought about this, but is there a `network.target` that could be referenced with an `After=` definition to help?

jack9603301:
`After =` usually means to express the dependence on the service. The service after this field must be started before this service. Vyos uses the method of automatic load configuration to manually execute the start and load commands of the service by using scripts. According to my understanding of Linux, I have never heard that the specific network interface name can be written after this field. Even if it can, it will only make it more Because of the particularity of dhcp6c, even if it can be written in this way, the file must be generated dynamically by using the template. In fact, it may not be possible to use the specific interface name after the field. Since the time to start dhcp6c may occur before its dependent interface, I used infinite failover for recovery.

c-po:
> I don't currently know much about the VyOS codebase, and I assume y'all have already thought about this, but is there a `network.target` that could be referenced with an `After=` definition to help?

Unfortunately this is not how VyOS curently works.

jack9603301:
@c-po After thinking about it, maybe the only thing I can think of is the word limit of a single line. I submitted another commit.

jack9603301:
Because the change is too small, I use the web page to submit. I may not have thought of this problem before, and I can't see how many words there are in a single line.

jack9603301:
Nobody?

c-po:
As you were not able to comply with the rules this PR is closed as it is superseeded by 992d356.

jack9603301:
Finally, I would like to know whether it is because the single line length limit does not meet the requirements? To prevent the same mistake next time?

c-po:
I have written all here already: https://github.com/vyos/vyos-1x/pull/438#issuecomment-636100998

You added three commits, which sometimes only changed a whitespace and the commit message was not very descriptive for such a behavioral change.

jack9603301:
Thank you. Let me see

PR #439, opened 2020-05-29
airbag :T2088: make airbag explicit

thomas-mangin:
airbag must now be explicitly installed.

the patch also allow to fully disables the installation of the logging
code at setup (and not just installing and doing nothing)


c-po:
Hi Thomas,

why cant this be enabled in vyos.config only once?

thomas-mangin:
vyos.config is imported by some op command. I do not think this feature should be included for operational commands AFAIU. So this patch matches the current behaviour. Happy to change it if the consensus is that it is fine to run all the time, but then vyos.config may not be the right file ...

c-po:
Ah missed that - you are totally right!

```
# git grep vyos.config src/op_mode/
src/op_mode/dns_forwarding_reset.py:from vyos.config import Config
src/op_mode/dns_forwarding_statistics.py:from vyos.config import Config
src/op_mode/dynamic_dns.py:from vyos.config import Config
src/op_mode/ipoe-control.py:from vyos.config import Config
src/op_mode/lldp_op.py:from vyos.config import Config
src/op_mode/ppp-server-ctrl.py:from vyos.config import Config
src/op_mode/restart_dhcp_relay.py:import vyos.config
src/op_mode/restart_dhcp_relay.py:    c = vyos.config.Config()
src/op_mode/show_acceleration.py:from vyos.config import Config
src/op_mode/show_dhcp.py:from vyos.config import Config
src/op_mode/show_dhcpv6.py:from vyos.config import Config
src/op_mode/show_igmpproxy.py:import vyos.config
src/op_mode/show_igmpproxy.py:    c = vyos.config.Config()
src/op_mode/show_openvpn.py:from vyos.config import Config
src/op_mode/show_users.py:from vyos.config import Config
src/op_mode/show_wireless.py:from vyos.config import Config
src/op_mode/snmp.py:from vyos.config import Config
src/op_mode/snmp_ifmib.py:from vyos.config import Config
src/op_mode/snmp_v3.py:from vyos.config import Config
src/op_mode/wireguard.py:from vyos.config import Config
```

I just came up with the question as it adds code to a lot of files and I just wanted to see if it could be done in an easier manner.

PR #445, opened 2020-06-07
T2558: Fix hypervisor detection bug, add CPU info

kroy-the-rabbit:
There was a bug that prevented the hypervisor info from displaying correctly.  

I was also interested in seeing some CPU stats, and at least for now, this seemed like a good place to drop it.

### DIFF
--- a/python/vyos/version.py
+++ b/python/vyos/version.py
@@ -31,6 +31,7 @@ Example of the version data dict::
 
 import os
 import json
+import logging
 
 import vyos.defaults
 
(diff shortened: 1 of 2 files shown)


thomas-mangin:
IMHO, If ninja is used, the function render should be adopted and the template moved to a file to keep in line with all the rest of the code. I am wondering why the dict is partially filled? Would it not be clearer to always have all the fields and the value “unknown” if it could not be extracted?

c-po:
@thomas-mangin Jinja2 templates for op-mode usually reside in the script itself

c-po:
```
Version:          VyOS 1.3-usb-202006071234
Release Train:    equuleus

Built by:         christian@poessinger.com
Built on:         Sun 07 Jun 2020 12:34 UTC
Build UUID:       dbd37db5-fd0d-45c6-a15d-41968caeebda
Build Commit ID:  972534c08225bc

Architecture:     x86_64
Boot via:         installed image
System type:      bare metal

CPU Vendor:       AuthenticAMD
Model:            AMD GX-412TC SOC
Total CPUs:       4
Sockets:          1
Cores:            4
Threads:          1
Current MHz:      807.880
Minimum MHz:      600.0000
Maximum MHz:      1000.0000

Hardware vendor:  PC Engines
Hardware model:   apu4
Hardware S/N:     xxxxx
Hardware UUID:    Unknown
```

Can we probably move the CPU stuff to `show system cpu` as we also have `show system memory` - I'd love to not see `show version` as a "catch all" information command.

PR #449, opened 2020-06-09
bonding: T2515: call 'interfaces-ethernet.py' for each interface that…

L6NqLW:
… leaves bond

### DIFF
--- a/src/conf_mode/interfaces-bonding.py
+++ b/src/conf_mode/interfaces-bonding.py
@@ -355,6 +355,10 @@ def apply(bond):
 
         # Some parameters can not be changed when the bond is up.
         if bond['shutdown_required']:
+            # Collect "old" and "new" bond members
+            slaves_old = list()
+            slaves_new = list()
+
             # Disable bond prior changing of certain properties
             b.set_admin_state('down')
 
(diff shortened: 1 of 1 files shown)


c-po:
Until today we lived with the design principle of not mixing Python scripts together and not calling them from each other - at least I'm not aware of any. We should keep this principle as there must be another way to fix this issue.

In my opinion the code for setting up interfaces is too fragmented at all as there is a lot of duplication here, too. We probably should once and for all migrate this to the object oriented pattern and handle everything - including setting the parameters on interfaces into the `ifconfig` package. Thus there is only one implementation dealing with all the interface stuff.

jjakob:
@c-po is it decided who will do the migration? I imagine it'll be a lot of work. bridge member interfaces have the same problem.

c-po:
Whoever signs up for it - I still have some tasks in the pipeline as of now - I can assist with concepts and ideas if you want.

c-po:
As intermediate workaround until there is this new shiny implementation (again 😄) why not do the following:
- On bond deletion (check previous assigned members by calling `conf.return_effective_value()`
- Ask config backend about desired interface state (enabled/disabled)
- Get handle to interface (`EthernetIf()`) and set interface to up/down?

jjakob:
The issue is other interface settings were also purged from it, not just state, that's why I suggested calling the whole interface script as a hacky workaround

c-po:
This problem which comes and goes should be resolved once and forever with the implementation following on https://phabricator.vyos.net/T2653

c-po:
Work on https://phabricator.vyos.net/T2653 is almost complete - closing this

PR #461, opened 2020-06-17
correct the key of fail_time and add support to config radius accounting port

abxuz:
correct the key of fail_time and add support to config radius accounting port

### DIFF
--- a/data/templates/accel-ppp/pppoe.config.tmpl
+++ b/data/templates/accel-ppp/pppoe.config.tmpl
@@ -93,7 +93,7 @@ chap-secrets={{ chap_secrets_file }}
 [radius]
 verbose=1
 {% for r in radius_server %}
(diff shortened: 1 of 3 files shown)


c-po:
Thank you for your contribution. Please note that we have a more or less good development process which you can read up here: https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch

Every change should be preceeded by a Phabricator task id. We use those IDs for release notes generation.

Your change affects the overall radius-setting for every radius instance (system login, wifi, l2tp, pptp, pppoe-server ...) but you only implemented it for the PPPoE server. It should be considered adding this for all Accel-PPP related services or move this to the PPPoE only CLI definition https://github.com/vyos/vyos-1x/blob/current/interface-definitions/service_pppoe-server.xml.in#L113

c-po:
The `fail_time` fix is implemented as of https://github.com/vyos/vyos-1x/commit/6d847cd6f.

Making the RADIUS accounting ports configurable is handled via https://phabricator.vyos.net/T2756

PR #465, opened 2020-06-18
T2562: Make VyOS able to be a server for a dhcp relay

kroy-the-rabbit:
Currently it's impossible for VyOS to be used as a server for a DHCP relay.  This should correct it.

[This line](https://github.com/vyos/vyos-1x/compare/current...kroy-the-rabbit:dhcp-fixes?expand=1#diff-ea15ed988b548fe6ac1faafe011b7403R599) is the one line that I'm not sure about on the logic, and will be the most breaking change, if any.

As mentioned in the task, I'm not 100% convinced on the naming of the node

### DIFF
--- a/src/conf_mode/dhcp_server.py
+++ b/src/conf_mode/dhcp_server.py
@@ -196,7 +196,8 @@ def get_config():
 
             if conf.exists('relay-networks'):
                 for relay_net in conf.return_values('relay-networks'):
-                    config['relay_networks'].append({'network_address':str(ip_network(relay_net).network_address), 'netmask':str(ip_network(relay_net).netmask)})
+                    temp_net = ip_network(relay_net)
+                    config['relay_networks'].append({'network_address':str(temp_net.network_address), 'netmask':str(temp_net.netmask)})
 
             # check for multiple subnet configurations in a shared network
             # config segment
--- a/src/conf_mode/dhcp_server.py
+++ b/src/conf_mode/dhcp_server.py
@@ -16,7 +16,7 @@
 
 import os
 
-from ipaddress import ip_address, ip_network
+from ipaddress import ip_address, ip_network, IPv4Network
 from socket import inet_ntoa
 from struct import pack
 from sys import exit
@@ -168,7 +168,8 @@ def get_config():
                 'description': '',
                 'disabled': False,
                 'network_parameters': [],
-                'subnet': []
+                'subnet': [],
+                'relay_networks': []
             }
             # check if DHCP server should be authoritative on this network
             if conf.exists('authoritative'):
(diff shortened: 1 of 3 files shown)


src/conf_mode/dhcp_server.py
```diff
@@ -193,6 +194,10 @@ def get_config():
             if conf.exists('shared-network-parameters'):
                 config['network_parameters'] = conf.return_values('shared-network-parameters')
 
+            if conf.exists('relay-networks'):
+                for relay_net in conf.return_values('relay-networks'):
+                    config['relay_networks'].append({'network_address':str(ip_network(relay_net).network_address), 'netmask':str(ip_network(relay_net).netmask)})
```
jjakob: it would make sense to initialize ip_network(relay_net) only once here and then call its methods, rather than initialize it twice

jjakob:
Why do we need to set an extra empty subnet when there is a subnet block created for every configured subnet already? What configuration is tnat needed in? Wouldn't it make more sense to allow creating an empty already existing subnet node rather than add an extra config node that achieves the same thing?

kroy-the-rabbit:
@jjakob 

It's required when none of the subnets being served are locally attached, as is the case when VyOS is the DHCP server for a dhcp relay.

The only issue with allowing you to create a blank subnet is it breaks all other existing validation. Like I mentioned, I'm definitely open to how else to represent this.  

The alternative here is allowing our systemd unit to pass in specific interfaces.  This is something the stock unit does but our custom one doesn't (/etc/default/isc-dhcp-server)

jjakob:
Does the empty subnet block need to be for the subnet the relay server is on, and the clients behind the relay would use one of the existing subnet blocks? As I see there are benefits and drawbacks with both approaches. With this one (empty subnet declaration) we need to either have a new config node to make empty subnet blocks or remove validators on an existing subnet node (which I wouldn't do as those are the subnet blocks that actually do the "work"). With the alternative - passing all listen interfaces to the systemd unit - we'd need to know all the listen interfaces, which requires a new config node too, plus a migrator (as we'd need to pass in all interfaces the existing subnets are on, it'd need to be something like the migrator I wrote in PR#452) so more disruptive for existing, non-relay configs.
So your solution may be the best out of all of them as it only requires new configuration for relay configs without impacting existing non-relay configs.

kroy-the-rabbit:
>So your solution may be the best out of all of them as it only requires new configuration for relay configs without impacting existing non-relay configs.

Yep, I basically played with all three of these configs and came to the same conclusion.


>Does the empty subnet block need to be for the subnet the relay server is on, and the clients behind the relay would use one of the existing subnet blocks?

Correct.  Basically it forces isc-dhcp to actually "respond" on the appropriate interface.  Otherwise isc-dhcp won't actually start since none of the referenced subnet blocks are attached to any local interfaces

There are two methods that isc-dhcp decides to respond on an interface:

- It responds on any interfaces that are attached to the subnet blocks it is serving.
- It responds on any interfaces that are passed in via the systemd unit.

I'm not totally happy with the config referenced in this PR, but it definitely seems to be least disruptive.

kroy-the-rabbit:
Made that quick change

jjakob (approved):

PR #467, opened 2020-06-22
Fix VXLAN cmdline mappings and allow for more versatile configurations.

SIN3R6Y:
There are numerous minor issues with the current creation of VXLAN interfaces...

In the command line mapping, "nolearning" should not be forced when a local address is specified. Many switches that have hardware vtep features require the learning packets in order to establish a tunnel link. If we want to make "nolearning" an option, it should have it's own configuration option.

"src_interface" is not a valid command line option, it should map to "dev"

as it sits now you can't actually set "remote" and "source-address" in your configuration, it will take the configuration just but it will only apply the "remote" configuration due to the if statement.

the multicast "group" and "source-interface" portion should only trigger when one of them is specified, then check if both are present. If not, present a slightly more descriptive error.

I'm not really sure how VXLANS have worked at all with the current code.

### DIFF
--- a/python/vyos/ifconfig/vxlan.py
+++ b/python/vyos/ifconfig/vxlan.py
@@ -66,24 +66,26 @@ class VXLANIf(Interface):
         'ifname': 'add',
         'vni':    'id',
         'port':   'dstport',
-        'src_address': 'nolearning local',
+        'src_address': 'local',
+        'src_interface': 'dev',
     }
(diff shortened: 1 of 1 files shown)


SIN3R6Y:
Going to fix to comply with https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch

PR #471, opened 2020-06-23
xml: T2588: fix mistake when determining if a node is multi

thomas-mangin:


thomas-mangin:
Sorry for the buggy previous PR !

PR #481, opened 2020-06-27
xml: T2582: use xml for is_tag and is_leaf

thomas-mangin:
POC - should just work as it but not tested with bridge

### DIFF
--- a/python/vyos/validate.py
+++ b/python/vyos/validate.py
@@ -19,6 +19,7 @@ import netifaces
 import ipaddress
 
 from vyos.util import cmd
+from vyos import xml
 
 # Important note when you are adding new validation functions:
 #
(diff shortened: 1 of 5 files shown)


dmbaturin:
I've been wondering if we should really be joining all XML files into one and storing it on the system, for everything that may want to know the current commands.

thomas-mangin:
If it is stored then it can be parsed at run time and not generated at package build. The current way does save the parsing when we import the code. An alternative would be to have a generic caching system at run time. Would be better but as it was more work and could not see a clear benefit so I ended up not going with code generation.

thomas-mangin:
Please do not merge I was trying editing the PR on a iPad to see and badly resolved the conflict

c-po:
If the big one XML file is generated automatically from all the single ones that would be fine. Havi g one big file to maintain will become horrible!

thomas-mangin:
The dict is indeed generated from parsing all the xml individual files.

thomas-mangin:
fixed via github but this time correctly.

thomas-mangin:
I will update following @jestabro changes to the configuration code.

jestabro:
Nice, @thomas-mangin I will update T2707/PR 499 accordingly, so is_tag and is_leaf are universally applicable; this is also useful for config_diff, where I had used a work-around for is_leaf in the interim.

thomas-mangin:
The code was updated to remove the conflict with @jestabro config_source changes.
This patch does not remove the code in the configuration but replaces the calls to use the xml.

If the patch is merged, then it would make sense to remove these function from the configuration API.

jestabro:
smoketests pass after removal of all Config.is_* references.

PR #485, opened 2020-06-29
xml: T2636: convert multi tag to list

thomas-mangin:
does pass smoketest and should just work but I did not check correctness on a router.

### DIFF
--- a/python/vyos/config.py
+++ b/python/vyos/config.py
@@ -67,6 +67,7 @@ import re
 import json
 from copy import deepcopy
 
+import vyos.xml
 import vyos.util
 import vyos.configtree
 from vyos.configsource import ConfigSource, ConfigSourceSession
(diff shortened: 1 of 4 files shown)


thomas-mangin:
@cpo if you want to have a look :-)

thomas-mangin:
passing multi as an option as suggested by @dmbaturin - updated to cache result better.
make iso; make test passing

python/vyos/config.py
```diff
@@ -193,50 +194,56 @@ def show_config(self, path=[], default=None, effective=False):
...
-            self._dict_cache[effective] = config_dict
+        self._dict_cache[(effective, False)] = config_dict
+        if multi:
+            config_dict = vyos.xml.multi_to_list([], config_dict)
+            self._dict_cache[(effective, True)] = config_dict
 
         return config_dict
 
-    def get_config_dict(self, path=[], effective=False, key_mangling=None, get_first_key=False):
+    def get_config_dict(self, path=[], effective=False, key_mangling=None, get_first_key=False, multi=False):
```
c-po, line 221: I'd like to see this as "always true" to avoid error by accident - also then the data model is always defined and strict.

thomas-mangin: Me too but this should be done in another patch where you can make sure all works as should

jestabro:
@thomas-mangin this is a nice solution, and since I know that @c-po has been waiting on it, I'm open to merging as is, but I think we should revisit directly to address two matters:
(1) c-po's comment
(2) have multi_to_list applied not to the cached dict, but only at output, as is done with key_mangling; this would mean that multi_to_list would only recurse once, namely, for the final subdict containing the value.

I understand your reasoning regarding (2), of course, as we've discussed the 'big picture' regarding caching recently: caching the root dict was a great suggestion of yours, however, there are two reasons to not recurse through the root_dict_cache: (a) the cost far outweighs the benefit when running as a standalone script outside of the daemon; as you know, we are preserving the orthogonality of running with or without the daemon (necessary in some cases; desired in all, so as to not foreclose on the next round of development); (b) the root_dict_cache is the 'raw' form, and representational tweaks should be the last step before output --- even under a daemon, it is rare that specific values are used twice.

As written, it is mildly expensive: ~10ms.

Finally, I appreciate the clean up of config; that is needed and elegant, but in the future, these should be independent commits in the PR, especially as it is critical code. c-po reminds us of best practices here, and I try to remember to practice it: even if one's code is bug-free, regressions are always possible, and it is easier to address with atomic commits.

thomas-mangin:
I will change the patch, the list of nodes which are tag/leaf and multi can be generated from the XML and therefore the check could be converted to a dict lookup.

thomas-mangin:
Regarding (1) which is about changing the default. No issues, I was trying to break any code until someone had more time than me to test it, as the devil is always in the details.

Performance seems to be more important than I initially thought: I always assumed that the data would remain cached between call for the configuration of a single module (and therefore cache would be easy to add later on). The code does indeed perform quite a lot of tree traversing which is not the cheapest action so I looked at having all the xml `is_` functions, including is_multi, using data generated at parse time (vs run time as it is now) and storing them like @c-po likes (joining the name with dots). It would allow checking for the answer with a single key lookup.

For this approach to be efficient the path need to not be analysed so it means that the benefit can only be achieved if key mangling and tagNode name are not in the path.

Currently, the mangling code is also generic, but only (minus to underscore is performed). Adding this genericity is another complication and slow down. A discussion on phabricator on how to best refactor this may be worth while.

However, For using the path without tagNodes. The current code does offer the option to ignore them if the option with_tag is set. If @c-po is happy to use the is_multi / and defaults having removed the tagNodes, then I can continue in that direction.

jestabro:
Superseded by PR536, following discussion with @thomas-mangin.

c-po:
Isn't this the same as https://github.com/vyos/vyos-1x/pull/536?

jestabro:
#536 made a few changes to this PR, signed off on by @thomas-mangin on the call last week ... so this one can be closed.

PR #486, opened 2020-07-01
Sub dict

jestabro:
T2667:
N.B.:
1) the utility function has been added to util.py instead of configdict.py, because until configdict.py is reorganized/rationalized, importing configidct.py in config.py leads to the nosetests pulling in extraneous build deps (e.g., netifaces).
2) @thomas-mangin may want to make use of these changes in interfaces-tunnel.py; the version as of yesterday required no change, but I see he has now reorganized, including a work-around for T2662; I reviewed and tested the earlier version for compatibility.
3) the reason for the form of the return value of get_sub_dict is in T2662; the reason that get_first_key is default False is leafNodes --- of course, it would be odd to get_config_dict with path a leafNode, but defaults should not allow corner-case error.

### DIFF
--- a/python/vyos/config.py
+++ b/python/vyos/config.py
@@ -71,7 +71,6 @@ import subprocess
 import vyos.util
 import vyos.configtree
 
-
 class VyOSError(Exception):
     """
     Raised on config access errors, most commonly if the type of a config tree node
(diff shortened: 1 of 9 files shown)


dmbaturin (approved):
Looks like a good idea to me.

PR #487, opened 2020-07-03
ids: T2659: Implement fastnetmon support

DmitriyEshenko:


src/conf_mode/service_ids_fastnetmon.py
```diff
@@ -0,0 +1,96 @@
...
+from vyos import airbag
+airbag.enable()
+
+config_file = r'/etc/fastnetmon.conf'
+networks_list = r'/etc/networks_list'
+
+def get_config():
+    conf = Config()
+    base = ['service', 'ids', 'ddos-protection']
+    if not conf.exists(base):
```
c-po: Conditional statement is not required as get_config_dict() will return an empty dict if not found.

src/conf_mode/service_ids_fastnetmon.py
```diff
@@ -0,0 +1,96 @@
...
+
+from sys import exit
+
+from vyos.config import Config
+from vyos.configdict import dict_merge
+from vyos import ConfigError
+from vyos.util import call
+from vyos.template import render
+from vyos.xml import defaults
+from stat import S_IRWXU, S_IXGRP, S_IXOTH, S_IROTH, S_IRGRP
```
c-po: I see no references of those imports?

src/conf_mode/service_ids_fastnetmon.py
```diff
@@ -0,0 +1,96 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+
+from sys import exit
+
+from vyos.config import Config
+from vyos.configdict import dict_merge
```
c-po: You do not merge dicts here

src/conf_mode/service_ids_fastnetmon.py
```diff
@@ -0,0 +1,96 @@
...
+from vyos import ConfigError
+from vyos.util import call
+from vyos.template import render
+from vyos.xml import defaults
```
c-po: I see no use of default values in XML so you do not need this.

src/conf_mode/service_ids_fastnetmon.py
```diff
@@ -0,0 +1,96 @@
...
+    render(config_file, 'ids/fastnetmon.tmpl', fastnetmon, trim_blocks=True)
+    render(networks_list, 'ids/fastnetmon_networks_list.tmpl', fastnetmon, trim_blocks=True)
+
+    return None
+
+def apply(fastnetmon):
+    if not fastnetmon:
+        # Stop fastnetmon service if removed
+        call('systemctl stop fastnetmon.service')
+
```
c-po: Why not else?

c-po (changes requested):

c-po (approved):

PR #491, opened 2020-07-07
config: T2689: cache the result of get_config_dict

thomas-mangin:
The code otherwise requires JSON parsing and parsing for the key replacement each time.

### DIFF
--- a/python/vyos/config.py
+++ b/python/vyos/config.py
@@ -67,6 +67,7 @@ import os
 import re
 import json
 import subprocess
+from copy import deepcopy
 
 import vyos.util
 import vyos.configtree
@@ -91,6 +92,8 @@ class Config(object):
     def __init__(self, session_env=None):
         self._cli_shell_api = "/bin/cli-shell-api"
         self._level = []
+        self._dict_cache = {}
+
         if session_env:
             self.__session_env = session_env
         else:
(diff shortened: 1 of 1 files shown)


thomas-mangin:
while this patch works will all the code, it assumed (as I missed the case) that two following call will have the same path. Change the patch to provide a naive per path caching (could be improved by loading [] and then looking up the path).

The forced push below changed that.

jestabro:
@thomas-mangin this is definitely much needed; two comments:
1) the line 304 will need to use _make_path for path to respect _level, and allow string paths in get_config_dict, which begs the following question
2) do we really want to cache per path? It is rare that the same path will be used more than once, and if running under a daemon, it will mean all paths for all conf_mode scripts will be cached; honestly asking, as I'm not sure gains vs overhead

jestabro:
Also, since we will cache the global dicts anyway, we should allow setting key_mangling globally --- I'm setting that up (cache would have unmangled and mangled, so one could re-mangle if needed); that avoids another need of caching by path.

--- likely a huge bottleneck to apply globally; strike that

thomas-mangin:
@jestabro I agree we should have a global dict cache and seek into it. Thanks for pointing the _make_path issue. I believe this would be worth implementing as it makes calling the function free'ish in term of CPU.

jestabro:
Committed the modified version discussed with @thomas-mangin .

PR #494, opened 2020-07-13
dns: T2675: fix recursor.vyos-hostsd.conf.lua

jack9603301:
When users use the standard fully qualified domain name writing method,
 there will be an extra point after the actual domain name. 
In order to ensure that the standard writing method is supported,
 it should not be mandatory to add this point in Lua script

### DIFF
--- a/data/templates/dns-forwarding/recursor.vyos-hostsd.conf.lua.tmpl
+++ b/data/templates/dns-forwarding/recursor.vyos-hostsd.conf.lua.tmpl
@@ -17,7 +17,7 @@ addNTA("{{ a }}.", "{{ tag }} alias")
 -- from 'service dns forwarding domain'
 {%- for zone, zonedata in forward_zones.items() %}
 {%- if zonedata['addNTA'] %}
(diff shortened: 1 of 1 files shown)


c-po:
@jjakob thoughts?

jack9603301:
Because I haven't dealt with it for a long time, I have to repair it by myself. However, it seems that the compatibility without. Is the best, and it also meets the requirements of the specification. Therefore, this point should not exist in Lua script

jack9603301:
Because I haven't dealt with it for a long time, I have to repair it by myself. However, the compatibility without this point seems to be the best at present, and this is also in line with the requirements of the specification. Therefore, this point should not exist in Lua scripts.

jjakob:
According to PowerDNS devs the dot isn't mandatory so this should work fine.

jack9603301:
According to the standard writing method of bind for zone name, the last point of zone name is not mandatory, which conforms to relevant specifications. If PDNS meets the specification, then it should work normally

jack9603301:
Originally I was hesitant, but after local test, it can work normally

PR #496, opened 2020-07-15
 ndptool: T2706: Support NDP protocol monitoring

jack9603301:


c-po:
Would it be possible to also have a `show` command just giving a snapshot of the current ndp table?

In addition it would be nice to have a command below `monitor ndp` to only show certain messages like rs, ra, ns, na.

jack9603301:
I will implement your second proposal first. As for the first one, I still need to study it.

jack9603301:
I will start to study how to improve it in a few hours. It's evening in China, and the current network is slow

jack9603301:
@c-po 
As far as I know, vyos has the following command to view the network neighbors, which seems to be the NDP table
```
show ipv6 neighbors
```

c-po:
Ah, thanks!

jack9603301:
In the document development section, I suggest that the XML framework formats of operation mode and configuration mode should be explained separately. I have to say that this makes me a little confused

c-po:
Can you please squash your commits into a single one with a proper commit message, see https://chris.beams.io/posts/git-commit/ for additional info. There are a lot of whitespaces - also a lot of commits for some lines of XML. Actually everything can be squashed into one commit and force-pushed to your branch.

Besides those cosmetic changes the commit works as expected.

jack9603301:
@c-po It has been compressed into a commit

PR #498, opened 2020-07-15
T1802: Wireguard QR code pkg dependencies

Maltahl:
This is used for adding QR codes for wireguard tunnels for mobile clients via the CLI for easy setup.
as mentioned on
https://phabricator.vyos.net/T1802

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -105,7 +105,10 @@ Depends: python3,
   nftables (>= 0.9.3),
   conntrack,
   libatomic1,
-  fastnetmon
+  fastnetmon,
+  qrencode,
+  libpng16-16,
+  libqrencode4
 Description: VyOS configuration scripts and data
  VyOS configuration scripts, interface definitions, and everything
 


c-po:
Thank you for this PR. Please follow our development process as described here: https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch

Also those commands only make real sense to me when an appropriate command for operation level exists, like `show interfaces wireguard wg0 qr-code` which is not included in this PR, only some more or less "random" dependencies.

Maltahl:
> Thank you for this PR. Please follow our development process as described here: https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch
> 
> Also those commands only make real sense to me when an appropriate command for operation level exists, like `show interfaces wireguard wg0 qr-code` which is not included in this PR, only some more or less "random" dependencies.

Hi c-po,

show wireguard wgX peer $peer-name qr-code

The QR code can only be generated when the peer config is made in VyOS.

I might need some assistance on getting the correct outputs from VyOS configuration to generate a proper QR code with the following formatting:

```
[Interface]
PrivateKey = abcdefghijklmnopqrstuvwxyz1234567890=+
Address = YOUR_VPN_PRIVATE_IP/24
DNS = 1.1.1.1, 1.0.0.1

[Peer]
PublicKey = YOUR_SERVER_PUBLIC_KEY
AllowedIPs = 0.0.0.0/0
Endpoint = YOUR_SERVER_WAN_IP:51820
```

c-po:
Well I guess then you would need to write a Python script which reads in the CLI configuration for your WireGuard peer and transforms it to the QR code.

Something like the following could be used for starters:

```python
c = Config()
base = ['interfaces', 'wireguard', 'wg0', 'peer', 'foo']
if c.exists_effective(base):
    pubkey = c.return_effective_values(base + ['pubkey'])
    address = c.return_effective_values(base + ['address'])
    port = c.return_effective_values(base + ['port'])
    endpoint = f'{address}:{port}'
```

Maltahl:
i guess this was more complicated than needed to be as i expected it to reside in proper syntax for wireguard to use in VyOS.

Closing this as im not strong enough in Python or VyOS's way of handling things.

PR #499, opened 2020-07-15
T2707: allow alternative initialization data for config

jestabro:
T2707: allow alternative initialization data for config

config: T2707: use ConfigSource and refactor Config methods
load-config: subclass ConfigSourceSession instead of Config
http-api: remove unneeded check for VyOSError
config: T2707: add ConfigSource classes for alternative init data


jestabro:
I will hold off on merging until adopting some suggestions of @thomas-mangin

PR #500, opened 2020-07-15
Improve readability of the port validation. Also fixes T2708.

dmbaturin:
There are two issues with the current `show flow-accounting` scripts: validation logic uses some really complicated regexes, and errors are displayed with usage information, as if the user ran the script by hand (which usually isn't the case).

This patch fixes both.


dmbaturin:
I take it that no one has objections. ;)

PR #501, opened 2020-07-16
pseudo-ethernet: T2626: Fix for changing interface mode

srividya0208:
Since it is not impossible to change MACVLAN interface mode on the fly, this fix adds a new one trigger to recreate it from scratch if the mode changed in system config

### DIFF
--- a/src/conf_mode/interfaces-pseudo-ethernet.py
+++ b/src/conf_mode/interfaces-pseudo-ethernet.py
@@ -36,7 +36,7 @@ default_config_data = {
     'ip_arp_cache_tmo': 30,
     'ip_proxy_arp_pvlan': 0,
     'source_interface': '',
-    'source_interface_changed': False,
+    'recreating_required': False,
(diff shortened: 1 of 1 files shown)


src/conf_mode/interfaces-pseudo-ethernet.py
```diff
@@ -79,11 +79,14 @@ def get_config():
         peth['source_interface'] = conf.return_value(['source-interface'])
         tmp = conf.return_effective_value(['source-interface'])
         if tmp != peth['source_interface']:
-            peth['source_interface_changed'] = True
+            peth['recreating_required'] = True
 
     # MACvlan mode
     if conf.exists(['mode']):
         peth['mode'] = conf.return_value(['mode'])
+        tmp = conf.return_effective_value(['mode'])
```
thomas-mangin, line 87: Could have called the variable mode and not tmp .. it would have been clearer.

PR #502, opened 2020-07-17
Admin State:T2673: Fix for admin state if the mac address is changed …

srividya0208:
…for an interface

	 Changes are made in the interface.py script in order to bring the admin state to 'UP' after the mac is manually added in system config.The script is marking the interface from up to down state(as the MAC address can only be changed if interface is in 'down' state) but it is not bringing it up after the change


c-po:
This is due to rewrite anyhow so merging this as fix. A test in https://github.com/vyos/vyos-smoketest/blob/master/scripts/cli/base_interfaces_test.py  would be great to avoid this issue in the future.

PR #504, opened 2020-07-18
monitor: ndp: T2706: Modify the file name to the correct file name

jack9603301:
Modify the file name to the correct file name, NTP 
should be changed to NDP


jack9603301:
It seems that this is not needed. It will be closed

PR #508, opened 2020-07-20
arping: T2715: Duplicate address detection option supporting ARP

jack9603301:


jack9603301:
@c-po Request merge

jack9603301:
Replace `send` with `discovery` because these tools are host information that can obtain MAC data by sending data, which can be regarded as an automatic discovery tool of host topology

jack9603301:
@c-po I have changed the top-level command, maybe it's better, can you help me see it, if possible, request a merge, otherwise, please reply?

c-po:
VyOS already supports a partial implementation for `arping` under the `force arp` command tree.

Please reuse the tree and add the missing features there. No need to introduce a new tree here. In addition our XML files use indent = 2 spaces.

c-po:
Why is the `duplicate` node explicitly required? Couldn't this be mangled together with the interface? You also always state "gratuitous ARP" but your nodes are for DAD.

Please also avoid the reverse merge in your PR, please rebase to latest `current` which has the exact same outcome but drops the reverse merge.

jack9603301:
@c-po 
> Couldn't this be mangled together with the interface?

I don't understand the meaning of this sentence

jack9603301:
@c-po I don't understand the meaning of this sentence

> Couldn't this be mangled together with the interface?

PR #510, opened 2020-07-21
vyos.template: T2720: Rework vyos.template Python library

bob1de:
Apart from code cleanup, this allows any module that wants to render a
template to register custom jinja2 template filters by using the new
@register_filter decorator. Examples can be found in vyos.template
itself.

### DIFF
--- a/python/vyos/template.py
+++ b/python/vyos/template.py
@@ -1,4 +1,4 @@
-# Copyright 2019 VyOS maintainers and contributors <maintainers@vyos.io>
+# Copyright 2019-2020 VyOS maintainers and contributors <maintainers@vyos.io>
 #
 # This library is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


thomas-mangin:
As I was the author the template.py: at first glances LGTM.

Just reminded me that we should use octal 0o755 (octal notation which is what the rwx is (3 bits, 0-7) with os.chmod and get rid of chmod_755.

bob1de:
> get rid of chmod_755.

Definitely. I think you had a reason to introduce these back in the day, but `chmod(..., 0o744)` seems a lot cleaner. I could check what is using them (is it vyos-1x only?) and remove the `chmod_*` functions in a subsequent PR.

c-po:
I like the `register_filter` decorator. Smoketests passed, thanks!

PR #513, opened 2020-07-21
xml: T2688: read the xml definition from /usr/share/vyos/

thomas-mangin:
not sure why the list of folders/files in debian/vyos-1x.install had to be extended (if someone can enlighten me, please do) but without the changes to the file the .deb will miss content.


thomas-mangin:
@dmbaturin the implementation of the idea you put forward in https://github.com/vyos/vyos-1x/pull/481

thomas-mangin:
Waiting for this to be merged before resuming work on generating defaults with tagNode names

thomas-mangin:
I would appreciate if this could be merged

thomas-mangin:
rebased and updated to work with new xml code (which caused an issue) make iso; make test passing fine.

debian/vyos-1x.install
```diff
@@ -8,3 +8,13 @@ etc/vyos
 lib/
 opt/
 usr/
+usr/lib/
```
c-po, line 11: Why is this needed? It's already covered by `usr/`

thomas-mangin: I would like to ask this exact question to the Debian developers as testing showed that without being explicit the files would not be included. Hence this exact question in the PR comment as I am also puzzled and wondered if there was some syntax I am not aware of to make this list shorter!

This is the smallest subset which worked and included all the files (comparing the content with the previously installed package and the new one using `dpkg -l` and `diff` ..

jestabro:
Following some tests, and discussion with @thomas-mangin , it may help to outline 3 separate ideas in this PR:
(1) new implementation of is_leaf/is_tag using runtime parsing of XML
--- this adds a notable increase in commit times; the previous implementation (by Thomas) with compile time parsing works quite well, so should be kept as is.
(2) adds XML definitions to /usr/share/vyos of the installed image
--- this was (part of?) the original suggestion of Daniil; useful as reference, and going forward may be necessary for (3), below; the issue mentioned by @c-po above would need to be cleaned up, as I believe dh_install provides.
(3) provides framework for runtime parsing of XML
--- Thomas's suggestion for a use case (not is_leaf/is_tag) is for the support of user-provided addons. This is a compelling idea, but should be integrated into the requirement/design docs for VyOS addons before being implemented: it is not clear at the outset whether this should be a capability of the SDK, or of the live system --- it would seem that it belong in the SDK, but the design needs to be mapped out.

thomas-mangin:
(3) the code allows for this feature to be added, it does not add it, but as XML merging code is now available it makes things easier should it become something. 

There is many way to keep the benefit of the PR but not cause initialisation speed impact:
(1) generating the cache at first execution (moving it from the source tree)
(2) disabling the loading of the XML (checking for the dict before the XML, or not installing the XML file)would be a trivial change to the PR while keeping the other improvements.

I feel that we would be better served doing (1)

PR #514, opened 2020-07-22
traceroute: T2723: Support tcptraceroute

jack9603301:


dmbaturin (changes requested):
I agree this can be a useful feature, but I think there may be better command names.

Zeroth, "send-mtr" references a specific tool. This isn't a good idea for a simple reason: we do change the underlying implementations from time to time. If we replace MTR with something else for that role, that reference will become misleading.
It's not completely unlikely, suppose original maintainer abandons MTR and/or someone makes a more actively maintained fork. We'll have to either rename the command or keep a misleading name—both options are confusing.
These things happened already, e.g. dnsmasq to pdns migration.
Besides, people who know what mtr is, can already just run it from the command line. People who don't know, won't get an idea to look for it in "send".

Now, first, I think "tcp-traceroute" would be easier on the eyes than "tcptraceroute".

Second, you are introducing a new top level word ("send"). This isn't a decision to take lightly, especially if there's just one command there. I can't think of anything else that we might put in the "send" subtree.

Since we already have a top level "traceroute" word, I believe to put tcp-traceroute there.

To summarize:
* remove the "send mtr" command
* rename "send tcptraceroute" to "traceroute $host tcp" or similar.

jack9603301:
I take your suggestion and I'm making a correction

jack9603301:
@dmbaturin There may be a series of tools under the command sending, such as arping and some network tools of IPv6.Therefore, we want to keep send as a unified identity.But I agree to amend the order to make it more reasonable

jack9603301:
I found a more suitable place. I didn't know there was a top-level command for route tracing

runborg:
i will also point out that the mtr command is already supported by the `monitor traceroute` command. if some options are missing, please extend that command instead of creating a new one. 

I'm also in line with @dmbaturin on the names, we already have a traceroute command and tcp support needs to be baked into that command and not added as a separate command.

jack9603301:
I will introduce tcptraceroute into the existing commands

jack9603301 (commented):
The latest submission uses an extension of the existing command and has been corrected

jack9603301:
@c-po Request merge

runborg (approved):
Looks good

PR #515, opened 2020-07-22
ipv6-tool: T2724: Support for IPv6 Toolset

jack9603301:
Support for IPv6 Toolset:

- ICMPv6 Neighbor Discovery tool
- ICMPv6 Router Discovery tool
- Send NDP Message(edit: Cancel implementation)

### DIFF
--- a/op-mode-definitions/force-ipv6-nd.xml
+++ b/op-mode-definitions/force-ipv6-nd.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0"?>
+<interfaceDefinition>
+  <node name="force">
+    <children>
+      <node name="ipv6-nd">
+        <properties>
+          <help>ICMPv6 Neighbor Discovery tool</help>
+        </properties>
+        <children>
+          <tagNode name="interface">
+            <properties>
+              <help>IPv6 Neighbor Discovery on a specific interface</help>
+              <completionHelp>
+                <script>${vyos_completion_dir}/list_interfaces.py</script>
+              </completionHelp>
+            </properties>
+            <children>
+              <tagNode name="address">
+                <properties>
+                  <help>The IP address of the target</help>
+                  <completionHelp>
+                    <list>&lt;h:h:h:h:h:h:h:h&gt;</list>
+                  </completionHelp>
+                </properties>
+                <command>ndisc6 $6 $4 -m</command>
+              </tagNode>
+            </children>
+          </tagNode>
+        </children>
+      </node>
+    </children>
+  </node>
+</interfaceDefinition>
--- a/op-mode-definitions/force-ipv6-rd.xml
+++ b/op-mode-definitions/force-ipv6-rd.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0"?>
+<interfaceDefinition>
+  <node name="force">
+    <children>
+      <node name="ipv6-rd">
+        <properties>
+          <help>IPv6 Router Discovery tool</help>
+        </properties>
+        <children>
+          <tagNode name="interface">
+            <properties>
+              <help>IPv6 Router Discovery tool on a specific interface</help>
+              <completionHelp>
+                <script>${vyos_completion_dir}/list_interfaces.py</script>
+              </completionHelp>
+            </properties>
+            <command>rdisc6 $4</command>
+            <children>
+              <tagNode name="address">
+                <properties>
+                  <help>The IP address of the target</help>
+                  <completionHelp>
+                    <list>&lt;h:h:h:h:h:h:h:h&gt;</list>
+                  </completionHelp>
+                </properties>
+                <command>rdisc6 $6 $4 -m</command>
+              </tagNode>
+            </children>
+          </tagNode>
+        </children>
+      </node>
+    </children>
+  </node>
+</interfaceDefinition>


jack9603301:
It may need to be corrected. Please do not merge for the time being

jack9603301:
The TCP route tracing command of IPv6 has been deleted and merged into #514

jack9603301:
The same top-level command is used in the #508

jack9603301:
@c-po Request merge

c-po:
Hi @jack9603301 your PRs are much appreciated but triggering my eMail inbox won't get your stuff merged faster - it actually will be quiet the opposite as I think you will trigger me again when it's urgent.

As most of the developers I have a regular job and are also busy with VyOS development.

I'm not to convinced with introducing a new `send` top level node. As asked in another PR how is Cisco/Juniper handling such cases? Did they already come up with a proper CLI? I'm asking b/c right now I have no other idea then `exec` - and `send` just feels wrong and leaves a bad feeling in my stomach. Please give us maintainers more time to think about such changes in the API/ABI.

jack9603301:
@c-po I have carefully checked the relevant documents. It may be better to change send to discovery?

jack9603301:
Replace `send` with `discovery`, because these tools can obtain the neighbor data of router or host by sending specific IPv6 NDP message, which can be regarded as automatic discovery tools of host topology.

The construction of NDP sends commands as a separate command structure.

jack9603301:
@c-po I have changed the top-level command, maybe it's better, can you help me see it, if possible, request a merge, otherwise, please reply?

c-po:
You again created a new top-level node called `ndp` - this is nothing we take lightly - why can't this be fitted in another tree? Also it should rather be `discover` over `discovery` - are we on board the USS Discovery (NCC-1031)? ;)

What also happened again is - you hardocde a tools name into the node definition - this is not what we want. We thus never can exchange the tool (ndisc6, rdisc6).

jack9603301:
@c-po It may be very simple to change the tool name to a more recognizable name, because we know the actual use of these two tools. What puzzles me most is the NDP command. Setting it seems that it simply sends some NDP instructions to induce the superior router or neighbor host to respond and actively send relevant requests. Therefore, this command does not return information. How to set it What about its orders?

jack9603301:
@c-po I don't know how to set the NDP command reasonably, so I'm looking forward to hearing suggestions from the community. These tools are used as IPv6 utility tools to support entering vyos, which should be helpful to vyos users.

jack9603301:
Perhaps, after the command of vyos, the command of sending NDP specific type message is not provided temporarily?

thomas-mangin:
Hi @jack9603301 

It is really nice to see your enthusiasm. We are currently considering what we could do to help new contributors such as yourself, who may benefit from some form of mentoring.

You seem to be a good programmer and have solid networking understanding so I am sure that we will be able to work with you.

I would suggest that you to speak with us on slack before spending time coding as I know that there is nothing more frustrating that a stuck pending PR after some hard work on it.

Also we tend to work on phabricator, and before starting new work, you could also indicate what you desire to do so there. It is were we try to keep the historic and knowledge about the project, not on GitHub.

Regarding your comments:

VyOS is used by a large number of people to run large networks, and our users expect us to do as good a job as known vendors.

It may indeed be simple to change the code to update a name but you must also think about users who will have learned the syntax as well as people with automated tools which will break on update.

Also we need to document features and all the documentation will also need to be changed too. We have limited time and therefore it is logical for us to try to get it as right as possible first time.

Were you able to have a look at other projects and how they how they implemented this?
This is what @cpo asked you. We can not move the discussion forwars until we know that.

Please reply on phabricator, where we will be able to discuss this feature there.

Have a good weekend.

jack9603301:
@thomas-mangin As I said in phabricator, providing some IPv6 utilities can help those users to a certain extent, and even be used for troubleshooting.Thank you for your reply

c-po:
@jack9603301 your implementation is too nested. Please move your commands to the following syntax:

* `force ipv6-nd <interface> address <address>`
* `force ipv6-rd <interface> [address <address>]`

Please also ensure that ipv6 router discovery does not require an address, it is optional!

jack9603301:
> @jack9603301 your implementation is too nested. Please move your commands to the following syntax:
> 
>     * `force ipv6-nd <interface> address <address>`
> 
>     * `force ipv6-rd <interface> [address <address>]`
> 
> 
> Please also ensure that ipv6 router discovery does not require an address, it is optional!


```
force ipv6-nd interface <interface> address <address>
force ipv6-rd interface <interface> [address <address>]
```

Can you design commands like this? In my opinion, the operation mode command of vyos is usually in front of the `interface`

c-po:
> Can you design commands like this? In my opinion, the operation mode command of vyos is usually in front of the interface

Yes you can, see e.g. `traceroute`implementation

jack9603301:
@c-po I updated the PR to refer to the implementation of `force ARP`. If there is any problem, please let me know and I will continue to revise it

c-po:
This PR is perfectly untested code! It can't work because your variable references are wrong: `<command>ndisc6 $7 $5 -m</command>` Those are still the old levels. Please TEST before PR!

Please add a test protocol/output of your CLI so I can see it worked for you - I will then look again.

jack9603301:
> Please add a test protocol/output of your CLI so I can see it worked for you - I will then look again.

I'm very sorry. I forgot to modify the variable reference when I modified it. Now I have finished the repair and updated PR again. But I don't know how to carry out the test mentioned here?

c-po:
> But I don't know how to carry out the test mentioned here?

Run your new op-mode commands and show the results.

jack9603301:
Here are the test results
```vyos@vyos:~$ force ipv6-rd interface br2
Soliciting ff02::2 (ff02::2) on br2...
Hop limit                 :           60 (      0x3c)
Stateful address conf.    :           No
Stateful other conf.      :           No
Mobile home agent         :           No
Router preference         :         high
Neighbor discovery proxy  :           No
Router lifetime           :         1800 (0x00000708) seconds
Reachable time            :  unspecified (0x00000000)
Retransmit time           :  unspecified (0x00000000)
 Prefix                   : 240e:fe:8c71:7f01::/64
  On-link                 :          Yes
  Autonomous address conf.:          Yes
  Valid time              :      2592000 (0x00278d00) seconds
  Pref. time              :        14400 (0x00003840) seconds
 Prefix                   : fc00:470:f1cd:101::/64
  On-link                 :          Yes
  Autonomous address conf.:          Yes
  Valid time              :      2592000 (0x00278d00) seconds
  Pref. time              :        14400 (0x00003840) seconds
 Recursive DNS server     : fc00:470:f1cd::ff00
  DNS server lifetime     :          600 (0x00000258) seconds
 Source link-layer address: 00:98:2B:F8:3F:11
 from fe80::298:2bff:fef8:3f11
vyos@vyos:~$ 
```

PR #520, opened 2020-08-01
nptv6: T2518: Initial support for nat66 (NPT)

jack9603301:
This submission actually incorporates the PR of #556.
In this new submission, the `ndppd` service configuration
file will be automatically configured and generated.


jack9603301:
This PR will provide experimental nat66 support, which needs to be tested

jack9603301:
@c-po According to H3C, the relevant operations are as follows:

```
nat66 prefix source original-ipv6-prefix prefix-length translated-ipv6-prefix prefix-length
nat66 prefix destination original-ipv6-prefix prefix-length translated-ipv6-prefix prefix-length
```

Nat66 can also be divided into SNAT and DNAT

According to the current configuration path of vyos, there are two solutions:

a) Change the command of NPT to add a subset, as follows:
```
set nat nptv6 source
set nat nptv6 destination
```
b) Introduce a new top-level command set nat66：
```
set nat66 source
set nat66 destination
```

Refer to the command structure of H3C, choose to use b scheme

thomas-mangin:
Hi @jack9603301 

Thank you for your work on VyOS. It is much appreciated. As you may have realised the team can sometimes need a little while to review patches.

Reading it, at first glance, the patch deletes a full section of the configuration to replace it. I would therefore have expected a migration script to convert the configuration format on upgrade to not break users' routers.

For this code I believe that you are right to ask @cpo for a review but I would wait for his feedback before jumping on that.

jack9603301:
@thomas-mangin In the 1.3 version before vyos, its NPT function has completely lost its function, and the previous command structure does not distinguish the NPT implementation of SNAT and DNAT. Therefore, the significance of migration may not be significant. To implement configuration migration, I may need to refer to the code to know how to correctly upgrade and configure the migration for vyos. If possible, can you give me a help?

jack9603301:
Starting to consider configuration migration

thomas-mangin:
Unfortunately, I am not an expert in this and will not be able to help you much.

Could we discuss this on phabricator. If something broke during the 1.3 development cycle we should have an issue on phabricator and add some code to test this so we can prevent regression.

Updating VyOS, even during the dev cycle, should never break a running configuration. Many many people run 1.3 in production, something which surprised me when I found it out.

As I read the patch, it changes the syntax, so it will require a migration. Also if you moved the configuration to the root.. 

CPO warned you that this should not be done without good reason and therefore consultation with the team.

CPO also told you that @ signing him was a bad practice but it looks like  you ignored his comment and continued. If you are not taking onboard the feedback provided you may risk to see people disengage and therefore your work to remain stuck.

We will politely ask but when a request has a reason, you may need to read this as a requirement more than a suggestion.

I will try to help you. But please understand that if you do not let us help you, you are unlikely to see your work incorporated, which would be a pitty for both you, us and you everyone using VyOS.

See you on phabricator

jack9603301:
Thank you. I have found the path of the merge script. I will try to solve it by myself and phabricator will discuss it. If there is any problem, move the discussion to phabricator

thomas-mangin:
I feel we are not understanding each other... I am not sure if it is a cultural or language barrier, - I am not a native English speaker - both or perhaps even more ☹️.

In my opinion, there is no point in writing a migrator until the location for the new section is right and the syntax was reviewed and agreed upon. It should be discussed on phabricator.

I really appreciate your enthusiasm and do not want to stop you but please understand that the VyOS team has a way of working and you need to understand it and follow it if you want to contribute.

Did you read https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch ?

If information is missing, we need to update it to help you and future contributors.

jack9603301:
@thomas-mangin thomas-mangin  It may be better to go to phabricator for discussion. I explained the reasons in detail there. I only submitted to my own library to keep pace with my own GitHub repo, not the final draft

https://phabricator.vyos.net/T2518

thomas-mangin:
@jack9603301 Are you on our slack ? 

I take that back, if you are located in China it probably won't work.

jack9603301:
> @jack9603301 Are you on our slack ?
> 
> I take that back, if you are located in China it probably won't work.

I have made it clear that I agree with you and discuss this issue in phabricator. In phabricator, I explain why it needs to be modified and refer to the source of equipment. If you have any questions, you can ask me there.

https://phabricator.vyos.net/T2518

jack9603301:
>  Are you on our slack

@thomas-mangin I would like to say that maybe I can log in to stack. Please send me the link of slack. I have an account of slack workspace in WordPress

jack9603301:
There is a problem with the PR script. Do not merge it for now

jack9603301:
Update PR and pass the test in the local environment, but more tests may be needed. For test cases, see https://phabricator.vyos.net/T2518

jack9603301:
@c-po Note that because the test requirements are not met for the time being (vyos does not use the linux kernel with a kernel version of 5.8+ for the time being), it cannot be tested well. The merged code has not undergone any testing. Please upgrade the vyos linux kernel to 5.8+ , Let me know that I can perform related tests, and merge after testing

jack9603301:
Today, I completed the test of merging `proxy-ndp` into `nat66`, the virtual machine test was basically completed, and the smoke test was modified.

c-po:
Is there a reason that the ndppd configuration is sort of re-assempled in  nat66_conf_to_ndp_proxy_conf? Can this not be done in the Jinja2 template?

jack9603301:
> Is there a reason that the ndppd configuration is sort of re-assempled in nat66_conf_to_ndp_proxy_conf? Can this not be done in the Jinja2 template?

Yes, in order to make configuration more convenient, use Python script for configuration conversion. Otherwise, I don't know how to generate ndppd configuration file by using nat66 configuration tree

jack9603301:
In a few days, I will consider testing this implementation again. After all, it is officially tested on Linux kernel 5.10

c-po:
Please migrate it to Jinja2 - we do not do Python config generation/alteration!

jack9603301:
> Please migrate it to Jinja2 - we do not do Python config generation/alteration!

Let me see, but there is a reason for using Python transformation, because the configuration trees of the two services are completely different, and configuration assembly must be carried out (either Python or jinja2)

Even so, does jinja2 still have to be used?

jack9603301:
PR update, generating ndppd configuration file is normal, but it still needs to be tested

jack9603301:
Many to many mapping supporting DNPT

jack9603301:
GNS test cases can be found at the following address:

https://phabricator.vyos.net/T2518#84736

data/templates/proxy-ndp/ndppd.conf.tmpl
```diff
@@ -0,0 +1,49 @@
...
+{% endif %}
+
+{% for i in global.ndppd_interfaces %}
+
+proxy {{ i }} {
+    router yes
+    timeout 500
+    ttl 30000
+{%  for map in global.ndppd_prefixs %}
+{%     if map['interface'] == i %}
```
c-po: Please use map.interface instead

src/systemd/ndppd.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=NDP Proxy Daemon
+After=network.target vyos-router.service
+ConditionPathExists=/run/ndppd/ndppd.conf
+StartLimitIntervalSec=0
+
+[Service]
+ExecStart=/usr/sbin/ndppd -d -p /var/run/ndppd/ndppd.pid -c /var/run/ndppd/ndppd.conf
+Type=forking
+PIDFile=/var/run/ndppd/ndppd.pid
```
c-po: Please use /run

src/systemd/ndppd.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=NDP Proxy Daemon
+After=network.target vyos-router.service
```
c-po: Please only use vyos-router service

src/conf_mode/nat66.py
```diff
@@ -0,0 +1,170 @@
...
+    return None
+
+def generate(nat):
+    render(iptables_nat_config, 'firewall/nftables-nat66.tmpl', nat, permission=0o755)
+    render(ndppd_config, 'proxy-ndp/ndppd.conf.tmpl', nat, permission=0o755)
+    return None
+
+def apply(nat):
+    cmd(f'{iptables_nat_config}')
+    cmd('systemctl restart ndppd')
```
c-po: Where is ndppd stopped?

c-po (changes requested):

jack9603301:
done

data/templates/proxy-ndp/ndppd.conf.tmpl
jack9603301: done

src/systemd/ndppd.service
jack9603301: done

src/systemd/ndppd.service
jack9603301: done

src/conf_mode/nat66.py
jack9603301: done

src/conf_mode/nat66.py
```diff
@@ -0,0 +1,176 @@
...
+
+def generate(nat):
+    render(iptables_nat_config, 'firewall/nftables-nat66.tmpl', nat, permission=0o755)
+    render(ndppd_config, 'proxy-ndp/ndppd.conf.tmpl', nat, permission=0o755)
+    return None
+
+def apply(nat):
+    if not nat:
+        return None
+    else:
```
c-po, line 155: As you `return` in line 155 there is no need for an else path, please drop the else which saves one level of indention.

src/conf_mode/nat66.py
```diff
@@ -0,0 +1,176 @@
...
+        cmd(f'{iptables_nat_config}')
+        if 'deleted' in nat or not dict_search('source.rule', nat):
+            cmd('systemctl stop ndppd')
```
c-po: Any reason to not remove not ndppd configuration=

src/conf_mode/nat66.py
jack9603301: done

src/conf_mode/nat66.py
jack9603301: done

src/migration-scripts/nat66/0-to-1
```diff
@@ -0,0 +1,76 @@
...
+
+    if config.exists(base + ['translation','prefix']):
+        tmp = config.return_value(base + ['translation','prefix'])
+        config.set(merge_base + ['translation','prefix'],value=tmp)
+
+
+if not config.exists(['nat']):
+    # Nothing to do
+    exit(0)
+else:
```
c-po: Why use an else path when the previous branch calls `exit`?

jack9603301: done

data/templates/proxy-ndp/ndppd.conf.tmpl
```diff
@@ -0,0 +1,48 @@
...
+{%             set global.ndppd_interfaces = global.ndppd_interfaces + [config.outbound_interface] %}
+{%         endif   %}
+{%         if config.translation.prefix is defined  %}
+{%             set global.ndppd_prefixs = global.ndppd_prefixs + [{'interface':config.outbound_interface,'rule':config.translation.prefix}] %}
+{%         endif    %}
+{%     endif    %}
+{%   endfor %}
+{% endif %}
+
+{% for i in global.ndppd_interfaces %}
```
c-po: Please use `interface` as iterator and drop the whitespace line below.

data/templates/proxy-ndp/ndppd.conf.tmpl
```diff
@@ -0,0 +1,48 @@
...
+
+proxy {{ i }} {
+    router yes
+    timeout 500
+    ttl 30000
+{%  for map in global.ndppd_prefixs %}
+{%     if map.interface == i %}
+{%         set p = map.rule  %}
+    rule {{ p }} {
```
c-po: Why not use `map.rule` directly?

c-po:
I wonder, if the NAT66 stuff can not be directly integrated into the `nat source` and `nat destination` tree to have a common script rendering the required nftable rules? This could make things a lot easier, as only the rules must be extended to also support IPv6 addresses.

data/templates/proxy-ndp/ndppd.conf.tmpl
jack9603301: There is a reason for this. `global.ndppd_prefixs` contains every rule, and if is needed to filter the rule interface

jack9603301:
Generally, integrating into NAT tree is not the best way. For configuration, users need to distinguish nat44 from nat66, and there are some differences in configuration, not including nat64 or nat64 that may be supported in the future

If you carefully look at the configuration trees of nat66 and nat, you will find that they are different, and the policy logic is also different. For example, dnat66 does not support port mapping at present (I don't think it is necessary)

I have considered various schemes. According to the H3C command setting case, using this scheme is better than confusing nat66 and nat

data/templates/proxy-ndp/ndppd.conf.tmpl
jack9603301: Oh, I see what you mean

jack9603301:
Associated with https://github.com/vyos/vyatta-cfg-system/pull/137

PR #522, opened 2020-08-06
tcptraceroute: T2723: Some repairs have been made

jack9603301:
- Support VRF
- Add dependency fixed tcptraceroute6
- Improve command line options, support execution without port

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -106,7 +106,8 @@ Depends: python3,
   conntrack,
   libatomic1,
   fastnetmon,
-  libndp-tools
+  libndp-tools,
+  tcptraceroute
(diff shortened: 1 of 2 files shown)


c-po:
Your commit message is well - very not like it should be. YOu combine two non relevant changes into one commit which is bad practice - see https://docs.vyos.io/en/latest/contributing/development.html#writing-good-commit-messages.

Please split to two commits.

c-po:
The TCP port in your implementation is mandatory - but in reality it is optional, please adjust your XML definition to make the port optional.

In addition your help string `Perform operations on TCP ports` should better be renamed to "TCP port to connect to for path tracing".

PR #523, opened 2020-08-06
nat: T2764: increase maximum rule number to 999999

jjakob:


dmbaturin (changes requested):
Do we need any limit at all here? Originally, Vyatta had an insane NAT syntax where rule number determined whether it's a source or destination NAT rule (<5000 was source and >=5000 was destination, or the other way around). EdgeOS is stuck with that syntax AFAIR.

But our rule numbers don't hold any significance, they are just for sorting. Is there a reason one cannot have rule 100200300 if they really like absurdly large numbers?

jjakob:
As far as I can see the only limitation is in the op-mode display code src/op_mode/show_nat_translations.py src/op_mode/show_nat_statistics.py which use fixed column widths. There is also definitely a nftables limitation but I can't find it immediately in the source: https://sources.debian.org/src/nftables/0.9.0-2/

jjakob:
Should we cap it at INT_MAX to be safe? We should probably also then test the script to see if it can handle that number of rules and not crash. (to clarify - talking about the total number of rules, not the rule number length, that only has an influence on the op-mode display script, which needs to be fixed to support this also) I guess if we go so far as to test it, we could find the maximum number of rules via testing. It's likely also limited by the Python script processing. Of course it's very highly unlikely that such a scenario would be reached in real life, so it may be enough to just pick an arbitrarily high number like INT_MAX.

c-po:
I'm fine with any rule number greater or equal then 0. I do not recap there is a INT_MAX in Python like in C, only sys.maxint - but anything greater or equal 0 should be fine.

Regarding the fixed limits in the op-mode scripts - maybe the "fixed part" should be gathered from the maximum rule number and adjust dynamically.

dmbaturin:
Why do we need fixed limits in op mode scripts? They use, or should use, `listNodes` for the completion.

c-po:
@dmbaturin I guess @jjakob is referring to the output column width.

c-po:
@jjakob it would be good if you could provide another PR removing the limit entirely.

PR #536, opened 2020-08-30
config: T2636: get_config_dict() returns a list on multi node by default

jestabro:
Unless no_multi_convert is True, a single valued multi node will be
returned as a list by get_config_dict(). Modification of Thomas Mangin's
version.

### DIFF
--- a/python/vyos/config.py
+++ b/python/vyos/config.py
@@ -67,6 +67,7 @@ import re
 import json
 from copy import deepcopy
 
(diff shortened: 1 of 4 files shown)


jestabro:
@c-po if you want to look before I merge, this is the revised version discussed with @thomas-mangin; multi node value returned as list by default.

c-po:
Please proceed

PR #540, opened 2020-09-04
vyos.client_server: T2854: Add classes for building client/server pairs

bob1de:
The classes in vyos.client_server should make building system daemons and associated clients a lot simpler. They care about all the low-level stuff such as message passing, data validation, sockets, logging, persistent state storage and CLI integration.

There also is vyos.ipsetd. For now, this mainly demonstrates how to use vyos.client_server. However, vyos.ipsetd is fully working, and it can probably be used as part of the firewall rewrite.


bob1de:
Ok, I think it's sufficiently complete for now.

thomas-mangin:
Not my role to do the review, but I like this work.

bob1de:
I had to rebase onto the current branch again because the state at which it originally diverted had a bug which made testing harder, so don't wonder about the extra commits.

bob1de:
Sorry, but the more time I have, the more things I find to fine-tune :).

jestabro:
I will echo Thomas' comment that this is very nice work; I am testing it in some current scenarios, and will update this PR with questions/feedback. Thank you!

dmbaturin:
One note: IPset will be irrelevant when we switch the firewall to nftables. We aren't planning to use IPset in the future.

I'll give the PR a deeper review, I think it's a good idea, but for the long term it will need an example other than IPsec...

bob1de:
@jestabro Thanks!

@dmbaturin 
> One note: IPset will be irrelevant when we switch the firewall to nftables. We aren't planning to use IPset in the future.

Ah, good to know. Never really "worked" with nftables so far. My original intention for this PR was [T2719](https://phabricator.vyos.net/T2719) which I already started fiddling with, so once that's implemented, it could serve as an example for using `vyos.client_server`. However, do you think we may keep `vyos.ipsetd` as reference for now?

jestabro (commented):
This was considered and discussed at the time. It is very nice work, and I tested it for performance and overhead, which revealed it comparable to, say, ZMQ based daemons. However, at the time it was seen as more than needed for our specific use cases: (1) the number of daemons that we will need (including the planned op-mode daemon, which inspired this work) is few, hence may not warrant a general framework; (2) the specific needs of a given VyOS daemon suggest a targeted, custom design, again not needing a general framework. That being said, with the new design of standardized op-mode scripts inviting us to revisit the design of an op-mode daemon, why should keep this good work in mind,

jestabro:
As mentioned in the previous comment, we will not make use of this as is, but will remain informed by the ideas contained.

PR #551, opened 2020-09-25
T2926: Missing import

kroy-the-rabbit:


dmbaturin (changes requested):
I think we shoud better use fully qualified names in these cases. Those module paths aren't that long, and it's immediately obvious where a function is from.

kroy-the-rabbit:
@dmbaturin So on a one-off like this, instead of importing it, just doing a `vyos.chmod_775` where it's called?

c-po:
But then we should change this everywhere to be consistent. Merging as it matches the current codebase, lets see and change later on.

PR #561, opened 2020-10-03
openvpn: T2957: show openvpn not returning anything

marcushoff:
Fix location of status files directory


c-po:
Thank you for the PR, could you please exclude .github/pull.yml from this PR?

We also try to keep a common style on our commit messages (https://docs.vyos.io/en/latest/contributing/development.html#prepare-patch-commit) - it would be nice if you could change your commit message, else I‘d do it when squashing it, but the pull.yml should still not be part of the PR.

Happy to merge afterwards.

marcushoff:
Is that better?

PR #564, opened 2020-10-06
Update search domain error message to match new validation logic

lucasec:
This should match the newer, stricter validation logic that was put in place.

I also converted the file to unix line endings. If we've already long given up hope on having consistent line endings, happy to remove that part in favor of diff readability.


dmbaturin:
Regarding line endings, we should probably add a linter script to detect inconsistencies. We need a linter script for interface definition files in general.

PR #567, opened 2020-10-13
pppoe-server: T2976: Add missing * to chap-secret file

DmitriyEshenko:


data/templates/accel-ppp/chap-secrets.config_dict.tmpl
```diff
@@ -2,10 +2,15 @@
 {% if authentication is defined and authentication.local_users is defined and authentication.local_users.username is defined %}
 {%   for user, user_config in authentication.local_users.username.items() %}
 {%     if user_config.disabled is not defined %}
+{%       if user_config.static_ip %}
```
c-po, line 5: Fix must be part of `configdict.py` which loads the defaults from the XML <defaultValue> node - this change would only scatter default values accross the codebase.

c-po (changes requested):

PR #569, opened 2020-10-13
Bridge's Priority: T2924: Bridge interface's commit priority is fixed for reboot

srividya0208:
This change is made in the "interfaces-bridge.xml.in" script where the priority is increased from 489 to 320 so that the bridge interface address could be added before the policy route-map(whose priority is 470) in a single commit.
Before the change was added, using bridge for 'set src' in a route-map invalidates it as part of a subsequent reboot and removes from the running configuration
However, higher priority has completely fixed for the ipv4 address whereas double commit for ipv6 works as a workaround.

```
vyos@vyos# set interfaces bridge br0 address '2001:db8::dead:beef/64'
[edit]
vyos@vyos# set interfaces bridge br0 member interface eth0.15
[edit]
vyos@vyos# set interfaces ethernet eth0 vif 15
[edit]
vyos@vyos# set policy route-map bgp-in rule 10 action 'permit'
[edit]
vyos@vyos# set policy route-map bgp-in rule 10 set src '2001:db8::dead:beef'
[edit]
vyos@vyos# commit
[ policy route-map bgp-in rule 10 set src 2001:db8::dead:beef ]
% not a local address

[[policy route-map bgp-in]] failed
Commit failed
[edit]
vyos@vyos# commit
[edit]

vyos@vyos:~$ sh conf comm | grep bgp-in
set policy route-map bgp-in rule 10 action 'permit'
set policy route-map bgp-in rule 10 set src '2001:db8::dead:beef'
```

### DIFF
--- a/interface-definitions/interfaces-bridge.xml.in
+++ b/interface-definitions/interfaces-bridge.xml.in
@@ -5,7 +5,7 @@
       <tagNode name="bridge" owner="${vyos_conf_scripts_dir}/interfaces-bridge.py">
         <properties>
           <help>Bridge Interface</help>
-          <priority>489</priority>
+          <priority>320</priority>
           <constraint>
             <regex>^br[0-9]+$</regex>
           </constraint>


c-po:
Why not use priority 469? It's still high enough to accommodate the route-map but also keep the interfaces which can be added to a bridge below the bridge interface.

Using priority 320 will make e.g. geneve and vxlan interfaces fail as their priority is higher then 320.

srividya0208:
Tested for vxlan interface and it fails when when the priority is set to 320.
Tried the suggested value **469**, but fails with this error:

```
[   40.332989] vyos-router[638]: Starting VyOS router: migrate rl-system firewall configure failed!
[   40.902497] vyos-config[700]: Configuration error
```

Tried couple of more priority numbers for bridge i.e 461,460 and 471 (where roupe-map priority was 472), still receiving the error. I will test for few more values.

c-po:
Instead of randomly shifting priorities, you should probably boot using the vyos-config-debug flag (https://docs.vyos.io/en/latest/contributing/debugging.html#system-startup) and the see the logfile why the bootup failed. Its better to determine the root cause the masking an issue - it else will pop up again at any point in time.

c-po:
Will be handled by https://phabricator.vyos.net/T2985

PR #585, opened 2020-10-27
mpls-conf: T915: Separate IPv4 and IPv6 hello timers, add IPv6 timers

Cheeze-It:
The commit has to do with separating the hello/hold timers from being only IPv4 to being both IPv4 and IPv6.

I renamed the existing hello and hold timers with an "-ipv4" and added ones that were "-ipv6". I did verify that the commands properly commit under FRR as well. I also added some room on the protocols_mpls.py file for the different variables as it seems we're might end up having longer names. Removed some spaces that I found too that weren't needed on ldpd.frr.tmpl as well.

### DIFF
--- a/data/templates/frr/ldpd.frr.tmpl
+++ b/data/templates/frr/ldpd.frr.tmpl
@@ -13,6 +13,7 @@ no neighbor {{neighbor_id}} password {{old_ldp.neighbors[neighbor_id].password}}
 {% for neighbor_id in ldp.neighbors -%}
 neighbor {{neighbor_id}} password {{ldp.neighbors[neighbor_id].password}}
 {% endfor -%}
+!
 address-family ipv4
 label local allocate host-routes
 {% if old_ldp.export_ipv4_exp -%}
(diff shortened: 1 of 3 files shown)


c-po:
Hi @Cheeze-It,

thanks for the contribution. Unfortunately there are some points which must be changed first. Splitting IPv4 and IPv6 makes sense but they still should be placed inside the same CLI node as the user must not care about the internal structure.

Other services like e.g. "dns forwarding" which uses PowerDNS recursor under the hood also has such a limitation for the "source-address" CLI node. You can overcome this by internally splitting the returned CLI values into IPv4 and IPv6 dictionary keys. See https://github.com/vyos/vyos-1x/blob/9cf806a8b60af3b9ce5f1d8141099879c8faef69/src/conf_mode/dns_forwarding.py#L75-L82 for the implementation.

On the other hand simply "renaming" a node in the CLI will break any succedding upgrade of VyOS. When CLI nodes are renamed, a so called migration script (https://github.com/vyos/vyos-1x/tree/9cf806a8b60af3b9ce5f1d8141099879c8faef69/src/migration-scripts) must be programmed which transforms the old CLI node to the new CLI format.

Cheeze-It:
Hello @c-po, 

Hmm, I am unsure of how to do that in regards to the hello/hold timers for IPv4 and IPv6. This unfortunately is something that I am unsure how to do. I can see two options.

We can keep them separate (as is right now, separate between IPv4 and IPv6....which is preferable but yes, it does change the configuration). 

Or we can make it to where we revert the commands and just have one hello/hold timer but it will affect both IPv4 and IPv6 unilaterally. We can make the command be enacted in address family IPv4 and IPv6 at the same time. It was only affecting IPv4 before.

c-po:
Sorry, in my first comment I was not fully up to speed.

I now wonder why the CLI is not a bit more like FRR and also our BGP implementation which uses address-families for those types of configuration items, but that is a separate thing to discuss and maybe re-align with a migrator.

We can keep it in two separate configurations but still you will require a migrator as you have renamed `hello-interval` to `hello-ipv4-interval` and `hello-holdtime` to `hello-ipv4-holdtime`. This will break all current configurations out there.

As MPLS and LDP is still a new beta feature it's up to @dmbaturin and @jestabro to decide if a migrator should be added or if it will be okay to break the current config.

Cheeze-It:
@c-po,

No worries :)

Please don't think I'm trying to push one "type" of CLI over another. My mindset when it comes to creating a CLI is to allow all options to be independent from each other and to offer as much power to the user as possible. I will admit, in my lack of experience I didn't think about breaking current configurations until you mentioned it. Although on one hand I can see why it was done the way it was done, but on the other hand it is still in development.

Since CLI hierarchies are "best effort" then sometimes we will program ourselves into a corner, and I think in this case it happened albeit unintentionally. Should we try to rip the band aid off now? or should we just make an assumption and aggregate the hello timers? Either would *probably* be reasonable, but they both have caveats.

dmbaturin (approved):
Since we've never released an LTS version with any MPLS CLI, or any _named_ version with it at all, I believe it's fine to break compatibility with older rolling release images.

PR #586, opened 2020-10-29
T3029 Correct generated nginx content

cjeanneret:
The redirection was using the wrong variable ($server_name),
making the browser going to https://_ instead of the right
variable.


jestabro:
Thanks, @cjeanneret

PR #597, opened 2020-11-08
mpls-conf: T915: Add targeted LDP neighbors with parameters

Cheeze-It:
The commit has to do with the addition of targeted LDP neighbors and parameters being added. FRR allows for this functionality and I just wanted to add it.

We have basically 4 options that are added. Enabling targeted LDP functionality, the targeted neighbor, the hello interval of targeted sessions, and the hold time of targeted sessions.  Both IPv4 and IPv6 has been coded in.


dmbaturin (approved):
Looks good to me.

PR #599, opened 2020-11-11
l2tpv3: T3059: Fix interface state which is permanently down

ernstjo:
I think that is a bug. Because interface is down at create and when you update interface configuration. But you have no function to enable interface permanently. If we change it to default state "up" everyone can disable interface permanently via shutdown/disable command. Without this change you have to use sudo ip link set {ifname} up to enable interface. But after change or reboot you have to down it again. Makes no sense for me because you have the option to disable interface or use boot script to disable but no way to enable it.

### DIFF
--- a/python/vyos/ifconfig/l2tpv3.py
+++ b/python/vyos/ifconfig/l2tpv3.py
@@ -68,8 +68,9 @@ class L2TPv3If(Interface):
         cmd += ' peer_session_id {peer_session_id}'
         self._cmd(cmd.format(**self.config))
 
-        # interface is always A/D down. It needs to be enabled explicitly
-        self.set_admin_state('down')
+        # No need for interface shut down. There exist no function to permanently enable tunnel.
+        # But you can disable interface permanently with shutdown/disable command.
+        self.set_admin_state('up')
 
     def remove(self):
         """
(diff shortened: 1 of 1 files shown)


DmitriyEshenko:
Hello @ernstjo, Thanks for contributing.
Can you create a bug report at first on https://phabricator.vyos.net/
I believe this article can explain a bit more why this needs https://docs.vyos.io/en/latest/contributing/development.html#prepare-patch-commit

ernstjo:
Hi Dmitriy,

bug report has been created.

DmitriyEshenko:
Thanks. Try to change the comment on your pull request with `git commit --amend`
Set comment `l2tpv3: T3059: Fix interface state which is permanently down`

c-po:
Thanks for the contribution. This is indeed a bug but the fix is improper.

The fix must be in `class L2TPv3If` - ofcourse - but rather this lacs the implementation of the `update()` method as found e.g. in https://github.com/vyos/vyos-1x/blob/current/python/vyos/ifconfig/dummy.py#L37-L54.

Can you please change your PR to implement the fix using the `update()` function I linked to?

ernstjo:
I'll have to check that. My fix work on a rolling installation. I have to check if i can adapt update() methode.

PR #604, opened 2020-11-13
bridge: T3042: Better fix implementation errors

jack9603301:
In #601, I provided a basic patch. Under this patch, I rely on vif to
detect the vlan id range that the bridge should flow through,
which may lead to greater redundancy in the configuration,
 so I am considering detecting effective vlan filters In setting the range
 of vlan id that is required to flow through the bridge,
I use `set()` to complete the deduplication of this vlan id
and set it to the bridge uniformly (at the same time,
 I slightly modified the smoke test script)

### DIFF
--- a/python/vyos/ifconfig/bridge.py
+++ b/python/vyos/ifconfig/bridge.py
@@ -255,6 +255,9 @@ class BridgeIf(Interface):
             if member in interfaces():
                 self.del_port(member)
         vlan_filter = 0
+        
+        vlan_del = set()
+        vlan_add = set()
 
         tmp = dict_search('member.interface', config)
         if tmp:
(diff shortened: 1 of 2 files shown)


c-po:
commit id subject should be prefixed with `bridge` and not `fix`.

jack9603301:
@c-po Sorry, I did not pay attention, I have already repaired the message

c-po:
The entire bridge code was able to run and not have any exception handling. You now introduce exception handling but just pass it on which my gut feeling tells me is wrong.

jack9603301:
@c-po The processing is very simple, but it is necessary to collect the set of vlanid numbers of congestion settings through 'set()', and then set them all to the bridge itself. The following command must be executed:
```
bridge vlan add dev br1 vid {vid} self
```

Otherwise, the bridge can not actually forward VLAN data. Because the test environment is not complete, this problem is not discovered until runtime. Although a patch has been merged, the previous patch is handled according to Vif, and there will be configuration redundancy (and it is not necessary to configure Vif for each VLAN)

Most of the processing can be effectively judged by `if-else`. If the command is correct, the exception mechanism should not be used. However, during the VLAN deletion process of the bridge itself, I added exception capture just in case, because the deletion of `VLAN 1` may fail (in fact, I can test several times to confirm whether the exception handling can be deleted)

jack9603301:
I removed the exception handling in the vlan deletion part of the bridge, and instead used `if-else`

PR #613, opened 2020-11-17
bridge:T3067: Fix VLAN aware setting failure under WLAN

jack9603301:
In the implementation of T3042, it will cause two problems:
    
1. Even if VLAN awareness is not enabled, the VLAN settings of the
       vlan filter will be modified. When the bridge member has a WLAN interface,
       the error is exposed, so repair it here. You should not modify the
      related settings when the VLAN awareness mode is not enabled
    
2. Even if VLAN awareness is not enabled, the VLAN settings of the
       vlan filter will be modified. When the bridge member has a WLAN interface,
        due to special settings, the bridge mode cannot be entered and the settings
       cannot be completed directly. Therefore, the WLAN interface should be rejected
       Enter the bridge with VLAN awareness

### DIFF
--- a/python/vyos/ifconfig/bridge.py
+++ b/python/vyos/ifconfig/bridge.py
@@ -309,15 +309,12 @@ class BridgeIf(Interface):
                     vlan_filter = 1
                     cmd = f'bridge vlan del dev {interface} vid 1'
                     self._cmd(cmd)
-                    vlan_del.add(1)
                     vlan_id = interface_config['native_vlan']
+                    if vlan_id != 1:
+                        vlan_del.add(1)
                     cmd = f'bridge vlan add dev {interface} vid {vlan_id} pvid untagged master'
                     self._cmd(cmd)
                     vlan_add.add(vlan_id)
-                else:
-                    cmd = f'bridge vlan del dev {interface} vid 1'
-                    self._cmd(cmd)
-                    vlan_del.add(1)
                 
                 if 'allowed_vlan' in interface_config:
                     vlan_filter = 1
@@ -325,6 +322,11 @@ class BridgeIf(Interface):
                         cmd = f'bridge vlan add dev {interface} vid {vlan} master'
                         self._cmd(cmd)
                         vlan_add.add(vlan)
+                
+                if vlan_filter:
+                    if 'native_vlan' not in interface_config:
+                        cmd = f'bridge vlan del dev {interface} vid 1'
+                        self._cmd(cmd)
         
         
         for vlan in vlan_del:
--- a/src/conf_mode/interfaces-bridge.py
+++ b/src/conf_mode/interfaces-bridge.py
@@ -123,11 +123,11 @@ def get_config(config=None):
             
             # VLAN-aware bridge members must not have VLAN interface configuration
             if 'native_vlan' in interface_config:
-                if 'disable' not in interface_config['native_vlan']:
-                    vlan_aware = True
+                vlan_aware = True
             
             if 'allowed_vlan' in interface_config:
                 vlan_aware = True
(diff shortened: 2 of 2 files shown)


dmbaturin (approved):

c-po:
Hi @jack9603301 you have two commits here both with the same headline and different changes.

Please be more precise when descibing your change in the individual commit messages. Proper commit messages help us in the future identifying the reason why code was added/removed.

c-po:
Good to go once the commit messages are properly rephrased.

```
cpo@LR4.wue3# brctl show
bridge name     bridge id               STP enabled     interfaces
br0             8000.000db95307ee       no              eth2
                                                        eth3
                                                        wlan0
```

c-po:
Okay, this was to fast unfortunately. As you have removed my code to not add wlan interfaces in `bridge.py` this will now trigger an exception on system bootup - looks like this was not tested.

```
  File "/usr/libexec/vyos/conf_mode/interfaces-wireless.py", line 288, in <module>
    apply(c)
  File "/usr/libexec/vyos/conf_mode/interfaces-wireless.py", line 268, in apply
    w.update(wifi)
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/wireless.py", line 74, in update
    super().update(config)
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/interface.py", line 1088, in update
    self.add_to_bridge(bridge_dict)
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/interface.py", line 858, in add_to_bridge
    Section.klass(bridge)(bridge, create=True).add_port(self.ifname)
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/bridge.py", line 196, in add_port
    return self.set_interface('add_port', interface)
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/control.py", line 184, in set_interface
    return self._set_command(self.config, name, value)
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/control.py", line 109, in _set_command
    return self._command_set[name].get('format', lambda _: _)(self._cmd(cmd))
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/control.py", line 51, in _cmd
    return cmd(command, self.debug)
  File "/usr/lib/python3/dist-packages/vyos/util.py", line 179, in cmd
    raise OSError(code, feedback)
FileNotFoundError: [Errno 2] failed to run command: ip link set dev wlan0 master br0
returned:
exit code: 2
```

jack9603301:
I will deal with your message in a few hours `bridge.py` Adding a wireless port is bound to be a bit troublesome, which means that VLAN filter cannot be set, and VLAN sensing cannot be set directly when there are WLAN ports

Why is it normal in manual configuration?

c-po:
Easiest thing would be to not allow wlan member interfaces in vlan mode, check this in verify()

jack9603301:
I will try to modify PR in a few hours. If VLAN aware is not supported, I will take your advice (VLAN aware is not supported for WLAN interface). I seem to have got some clues in your error information. I'll take a test first.

Thanks

jack9603301:
@c-po I banned the WLAN port bridge from entering the sensing bridge mode

python/vyos/ifconfig/bridge.py
```diff
@@ -287,6 +287,9 @@ def update(self, config):
                     value = interface_config.get('priority')
                     lower.set_path_priority(value)
                 
+                if 'wlan' in interface:
```
c-po: Why not leave the code as it was before? This feels like an unnecessary change which probably does more harm then good.

c-po (changes requested):

PR #618, opened 2020-11-25
pim: igmp: igmp-proxy: T2744: Add check to prevent pimd and igmp-prox…

DmitriyEshenko:
…y conflict

### DIFF
--- a/src/conf_mode/protocols_igmp.py
+++ b/src/conf_mode/protocols_igmp.py
@@ -21,8 +21,9 @@ from sys import exit
 
 from vyos import ConfigError
 from vyos.config import Config
-from vyos.util import call
+from vyos.util import call, process_named_running
 from vyos.template import render
+from signal import SIGTERM
 
 from vyos import airbag
 airbag.enable()
@@ -36,11 +37,19 @@ def get_config(config=None):
         conf = Config()
     igmp_conf = {
         'igmp_conf' : False,
+        'pim_conf'  : False,
+        'igmp_proxy_conf' : False,
         'old_ifaces'    : {},
         'ifaces'    : {}
     }
     if not (conf.exists('protocols igmp') or conf.exists_effective('protocols igmp')):
         return None
+
+    if conf.exists('protocols igmp-proxy'):
+        igmp_conf['igmp_proxy_conf'] = True
+
+    if conf.exists('protocols pim'):
+        igmp_conf['pim_conf'] = True
 
     if conf.exists('protocols igmp'):
         igmp_conf['igmp_conf'] = True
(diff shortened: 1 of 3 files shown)


src/conf_mode/protocols_igmp.py
```diff
@@ -79,6 +88,10 @@ def verify(igmp):
         return None
 
     if igmp['igmp_conf']:
+        # Check conflict with IGMP-Proxy
+        if igmp['igmp_proxy_conf']:
+            raise ConfigError(f"Does not possible to configure PIM/IGMP and IGMP-Proxy together!")
```
dmbaturin: This is ungrammatical. I would use "IGMP proxy and PIM cannot be both configured at the same time".

src/conf_mode/protocols_igmp.py
```diff
@@ -99,9 +112,16 @@ def apply(igmp):
     if igmp is None:
         return None
 
-    if os.path.exists(config_file):
-        call(f'vtysh -d pimd -f {config_file}')
-        os.remove(config_file)
+    pim_pid = process_named_running('pimd')
+    if igmp['igmp_conf'] or igmp['pim_conf']:
+        if not pim_pid:
+            call(f'/usr/lib/frr/pimd -d -F traditional --daemon -A 127.0.0.1')
```
dmbaturin: Who knows if the path changes in the future. Let's just use `pimd -d ...`.

src/conf_mode/protocols_pim.py
```diff
@@ -92,6 +101,10 @@ def verify(pim):
         return None
 
     if pim['pim_conf']:
+        # Check conflict with IGMP-Proxy
+        if pim['igmp_proxy_conf']:
+            raise ConfigError(f"Does not possible to configure PIM/IGMP and IGMP-Proxy together!")
```
dmbaturin: Grammar comment above applies here too.

src/conf_mode/protocols_pim.py
```diff
@@ -126,9 +139,16 @@ def apply(pim):
     if pim is None:
         return None
 
-    if os.path.exists(config_file):
-        call("vtysh -d pimd -f " + config_file)
-        os.remove(config_file)
+    pim_pid = process_named_running('pimd')
+    if pim['igmp_conf'] or pim['pim_conf']:
+        if not pim_pid:
+            call(f'/usr/lib/frr/pimd -d -F traditional --daemon -A 127.0.0.1')
```
dmbaturin: Same hardcoded path.

dmbaturin (changes requested):
The logic goods good to me, other than these to nitpicks.

PR #619, opened 2020-11-26
mpls-conf: T915: Add ethernet vif MPLS enable

Cheeze-It:
In this commit I did some work to try to enable ethernet sub interface MPLS enablement. Per request by @bbs2web, I think I got it to work.

I am unsure of how to make this look any prettier. I really don't know how to. But this *should* work as I gave it a test and everything seemed to come out proper.

We can refactor this later but I don't know how to do it for now.

### DIFF
--- a/src/conf_mode/protocols_mpls.py
+++ b/src/conf_mode/protocols_mpls.py
@@ -121,19 +121,31 @@ def apply(mpls):
         system_interfaces = []
         system_interfaces.append(((os.popen('sysctl net.mpls.conf').read()).split('\n')))
         del system_interfaces[0][-1]
-        for configured_interface in mpls['interface']:
-            for system_interface in system_interfaces[0]:
-                if configured_interface in system_interface:
-                    call(f'sysctl -wq net.mpls.conf.{configured_interface}.input=1')
-                elif system_interface.endswith(' = 1'):
-                    system_interface = system_interface.replace(' = 1', '=0')
-                    call(f'sysctl -wq {system_interface}')
+        system_interfaces = system_interfaces[0]
+        for system_interface in system_interfaces:
+            if system_interface.endswith(' = 1'):
+                replaced_once_system_interface = system_interface.replace('net.mpls.conf.', '')
+                replaced_twice_system_interface = replaced_once_system_interface.replace('.input = 1', '')
+                parsed_system_interface = replaced_twice_system_interface.replace('/', '.')
+                if parsed_system_interface in mpls['interface']:
+                    pass
+                else:
+                    call(f'sysctl -wq net.mpls.conf.{replaced_twice_system_interface}.input=0')
+            elif system_interface.endswith(' = 0'):
+                replaced_once_system_interface = system_interface.replace('net.mpls.conf.', '')
+                replaced_twice_system_interface = replaced_once_system_interface.replace('.input = 0', '')
+                parsed_system_interface = replaced_twice_system_interface.replace('/', '.')
+                if parsed_system_interface in mpls['interface']:
+                    call(f'sysctl -wq net.mpls.conf.{replaced_twice_system_interface}.input=1')
+                else:
+                    call(f'sysctl -wq net.mpls.conf.{replaced_twice_system_interface}.input=0')
     else:
         # If MPLS interfaces are not configured, set MPLS processing disabled
         system_interfaces = []
         system_interfaces.append(((os.popen('sysctl net.mpls.conf').read()).replace(" = 1", "=0")).split('\n'))
         del system_interfaces[0][-1]
-        for interface in (system_interfaces[0]):
+        system_interfaces = system_interfaces[0]
+        for interface in system_interfaces:
             call(f'sysctl -wq {interface}')
 
     return None
--- a/src/conf_mode/protocols_mpls.py
+++ b/src/conf_mode/protocols_mpls.py
@@ -123,10 +123,10 @@ def apply(mpls):
         system_interfaces = []
         # Populate system interfaces list with local MPLS capable interfaces
         for interface in glob('/proc/sys/net/mpls/conf/*'):
-            system_interfaces.append(interface.replace('/proc/sys/net/mpls/conf/', ''))    
+            system_interfaces.append(os.path.basename(interface))   
         # This is where the comparison is done on if an interface needs to be enabled/disabled.
         for system_interface in system_interfaces:
-            interface_state = read_file(f'/proc/sys/net/mpls/conf/%s/input' %(system_interface))
+            interface_state = read_file(f'/proc/sys/net/mpls/conf/{system_interface}/input')
             if '1' in interface_state:
                 if system_interface not in mpls['interface']:
                     system_interface = system_interface.replace('.', '/')
@@ -139,7 +139,7 @@ def apply(mpls):
         system_interfaces = []
         # If MPLS interfaces are not configured, set MPLS processing disabled
         for interface in glob('/proc/sys/net/mpls/conf/*'):
-            system_interfaces.append(interface.replace('/proc/sys/net/mpls/conf/', '')) 
+            system_interfaces.append(os.path.basename(interface)) 
         for system_interface in system_interfaces:
             system_interface = system_interface.replace('.', '/')
             call(f'sysctl -wq net.mpls.conf.{system_interface}.input=0')
--- a/src/conf_mode/protocols_mpls.py
+++ b/src/conf_mode/protocols_mpls.py
@@ -18,11 +18,13 @@ import os
 
 from sys import exit
 
+from glob import glob
 from vyos.config import Config
 from vyos.configdict import node_changed
 from vyos.template import render_to_string
 from vyos.util import call
 from vyos.util import dict_search
+from vyos.util import read_file
 from vyos import ConfigError
 from vyos import frr
 from vyos import airbag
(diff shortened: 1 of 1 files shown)


src/conf_mode/protocols_mpls.py
```diff
@@ -121,19 +121,31 @@ def apply(mpls):
         system_interfaces = []
         system_interfaces.append(((os.popen('sysctl net.mpls.conf').read()).split('\n')))
```
c-po: Please use either `vyos.util.read_file()` or walk the folder with Python hear:

```
>>> import os
>>> from glob import glob
>>> for tmp in glob('/proc/sys/net/mpls/conf/*'):
...   os.path.basename(tmp)
...
'dum0'
'eth0'
'eth0.201'
'eth0.202'
'eth1'
'eth2'
'ip6gre0'
'lo'
```

src/conf_mode/protocols_mpls.py
```diff
@@ -121,19 +121,31 @@ def apply(mpls):
...
         del system_interfaces[0][-1]
-        for configured_interface in mpls['interface']:
-            for system_interface in system_interfaces[0]:
-                if configured_interface in system_interface:
-                    call(f'sysctl -wq net.mpls.conf.{configured_interface}.input=1')
-                elif system_interface.endswith(' = 1'):
-                    system_interface = system_interface.replace(' = 1', '=0')
-                    call(f'sysctl -wq {system_interface}')
+        system_interfaces = system_interfaces[0]
```
c-po: Why not merge line 124 and 125?

```
for system_interface in system_interfaces[0]:
```

src/conf_mode/protocols_mpls.py
```diff
@@ -121,19 +121,31 @@ def apply(mpls):
...
+        for system_interface in system_interfaces:
+            if system_interface.endswith(' = 1'):
+                replaced_once_system_interface = system_interface.replace('net.mpls.conf.', '')
+                replaced_twice_system_interface = replaced_once_system_interface.replace('.input = 1', '')
+                parsed_system_interface = replaced_twice_system_interface.replace('/', '.')
+                if parsed_system_interface in mpls['interface']:
```
c-po: Please use a `not` condition instead of `pass` - makes it easier to read.

```
if parsed_system_interface not in mpls['interface']:
    call(f'sysctl -wq net.mpls.conf.{replaced_twice_system_interface}.input=0')
```

src/conf_mode/protocols_mpls.py
```diff
@@ -121,19 +121,31 @@ def apply(mpls):
...
+                    pass
+                else:
+                    call(f'sysctl -wq net.mpls.conf.{replaced_twice_system_interface}.input=0')
+            elif system_interface.endswith(' = 0'):
+                replaced_once_system_interface = system_interface.replace('net.mpls.conf.', '')
+                replaced_twice_system_interface = replaced_once_system_interface.replace('.input = 0', '')
+                parsed_system_interface = replaced_twice_system_interface.replace('/', '.')
+                if parsed_system_interface in mpls['interface']:
```
c-po: Please make this more future proof by writing:

```
input = '0'
if parsed_system_interface in mpls['interface']:
    input = '1'
call(f'sysctl -wq net.mpls.conf.{replaced_twice_system_interface}.input={input}')
```

Thus it is a) more readable and b) if we need to refactore this we have a 50% lesser chance in breaking things by bad copy/pasting 😄

src/conf_mode/protocols_mpls.py
```diff
@@ -121,19 +121,31 @@ def apply(mpls):
...
+                    call(f'sysctl -wq net.mpls.conf.{replaced_twice_system_interface}.input=1')
+                else:
+                    call(f'sysctl -wq net.mpls.conf.{replaced_twice_system_interface}.input=0')
     else:
         # If MPLS interfaces are not configured, set MPLS processing disabled
         system_interfaces = []
         system_interfaces.append(((os.popen('sysctl net.mpls.conf').read()).replace(" = 1", "=0")).split('\n'))
         del system_interfaces[0][-1]
-        for interface in (system_interfaces[0]):
+        system_interfaces = system_interfaces[0]
```
c-po: Why not use `for system_interface in system_interfaces[0]:` - see comment above

c-po (changes requested):

src/conf_mode/protocols_mpls.py
Cheeze-It: I can do so, I just don't like referencing the element. If you'd prefer "for system_interface in system_interfaces[0]:" then I absolutely will defer to that.

src/conf_mode/protocols_mpls.py
Cheeze-It: So, while I want to say thank you for this info here as it's very useful...it doesn't give me what I am trying to do.

So the first 4 lines I have give this:

>>> system_interfaces = []
>>> system_interfaces.append(((os.popen('sysctl net.mpls.conf').read()).split('\n')))
>>> del system_interfaces[0][-1]
>>> system_interfaces = system_interfaces[0]
>>>
>>>
>>> print (system_interfaces)
['net.mpls.conf.eth0.input = 0', 'net.mpls.conf.eth0/10.input = 0', 'net.mpls.conf.eth0/11.input = 0', 'net.mpls.conf.eth0/12.input = 0', 'net.mpls.conf.eth0/13.input = 0', 'net.mpls.conf.eth0/14.input = 0', 'net.mpls.conf.eth0/15.input = 0', 'net.mpls.conf.lo.input = 0']


What is super useful here is, it tells you the interface name and the state that it is in. When using glob('/proc/sys/net/mpls/conf/*'), I only get an actual raw dump of interfaces that looks exactly like you have above. When I did it, I got it too:

>>> system_interfaces = []
>>> for interface in glob('/proc/sys/net/mpls/conf/*'):
...     system_interfaces.append(interface.replace('/proc/sys/net/mpls/conf/', ''))
...
>>>
>>> print (system_interfaces)
['eth0', 'eth0.10', 'eth0.11', 'eth0.12', 'eth0.13', 'eth0.14', 'eth0.15', 'lo']

So how should I go about verifying the actual state the interface is in and then making a decision on enabling/disabling?

The reason I ask is because, I would only enable/disable the interfaces that need to have their configuration status changed. I don't want to set all interfaces every time an interface commit happens. So....I am unsure how to proceed.

src/conf_mode/protocols_mpls.py
Cheeze-It: Ahh, sure. I can make this change. Thank you :)

src/conf_mode/protocols_mpls.py
Cheeze-It: Sure, absolutely. Will make this change. Thank you :)

src/conf_mode/protocols_mpls.py
Cheeze-It: Will change over to format you prefer.

src/conf_mode/protocols_mpls.py
c-po: If you wan't the get the current state of all interfaces then please use the `vyos.util.read_file()` wrapper instead of your own `popen()` implementation.

src/conf_mode/protocols_mpls.py
c-po: Well "under the hood" it does the same - it only spares one line of code which makes it simpler to read.

Cheeze-It: Yeah, and that's fine. It does save a line. I just didn't know which was preferred. Either works for me. Thank you :)

Cheeze-It:
Let me know what you think C-Po. Here's hoping this one looks better :)

src/conf_mode/protocols_mpls.py
```diff
@@ -119,22 +121,28 @@ def apply(mpls):
...
-                    call(f'sysctl -wq net.mpls.conf.{configured_interface}.input=1')
-                elif system_interface.endswith(' = 1'):
-                    system_interface = system_interface.replace(' = 1', '=0')
-                    call(f'sysctl -wq {system_interface}')
+        # Populate system interfaces list with local MPLS capable interfaces
+        for interface in glob('/proc/sys/net/mpls/conf/*'):
+            system_interfaces.append(interface.replace('/proc/sys/net/mpls/conf/', ''))    
+        # This is where the comparison is done on if an interface needs to be enabled/disabled.
+        for system_interface in system_interfaces:
+            interface_state = read_file(f'/proc/sys/net/mpls/conf/%s/input' %(system_interface))
```
c-po: Tis is a mix of Python2 and Python3 syntax. Please stick to Python3

src/conf_mode/protocols_mpls.py
```diff
@@ -119,22 +121,28 @@ def apply(mpls):
...
-        for configured_interface in mpls['interface']:
-            for system_interface in system_interfaces[0]:
-                if configured_interface in system_interface:
-                    call(f'sysctl -wq net.mpls.conf.{configured_interface}.input=1')
-                elif system_interface.endswith(' = 1'):
-                    system_interface = system_interface.replace(' = 1', '=0')
-                    call(f'sysctl -wq {system_interface}')
+        # Populate system interfaces list with local MPLS capable interfaces
+        for interface in glob('/proc/sys/net/mpls/conf/*'):
+            system_interfaces.append(interface.replace('/proc/sys/net/mpls/conf/', ''))
```
c-po: Please use the basename optim in one of my earlier comments over replace, it was build for this exact purpose.

c-po (changes requested):

Cheeze-It:
Ugh, yeah that was my fault. I didn't notice that I left the Python2 and Python3 syntax in. I thought I removed it, good catch.

I also changed to use the "os.path.basename" option as well. It seems to work the same. Thankfully.

c-po:
Can you please either squash the commits on your side or update the commit messages according to the guidelines at https://docs.vyos.io/en/latest/contributing/development.html#writing-good-commit-messages

Cheeze-It:
I'll re-issue another PR. I completely messed this one up. My bad :(

PR #638, opened 2020-12-08
interfaces: T3114: Fix VLAN-aware bridge setting failure

jack9603301:
## Change Summary

Fix VLAN-aware bridge setting failure

## Types of changes

 
## Related Task(s)

https://phabricator.vyos.net/T3114

## Component(s) name

bridge

## Proposed changes

Due to dependency issues, if a non-existent interface appears when creating a VLAN-aware bridge, the VLAN-aware parameters of this interface cannot be set. Therefore, the interface will complete the setting when waiting for the interface to be created.

## How to test

```
set interfaces bonding bond1 member interface eth0
set interfaces bonding bond1 member interface eth1
set interfaces bridge br1 member interface bond1 native-vlan 3
set interfaces bridge br1 member interface bond1 allowed-vlan 2
commit
run show int br vlan
```

## Checklist:

### DIFF
--- a/python/vyos/ifconfig/interface.py
+++ b/python/vyos/ifconfig/interface.py
@@ -950,10 +950,6 @@ class Interface(Control):
                         if int(vlan_id) not in vlan_bridge_ids:
                             cmd = f'bridge vlan add dev {bridge} vid {vlan_id} self'
                             self._cmd(cmd)
-                else:
-                    if vlan not in vlan_bridge_ids:
-                        cmd = f'bridge vlan add dev {bridge} vid {vlan} self'
-                        self._cmd(cmd)
             
             # enable/disable Vlan Filter
             Section.klass(bridge)(bridge, create=True).set_vlan_filter(vlan_filter)
--- a/python/vyos/ifconfig/interface.py
+++ b/python/vyos/ifconfig/interface.py
@@ -918,13 +918,9 @@ class Interface(Control):
 
             if 'allowed_vlan' in bridge_config:
                 vlan_filter = 1
-
-            if vlan_filter:
                 if 'native_vlan' not in bridge_config:
                     cmd = f'bridge vlan del dev {self.ifname} vid 1'
                     self._cmd(cmd)
-
-            if 'allowed_vlan' in bridge_config:
                 for vlan in bridge_config['allowed_vlan']:
                     cmd = f'bridge vlan add dev {self.ifname} vid {vlan} master'
                     self._cmd(cmd)
--- a/python/vyos/ifconfig/interface.py
+++ b/python/vyos/ifconfig/interface.py
@@ -928,6 +928,8 @@ class Interface(Control):
             
             vlan_bridge_ids = Section.klass(bridge)(bridge, create=True).get_vlan_ids()
             
+            apply_vlan_ids = set()
+            
             # Delete VLAN ID for the bridge
             for vlan in vlan_del:
                 if int(vlan) == 1:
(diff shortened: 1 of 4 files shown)


c-po:
Randomly changing the priority is a very, very bad idea. Usually the problem lies in the configuration itself and some dependencies. The dependencies must be resolved instead and changing those parameters must probably be done here: 

https://github.com/vyos/vyos-1x/blob/current/python/vyos/ifconfig/interface.py#L1178-L1181

This is called for every interface which might be part of the bridge. So the `is_bridge_member` key holds the entire interface configuration of the member interface.

jack9603301:
@c-po Sorry, I have changed the repair implementation

jack9603301:
The fix is ready to merge

python/vyos/ifconfig/interface.py
```diff
@@ -898,6 +899,82 @@ def add_to_bridge(self, bridge_dict):
...
+                    self._cmd(cmd)
+                    vlan_add.add(vlan)
+            
+            vlan_bridge_ids = Section.klass(bridge)(bridge, create=True).get_vlan_ids()
+            
+            # Delete VLAN ID for the bridge
+            for vlan in vlan_del:
+                if isinstance(vlan,str) and vlan.isnumeric():
+                    if int(vlan) == 1 and int(vlan) not in vlan_bridge_ids:
+                        cmd = f'bridge vlan del dev {bridge} vid {vlan} self'
```
c-po: Why is this line required four times? This is too redundant. It must be shrinked to one occurance.

image

c-po (changes requested):

c-po:
Please also provide a smoketest validating this change. It must fail on the current ISO image and not fail within this PR.

smoketest/scripts/cli/test_interfaces_bridge.py
```diff
@@ -45,17 +45,33 @@ def setUp(self):
             for tmp in Section.interfaces("ethernet"):
                 if not '.' in tmp:
                     self._members.append(tmp)
+            
+            self.session.set(['interfaces','dummy','dum0'])
+            self.session.set(['interfaces','dummy','dum1'])
+            self.session.commit()
```
c-po: Why is this `commit()` required?

smoketest/scripts/cli/test_interfaces_bridge.py
```diff
@@ -45,17 +45,33 @@ def setUp(self):
...
+            self.session.set(['interfaces','bonding','bond1','member','interface','dum0'])
+            self.session.set(['interfaces','bonding','bond1','member','interface','dum1'])
+            self.session.commit()
```
c-po: Why is this `commit()` required?

smoketest/scripts/cli/test_interfaces_bridge.py
```diff
@@ -45,17 +45,33 @@ def setUp(self):
...
+        self.session.delete(['interfaces','dummy'])
+        super().tearDown()
 
     def test_add_remove_member(self):
         """ Add member interfaces to bridge and set STP cost/priority """
         for interface in self._interfaces:
             base = self._base_path + [interface]
             self.session.set(base + ['stp'])
             self.session.set(base + ['address', '192.0.2.1/24'])
+            self.session.commit()
```
c-po: Why is this `commit()` required?

c-po (changes requested):

smoketest/scripts/cli/test_interfaces_bridge.py
jack9603301: Mainly to make the required interface effective during testing

smoketest/scripts/cli/test_interfaces_bridge.py
jack9603301: Mainly to make the required interface effective during testing

smoketest/scripts/cli/test_interfaces_bridge.py
jack9603301: I remember that I have not modified irrelevant functions, maybe I added them inadvertently

smoketest/scripts/cli/test_interfaces_bridge.py
jack9603301: This `commit` must exist, otherwise the following code will not take effect:

```
            for tmp in Section.interfaces("bonding"):
                if not '.' in tmp:
                    self._members.append(tmp)
```

c-po: `setUp()` is called for every testcase and within every testcase we call `commit()`, so calling `commit` here is redundant, please remove.

See https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp for more info.

c-po (changes requested):

python/vyos/ifconfig/interface.py
```diff
@@ -898,6 +899,60 @@ def add_to_bridge(self, bridge_dict):
...
+                if int(vlan_id) != 1:
+                    vlan_del.add(1)
+                cmd = f'bridge vlan add dev {self.ifname} vid {vlan_id} pvid untagged master'
+                self._cmd(cmd)
+                vlan_add.add(vlan_id)
+
+            if 'allowed_vlan' in bridge_config:
+                vlan_filter = 1
+
+            if vlan_filter:
```
c-po: Why is this `if vlan_filter` required? Just executing the commands below under `if 'allowed_vlan' in bridge_config:` have the same effect.

c-po (changes requested):

PR #644, opened 2020-12-13
interfaces: mirror: T3089: Fix the dependency problem between interfaces

jack9603301:
## Types of changes

 
## Related Task(s)

https://phabricator.vyos.net/T3089

## Component(s) name

mirror

## Proposed changes

Since the dependency problem has not been solved before,if the monitoring interface does not exist when the mirror rule is created, the execution will be abnormal

## How to test

Router(eth0)
```
config
set interfaces dummy dum1
set interfaces dummy dum2
set interfaces bonding bond1 member interface dum1
set interfaces bonding bond1 member interface dum2
set interfaces ethernet eth0 address 192.168.122.2/24
set interfaces ethernet eth0 mirror ingress bond1
set interfaces ethernet eth0 mirror egress bond1
commit
tcpdump -i bond1 -nvvve
```
PC(Connect to router eth0, address 192.168.122.1)
```
ping 192.192.122.1
```

## Checklist:

### DIFF
--- a/python/vyos/ifconfig/interface.py
+++ b/python/vyos/ifconfig/interface.py
@@ -1044,34 +1044,33 @@ class Interface(Control):
         
         # Setting up packet mirroring
         ingress_mirror = dict_search('mirror.ingress', self._config)
-        if ingress_mirror:
-            # if interface does yet not exist bail out early and
-            # add it later
-            if ingress_mirror in interfaces():
-                # Mirror ingress traffic
-                mirror_cmd = f'tc qdisc add dev {ifname} handle ffff: ingress'
-                self._cmd(mirror_cmd)
-                # Export the mirrored traffic to the interface
-                mirror_cmd = f'tc filter add dev {ifname} parent ffff: protocol all prio 10 u32 match u32 0 0 flowid 1:1 action mirred egress mirror dev {ingress_mirror}'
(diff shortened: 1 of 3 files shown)


python/vyos/ifconfig/interface.py
```diff
@@ -1227,7 +1266,12 @@ def update(self, config):
         if 'is_bridge_member' in config:
             bridge_dict = config.get('is_bridge_member')
             self.add_to_bridge(bridge_dict)
-
+        
+        # Re-set rules for the mirror monitoring interface
+        if 'is_monitor_intf' in config:
+            mirror_rules = config.get('is_monitor_intf')
+            self.apply_mirror_of_monitor(mirror_rules)
```
c-po: When we have `self._config` avaliable as class member, why pass a sub key of the dict as argument?

python/vyos/ifconfig/interface.py
```diff
@@ -1044,21 +1045,59 @@ def apply_mirror(self):
         # Setting up packet mirroring
         ingress_mirror = dict_search('mirror.ingress', self._config)
         if ingress_mirror:
```
c-po: Why is this `if` required when the next line is `if ingress_mirror in interfaces():`?

c-po (changes requested):

PR #645, opened 2020-12-15
Revert "interfaces: mirror: T3089: Fix the dependency problem between interfaces"

c-po:
Reverts vyos/vyos-1x#644

Smoketests are broken


jack9603301:
Please resume this submission, the problem does not lie in #644

jack9603301:
@c-po The problem seems to be due to the following verification code not passing

https://github.com/vyos/vyos-1x/blob/current/smoketest/scripts/cli/test_interfaces_bonding.py#L58-L60

PR #646, opened 2020-12-15
Revert "Revert "interfaces: mirror: T3089: Fix the dependency problem between interfaces""

jack9603301:
Reverts vyos/vyos-1x#645

Please resume this submission, the problem does not lie in #644


jack9603301:
@c-po the error message of the test case has nothing to do with the function

PR #650, opened 2020-12-19
bridge: T3137: When setting a VLAN-aware bridge, the VLAN ID of the parent interface is always 1

jack9603301:
## Change Summary

When setting a VLAN-aware bridge, the VLAN ID of the parent interface is always 1

## Types of changes

 
## Related Task(s)

https://phabricator.vyos.net/T3137

## Component(s) name

bridge

## Proposed changes

When setting a VLAN-aware bridge, the VLAN ID of the parent interface is always 1

## How to test

```
set interfaces bridge br1 member interface eth0 allowed-vlan 2
set interfaces bridge br1 member interface eth0 native-vlan 1
set interfaces bridge br1 address 1.1.1.1/32
```

## Checklist:

### DIFF
--- a/python/vyos/ifconfig/bridge.py
+++ b/python/vyos/ifconfig/bridge.py
@@ -368,6 +368,10 @@ class BridgeIf(Interface):
         for vlan in vlan_add:
             cmd = f'bridge vlan add dev {ifname} vid {vlan} self'
             self._cmd(cmd)
+        
+        # Always set VLAN 1 to the parent bridge interface
+        cmd = f'bridge vlan add dev {ifname} vid 1 pvid untagged self'
+        self._cmd(cmd)
 
         # enable/disable Vlan Filter
         self.set_vlan_filter(vlan_filter)
--- a/python/vyos/ifconfig/interface.py
+++ b/python/vyos/ifconfig/interface.py
@@ -938,6 +938,10 @@ class Interface(Control):
                 for vlan in vlan_add:
                     cmd = f'bridge vlan add dev {bridge} vid {vlan} self'
                     self._cmd(cmd)
+                
+                # Always set VLAN 1 to the parent bridge interface
+                cmd = f'bridge vlan add dev {bridge} vid 1 pvid untagged self'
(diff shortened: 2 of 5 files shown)


c-po:
This behavior is different to all other interfaces which support VLANs. Please make this symetric.

jack9603301:
Due to the migration command line, you need to configure the migration device, please handle it at the same time https://github.com/vyos/vyatta-cfg-system/pull/134

jack9603301:
@c-po Can we consider reviewing the PR

c-po:
Bridge interface should mimic the ethernet behavior 1:1 https://phabricator.vyos.net/T3189

* `set interfaces bridge br0 address 192.0.2.1/24` generates untagged packets
* `set interfaces bridge br0 vif 1 address 192.0.2.1/24` generates tagged packets in vlan 1

jack9603301:
VLAN aware bridge is a special bridge use case. In general, the default setting in the system is that the parent interface of the bridge is equivalent to `native-vlan 1`. In the original implementation, I canceled this default setting. I need to test whether the parent interface can be used to generate packets without VLAN tag in this case

jack9603301:
@c-po  First, the test results are published. After testing, it is found that when the VLAN aware bridge is started, any packets other than the VLAN ID cannot be set

VLAN-aware bridge is a special use case. We should design additional special rules for VLAN-aware bridges. This rule should be consistent with the behavior of other routers and switch manufacturers, combined with https://phabricator.vyos.net/ According to T1354, can the following changes be considered:

1. Maybe we can set `set interfaces bridge br1 vlan-aware` to activate the VLAN aware bridge function
2. Set the parent interface and child members of the bridge. When the VLAN-aware bridge is turned on, `native-vlan 1` is the default value, unless the `native-vlan` is manually modified, which means that when VLAN-aware is enabled by default, all Untagged packets are all members of VLAN 1

jack9603301:
Due to the consensus on the characteristics of the new bridge implementation, the PR will be closed and waiting for the new PR to be submitted

PR #658, opened 2020-12-24
mirror: T3150: When configuring QoS, the setting procedure of port mirroring is wrong

jack9603301:
## Change Summary

When configuring QoS, the setting procedure of port mirroring is wrong

## Types of changes

 
## Related Task(s)

https://phabricator.vyos.net/T3150

## Component(s) name

interfaces: mirror

## Proposed changes

In e8957b57, we used json to parse the `tc qdisc` filter to determine whether it needs to be deleted (reduction of exception mechanism), but now we find that the json output by this command will output unparsed json in some cases, so We have to go back to the processing of the exception mechanism

## How to test

```
set traffic-policy shaper FQ-SHAPER default bandwidth 100%
set traffic-policy shaper MY-SHAPER class 30 match HTTP
set traffic-policy shaper MY-SHAPER class 30 match HTTPs
set interfaces ethernet eth4 traffic-policy out MY-SHAPER
```

## Checklist:

### DIFF
--- a/python/vyos/ifconfig/interface.py
+++ b/python/vyos/ifconfig/interface.py
@@ -23,8 +23,6 @@ from copy import deepcopy
 from glob import glob
 
 from ipaddress import IPv4Network
-from ipaddress import IPv6Address
-from ipaddress import IPv6Network
 from netifaces import ifaddresses
 # this is not the same as socket.AF_INET/INET6
 from netifaces import AF_INET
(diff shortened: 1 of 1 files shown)


c-po:
What is the behavior using the old implementation from Perl? Did this error occur in the old implementation, too?

jack9603301:
I haven't tested the old implementation, which seems to be similar to the current implementation, which is to force execution without parsing JSON

jack9603301:
`tc -j qdisc show` check seems to be the most reliable in programming logic, because it does not need to throw an exception, but now it seems to be unreliable, because when it returns an error JSON, it will throw an exception to the JSON parser and have to go back to the original implementation

jack9603301:
I think this fix is an urgent implementation, because this bug will destroy the QoS configurator

And this implementation has been verified in the actual vyos system

jack9603301:
@c-po Can we consider reviewing the PR

jack9603301:
@c-po Can we consider reviewing the PR

PR #664, opened 2020-12-30
pppoe-server: T3162: Generate pado-delay string

DmitriyEshenko:
## Types of changes

 
## Checklist:


c-po:
We should not handle this in the Python code and rather move the stringification into the Jinja2 template itself.

Jinja2 is more or less Python and that would keep all the stuff in a single location.

DmitriyEshenko:
At first, I tried to implement this code in jinja2, but I see a lot of jinja2 restrictions. Does not possible to update sting in a loop without some jinja2 magic and additional list definition. Could you help me with the implementation on jinja2?

PR #666, opened 2021-01-01
T3171: Ethernet RPS support

c-po:
## Change Summary

RPS is not enabled by default, same for VyOS 1.2, thus add a new CLI node set interfaces ethernet <interface> offload rps

## Types of changes

 
## Related Task(s)

https://phabricator.vyos.net/T3171

## Component(s) name

ethernet

## Checklist:

### DIFF
--- a/interface-definitions/interfaces-ethernet.xml.in
+++ b/interface-definitions/interfaces-ethernet.xml.in
@@ -77,6 +77,12 @@
                   <valueless/>
                 </properties>
               </leafNode>
+              <leafNode name="rps">
+                <properties>
+                  <help>Enable Receive Packet Steering</help>
+                  <valueless/>
+                </properties>
+              </leafNode>
               <leafNode name="sg">
                 <properties>
                   <help>Enable Scatter-Gather</help>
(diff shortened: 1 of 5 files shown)


hagbard-01:
Your RPS implementation won't work on every system, the f in the bitmask represents a core on a CPU, but it also depends on the driver capabilities. So on big systems you will have '000000,00000000' (it's from a 56 core OS router), your formula will fail in that case. Same for a 4 core and you have just '0' in rps_cpus, you auto setting will leave it at 0 and traffic won't be pinned to a core, it will stay disabled. On a bnx2 with 16 cores rps_cpus looks like '00000000' (no NUMA domain thus no comma). You need to read the available mask from sysfs, the default is 0 but xor will zero it too (value ^ value = 0), the shift it and then I would to a logical or with cpu_count -1 or -2, that leaves you a core for interrupt handling and another one for running apps like frr. I don't know if frr can be that busy, I would think it may be quite busy on route changes if a link fails or so, but not too sure about it.

c-po:
```
>>> rps_cpus = (1 << 4) -1
>>> rps_cpus &= ~1
>>> f'{rps_cpus:x}'
'e'
```

PR #667, opened 2021-01-03
ISIS: T3156: Adding segment routing for ISIS

Cheeze-It:
In this commit we add the segment routing portion for ISIS.
There's also an additional check that is added so that the
global block label ranges are properly configured.

## Change Summary

We are adding the ISIS segment routing configurations here. We're also adding a global block check as well to make sure the label values provided aren't improper. Traffic engineering configurations are added as well, but those really don't do much yet as RSVP isn't implemented in FRR.

## Types of changes

 
## Related Task(s)

https://phabricator.vyos.net/T3156

## Component(s) name

ISIS

## Proposed changes

Not much different than the change summary up above.

There are some things that are disabled though in the XML file because FRR doesn't currently support them yet. These I have added in anticipation for the future addition within FRR. Should be fairly easy to just turn them on, and off we go.

## How to test

I did my testing in a VM that is running inside of GNS3. It seems that things are properly committing within ISIS and deleting. 

## Checklist:

ISIS documentation page hasn't been made yet. Will absolutely add when said page is created.

### DIFF
--- a/data/templates/frr/isis.frr.tmpl
+++ b/data/templates/frr/isis.frr.tmpl
@@ -31,6 +31,67 @@ router isis {{ process }}
 {% if spf_interval is defined and spf_interval is not none %}
  spf-interval {{ spf_interval }}
 {% endif %}
+{% if traffic_engineering is defined and traffic_engineering is not none %}
+{%   if traffic_engineering.enable is defined %}
+ mpls-te on
+{%   endif %}
+{%   if traffic_engineering.address is defined %}
+ mpls-te router-address {{ traffic_engineering.address }}
+{%   endif %}
+{%   if traffic_engineering.inter_as is defined %}
+{%     if traffic_engineering.inter_as.level_1 is defined %}
+ mpls-te inter-as level-1
+{%     endif %}
+{%     if traffic_engineering.inter_as.level_1_2 is defined %}
+ mpls-te inter-as level-1-2
+{%     endif %}
+{%     if traffic_engineering.inter_as.level_2 is defined %}
+ mpls-te inter-as level-2-only
+{%     endif %}
+{%   else %}
+ mpls-te inter-as
+{%   endif %}
+{% endif %}
+{% if segment_routing is defined %}
+{%   if segment_routing.enable is defined %}
+ segment-routing on
+{%   endif %}
+{%   if segment_routing.maximum_label_depth is defined %}
+ segment-routing node-msd {{ segment_routing.maximum_label_depth }}
+{%   endif %}
+{%   if segment_routing.global_block is defined %}
+ segment-routing global-block {{ segment_routing.global_block.low_label_value }} {{ segment_routing.global_block.high_label_value }}
+{%   endif %}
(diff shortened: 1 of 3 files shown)


data/templates/frr/isis.frr.tmpl
```diff
@@ -31,6 +31,67 @@ router isis {{ process }}
 {% if spf_interval is defined and spf_interval is not none %}
  spf-interval {{ spf_interval }}
 {% endif %}
+{% if traffic_engineering is defined %}
```
c-po: Please use: `{% if traffic_engineering is defined and traffic_engineering is not none %}`, thus if TE is not enabled the entire tree won't be run which is faster.

The difference is that if a dict contains an empty `traffic_engineering` key this evaluates to true, but if there are no more sub-keys, it evaluates to false entirely at once.

src/conf_mode/protocols_isis.py
```diff
@@ -85,6 +85,24 @@ def verify(isis):
                 if proc_level and proc_level != 'level_1_2' and proc_level != redistribute_level:
                     raise ConfigError('\"protocols isis {0} redistribute ipv4 {2} {3}\" cannot be used with \"protocols isis {0} level {1}\"'.format(process, proc_level, proto, redistribute_level))
 
+        # Segment routing checks
+        if 'segment_routing' in isis_config:
```
c-po: Please try to make use of `dict_search()` function from `vyos.utils`

c-po (changes requested):

data/templates/frr/isis.frr.tmpl
Cheeze-It: Ahh yes, that makes sense then. Will make said changes.

src/conf_mode/protocols_isis.py
Cheeze-It: Will make said changes.

src/conf_mode/protocols_isis.py
```diff
@@ -85,6 +86,34 @@ def verify(isis):
...
+                    raise ConfigError('Segment routing global block low value must not be left blank')
+                # If segment routing global block low value is higher than the high value, throw error
+                if int(low_label_value) > int(high_label_value):
+                    raise ConfigError('Segment routing global block low value must be lower than high value')
+                
+            if dict_search('segment_routing.local_block', isis_config):
+                high_label_value = dict_search('segment_routing.local_block.high_label_value', isis_config)
+                low_label_value = dict_search('segment_routing.local_block.low_label_value', isis_config)
+                # If segment routing local block high value is blank, throw error
+                if high_label_value is None and low_label_value is not None:
```
c-po: Can we write: `if low_label_value and not high_label_value`: makes it easier for humans to read.

c-po (changes requested):

src/conf_mode/protocols_isis.py
Cheeze-It: Yes, absolutely we can.

PR #673, opened 2021-01-07
T3165: use a separate file for "base" CLI nodes.

dmbaturin:
Needs a bit of a clean-up to move all top level nodes to the base file.


c-po:
Unfortunately this PR broke NAT. the previous NAT `node.def` called the nat related Python script. Now the node.def is empty.

https://github.com/vyos/vyos-1x/blob/current/interface-definitions/nat.xml.in#L3

PR #725, opened 2021-02-09
ethernet: T3291: Disable RPS on single-core processors(KVM Only)

jack9603301:
## Change Summary

Disable RPS on single-core processors(KVM Only)

## Types of changes

 
## Related Task(s)

https://phabricator.vyos.net/T3291

## Component(s) name

ethernet, rps

## Proposed changes

In existing RPS implementations, the mask is set to not use CPU0,which causes the RPS to be unable to set when the user only has CPU0,so add validation logic to disable RPS(KVM Only)

## Checklist:

### DIFF
--- a/src/conf_mode/interfaces-ethernet.py
+++ b/src/conf_mode/interfaces-ethernet.py
@@ -18,7 +18,6 @@ import os
 
 from glob import glob
 from sys import exit
-from multiprocessing import cpu_count
 
 from vyos.config import Config
 from vyos.configdict import get_interface_dict
@@ -78,7 +77,9 @@ def verify(ethernet):
 
     # verify offloading capabilities
     if 'offload' in ethernet and 'rps' in ethernet['offload']:
-        if not os.path.exists(f'/sys/class/net/{ifname}/queues/rx-0/rps_cpus') or cpu_count() < 2:
+        if os.cpu_count() < 2:
+            raise ConfigError('RPS cannot be enabled on a system with only 1 logical CPU!')
+        if not os.path.exists(f'/sys/class/net/{ifname}/queues/rx-0/rps_cpus'):
             raise ConfigError('Interface does not suport RPS!')
 
     # XDP requires multiple TX queues
--- a/src/conf_mode/interfaces-ethernet.py
+++ b/src/conf_mode/interfaces-ethernet.py
@@ -18,6 +18,7 @@ import os
 
 from glob import glob
 from sys import exit
+from multiprocessing import cpu_count
 
 from vyos.config import Config
 from vyos.configdict import get_interface_dict
@@ -77,7 +78,7 @@ def verify(ethernet):
 
     # verify offloading capabilities
     if 'offload' in ethernet and 'rps' in ethernet['offload']:
-        if not os.path.exists(f'/sys/class/net/{ifname}/queues/rx-0/rps_cpus'):
+        if not os.path.exists(f'/sys/class/net/{ifname}/queues/rx-0/rps_cpus') or cpu_count() < 2:
             raise ConfigError('Interface does not suport RPS!')
 
     # XDP requires multiple TX queues
--- a/src/conf_mode/interfaces-ethernet.py
+++ b/src/conf_mode/interfaces-ethernet.py
@@ -36,6 +36,7 @@ from vyos.util import call
 from vyos.util import dict_search
 from vyos import ConfigError
 from vyos import airbag
+from vyos.version import get_full_version_data
 airbag.enable()
 
 # XXX: wpa_supplicant works on the source interface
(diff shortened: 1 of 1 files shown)


c-po:
The `Interface does not suport RPS!` erro message would be missleading by your change as this is clearly not the case.

```bash
cpo@LR1.wue3# set interfaces ethernet eth1 offload rps
[edit]
cpo@LR1.wue3# commit
[edit]

cpo@LR1.wue3# cat /proc/cpuinfo
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 63
model name      : Intel(R) Xeon(R) CPU E5-2630L v3 @ 1.80GHz
stepping        : 2
microcode       : 0x43
cpu MHz         : 1797.917
cache size      : 20480 KB
physical id     : 0
siblings        : 1
core id         : 0
cpu cores       : 1
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 15
wp              : yes
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon nopl xtopology tsc_reliable nonstop_tsc cpuid pni pclmulqdq vmx ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm abm cpuid_fault invpcid_single pti ssbd ibrs ibpb stibp tpr_shadow vnmi ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 invpcid xsaveopt arat md_clear flush_l1d arch_capabilities
vmx flags       : vnmi invvpid ept_x_only ept_ad tsc_offset vtpr mtf ept vpid unrestricted_guest ple
bugs            : cpu_meltdown spectre_v1 spectre_v2 spec_store_bypass l1tf mds swapgs itlb_multihit
bogomips        : 3595.83
clflush size    : 64
cache_alignment : 64
address sizes   : 43 bits physical, 48 bits virtual
power management:

[edit]

```
```python
Python 3.7.3 (default, Jul 25 2020, 13:03:44)
[GCC 8.3.0] on linux
>>> import os
>>> os.cpu_count()
1
```

There is no issue with only one CPU.

jack9603301:
图片

There is a problem in single core single thread of virtual machine

hagbard-01:
It looks to me that the error comes from `tc qdisc del...` instead of rps. If you only have a single core it already handles the interrupt as well as it processes the connection, so it may be questionable to set rps_cpus to 0x1, but won't do anything worse. Per default rps is disabled on any OS per default, may just log a line to syslog if anyone enables it?

jack9603301:
@hagbard-01 I'm very sorry to say that this is actually an overlap of two bugs. Please see 

https://github.com/vyos/vyos-1x/pull/727

The error does come from the failure of setting the parameters of `rps_cpus` file, but when displaying the error stack, another bug is triggered, causing the error to display the error of the last system command (in fact, the execution error of TC is expected)

jack9603301:
When users enable RPS, the current mask will not use cpu0. I have repeated this error on multiple single core virtual machine systems. In single core processors, there seems to be a problem with this mask

图片

This problem appears to be limited to single-core processor systems

c-po:
Well this issue does not persist with Vmware based VMs. Which hypervisor do you use? How can we reproduce it? What happens if a multiprocessor system is booted with cpu/cpus=1 to not initialize additional cores?

jack9603301:
@c-po I use KVM virtualization,

Simply start VyOS on a single-core processor and execute the following command to reproduce

```
config
set interfaces ethernet eth0 offload rps
commit
```

Or simply execute the following command

```
sudo passwd root
su root
echo fffe > /sys/class/net/eth0/queues/rx-0/rps_cpus
```

It's very easy to reproduce, and in my case, I've been able to reproduce the bug almost every time

jack9603301:
@c-po 
图片

The same happens when the `nosmp` parameter is used to force the kernel to boot as a single processor

jack9603301:
@c-po Both my local KVM manager and Promox ve have replicated the following link bug without exception.

https://phabricator.vyos.net/T3291

src/conf_mode/interfaces-ethernet.py
```diff
@@ -77,6 +77,8 @@ def verify(ethernet):
 
     # verify offloading capabilities
     if 'offload' in ethernet and 'rps' in ethernet['offload']:
+        if os.cpu_count() < 2:
+            raise ConfigError('RPS cannot be enabled on a system with only 1 logical CPU!')
```
c-po: This is incorrect - works on VMware with only one vCPU. The problem must be of a different kind.

jack9603301: @c-po I have a question, is it possible to judge only for KVM?

jack9603301: done

c-po:
This is clearly an issue with the Linux Kernel (incompatible ABI change) as the error is not present on 1.3-beta which uses 5.4 kernel series but it exists on 1.4-rolling which uses the 5.10 series.

It affects virtio, e1000 and vmxnet drivers on 5.10 Kernel and can be reproduced as easy as:
```bash
$ echo ffffffff,ffffffff,ffffffff,fffffffe > /sys/class/net/eth0/queues/rx-0/rps_cpus
```

Please investigate what the root-cause is and if this is a known Kernel issue, should be fixed or is beyond fixing. We then can decide what to do.

PR #727, opened 2021-02-10
mirror: T3297: Optimize irrelevant error stack hints

jack9603301:
## Change Summary

Optimize irrelevant error stack hints

## Types of changes

 
## Related Task(s)

https://phabricator.vyos.net/T3297

## Component(s) name

mirror

## Proposed changes

When setting the interface, the rule of port image will always be cleared.Because the TC command will return an error, when the actual error occurs,the error stack will display the execution information of
the last error command. This is an error.Now, we force the return value of these two commands to be overridden to ensure that they always return 0

## Checklist:

### DIFF
--- a/python/vyos/ifconfig/interface.py
+++ b/python/vyos/ifconfig/interface.py
@@ -1015,9 +1015,11 @@ class Interface(Control):
             source_if = next(iter(self._config['is_mirror_intf']))
             config = self._config['is_mirror_intf'][source_if].get('mirror', None)
 
+        # Please don't clear 'set $?=0'. It will return the value when executing error
         # Remove existing mirroring rules
-        delete_tc_cmd  = f'tc qdisc del dev {source_if} handle ffff: ingress; '
-        delete_tc_cmd += f'tc qdisc del dev {source_if} handle 1: root prio'
+        delete_tc_cmd  = f'tc qdisc del dev {source_if} handle ffff: ingress 2> /dev/null;'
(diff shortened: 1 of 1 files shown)


python/vyos/ifconfig/interface.py
```diff
@@ -1015,9 +1015,11 @@ def set_mirror(self):
             source_if = next(iter(self._config['is_mirror_intf']))
             config = self._config['is_mirror_intf'][source_if].get('mirror', None)
 
+        # Please don't clear 'set $?=0'. It will return the value when executing error
```
c-po: I do not understand this comment, from my understanding this forces the return code to be 0.

c-po (changes requested):

c-po:
The redirect of stderr mimics the old implementation: https://github.com/vyos/vyatta-cfg-qos/blob/crux/scripts/vyatta-qos.pl#L332 so this is okay.

jack9603301:
@c-po I actually used the set command, which always executed successfully, overriding the return value

python/vyos/ifconfig/interface.py
jack9603301: To simulate its normal return, one is no stderr, and two is its return value 0

jack9603301: In fact, any set command can only result in a return value of 0, but I write this sentence just to make sense

PR #729, opened 2021-02-11
dhcp: T3300: add DHCP default route distance

bstepler:
## Change Summary

Add DHCP default route distance.

## Types of changes

 
## Related Task(s)

https://phabricator.vyos.net/T3300

## Component(s) name

dhcp

## How to test

```
configure
commit
exit
show ip route
configure
set interfaces ethernet eth0 dhcp-options default-route-distance 220
commit
exit
show ip route
```

## Checklist:

- [X] I have updated the documentation accordingly


data/templates/dhcp-client/daemon-options.tmpl
```diff
@@ -1,4 +1,4 @@
 ### Autogenerated by interface.py ###
 
-DHCLIENT_OPTS="-nw -cf /var/lib/dhcp/dhclient_{{ifname}}.conf -pf /var/lib/dhcp/dhclient_{{ifname}}.pid -lf /var/lib/dhcp/dhclient_{{ifname}}.leases {{ifname}}"
+DHCLIENT_OPTS="-nw -cf /var/lib/dhcp/dhclient_{{ifname}}.conf -pf /var/lib/dhcp/dhclient_{{ifname}}.pid -lf /var/lib/dhcp/dhclient_{{ifname}}.leases{{" -e IF_METRIC=" ~ dhcp_options.default_route_distance if dhcp_options.default_route_distance is defined and dhcp_options.default_route_distance is not none}} {{ifname}}"
```
c-po, line 3: What does the `~` do? I thought it's `+` to append a string.

dev00: https://jinja.palletsprojects.com/en/2.10.x/templates/?highlight=operator#math
https://jinja.palletsprojects.com/en/2.10.x/templates/?highlight=operator#other-operators

`~` is "convert to str and concatenate", `+` is more or less "Add or concat if add makes no sense".

src/etc/dhcp/dhclient-enter-hooks.d/03-vyos-ipwrapper
```diff
@@ -1,5 +1,8 @@
 # redefine ip command to use FRR when it is available
 
+# default route distance
+IF_METRIC=${IF_METRIC:-210}
```
zdc, line 4: It looks like having this here makes checks for the `if_metric` value in the `01-vyos-cleanup` useless because `if_metric` will always be defined via `if_metric="$IF_METRIC"`.
I assume that this definition here is excessive and default value should be kept in the `local VTYSH_DISTANCE`.

python/vyos/ifconfig/interface.py
```diff
@@ -951,6 +951,9 @@ def set_dhcp(self, enable):
         pid_file = f'{config_base}_{ifname}.pid'
         lease_file = f'{config_base}_{ifname}.leases'
 
+        # Stop client with old config files to get the right IF_METRIC.
+        self._cmd(f'systemctl stop dhclient@{ifname}.service')
```
zdc, line 955: It does not look safe because it stops even before the first start and when `DHCLIENT_OPTS` still defaults.

src/etc/dhcp/dhclient-enter-hooks.d/03-vyos-ipwrapper
bstepler: The code in `01-vyos-cleanup` was copied from lines 261-265 of `/usr/sbin/dhclient-script`. This is purposely overriding that so we don't default to a distance of 1 if there are multiple routers. I'll remove the redundant code from `01-vyos-cleanup`.

python/vyos/ifconfig/interface.py
bstepler: It does not execute the stop command if the service is not started. Tested with:

```
sudo systemctl stop dhclient@eth0.service
sudo tee /etc/systemd/system/dhclient@.service <<'EOF' >/dev/null
[Unit]
Description=DHCP client on %i
Documentation=man:dhclient(8)
ConditionPathExists=/var/lib/dhcp/dhclient_%i.conf
ConditionPathExists=/var/lib/dhcp/dhclient_%i.options
After=vyos-router.service

[Service]
WorkingDirectory=/var/lib/dhcp
Type=exec
EnvironmentFile=-/var/lib/dhcp/dhclient_%i.options
PIDFile=/var/lib/dhcp/dhclient_%i.pid
ExecStart=/sbin/dhclient -4 $DHCLIENT_OPTS
ExecStop=/bin/logger "Testing: stopping DHCP client on %i"
ExecStop=/sbin/dhclient -4 $DHCLIENT_OPTS -r
Restart=always

[Install]
WantedBy=multi-user.target
EOF
sudo systemctl daemon-reload
sudo systemctl stop dhclient@eth0.service
grep "Testing:" /var/log/messages
sudo systemctl start dhclient@eth0.service
sudo systemctl stop dhclient@eth0.service
grep "Testing:" /var/log/messages
```

c-po:
Lets give this a test drive in 1.4

PR #759, opened 2021-03-04
grub: T3271: don't write grub.cfg if it hasn't changed

bstepler:
## Change Summary

Don't write `grub.cfg` if it hasn't changed.

## Types of changes

## Related Task(s)

https://phabricator.vyos.net/T3271

## Component(s) name

grub

## How to test

Boot an installed system, then power it off without properly shutting it down. Repeat if necessary. Without this fix, it will eventually boot to a `grub` prompt.

## Checklist:

### DIFF
--- a/src/conf_mode/system_console.py
+++ b/src/conf_mode/system_console.py
@@ -17,9 +17,8 @@
 import os
 import re
 
-from fileinput import input as replace_in_file
 from vyos.config import Config
-from vyos.util import call
+from vyos.util import call, read_file, write_file
 from vyos.template import render
 from vyos import ConfigError, airbag
 airbag.enable()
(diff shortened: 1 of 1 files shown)


src/conf_mode/system_console.py
```diff
@@ -98,15 +97,29 @@ def generate(console):
...
+            newline = line
+
+        if newline != line:
+            write = True
+
+        newlines.append(newline)
 
-        print(line, end='')
+    if write:
+        with open(grub_config, 'w') as f:
```
c-po: As I understand it you do an in memory replacement instead of mangling the file on the filesystem line by line, correct?

If this is the case can we please also wrap the write with `vyos.util.write_file()`?

c-po (changes requested):

src/conf_mode/system_console.py
bstepler: Correct.

Done.

PR #776, opened 2021-03-16
nat66: T2518: Correct the wrong logic

jack9603301:
## Change Summary

Correct the wrong logic

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T2518

## Component(s) name

nat66

## Proposed changes

For nat66, the previous processing of f0d0a572 (NAT: nat66: t2518: support operation...) has errors.If there is no index 3, we think that this is not the record we need

## Checklist:

### DIFF
--- a/src/op_mode/show_nat66_rules.py
+++ b/src/op_mode/show_nat66_rules.py
@@ -33,9 +33,9 @@ if args.source or args.destination:
     tmp = cmd('sudo nft -j list table ip6 nat')
     tmp = json.loads(tmp)
     
-    format_nat66_rule = '%-10s %-50s %-50s %-10s'
-    print(format_nat66_rule % ("Rule", "Source" if args.source else "Destination", "Translation", "Outbound Interface" if args.source else "Inbound Interface"))
-    print(format_nat66_rule % ("----", "------" if args.source else "-----------", "-----------", "------------------" if args.source else "-----------------"))
+    format_nat66_rule = '{0: <10} {1: <50} {2: <50} {3: <10}'
+    print(format_nat66_rule.format("Rule", "Source" if args.source else "Destination", "Translation", "Outbound Interface" if args.source else "Inbound Interface"))
+    print(format_nat66_rule.format("----", "------" if args.source else "-----------", "-----------", "------------------" if args.source else "-----------------"))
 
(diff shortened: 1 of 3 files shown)


src/op_mode/show_nat66_rules.py
```diff
@@ -40,7 +40,14 @@
     data_json = jmespath.search('nftables[?rule].rule[?chain]', tmp)
     for idx in range(0, len(data_json)):
         data = data_json[idx]
+        
+        # If there is no index 3, we don't think this is the record we need to check
+        if len(data['expr']) <= 3:
+            continue
+        
         comment = data['comment']
+        rule = comment.replace('SRC-NAT66-','')
```
c-po, line 49: This looks very odd, what do you plan to acciebe hete?

src/op_mode/show_nat66_rules.py
```diff
@@ -60,12 +68,10 @@
         else:
             if 'masquerade' in data['expr'][3]:
                 tran_addr = 'masquerade'
-            elif 'log' in data['expr'][3]:
-                continue
             else:
                 tran_addr = dict_search('snat.addr' if args.source else 'dnat.addr', data['expr'][3])
         
-        print(format_nat66_rule % (comment, srcdest, tran_addr, interface))
+        print(format_nat66_rule % (rule, srcdest, tran_addr, interface))
```
c-po: Can we mobe to Python3 format identifiers?

c-po (changes requested):

src/op_mode/show_nat66_rules.py
jack9603301: As the other script does, take the rule ID from the comment

c-po: Why not use something like: https://stackoverflow.com/a/26825833

jack9603301: For IPv4, NAT is OK, but if it is used for IPv6, it will result in unwanted numbers

src/op_mode/show_nat66_rules.py
jack9603301: done

src/op_mode/show_nat66_rules.py
jack9603301: For nat only, nat66 is maintained

PR #789, opened 2021-03-25
upnpd: T3420: Support UPNP protocol

jack9603301:
## Change Summary

Support UPNP protocol

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3420

## Component(s) name

upnpd

## Checklist:

### DIFF
--- a/data/templates/firewall/upnpd.conf.tmpl
+++ b/data/templates/firewall/upnpd.conf.tmpl
@@ -1,0 +1,172 @@
+# This is the UPNP configuration file
+
+# WAN network interface
+ext_ifname={{ wan_interface }}
+{% if wan_ip is defined %}
+# If the WAN interface has several IP addresses, you
+# can specify the one to use below
+{%   for addr in wan_ip %}
+ext_ip={{ addr }}
+{%   endfor  %}
+{% endif %}
+
+# LAN network interfaces IPs / networks
+{% if listen is defined %}
+# There can be multiple listening IPs for SSDP traffic, in that case
+# use multiple 'listening_ip=...' lines, one for each network interface.
+# It can be IP address or network interface name (ie. "eth0")
+# It is mandatory to use the network interface name in order to enable IPv6
+# HTTP is available on all interfaces.
+# When MULTIPLE_EXTERNAL_IP is enabled, the external IP
+# address associated with the subnet follows. For example:
+#  listening_ip=192.168.0.1/24 88.22.44.13
+{%   for addr in listen %}
+{%     if addr | is_ipv4  %}
+listening_ip={{ addr }}
+{%     elif addr | is_ipv6  %}
+ipv6_listening_ip={{ addr }}
+{%     else %}
+listening_ip={{ addr }}
+{%     endif  %}
+{%   endfor  %}
+{% endif %}
+
+# CAUTION: mixing up WAN and LAN interfaces may introduce security risks!
+# Be sure to assign the correct interfaces to LAN and WAN and consider
+# implementing UPnP permission rules at the bottom of this configuration file
+
+# Port for HTTP (descriptions and SOAP) traffic. Set to 0 for autoselect.
+#http_port=0
+# Port for HTTPS. Set to 0 for autoselect (default)
+#https_port=0
+
+# Path to the UNIX socket used to communicate with MiniSSDPd
+# If running, MiniSSDPd will manage M-SEARCH answering.
+# default is /var/run/minissdpd.sock
+#minissdpdsocket=/var/run/minissdpd.sock
+
+{% if nat_pmp is defined %}
+# Enable NAT-PMP support (default is no)
+enable_natpmp=yes
+{% endif %}
+
+# Enable UPNP support (default is yes)
+enable_upnp=yes
+
+{% if pcp_lifetime is defined %}
+# PCP
+# Configure the minimum and maximum lifetime of a port mapping in seconds
+# 120s and 86400s (24h) are suggested values from PCP-base
+{% if pcp_lifetime.max is defined %}
+max_lifetime={{ pcp_lifetime.max }}
+{% endif %}
+{% if pcp_lifetime.min is defined %}
+min_lifetime={{ pcp_lifetime.min }}
+{% endif %}
+{% endif %}
+
+
+# To enable the next few runtime options, see compile time
+# ENABLE_MANUFACTURER_INFO_CONFIGURATION (config.h)
+
+{% if friendly_name is defined %}
+# Name of this service, default is "`uname -s` router"
+friendly_name= {{ friendly_name }}
+{% endif  %}
+
+# Manufacturer name, default is "`uname -s`"
+manufacturer_name=VyOS
+
+# Manufacturer URL, default is URL of OS vendor
+manufacturer_url=https://vyos.io/
+
+# Model name, default is "`uname -s` router"
+model_name=VyOS Router Model
+
+# Model description, default is "`uname -s` router"
+model_description=Vyos open source enterprise router/firewall operating system
+
+# Model URL, default is URL of OS vendor
+model_url=https://vyos.io/
+
+{% if secure_mode is defined %}
+# Secure Mode, UPnP clients can only add mappings to their own IP
+secure_mode=yes
+{% else %}
+# Secure Mode, UPnP clients can only add mappings to their own IP
+secure_mode=no
+{% endif %}
+
+{% if presentation_url is defined %}
+# Default presentation URL is HTTP address on port 80
+# If set to an empty string, no presentationURL element will appear
+# in the XML description of the device, which prevents MS Windows
+# from displaying an icon in the "Network Connections" panel.
+#presentation_url= {{ presentation_url }}
+{% endif %}
+
+# Report system uptime instead of daemon uptime
+system_uptime=yes
+
+# Unused rules cleaning.
+# never remove any rule before this threshold for the number
+# of redirections is exceeded. default to 20
+clean_ruleset_threshold=10
+# Clean process work interval in seconds. default to 0 (disabled).
+# a 600 seconds (10 minutes) interval makes sense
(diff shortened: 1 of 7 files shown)


jack9603301:
associated: https://github.com/vyos/vyos-build/pull/153

jack9603301:
associated: https://github.com/vyos/vyatta-cfg-system/pull/142

c-po:
Hi @jack9603301,

thanks for this contribution. Unfortunately I have had yet no time to dig myself through UPnP and your commit. Will do in the next couple of days.

src/conf_mode/service_upnp.py
```diff
@@ -0,0 +1,155 @@
...
+        conf = Config()
+    base = ['service', 'upnp']
+    upnpd = conf.get_config_dict(base, key_mangling=('-', '_'), get_first_key=True)
+    
+    if not upnpd:
+        return None
+    
+    if dict_search('rule', upnpd):
+        default_member_values = defaults(base + ['rule'])
+        for rule,rule_config in upnpd['rule'].items():
```
sever-sever, line 54: Add whitespace after the comma.

src/conf_mode/service_upnp.py
```diff
@@ -0,0 +1,155 @@
...
+
+def verify(upnpd):
+    if not upnpd:
+        return None
+    
+    if 'wan_interface' not in upnpd:
+        raise ConfigError('To enable UPNP, you must have the "wan-interface" option!')
+    
+    if dict_search('rules', upnpd):
+        for rule,rule_config in upnpd['rule'].items():
```
sever-sever, line 101: Add whitespace after the comma.

c-po:
Given the recent flaws and "problems" once more uncovered by UPNP, I am sorry but I must say: -1

Thank you for this PR anyway and your time spend on it.

KittyFort:
Realistically, uPnP - like any other tool - still has value despite being a potential attack vector. _Especially_ if it's not enabled by default. Let's talk about ways to get the most out of uPnP:

1. Being able to lock uPnP support to specific interfaces, IP ranges or zones will prevent concerns with untrusted devices being able to punch holes in the NAT. We're all aware uPnP isn't the _only_ way to allow incoming traffic, but its simplicity makes targeting it substantially easier.
2. Only allowing devices to do uPnP configuration for their own IPs prevents some of the worst vulnerabilities around uPnP. This is one of the worst of the uPnP implementation issues seen.
3. Limiting the number of ports that a device can open is another simple, sane restriction. Similarly, a range of ports could be reserved up front for uPnP applications.
4. Malware attempting to utilize uPnP doesn't go away just because uPnP is disabled. Compromised systems will still be compromised.
5. uPnP doesn't have to actually _open the port_ or _save the configuration_. I think it's greatest potential is as a draft port mapping tool. A uPnP service could actually be used to auto-discover ports used by applications. Some of this can be done with firewall rules, and service implementations aren't there yet. But potential should be a good reason to not unilaterally block an optional service.
6. uPnP is a consumer technology used on consumer routers. Which, in the end, are a massive security risk. An optional and _limited_ implementation of uPnP could be a means to push consumers away from terrible turnkey router e-waste that, despite being insecure, actually lets them play Xbox or whatever without much hassle. If we won't support anything that isn't perfect, we miss the opportunity to improve the already terrible market.
7. uPnP actually has some potential use in DMZ zones as well, as long as traffic between zones is properly firewalled.
8. When we actually get a GUI in place, uPnP could integrate with the UI and with admin notification systems to deal with the security issues of eternal access to immediate port mapping. Temporary uPnP enablement windows, mapping approval procedures, etc.
9. No one's going to work on betting uPnP (or any other technology, for that matter) if we refuse to even allow it to be optionally enabled until it's perfect. Especially when _existing_ services and kernel interfaces don't have all their options available through the config interface.

I would strongly dissent to making uPnP a default service. But, then again, Vyos doesn't really have default services. I'd rather have the option there and _documented_ for its potential security risks than just take a purist approach that ignores existing standards that can be misused.

uPnP is hardly the only thing a sloppy IT admin could do to put the network at risk. And having the ability to configure it but leave it disabled until someone needs to start up some random P2P application service would make it significantly easier to deploy and manage among less tech-savvy locations. That could change a port-mapping chore from an hour-long fiasco of googling for documentation to a five-minute affair.

lucasec:
@jack9603301 Thanks for working on this, UPnP was also on my plate to look into contributing. This looks pretty similar to what I was planning to implement, will try it out tomorrow and see if I have any feedback on how it was implemented.

I think this could still be enhanced by adding a run mode command to show active UPnP leases. I also planned to do some testing on whether upnp could stomp on existing DNAT/firewall rules and if so whether we could autogenerate "exclude" rules in the daemon config to prevent that.

As to arguments against merging UPnP, I think @KittyFort did a good job rebutting most. `miniupnpd` is pretty mature at this point, and by nature of being widely used, bugs get found and patched pretty quickly. It also has a good variety of hardening options, such as `secure_mode` (can we default this to `yes` and require an opt-out `insecure-mode` instead?) and the ACL-style "rules" list that lets you prohibit clients mapping certain ports.

I suppose the real question is whether we want to exclude some use cases from VyOS. To me there is real value to having one platform that can serve many uses. For example, I run VyOS as a VPN gateway in the cloud but I also use VyOS as a CPE in high end residential and certain small business usecases. UPnP is sort of a necessary evil as unfortunately some devices targeted at those segments simply don't function fully without it. Game consoles get mentioned frequently, but also some VoIP devices, camera systems, and random IoT devices (yes, I know, but I can't stop clients from installing these and I get the call when they don't work).

VyOS supports dynamic dns, which certainly has its share of insecure implementations and arguably is not needed in a tightly-managed IT environment. I would argue UPnP falls in the same camp. I'd rather put effort towards an in-tree solution (where we can ensure `miniupnpd` gets updated regularly) than maintaining a fork or patch.

jack9603301:
> As to arguments against merging UPnP, I think @KittyFort did a good job rebutting most. miniupnpd is pretty mature at this point, and by nature of being widely used, bugs get found and patched pretty quickly. It also has a good variety of hardening options, such as secure_mode (can we default this to yes and require an opt-out insecure-mode instead?) and the ACL-style "rules" list that lets you prohibit clients mapping certain ports.

Perhaps, it’s a good idea to require upnp to enter secure mode by default.

fegauthier:
Hi @jack9603301,

I just tried your PR with the latest current build. I got an error when I commit. I think it's about IPV6 interfaces. The only IPV6 address that I have is on the loopback interface.

image

There is the error I got

image


Do you know what is the problem?

fegauthier:
I fixed it! 

Changing the line 85 in the service_upnp.py

`bits = bin(int(addr['netmask'].replace(':', ''), 16)).count('1')`

by

`bits = bin(int(addr['netmask'].replace(':', '').split('/')[0], 16)).count('1')`

Fixed the issue

jack9603301:
> bits = bin(int(addr['netmask'].replace(':', '').split('/')[0], 16)).count('1')

thanks

fegauthier:
Hi @jack9603301 

It doesn't seems to work. Do you know how can I get miniupnpd logs? If I run miniupnpd standalone, I got these errors

```
invalid option in file /run/upnp/miniupnp.conf line 74 : anchor=VyOS
invalid option in file /run/upnp/miniupnp.conf line 79 : lease_file=/config/upnp.leases
miniupnpd[49644]: system uptime is 112133 seconds
miniupnpd[49644]: mnl_socket bound, port_id=49644
miniupnpd[49644]: send_batch: mnl_cb_run returned -1
miniupnpd[49644]: table_op(0, 1, miniupnpd) send_batch failed -4
miniupnpd[49644]: Failed to init redirection engine. EXITING

```

Do you know what could be wrong?

fegauthier:
@jack9603301 

The actual version of miniupnpd is 2.2.1. I compiled the new version which is 2.2.3 and the problem is the same.

jack9603301:
> @jack9603301
> 
> The actual version of miniupnpd is 2.2.1. I compiled the new version which is 2.2.3 and the problem is the same.

You can freely consult the configuration file of UPnP and compare it. If necessary, you can modify it and run upnpd manually or automatically. If you have any new harvest, please let me know.

jack9603301:
> @jack9603301
> 
> The actual version of miniupnpd is 2.2.1. I compiled the new version which is 2.2.3 and the problem is the same.

If you run it in the same way as the implementation, it should get the correct results. For details, see:

https://www.freebsd.org/cgi/man.cgi?query=miniupnpd&sektion=8

fegauthier:
I finally get it worked, but there is some changes that we need to do.

1. Validator port-range doesn't exist. I got this error

```
vyos@vyos# set service upnp rules 1 external-port-range 1024-65535

  Could not execute validator /usr/libexec/vyos/validators/port-range

  Invalid value
  Value validation failed
  Set failed
```

2. We have to remove anchor option in the config file because it's useless in our case because we use nftables.

dmbaturin (approved):
I'm inclined to merge this, now that 1.3 is out and 1.4 is entering active development. There will be time to refine the implementation.

data/templates/firewall/upnpd.conf.tmpl
```diff
@@ -0,0 +1,172 @@
...
+# Secure Mode, UPnP clients can only add mappings to their own IP
+secure_mode=no
+{% endif %}
+
+{% if presentation_url is defined %}
+# Default presentation URL is HTTP address on port 80
+# If set to an empty string, no presentationURL element will appear
+# in the XML description of the device, which prevents MS Windows
+# from displaying an icon in the "Network Connections" panel.
+#presentation_url= {{ presentation_url }}
```
sever-sever, line 105: Should it be uncommented?

sever-sever (changes requested):

data/templates/firewall/upnpd.conf.tmpl
jack9603301: Do you think I should uncomment it?

PR #838, opened 2021-05-13
vlan: T3532: Fix an error where VLAN protocol could not be modified

jack9603301:
## Change Summary

Fix an error where VLAN protocol could not be modified

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3532

## Component(s) name

vlan

## Proposed changes

The VLAN interface needs to be re created because of the change of VLAN protocol state. Therefore, the VLAN state cannot be changed without the VLAN deletion function (the actual test confirms my conjecture). Every time we create a VLAN, we always need to delete the existing VLAN interface

## How to test

Execution configuration:

```
set interfaces ethernet eth0 vif-s 700 vif-c 20 address 10.0.0.1/24
commit

set interfaces ethernet eth0 vif-s 700 protocol 802.1q
commit

sudo ip -d link show dev eth0.700
15: eth0.700@eth0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state LOWERLAYERDOWN mode DEFAULT group default qlen 1000
    link/ether 00:0d:b9:5a:2e:54 brd ff:ff:ff:ff:ff:ff promiscuity 0 minmtu 0 maxmtu 65535
    vlan protocol 802.1Q id 700 <REORDER_HDR> addrgenmode none numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535
```

## Checklist:

### DIFF
--- a/python/vyos/configdict.py
+++ b/python/vyos/configdict.py
@@ -315,6 +315,33 @@ def is_source_interface(conf, interface, intftype=None):
     old_level = conf.set_level(old_level)
     return ret_val
 
+def node_diff(conf, path):
+    """
+    Returns True when a node has been changed, False otherwise
+    """
+    from vyos.configdiff import get_config_diff
+    D = get_config_diff(conf, key_mangling=('-', '_'))
+    D.set_level(conf.get_level())
+    (new, old) = D.get_value_diff(path)
+    if new != old: 
+        return True
+    else:
+        return False
+
+def T3532_set_encapsulation_changed(conf, base, vlan_config_dict):
+    """
+    This help function is responsible for detecting changes in VLAN encapsulation and setting the
+    """
+    
+    protocol_change = node_diff(conf, base + ['protocol'])
+    ingress_qos_change = node_diff(conf, base + ['ingress_qos'])
+    egress_qos_change = node_diff(conf, base + ['egress_qos'])
+    
+    if protocol_change or ingress_qos_change or egress_qos_change:
+        vlan_config_dict.update({'encapsulation_changed': {}})
+    
+    return vlan_config_dict
+
 def get_interface_dict(config, base, ifname=''):
     """
     Common utility function to retrieve and mangle the interfaces configuration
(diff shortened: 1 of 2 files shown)


c-po:
This change will make the vif interface flap on every configuration change. This is very bad - even changing a device description could cause a BGP session to flap.

jack9603301:
@c-po Let me see if it's possible to modify without deleting the interface...

PS: in my previous test, it was found that even if the setting command did not report errors, there was no actual setting effect on the interface

c-po:
The logic must be changed to detect if the encapsulation mode was changed and only remove the VLAN interface on encapsulation mode change.

See `encapsulation_changed` https://github.com/vyos/vyos-1x/blob/44f766a7880e9264487e62f4aacc9f4f635219ed/src/conf_mode/interfaces-tunnel.py#L127-L131 for reference

c-po:
I do not understand your change. The task/bug is only about `protocol` which is only applicable for `vif-s` interfaces.

jack9603301:
> I do not understand your change. The task/bug is only about `protocol` which is only applicable for `vif-s` interfaces.

I'm still testing, and it's scheduled to cover `vif` and `vif-s`. If the VLAN interface information of the VLAN sub interface is changed, it needs to be re created (such as ingress_qos, egress_qos and protocol)

c-po:
Please only submit a PR when you thing it's ready for testing as otherwise this only consumes time. I actually also started to work on this (that's why I set myself as the assignee in Phabricator) and my solution is somehow much shorter. Maybe I miss some pieces?

jack9603301:
> Please only submit a PR when you thing it's ready for testing as otherwise this only consumes time. I actually also started to work on this (that's why I set myself as the assignee in Phabricator) and my solution is somehow much shorter. Maybe I miss some pieces?

Okay 
Sorry to see the message now, when I finish the test, I will reply

jack9603301:
@c-po It is ready for audit, which means that the latest commit will be able to resolve the problem of invalid VLAN parameter changes.

c-po:
Hi @jack9603301 please check my fix from commit https://github.com/vyos/vyos-1x/commit/cd504035015dca62149b57bc07d8e002bd8723b1 - as this is a different approach. In my testing I discovered that simply checking for a change leaf_node_protocol this is not enough as the testcase would still fail.

We also already have an implementation of `node_diff()` called `leaf_node_changed()`.

Please check if I made a mistake but comparing against the actual driver configuration is more efficient then I initially said.

You can find my testcase here: https://github.com/vyos/vyos-1x/commit/7eedd045534c1375e3a02752412d1652d10b0fa1

If you think my implementation is wrong please submit a new fixup PR.

jack9603301:
@c-po Seems simpler than my implementation, well done

PR #843, opened 2021-05-16
vlan: T3551: Repair VLAN sub interface QoS control failure

jack9603301:
## Change Summary

Repair VLAN sub interface QoS control failure

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3551

## Component(s) name

vif

## Proposed changes

Repair VLAN sub interface QoS control failure

## Relation

https://github.com/vyos/vyatta-cfg-system/pull/147

## Checklist:


interface-definitions/include/interface/vif.xml.i
```diff
@@ -22,25 +22,29 @@
...
+          <format>from:to</format>
+          <description>The format is FROM:TO with multiple mappings separated by spaces.</description>
+        </valueHelp>
         <constraint>
-          <regex>[:0-7 ]+$</regex>
+          <regex>[:0-7]+$</regex>
         </constraint>
-        <constraintErrorMessage>QoS mapping should be in the format of '0:7 2:3' with numbers 0-9</constraintErrorMessage>
+        <constraintErrorMessage>QoS mapping should be in the format of '0:7' with numbers 0-9</constraintErrorMessage>
+        <multi />
```
c-po: If multiple values are supported by a space delimiter, why use a multi node?

interface-definitions/include/interface/vif.xml.i
```diff
@@ -22,25 +22,29 @@
...
       </properties>
     </leafNode>
     <leafNode name="ingress-qos">
       <properties>
         <help>VLAN ingress QoS</help>
-        <completionHelp>
-          <script>echo Format for qos mapping '0:1 1:6 7:6'</script>
-        </completionHelp>
+        <valueHelp>
+          <format>from:to</format>
```
c-po: Much better then the completion helper, very Good!

python/vyos/ifconfig/interface.py
```diff
@@ -1345,12 +1345,53 @@ def update(self, config):
...
+            vif_config['ifname'] = vif_ifname
+            
             tmp = deepcopy(VLANIf.get_config())
             tmp['source_interface'] = ifname
             tmp['vlan_id'] = vif_id
+            
+            if 'egress_qos' in vif_config:
+                egress_qos_str = ''
+                for qos_map in vif_config['egress_qos']:
+                    egress_qos_str += f'{qos_map} '
```
c-po: Any reason to not use Python join() function?

c-po (changes requested):

interface-definitions/include/interface/vif.xml.i
jack9603301: The reason why null values are not used is to ensure format consistency. When a user uses more than one space in a string, adding or removing a space will trigger remove. In fact, there is no need to change the VLAN wrapper interface

python/vyos/ifconfig/interface.py
jack9603301: done

c-po:
Please ensure first that the functionality of the old VyOS 1.2 implementation is honored before changing the CLI and making this a multi node. To me changing this to a multi node seems to not be the right way at the moment. We should fix the behavior first!

```
vyos@vyos# run show ver
Version:          VyOS 1.2.6
Release Train:    crux

Built by:         Sentrium S.L.
Built on:         Fri 11 Sep 2020 09:26 UTC
Build UUID:       b7236651-5383-482d-93e5-043470b5e2e0
Build Commit ID:  706d01f247bb83
```

```
vyos@vyos# set interfaces ethernet eth0 vif 10 ingress-qos "0:1 1:6 7:6"
vyos@vyos# commit
```

```
vyos@vyos# ip -d link show eth0.10
6: eth0.10@eth0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state LOWERLAYERDOWN mode DEFAULT group default qlen 1000
    link/ether 50:00:00:02:00:00 brd ff:ff:ff:ff:ff:ff promiscuity 0 minmtu 0 maxmtu 65535 
    vlan protocol 802.1Q id 10 <REORDER_HDR> 
      ingress-qos-map { 0:1 1:6 7:6 } addrgenmode eui64 numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535
```

jack9603301:
@c-po Do you want to maintain the CLI format at this stage? Usually, this is feasible. However, because it is uncertain how many spaces the user will input in the string, it is necessary to split the set string into an array and then splice it into a string. Otherwise, there will be problems in encapsulation detection

Changing the CLI format will result in one step less processing

c-po:
Please maintain the format. Why is the amount of spaces uncertain? Just enforce one whitespace by regex (like in the past)

https://github.com/vyos/vyatta-cfg-system/blob/crux/templates/interfaces/ethernet/node.tag/vif/node.tag/ingress-qos/node.def#L4

jack9603301:
> Please maintain the format. Why is the amount of spaces uncertain? Just enforce one whitespace by regex (like in the past)
> 
> https://github.com/vyos/vyatta-cfg-system/blob/crux/templates/interfaces/ethernet/node.tag/vif/node.tag/ingress-qos/node.def#L4

@c-po I'll try the solution you mentioned and see if I can fix it by modifying the regular expression, which is exactly what I'm hesitating about right now, and to be honest, I was thinking about it before implementing the solution.

Using regular `[:0-7 ]+$` seems the easiest, but if the following two values exist:

```
a) 0:7 1:6 3:5
b) 0:7 1:6     3:5
```

Obviously, user `b` has just `a` little more space than user `a`, but in the configurator, two completely different Settings are detected. As I said earlier, to ensure that the old, outdated Settings are always deleted, the VLAN subinterface will be deleted and rebuilt when the values change. But now, the user has just added a few Spaces without actually changing the Settings, which will trigger remove as well.

To solve this problem, the configurator needs to split the string into an array and then assemble it again. While it doesn't take much code (just one sentence), changing the CLI to be many-valued obviously saves a step

jack9603301:
@c-po  Keep the old CLI format 

done

python/vyos/ifconfig/interface.py
```diff
@@ -1373,7 +1416,7 @@ def _create(self):
             cmd += ' egress-qos-map {egress_qos}'
 
         self._cmd(cmd.format(**self.config))
-
+
```
c-po: Please do not add empty whitespace lines.

interface-definitions/include/interface/vif.xml.i
```diff
@@ -22,9 +22,10 @@
     <leafNode name="egress-qos">
       <properties>
         <help>VLAN egress QoS</help>
-        <completionHelp>
-          <script>echo Format for qos mapping, e.g.: '0:1 1:6 7:6'</script>
-        </completionHelp>
+        <valueHelp>
+          <format>txt</format>
+          <description>Format for qos mapping, e.g.: '0:1 1:6 7:6'.</description>
```
c-po: Please remove trailing dot

interface-definitions/include/interface/vif.xml.i
```diff
@@ -34,9 +35,10 @@
     <leafNode name="ingress-qos">
       <properties>
         <help>VLAN ingress QoS</help>
-        <completionHelp>
-          <script>echo Format for qos mapping '0:1 1:6 7:6'</script>
-        </completionHelp>
+        <valueHelp>
+          <format>txt</format>
+          <description>Format for qos mapping, e.g.: '0:1 1:6 7:6'.</description>
```
c-po: Please remove trailing dot

smoketest/scripts/cli/base_interfaces_test.py
```diff
@@ -278,18 +278,119 @@ def test_vif_8021q_interfaces(self):
...
             self.cli_commit()
 
             for intf in self._interfaces:
                 for vlan in self._vlan_range:
                     vif = f'{intf}.{vlan}'
+                    tmp = get_interface_config(f'{vif}')
+                    
+                    tmp2 = dict_search('linkinfo.info_data.ingress_qos', tmp)
+                    if tmp2:
+                        from_key = tmp2[0]['from']
```
c-po: Why not use the same code snipped as in `interface.py`?

```python
for item in tmp2:
    from_key = item['from']
    to_key = item['to']
```

c-po (changes requested):

python/vyos/ifconfig/interface.py
jack9603301: done

interface-definitions/include/interface/vif.xml.i
jack9603301: done

interface-definitions/include/interface/vif.xml.i
jack9603301: done

smoketest/scripts/cli/base_interfaces_test.py
jack9603301: done

jack9603301:
@c-po done

smoketest/scripts/cli/base_interfaces_test.py
c-po: There are other places the old syntax still persists.

jack9603301: done

PR #846, opened 2021-05-18
T3508, T3356: remote: Friendly download procedure for user-facing scripts

erkin:
## Change Summary

This commit introduces a `friendly_download()` to be used in interactive scripts (mainly `install-image`). See: vyos/vyatta-cfg-system#146.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3356
* https://phabricator.vyos.net/T3508

## Component(s) name

remote

## Proposed changes

`friendly_download()` checks the file size in advance before downloading and compares it against the available storage and prints errors without the lengthy stack trace.

## How to test

Mostly intended to be called from scripts but it can be used directly like
```
> vyos.remote.friendly_download('/tmp/foobar', 'ftp://ftp.example.net/foobar', authentication=('username', 'password'))
```

## Checklist:

### DIFF
--- a/python/vyos/remote.py
+++ b/python/vyos/remote.py
@@ -15,6 +15,7 @@
 
 from ftplib import FTP
 import os
+import shutil
 import socket
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


dmbaturin (commented):
As I mentioned in the other PR, I'm concerned about passing plaintext passwords as process arguments since it may end up visible in `ps`. The AWS SDK and similar projects usually pass sensitive data in environment variables.

erkin:
That's a good point, especially considering the process runs for a nontrivial time. Converted to draft for the time being.

erkin:
It's also trivial to get environment variables out of a running process.
```
% PASSWORD=swordfish sleep 500 &
% strings /proc/$(pgrep sleep)/environ | grep PASSWORD
PASSWORD=swordfish
```
Better than passing it directly though.

Edit: Ah, wait, default user has no permissions for that. All right, it is decided then.

c-po:
When uploading/downloading data, can't the Python3 requests library be used to setup a session which hold credentials?

dmbaturin:
@c-po Since the code supports multiple protocols and thus uses multiple unrelated libraries apart from `requests`, a mechanism unique to `requests` is of little help in the big picture.

erkin:
`requests` would be helpful with authentication (which we handle by installing a password manager each time we do HTTP authentication, which is only somewhat better than manually adding the HTTP header) and would give us a non-painful way to do HTTP uploading (which is a would-be-nice feature). I opted to rely on `urllib.request` alone for the time being to avoid adding another dependency. (Speaking of which, `urllib.request` has a superficial FTP support so it might be a good idea to migrate the FTP implementation from `ftplib` to `urllib.request` for the sake of uniformity.)

The current pain point with what we have so far is the lack of a progress bar to replace `curl`'s, which is rather painful to do manually across different protocols. The user installing a new image has no feedback from the system about the progress of the download.

c-po:
As https://github.com/vyos/vyatta-cfg-system/pull/146 was merged early we also need to merge this. Any changes to authentication should be added in a subsequent PR. Sorry for the trouble

PR #863, opened 2021-06-03
dhcp6relay: T3596: Replace ISC DHCPv6 relay with wide-dhcp6-relay

jack9603301:
## Change Summary

Replace ISC DHCPv6 relay with wide-dhcp6-relay

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3596

## Component(s) name

dhcp6relay

## Proposed changes

I plan to use wide-dhcp6-relay to support DHCPv6 relay, which may break the limitation of ISC DHCP on PPPoE interface.

## How to test

Since the test topology has not been designed yet, only packet capture test is carried out (to confirm whether DHCPv6 PD request packets are sent)

## Checklist:

### DIFF
--- a/data/templates/dhcp-relay/dhcp6relay.conf.tmpl
+++ b/data/templates/dhcp-relay/dhcp6relay.conf.tmpl
@@ -1,0 +1,22 @@
+{% set option = namespace(value='')  %}
+
+{% if dhcpv6_options is defined and dhcpv6_options.dhcp6relay is defined %}
+{%    if dhcpv6_options.dhcp6relay.upstream_address is defined %}
+{%      set option.value = option.value + '-s ' + dhcpv6_options.dhcp6relay.upstream_address + ' ' %}
+{%    endif  %}
+{%    if dhcpv6_options.dhcp6relay.boundaddr is defined %}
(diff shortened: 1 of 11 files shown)


c-po:
Hi Jack,

I wonder why this is flagged only as refactoring bit you open an entire new CLI tree. I also wo der why this is now under the interface secrion and no longer under service.

Please explain in a more detailed manner.

jack9603301:
@c-po 

In fact, I designed it as fix at the beginning, but I found that it almost needs to modify the existing service logic (I can't reuse the old logic, I need to modify the CLI), so I changed it to refactoring

> I also wo der why this is now under the interface secrion and no longer under service.

This is mainly because the configuration structure of the interface is convenient and friendly, and the PD configuration is in the same path. This is just a draft. If necessary, you can change the CLI

Traditionally, only one configuration is needed:

```
set interfaces ethernet eth0 dhcpv6-option dhcp6relay interface eth1
```

The above configuration takes `eth0` as the upstream and provides services for `eth1`

dmbaturin:
We'll need to re-evaluate this in 1.5. Closing for now.

PR #865, opened 2021-06-04
nhrp: T3599: Migrate NHRP to XML/Python

sarthurdev:
## Change Summary

Migrates NHRP to XML/Python

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3599

## Component(s) name

nhrp

## Proposed changes

Migrate code to XML/Python, still uses opennhrp instead of integrating with FRR due to shortcomings in the implementation. Removes dependency of `vyos-nhrp` package.

I am watching developments with FRR's nhrpd implementation and hope in the near future it will have feature parity with opennhrp and VyOS can switch over.

## How to test

Config syntax is verified to being generated identically. Functionality has been verified in a DMVPN configuration.

## Checklist:

### DIFF
--- a/src/conf_mode/protocols_nhrp.py
+++ b/src/conf_mode/protocols_nhrp.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) 2021 VyOS maintainers and contributors
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or later as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from vyos.config import Config
+from vyos.configdict import node_changed
+from vyos.template import render
+from vyos.util import process_named_running
+from vyos.util import run
+from vyos import ConfigError
+from vyos import airbag
+airbag.enable()
+
+opennhrp_conf = '/run/opennhrp/opennhrp.conf'
+
+def get_config(config=None):
+    if config:
+        conf = config
+    else:
+        conf = Config()
+    base = ['protocols', 'nhrp']
+
+    nhrp = conf.get_config_dict(base, key_mangling=('-', '_'),
+                                get_first_key=True, no_tag_node_value_mangle=True)
+    nhrp['del_tunnels'] = node_changed(conf, base + ['tunnel'], key_mangling=('-', '_'))
+
+    if not conf.exists(base):
+        return nhrp
+
+    nhrp['if_tunnel'] = conf.get_config_dict(['interfaces', 'tunnel'], key_mangling=('-', '_'),
+                                get_first_key=True, no_tag_node_value_mangle=True)
+
+    nhrp['profile_map'] = {}
+    profile = conf.get_config_dict(['vpn', 'ipsec', 'profile'], key_mangling=('-', '_'),
+                                get_first_key=True, no_tag_node_value_mangle=True)
+
+    for name, profile_conf in profile.items():
+        if 'bind' in profile_conf and 'tunnel' in profile_conf['bind']:
+            interfaces = profile_conf['bind']['tunnel']
+            if isinstance(interfaces, str):
+                interfaces = [interfaces]
+            for interface in interfaces:
+                nhrp['profile_map'][interface] = name
+
+    return nhrp
+
+def verify(nhrp):
+    if 'tunnel' in nhrp:
+        for name, nhrp_conf in nhrp['tunnel'].items():
+            if not nhrp['if_tunnel'] or name not in nhrp['if_tunnel']:
+                raise ConfigError(f'Tunnel interface "{name}" does not exist')
+
+            tunnel_conf = nhrp['if_tunnel'][name]
+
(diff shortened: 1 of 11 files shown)


src/conf_mode/protocols_nhrp.py
```diff
@@ -0,0 +1,139 @@
...
+                for map_name, map_conf in nhrp_conf['dynamic_map'].items():
+                    if 'nbma_domain_name' not in map_conf:
+                        raise ConfigError(f'nbma-domain-name missing on dynamic-map {map_name} on tunnel {name}')
+    return None
+
+def generate(nhrp):
+    render('/etc/opennhrp/opennhrp.conf', 'nhrp/opennhrp.conf.tmpl', nhrp)
+    return None
+
+def apply_iptables(iface, tunnel_conf):
```
c-po: As `apply_iptables` and `remove_iptables` is only called once, please not move it to a function.

Also we try to not add extra `try` / `catch` blocks to our code. If you really really on the error code, check it.

src/conf_mode/protocols_nhrp.py
```diff
@@ -0,0 +1,139 @@
...
+    if not tunnel_conf or 'source_address' not in tunnel_conf:
+        return None
+
+    chain = f'VYOS_NHRP_{iface}_OUT_HOOK'
+    source_addr = tunnel_conf['source_address']
+    try:
+        cmd(f'sudo iptables -N {chain}')
+        cmd(f'sudo iptables -A {chain} -p gre -s {source_addr} -d 224.0.0.0/4 -j DROP')
```
c-po: This could become a problem - I had the same in `conntrack.py`. If only minor parameters of the tunnel are adjusted, the iptables rule is installed multiple times, thus you should `iptables --check` if the rule probably already exists.

Please also use the long option names as it makes it easier to read.

src/conf_mode/protocols_nhrp.py
```diff
@@ -0,0 +1,139 @@
...
+        tunnel_conf = nhrp['if_tunnel'][tunnel]
+        apply_iptables(tunnel, tunnel_conf)
+
+    for tunnel in nhrp['del_tunnels']:
+        remove_iptables(tunnel)
+
+    action = 'stop'
+
+    if 'tunnel' in nhrp:
+        action = 'restart' if process_named_running('opennhrp') else 'start'
```
c-po: Well, we usually always simply `restart` the service as it will auto-start if not already running.

c-po (changes requested):

src/conf_mode/protocols_nhrp.py
sarthurdev: Done

src/conf_mode/protocols_nhrp.py
```diff
@@ -0,0 +1,124 @@
...
+    if config:
+        conf = config
+    else:
+        conf = Config()
+    base = ['protocols', 'nhrp']
+
+    nhrp = conf.get_config_dict(base, key_mangling=('-', '_'),
+                                get_first_key=True, no_tag_node_value_mangle=True)
+
+    diff = get_config_diff(conf, key_mangling=('-', '_'))
```
c-po: Can you please use an already established pattern which can be found here: https://github.com/vyos/vyos-1x/blob/current/src/conf_mode/interfaces-bridge.py#L55-L61

That would make all our lifes easier ;)

src/conf_mode/protocols_nhrp.py
```diff
@@ -0,0 +1,124 @@
...
+    return None
+
+def apply(nhrp):
+    for tunnel in nhrp['new_tunnels']:
+        tunnel_conf = nhrp['if_tunnel'][tunnel]
+        if 'source_address' in tunnel_conf:
+            chain = f'VYOS_NHRP_{tunnel}_OUT_HOOK'
+            source_address = tunnel_conf['source_address']
+
+            chain_exists = run(f'sudo iptables --check {chain} -j RETURN')
```
c-po: This reads like inverse logic to me. chain_exists = 1 -> iptables chain does not exist. chain_exists  = 0, iptables chain exists. COuld you please reword the variable name? Or simply add a comment and name the variable `rc`, `ret` or `tmp`.

src/conf_mode/protocols_nhrp.py
```diff
@@ -0,0 +1,124 @@
...
+                        raise ConfigError(f'nbma-address missing on map {map_name} on tunnel {name}')
+
+            if 'dynamic_map' in nhrp_conf:
+                for map_name, map_conf in nhrp_conf['dynamic_map'].items():
+                    if 'nbma_domain_name' not in map_conf:
+                        raise ConfigError(f'nbma-domain-name missing on dynamic-map {map_name} on tunnel {name}')
+    return None
+
+def generate(nhrp):
+    render('/etc/opennhrp/opennhrp.conf', 'nhrp/opennhrp.conf.tmpl', nhrp)
```
c-po: Please assign the configuration name to a variable and ensure it is unlinked from the filesystem on service removal. Please also render it to `/run` so the configuration will never survive a reboot.

src/systemd/opennhrp.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=OpenNHRP
+After=vyos-router.service
+ConditionPathExists=/etc/opennhrp/opennhrp.conf
+StartLimitIntervalSec=0
+
+[Service]
+Type=forking
+ExecStart=/usr/sbin/opennhrp -d -v -a /run/opennhrp.socket -c /etc/opennhrp/opennhrp.conf -s /etc/opennhrp/opennhrp-script.py -p /run/opennhrp.pid
```
c-po: Please render config to `/run/opennhrp`

c-po (changes requested):

c-po:
Besides this last round of requested changed this look good to me. Thank you very much for your contributions.

PR #887, opened 2021-06-19
mdns: vrrp: T3635: Add ability to use mDNS repeater with VRRP

sarthurdev:
## Change Summary

This allows the mDNS repeater to be enabled on VRRP-enabled routers as only the master router will repeat the packets.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3635

## Component(s) name

mdns, vrrp

## Proposed changes

When the new setting is active, mDNS repeater will disable mDNS on interfaces where VRRP is active and the state is not MASTER.
A hook is also added to keepalived_fifo to keep the mDNS repeater updated as VRRP states change.

## How to test

```
set service mdns repeater interface eth0
set service mdns repeater interface eth1
set service mdns repeater disable-on-vrrp
```
Assuming VRRP is enabled on eth0 and eth1, if the router is not the master router mDNS will be disabled to prevent a packet storm. If the state changes to master, mDNS will then be enabled on each master interface.

## Checklist:

### DIFF
--- a/src/conf_mode/service_mdns-repeater.py
+++ b/src/conf_mode/service_mdns-repeater.py
@@ -16,10 +16,12 @@
 
 import os
 
+from json import loads
 from sys import exit
 from netifaces import ifaddresses, interfaces, AF_INET
 
 from vyos.config import Config
+from vyos.ifconfig.vrrp import VRRP
 from vyos.template import render
 from vyos.util import call
 from vyos import ConfigError
@@ -27,6 +29,7 @@ from vyos import airbag
 airbag.enable()
 
 config_file = r'/etc/default/mdns-repeater'
+vrrp_running_file = '/run/mdns_vrrp_active'
 
 def get_config(config=None):
     if config:
@@ -35,6 +38,9 @@ def get_config(config=None):
         conf = Config()
     base = ['service', 'mdns', 'repeater']
     mdns = conf.get_config_dict(base, key_mangling=('-', '_'), get_first_key=True)
+
+    if mdns:
+        mdns['vrrp_exists'] = conf.exists('high-availability vrrp')
     return mdns
 
 def verify(mdns):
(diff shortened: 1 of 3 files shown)


src/conf_mode/service_mdns-repeater.py
```diff
@@ -16,17 +16,20 @@
...
 from vyos.config import Config
+from vyos.ifconfig.vrrp import VRRP
 from vyos.template import render
 from vyos.util import call
 from vyos import ConfigError
 from vyos import airbag
 airbag.enable()
 
 config_file = r'/etc/default/mdns-repeater'
+vrrp_running_file = '/run/mdns_vrrp_active'
```
c-po, line 32: Why not simply check if keepalived is running?

`vyos.util.process_named_running('keepalived')`

src/helpers/vyos-vrrp-mdns.sh
```diff
@@ -0,0 +1,24 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+case "$3" in
+    "MASTER"|"BACKUP"|"FAULT")
+        if [ -f /run/mdns_vrrp_active ]; then
```
c-po: There are already VRRP/keepalived state transition scripts but I'm not into detail. Maybe @DmitriyEshenko knows more?

src/conf_mode/service_mdns-repeater.py
sarthurdev: That file is only created when `disable-on-vrrp` is set and there are VRRP groups. Only if it exists will the handler call the mdns repeater conf script on state changes.

src/helpers/vyos-vrrp-mdns.sh
sarthurdev: This handler is called by the existing transition script: https://github.com/vyos/vyos-1x/pull/887/commits/3a77a1f25aabd92a1785fcdc8c89d65a2cf8d5de#diff-33fc5560ca57968dd54276c71f0a4c5e29f2ce34ff4d937b58672e112a19c75aR127

c-po:
Can we name it `disable-vrrp` or `vrrp-disable`?

sarthurdev:
> 
> 
> Can we name it `disable-vrrp` or `vrrp-disable`?

Updated name to `vrrp-disable` so it doesn't also share autocomplete with `disable`.

Also removed the handler shell script and let keepalive_fifo call the python script directly, as it doesn't really need to go through a shell script.

PR #889, opened 2021-06-20
T3506: loadkey: Add `generate public-key-command` command

erkin:
## Change Summary

A new command for adding OpenSSH public keys from files, consistent with the VyOS CLI to deprecate `loadkey`. See vyos/vyatta-cfg-system#150

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3506

## Component(s) name

loadkey

## Proposed changes

`loadkey` has always been a wart in conf-mode. Per Daniil's suggestion, instead of automatically adding the key, this replacement simply prints out the commands for the user to manually add the public key. The instruction line is in the form of a shell comment to allow the user to pipe it to shell, should the user prefer to add it automatically anyway.

## How to test

```
$ generate public-key-command user alyssa path sftp://example.net/home/alyssa/.ssh/id_rsa.pub
# To add this key as an embedded key, run the following commands:
configure
set system login user alyssa authentication public-keys alyssa@example.net key AAA...
set system login user alyssa authentication public-keys alyssa@example.net type ssh-rsa
commit
exit

$ generate public-key-command user ben path ~/.ssh/id_rsa.pub
# To add this key as an embedded key, run the following commands:
configure
set system login user ben authentication public-keys ben@vyos key AAA...
set system login user ben authentication public-keys ben@vyos type ssh-dss
commit
exit
```

## Checklist:


erkin:
It's strange that CI isn't building the interface definition. I used an XML validator to check its correctness and `build-command-op-templates op-mode-definitions/generate-public-key-command.xml.in schema/op-mode-definition.rng .` produces the correct node files that work correctly in a clean install. Perhaps, somewhere down the line, RelaxNG's `text` attribute chokes on the square brackets. I can try using `CDATA` attributes to properly quote the URL templates, which would be much more readable than this character entity reference mess.

Edit: See vyos/vyos-1x#891

PR #909, opened 2021-07-04
containers: T3662: Repair system damage during upgrade

jack9603301:
## Change Summary

Repair system damage during upgrade

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3662

## Component(s) name

containers

## Proposed changes

Because the new vyos system may not have the OCI runtime `runc` used by podman by default, but there is a crun, resulting in the container configuration can not be used normally. We need to point to the `crun` through the soft link when the system can not find the `runc`.

This patch will fix the above problem.

## How to test

When upgrading a vyos system containing container configuration, the configuration will be destroyed, such as the following configuration:

```
set interfaces ethernet eth0 ipv6 address autoconf
set system name-service 114.114.114.114
set container name nginx image nginx
set container name nginx allow-host-networks
commit
save
exit
```

After upgrading to vyos compiled by this branch, Restart after the first upgrade:
图片

## Checklist:

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -134,6 +134,7 @@ Depends:
   python3-zmq,
   qrencode,
   radvd,
+  runc,
   salt-minion,
   sed,
   smartmontools,
--- a/smoketest/scripts/cli/test_container.py
+++ b/smoketest/scripts/cli/test_container.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
 import unittest
 import json
 
@@ -75,6 +76,10 @@ class TesContainer(VyOSUnitTestSHIM.TestCase):
 
         self.assertEqual(json_subnet, prefix)
         self.assertEqual(json_ip, cont_ip)
+    
+    def test_oci_runtime_exists(self):
+        self.assertTrue(os.path.exists('/usr/bin/runc'))
+            
 
 if __name__ == '__main__':
     unittest.main(verbosity=2)
--- a/src/conf_mode/containers.py
+++ b/src/conf_mode/containers.py
@@ -36,6 +36,8 @@ airbag.enable()
 config_containers_registry = '/etc/containers/registries.conf'
 config_containers_storage = '/etc/containers/storage.conf'
 
+runtime = '/usr/bin/runc'
+
 def _cmd(command):
     if os.path.exists('/tmp/vyos.container.debug'):
         print(command)
@@ -244,18 +246,18 @@ def apply(container):
                         volume += f' -v {svol}:{dvol}'
 
                 if 'allow_host_networks' in container_config:
-                    _cmd(f'podman run -dit --name {name} --net host {port} {volume} {env_opt} {image}')
+                    _cmd(f'podman --runtime {runtime} run -dit --name {name} --net host {port} {volume} {env_opt} {image}')
                 else:
                     for network in container_config['network']:
                         ipparam = ''
                         if 'address' in container_config['network'][network]:
                             ipparam = '--ip ' + container_config['network'][network]['address']
-                        _cmd(f'podman run --name {name} -dit --net {network} {ipparam} {port} {volume} {env_opt} {image}')
+                        _cmd(f'podman --runtime {runtime} run --name {name} -dit --net {network} {ipparam} {port} {volume} {env_opt} {image}')
 
             # Else container is already created. Just start it.
             # It's needed after reboot.
             elif container_status(name) != 'running':
-                _cmd(f'podman start {name}')
+                _cmd(f'podman --runtime {runtime} start {name}')
 
     return None
 


sever-sever:
If you place storage to **/var** how it will copy container images/contents/metadata between updates?

jack9603301:
> If you place storage to **/var** how it will copy container images/contents/metadata between updates?

Volume map can be used to map data area persistence to `/config`, which is in line with the best practice of container microservice technology

Only valid persistent data needs to be copied, and no other data and metadata need to be copied. When the new system is configured for the first time, the image will be pulled directly from the docker. The container itself is suitable for stateless services. If the persistent data mapping is valid and undamaged, it should be able to be used by the newly pulled container image.

sever-sever:
Check this behavior.
Update system image, Disable interfaces or remove the default route. To simulate not internet connection.
After that, reboot the router to the updated version.

jack9603301:
At the same time, I increased the priority, which will make the container configuration execute after the network configuration.

After the system is upgraded and rebooted for the first time, all configurations will be migrated to the new system. Due to the first pull, it is necessary to connect to the Internet. This is a normal behavior. If enterprises need to pull from the intranet, they can easily establish a private image server such as harbor. Podman can pull from the intranet server instead of the public network

Therefore, this situation generally does not need to be considered

jack9603301:
In the case of no network, the network configuration is normal due to priority, but the service of container (priority > = 1200) cannot be configured

图片
图片
图片

Before applying this patch, when a user using container configuration upgrades to a new image, the system will not be able to configure the network (lack of priority) due to container configuration failure. At the same time, the container configuration appears to have been corrupted, so you must delete and reconfigure the container

data/templates/containers/storage.tmpl
```diff
@@ -2,4 +2,4 @@
 
 [storage]
   driver = "vfs"
-  graphroot = "/config/containers/storage"
+  graphroot = "/var/containers/storage"
```
c-po: This will not survive an image upgrade

c-po (changes requested):

data/templates/containers/storage.tmpl
jack9603301: For containers that support volume mapping, it does not need to exist

jack9603301: After the upgrade is started for the first time, the image will be pulled again, and the persistent storage can be used normally through volume mapping (yes, as long as the user sets the container correctly, this is in line with the best practice of container and microservice Technology)

sever-sever:
As for me, it needs to find out the root cause.
Maybe needs to add some additional keys/values for section **[storage]** or use another **driver**.

jack9603301:
@c-po @sever-sever Sorry, I didn't mean anything else. I just think that this problem can be solved according to normal container best practice. If you persist metadata permanently, I have a feeling that volume persistence mapping has lost its role

Persistent volume mapping has been implemented here:

https://github.com/vyos/vyos-1x/pull/818

Example, Use a similar command line:

```
set container name adguardhome allow-host-networks
set container name adguardhome description 'AdGuardHome'
set container name adguardhome image 'adguard/adguardhome'
set container name adguardhome volume conf destination '/opt/adguardhome/conf'
set container name adguardhome volume conf source '/config/adguardhome/conf'
set container name adguardhome volume work destination '/opt/adguardhome/work'
set container name adguardhome volume work source '/config/adguardhome/work'
set container name pac allow-host-networks
set container name pac description 'PAC-NGINX'
set container name pac image 'nginx'
set container name pac volume nginx-conf destination '/etc/nginx/conf.d'
set container name pac volume nginx-conf source '/config/pac/vhost'
set container name pac volume pac destination '/www/root/default'
set container name pac volume pac source '/config/pac/www'
```

For the home network, it can be pulled from the external network. When upgrading the system in the enterprise and it needs to be pulled from the internal network, a dedicated server such as Harhor can be used to store and pull the image.

:)

https://goharbor.io/

sever-sever:
Most likely it will work.
```
root@r1-roll:/home/vyos# podman --runtime /usr/bin/crun run --net host --rm fedora echo it works
Error: OCI runtime error: writing file `/sys/fs/cgroup/cgroup.subtree_control`: Invalid argument
root@r1-roll:/home/vyos# 

root@r1-roll:/home/vyos# podman --runtime /usr/bin/runc run --net host --rm fedora echo it works
it works
root@r1-roll:/home/vyos#

```

sever-sever (approved):

sever-sever (commented):
Most likely it will work.
```
root@r1-roll:/home/vyos# podman --runtime /usr/bin/crun run --net host --rm fedora echo it works
Error: OCI runtime error: writing file `/sys/fs/cgroup/cgroup.subtree_control`: Invalid argument
root@r1-roll:/home/vyos# 

root@r1-roll:/home/vyos# podman --runtime /usr/bin/runc run --net host --rm fedora echo it works
it works
root@r1-roll:/home/vyos#

```
But in my test vm by default, after install runc it doesn't use runc
```
root@r1-roll:/home/vyos# podman  run --net host --rm fedora echo it works
Error: OCI runtime error: writing file `/sys/fs/cgroup/cgroup.subtree_control`: Invalid argument
root@r1-roll:/home/vyos# 

```
ociRuntime still **crun**
```
root@r1-roll:/home/vyos# podman info | grep ociRuntime -A 8
  ociRuntime:
    name: crun
    package: 'crun: /usr/bin/crun'
    path: /usr/bin/crun
    version: |-
      crun version 0.17
      commit: 0e9229ae34caaebcb86f1fde18de3acaf18c6d9a
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL

```

jack9603301:
@sever-sever 
图片
Strangely enough, I can't reproduce this mistake

sever-sever:
The root cause in changing in **cgroups**/cgroup2 with disabled option **--cgroups disabled** I don't see any errors, even with **crun**
```
root@r1-roll:/home/vyos# podman  run --cgroups disabled --net host --rm fedora echo it works
it works
root@r1-roll:/home/vyos#
```

Debian 10 use cgroup v1
```
vyos@r1-roll:~$ sudo mount | grep cgroup
tmpfs on /sys/fs/cgroup type tmpfs (ro,nosuid,nodev,noexec,mode=755)
cgroup2 on /sys/fs/cgroup/unified type cgroup2 (rw,nosuid,nodev,noexec,relatime,nsdelegate)
cgroup on /sys/fs/cgroup/systemd type cgroup (rw,nosuid,nodev,noexec,relatime,xattr,name=systemd)
cgroup on /sys/fs/cgroup/net_cls,net_prio type cgroup (rw,nosuid,nodev,noexec,relatime,net_cls,net_prio)
cgroup on /sys/fs/cgroup/cpuset type cgroup (rw,nosuid,nodev,noexec,relatime,cpuset)
cgroup on /sys/fs/cgroup/cpu,cpuacct type cgroup (rw,nosuid,nodev,noexec,relatime,cpu,cpuacct)
cgroup on /sys/fs/cgroup/blkio type cgroup (rw,nosuid,nodev,noexec,relatime,blkio)
cgroup on /sys/fs/cgroup/devices type cgroup (rw,nosuid,nodev,noexec,relatime,devices)
cgroup on /sys/fs/cgroup/pids type cgroup (rw,nosuid,nodev,noexec,relatime,pids)
cgroup on /sys/fs/cgroup/memory type cgroup (rw,nosuid,nodev,noexec,relatime,memory)
cgroup on /sys/fs/cgroup/perf_event type cgroup (rw,nosuid,nodev,noexec,relatime,perf_event)
cgroup on /sys/fs/cgroup/freezer type cgroup (rw,nosuid,nodev,noexec,relatime,freezer)
vyos@r1-roll:~$
root@r1-roll:/home/vyos# podman info
host:
  arch: amd64
  buildahVersion: 1.19.6
  cgroupManager: systemd
  cgroupVersion: v1
```
Debian 11 use cgroup v2
```
vyos@r1-roll:~$ sudo mount | grep cgroup
cgroup2 on /sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime,nsdelegate,memory_recursiveprot)
vyos@r1-roll:~$ 
root@r1-roll:/home/vyos# podman info
host:
  arch: amd64
  buildahVersion: 1.19.6
  cgroupManager: systemd
  cgroupVersion: v2
```

sever-sever:
In the latest rolling release all works fine without any changes
```
vyos@r1-roll:~$ show version 

Version:          VyOS 1.4-rolling-202107010537
Release Train:    sagitta

vyos@r1-roll:~$ sudo podman  run --net host --rm fedora echo it works
it works
vyos@r1-roll:~$ 
```

jack9603301:
@sever-sever 
Why do you have problems after adding runc?

jack9603301:
note: Record the process of upgrading from 1.4-rolling-202107010537 to 1.4-rolling-202107122017

After the first startup, the container fails to start. Restart the operating system and start normally.

Wait for the next test to determine if this problem is resolved!

jack9603301:
图片

Unfortunately, it seems that the same problem still happened. Although it seems that the container can be started normally after restart due to priority adjustment and some factors, it is not a normal operation result!

c-po:
I do not understand why adding the `runc` dependency should fix this issue. I just upgraded to the latest 1.4 rolling release and the container was starting up as expected with all the data present.

I propose to close this PR.

jack9603301:
@c-po I will re evaluate whether the problem disappears. If it does, the PR will be closed

jack9603301:
图片
@c-po This problem still exists and needs to be normal after a second full restart, but one of the errors may be caught

c-po:
I have just tried this PR out of curiosity but as expected it does not work, `runc` is like `podman` a program to run containers and used by e.g. Docker.

The issue is that podman does not find the container image on first invocation on the first boot but on any other subsequent boot.

c-po:
Closing this as this does not solve the root cause. The root cause "seems" to be the fact that on system bootup and container initialisation there might still not be a full propagated routing table so podman can reach out to the registry.

This might require a conceptual change in how Containers are presented.

PR #921, opened 2021-07-14
dhcp: T3680: fix dhcp-interface static route flakyness

bstepler:
## Change Summary

Fix dhcp-interface static route flakyness.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3680

## Component(s) name

* dhcp

## Proposed changes

1. Change get_dhcp_router template filter to read from the Vyatta-custom lease file because the dhclient leases file is not written until after the DHCP hooks are run.
2. Add static-dhclient-hook to refresh staic routes on DHCP events.
3. Rename vyatta-dhclient-hook so it runs before the new static-dhclient-hook.
4. Change exit to return in ipsec-dhclient-hook so subsequent hooks work (it is sourced by dhclient-script, not executed).

## How to test

```
configure
set protocols static route 10.1.0.0/24 dhcp-interface eth0
commit
save
exit
show ip route
```

Try renumbering the network and renewing or rebooting.

## Checklist:


c-po (approved):

zdc:
I would ask to not merge this before my review. We have multiple problems with DHCP client in 1.4 now, and I would like to check if this PR will not conflict with planned solutions for them.

python/vyos/template.py
```diff
@@ -341,19 +341,21 @@ def get_dhcp_router(interface):
     """ Static routes can point to a router received by a DHCP reply. This
     helper is used to get the current default router from the DHCP reply.
 
-    Returns False of no router is found, returns the IP address as string if
+    Returns None if no router is found, returns the IP address as string if
     a router is found.
     """
-    interface = interface.replace('.', '_')
-    lease_file = f'/var/lib/dhcp/dhclient_{interface}.leases'
+    lease_file = f'/var/lib/dhcp/dhclient_{interface}.lease'
```
zdc, line 347: Not a problem but the `/var/lib/dhcp/dhclient_X.lease` file is abandoned in new implementations and used only in old scripts. I would prefer to not reuse it again.

src/etc/dhcp/dhclient-exit-hooks.d/static-dhclient-hook
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
zdc, line 1: If you are creating a new hook, it should be named with order respect (`NN-myhook`). You are fixing the order issue with one hook and creating the same potential issue with another in the same commit. :)

src/etc/dhcp/dhclient-exit-hooks.d/static-dhclient-hook
```diff
@@ -0,0 +1,31 @@
...
+
+python3 - <<PYEND
+import os
+from time import sleep
+from vyos.util import call
+from vyos.util import get_all_vrfs
+
+# Wait for the system to be fully configured.
+while not os.path.isfile('/tmp/vyos-config-status'):
+    sleep(0.2)
```
zdc, line 25: We cannot do this, because dhclient should work normally even before any of VyOS services, for example during early boot to provide access to remote configuration data sources.

zdc (changes requested):
Sorry for the long waiting. I did several attempts to review this in detail but each time something distracted me from the process. Anyway, now I am ready. :)

The DHCP client is tightly connected to many others components in the system, and it is very hard to recall everything actually. Obvious problems that exist in the PR are noted in the code comments.

About the idea of the change in general. With some modifications, it can solve the issue from https://phabricator.vyos.net/T3680, but I believe that exists the better way. Let me explain.

Now, we already have two different ways on how to configure routes via DHCP - via kernel (`ip` command) and FRR (`vtysh`). With this PR we adding the third one - triggering the static protocol config handler. This definitely does not looks good.

I would suggest doing another thing instead: since the current FRR configuration via `vtysh` is not fully compatible with `protocols_static.py`, it would be much better to rework the `03-vyos-ipwrapper` to fix this. And `dhclient-script` apply all necessary routes using the `ip` command wrapper.

How this can be achieved (an example, I am not sure if this is the best possible option):

- Instead of executing commands directly via `vtysh`, routes may be stored in someplace, like a temporary file.
- If FRR configuration change is necessary (add/delete routes), the `protocols_static.py` handler triggered, parses this temporary file and adds routes from there to what is already configured via CLI.
- Changes applied via CLI will not remove routes received via DHCP as they do now, thanks to integrating the logic into the `protocols_static.py`.

This will allow solving all obvious DHCP+FRR problems in one change, instead of adding more and more new tricks to mitigate specific bugs.

I hope that this makes sense.

c-po:
My thoughts on this for a proper implementation would be to change how static routes are installed in the system.

For this issue, it would be a good idea to determine which interface actually has DHCP enabled by traversing the configuration and for those that have a DHCP configuration, when we update the static routes, also those FRR statements must be rendered properly resolving the next-hop address.

As `static_routes_macro.j2` already uses `vyos.template.get_dhcp_router` this should be fairly easy to implement.

c-po:
Hi @bstepler,

thanks for this and all your [other](https://github.com/vyos/vyos-1x/pulls?q=is%3Apr+author%3Abstepler) contributions to the project. I will re-design the Python logic to no longer drop DHCP interface routes by default.

See progress tracked via [T3680](https://phabricator.vyos.net/T3680).

PR #927, opened 2021-07-17
VRF: T3655: proper connection tracking for VRFs

zdc:
## Change Summary

Proper connection tracking for VRFs

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3655

## Component(s) name

VRF, nftables, conntrack

## Proposed changes

Currently, all VRFs share the same connection tracking table, which can
lead to problems:

- traffic leaks to a wrong VRF
- improper NAT rules handling when multiple VRFs contain the same IP
networks
- stateful firewall rules issues

The commit implements connection tracking zones support. Each VRF
utilizes its own zone, so connections will never mix up.

It also adds some restrictions to VRF names and assigned table numbers,
because of nftables and conntrack requirements:

- VRF name should always start from a letter (interfaces that start from
numbers are not supported in nftables rules)
- table number must be in the 100-65535 range because conntrack supports
only 65535 zones

## How to test

Topology for test:

- eth0 and eth1 are connected to the same L2 domain;
- eth2 is connected to Internet via gateway with IP `192.168.122.1`

Minimum config to reproduce the original problem:
```
set interfaces ethernet eth0 address '192.168.0.101/24'
set interfaces ethernet eth0 vrf 'VRF1'
set interfaces ethernet eth1 address '192.168.0.102/24'
set interfaces ethernet eth1 vrf 'VRF2'
set interfaces ethernet eth2 address '192.168.122.103/24'
set interfaces ethernet eth2 vrf 'VRF2'
set nat source rule 100 outbound-interface 'eth2'
set nat source rule 100 translation address 'masquerade'
set vrf name VRF1 protocols static route 0.0.0.0/0 next-hop 192.168.0.102 interface 'eth0'
set vrf name VRF1 table '101'
set vrf name VRF2 protocols static route 0.0.0.0/0 next-hop 192.168.122.1
set vrf name VRF2 table '102'
```
After trying to ping something from VRF1 (`ping 8.8.8.8 vrf VRF1 count 1`), we may see the next conntrack table:
```
icmp     1 26 src=192.168.0.101 dst=8.8.8.8 type=8 code=0 id=3446 [UNREPLIED] src=8.8.8.8 dst=192.168.0.101 type=0 code=0 id=3446 mark=0 use=1
```
There is only single entry for this ICMP exchange and NAT actually does not performs, because of already existing conntrack entry.

After this PR situation will be different:
```
icmp     1 28 src=192.168.0.101 dst=8.8.8.8 type=8 code=0 id=51761 src=8.8.8.8 dst=192.168.122.103 type=0 code=0 id=51761 mark=0 zone=26112 use=1
icmp     1 28 src=192.168.0.101 dst=8.8.8.8 type=8 code=0 id=51761 src=8.8.8.8 dst=192.168.0.101 type=0 code=0 id=51761 mark=0 zone=25856 use=1
```
There will be two entries - for each VRF.

## Checklist:

### DIFF
--- a/python/vyos/ifconfig/interface.py
+++ b/python/vyos/ifconfig/interface.py
@@ -311,6 +311,28 @@ class Interface(Control):
         cmd = 'ip link del dev {ifname}'.format(**self.config)
         return self._cmd(cmd)
 
+    def _set_vrf_ct_zone(self, vrf):
+        """
+        Add/Remove rules in nftables to associate traffic in VRF to an
+        individual conntack zone
+        """
+        if vrf:
+            # Get routing table ID for VRF
+            vrf_table_id = get_interface_config(vrf).get('linkinfo', {}).get(
+                'info_data', {}).get('table')
+            # Add map element with interface and zone ID
+            if vrf_table_id:
+                self._cmd(
+                    f'nft add element inet vrf_zones ct_iface_map {{ "{self.ifname}" : {vrf_table_id} }}'
+                )
+        else:
+            nft_del_element = f'delete element inet vrf_zones ct_iface_map {{ "{self.ifname}" }}'
+            # Check if deleting is possible first to avoid raising errors
+            _, err = self._popen(f'nft -c {nft_del_element}')
+            if not err:
+                # Remove map element
+                self._cmd(f'nft {nft_del_element}')
+
     def get_min_mtu(self):
         """
         Get hardware minimum supported MTU
@@ -401,6 +423,7 @@ class Interface(Control):
         >>> Interface('eth0').set_vrf()
         """
         self.set_interface('vrf', vrf)
+        self._set_vrf_ct_zone(vrf)
 
     def set_arp_cache_tmo(self, tmo):
         """
--- a/data/templates/firewall/nftables-vrf-zones.tmpl
+++ b/data/templates/firewall/nftables-vrf-zones.tmpl
@@ -1,0 +1,17 @@
+table inet vrf_zones {
+  # Map of interfaces and connections tracking zones
+  map ct_iface_map {
+    typeof iifname : ct zone
+  }
+  # Assign unique zones for each VRF
+  # Chain for inbound traffic
+  chain vrf_zones_ct_in {
+    type filter hook prerouting priority raw; policy accept;
+    counter ct zone set iifname map @ct_iface_map
+  }
+  # Chain for locally-generated traffic
+  chain vrf_zones_ct_out {
+    type filter hook output priority raw; policy accept;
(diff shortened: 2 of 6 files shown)


dmbaturin (approved):

python/vyos/ifconfig/interface.py
```diff
@@ -311,6 +312,28 @@ def _delete(self):
         cmd = 'ip link del dev {ifname}'.format(**self.config)
         return self._cmd(cmd)
 
+    def _set_vrf_ct_zone(self, vrf):
+        """
+        Add/Remove rules in nftables to associate traffic in VRF to an
+        individual conntack zone
+        """
+        conf = Config()
```
c-po: @dmbaturin is it allowed/wise to instantiate a new `Config()` object here?

It would be easier and no object of type `Config()` would be needed, if the VRF table ID is simply retrieved from the operating system, as VRFs are always created in advance to assigning an interface to a VRF.

```
vyos@vyos:~$ python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44)
[GCC 10.2.1 20210110] on linux
>>> from vyos.util import get_interface_config
>>> import pprint
>>> pprint.pprint(get_interface_config('red'))
{'address': '62:7b:9c:1b:b1:23',
 'broadcast': 'ff:ff:ff:ff:ff:ff',
 'flags': ['NOARP', 'MASTER', 'UP', 'LOWER_UP'],
 'group': 'default',
 'gso_max_segs': 65535,
 'gso_max_size': 65536,
 'ifindex': 142,
 'ifname': 'red',
 'inet6_addr_gen_mode': 'none',
 'link_type': 'ether',
 'linkinfo': {'info_data': {'table': 2000}, 'info_kind': 'vrf'},
 'linkmode': 'DEFAULT',
 'max_mtu': 65575,
 'min_mtu': 1280,
 'mtu': 65575,
 'num_rx_queues': 1,
 'num_tx_queues': 1,
 'operstate': 'UP',
 'promiscuity': 0,
 'qdisc': 'noqueue',
 'txqlen': 1000}
>>>
```

VRF red has table ID 2000.

c-po (changes requested):

python/vyos/ifconfig/interface.py
zdc: I think you are right, let me change this.

dmbaturin: I think retrieving it from the kernel is a better idea indeed.

Creating another `Config` instance _should_ be safe, but retrieving VRFs from the kernel is going to be much faster.

zdc: Updated.

PR #928, opened 2021-07-17
openvpn: T56: check for alternative authentication methods

yunzheng:
## Change Summary

My second attempt of https://github.com/vyos/vyos-1x/pull/917, because this PR was reverted due to failing smoketest. I didn't remove these strict checks for the smoketests, so the test failed because it checked if these (removed) checks were working properly.

However, In this new PR, the tls cert-file and key-file config checks are still done but only if no alternative authentication methods are defined. Such as --pks12 or --pkcs11-providers or username/password authentication. I have checked the openvpn source code, and these are currently the alternative methods that can be configured instead of --cert and --key.

This change allows advanced users to configure these settings via "openvpn-option" and not be limited by the tls cert-file and key-file configuration check.

## Types of changes

Only check for TLS cert-file and key-file if no alternative authentication methods are defined. This should not impact any functionality or usability.

## Related Task(s)

* https://phabricator.vyos.net/T56

## Component(s) name

openvpn

## How to test

Given a valid `/config/auth/vpn/ca.crt` file, 

This should result in Missing "tls cert-file"
```
del interfaces openvpn vtun1
set interfaces openvpn vtun1 remote-host 127.0.0.1
set interfaces openvpn vtun1 mode client
set interfaces openvpn vtun1 tls ca-cert-file /config/auth/vpn/ca.crt
commit
```

This should pass (--pkcs11-providers):
```
del interfaces openvpn vtun1
set interfaces openvpn vtun1 remote-host 127.0.0.1
set interfaces openvpn vtun1 mode client
set interfaces openvpn vtun1 tls ca-cert-file /config/auth/vpn/ca.crt
set interfaces openvpn vtun1 openvpn-option "--pkcs11-providers /usr/lib/libeToken.so"
commit
```

And this should also pass (--pkcs12):

```
del interfaces openvpn vtun1
set interfaces openvpn vtun1 remote-host 127.0.0.1
set interfaces openvpn vtun1 mode client
set interfaces openvpn vtun1 tls ca-cert-file /config/auth/vpn/ca.crt
set interfaces openvpn vtun1 openvpn-option "--pkcs12 /config/auth/vpn/client.p12"
commit
```

And this should also pass (username/password auth only):

```
del interfaces openvpn vtun1
set interfaces openvpn vtun1 remote-host 127.0.0.1
set interfaces openvpn vtun1 mode client
set interfaces openvpn vtun1 tls ca-cert-file /config/auth/vpn/ca.crt
set interfaces openvpn vtun1 authentication username user
set interfaces openvpn vtun1 authentication password pass
commit
```

## Checklist:


c-po:
Instead of validating a raw openvpn-option on the CLI which we do not do I rather recommend that we add proper pkcs12 support to the CLI.

Do you happen to be on the VyOS SLACK channel so we can discuss this in a close loop? Just ping me (c-po)

sarthurdev:
I agree with @c-po, openvpn should probably follow the `authentication mode` format of other VPN configurations to define the method to be used and configure it via cli, then validate based on that mode.

yunzheng:
Discussed with @c-po that I will fix the 'authentication' issue to match with the old VyOS configuration checking.

Adding pkcs12 support is not really worth it, and pkcs11 usage is probably also not very common. However by fixing the above issue, advanced users can set a dummy username/password so they can workaround the strict certificate checks that VyOS imposes on the user.

I will close this issue in favour for a new PR.

PR #951, opened 2021-07-31
bridge: T3715: Standby port supporting

jack9603301:
## Change Summary

Standby port supporting

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3715

## Component(s) name

bridge

## Proposed changes

Standby port supporting

## Checklist:

### DIFF
--- a/python/vyos/ifconfig/bridge.py
+++ b/python/vyos/ifconfig/bridge.py
@@ -296,10 +296,10 @@ class BridgeIf(Interface):
             cmd = f'bridge vlan add dev {ifname} vid 1 pvid untagged self'
             self._cmd(cmd)
 
-        tmp = dict_search('member.interface', config)
-        if tmp:
+        member_intf = dict_search('member.interface', config)
+        if member_intf:
 
-            for interface, interface_config in tmp.items():
+            for interface, interface_config in member_intf.items():
                 # if interface does yet not exist bail out early and
                 # add it later
                 if interface not in interfaces():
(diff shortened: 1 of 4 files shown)


python/vyos/ifconfig/bridge.py
```diff
@@ -365,6 +365,21 @@ def update(self, config):
                     if native_vlan_id:
                         cmd = f'bridge vlan add dev {interface} vid {native_vlan_id} pvid untagged master'
                         self._cmd(cmd)
+            
+            # When the bridge port is initialized, set the standby port now,
+            # This reports an error if the bridge did not complete the member
+            # port addition before setting the standby port
+            for interface, interface_config in member_intf.items():
```
c-po, line 372: Duplicated code in interface.py and bridge.py

jack9603301: done

src/conf_mode/interfaces-bridge.py
```diff
@@ -123,6 +130,14 @@ def verify(bridge):
             if 'is_bond_member' in interface_config:
                 tmp = next(iter(interface_config['is_bond_member']))
                 raise ConfigError(error_msg + f'it is already a member of bond "{tmp}"!')
+            
+            if 'backup_port' in interface_config and 'backup_port_exist' not in interface_config:
+                tmp = interface_config['backup_port']
+                raise ConfigError(error_msg + f'Device {tmp} is not a bridge member and cannot be used as an alternate port!')
+            
+            if 'backup_port' in interface_config and 'backup_port_exist' in interface_config:
+                if not re.search('^((eth|lan)[0-9]+|(eno|ens|enp|enx).+)$', interface):
```
c-po: This is redundant code - we have an interface name validator which should be used from the XML level. This prevents invalid input to pass down into the Python script.

The validator is named `interface-name` and also supports VLAN interfaces - those can not be used for a bridge backup port, but the validator can be adjusted to validate with and without vlans.

c-po (changes requested):

src/conf_mode/interfaces-bridge.py
jack9603301: done
But now I want to limit the use to the default Ethernet network card in the system, so I add a verifier based on this verifier.

c-po: Please rename the validator to `interface-name-eth`

jack9603301: done

jack9603301:
@c-po Can you review the following

c-po:
Hi Jack - what is the purpose of this feature? I'd rather stick with a proper STP solution.

jack9603301:
@c-po Generally, Linux Bridge supports a standby port. According to the man manual, when a specific primary port stops working, traffic is automatically redirected to the standby port

Linux Bridge supports this, and if it does, it might be useful in some cases.

c-po:
Closing this as I unfortunately still did not got the use-case that I can solve using STP.

jack9603301:
> Closing this as I unfortunately still did not got the use-case that I can solve using STP.

Sorry I haven't been following this for a long time, I don't quite understand what you mean

PR #964, opened 2021-08-12
T3275: conntrack: Migrate 'disable' syntax to 'enable' syntax for the new default behavior

erkin:
## Change Summary

Conntrack syntax v3

## Types of changes

- [x] Other (please describe): Migration
## Related Task(s)

* https://phabricator.vyos.net/T3275

## Component(s) name

conntrack

## Proposed changes

Delete all `disable` statements and add `enable` statements for other conntrack modules.

## Checklist:

### DIFF
--- a/src/migration-scripts/conntrack/2-to-3
+++ b/src/migration-scripts/conntrack/2-to-3
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+
+# Conntrack syntax version 3
+# Enables all conntrack modules (previous default behaviour) and omits manually disabled modules.
+
+import os
+import sys
+
+from vyos.configtree import ConfigTree
+from vyos.version import get_version
+
+if len(sys.argv) < 1:
+    print('Must specify file name!')
+    sys.exit(1)
+
+filename = sys.argv[1]
+
+with open(filename, 'r') as f:
+    config = ConfigTree(f.read())
+
+# Get all conntrack modules supported through the CLI.
+module_templates_dir = '/opt/vyatta/share/vyatta-cfg/templates/system/conntrack/modules'
+module_path = ['system', 'conntrack', 'modules']
+
+for module in [f.name for f in os.scandir(module_templates_dir) if f.is_dir()]:
+    # 'disable' is being phased out.
+    if config.exists(module_path + [module, 'disable']):
+        config.delete(module_path + [module])
+    # If it wasn't manually 'disable'd, it was enabled by default.
+    # So we 'enable' it in the new syntax.
+    else:
+        config.set(module_path + [module, 'enable'])
+
+try:
+    if config.exists(module_path):
+        with open(filename, 'w') as f:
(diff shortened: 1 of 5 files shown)


src/migration-scripts/conntrack/2-to-3
```diff
@@ -0,0 +1,40 @@
...
+    print('Must specify file name!')
+    sys.exit(1)
+
+filename = sys.argv[1]
+
+with open(filename, 'r') as f:
+    config = ConfigTree(f.read())
+
+# Get all conntrack modules supported through the CLI.
+module_templates_dir = '/opt/vyatta/share/vyatta-cfg/templates/system/conntrack/modules'
```
dmbaturin: The `/opt/vyatta/` dir is to eventually get removed, but the migration script will likely last much longer. The `/opt/vyatta/share/vyatta-cfg/templates/system/conntrack/modules` can also go away much sooner when that CLI is rewritten.

Since the number of those modules is very small, we can just hardcode it here. The list of modules _supported before the 1.3_ release also can't change in the future, so we don't need to account for modules that weren't supported at that time, since an old config couldn't have possibly relied on their existence.

dmbaturin (changes requested):

c-po:
+1 for hardcoding the modules names.

That specific part is actually already rewritten for the 1.4 developmnet cycle.
https://phabricator.vyos.net/T3579

dmbaturin (approved):

c-po:
@erkin I am missing the change for the increased version file. Can you please also submit a PR in vyatta-cfg-system? https://github.com/vyos/vyatta-cfg-system/blob/current/cfg-version/conntrack%402

dmbaturin:
Also, am I missing the PR where the actual command syntax was changed? Or it's not done yet?

erkin:
You're right it seems. I went an added it.

dmbaturin (approved):

erkin:
See vyos/vyatta-cfg-system#162

c-po:
If we now have only `enable` wouldn't it make even more sense to simplify the CLI by specifying:
Because an additional `enable` node simply feels redundant to me.
```xml
          <node name="modules">
            <properties>
              <help>Connection tracking modules settings</help>
            </properties>
            <children>
              <leafNodename="ftp">
                <properties>
                  <help>FTP connection tracking settings</help>
                  <valueless/>
                </properties>
              </leafNode>
              <leafNode name="h323">
                <properties>
                  <help>H.323 connection tracking settings</help>
                  <valueless/>
                </properties>
              </leafNode>
```

dmbaturin:
@c-po That's a good point actually. It will not prevent us from adding module-specific options either: existence of a node will imply that the module is enabled. I think we should not use an extra "enable" command and just check if a node exists.

erkin:
Good idea. If no one has any complaints, I'm going to update it as such.

PR #976, opened 2021-08-22
T3769

kroy-the-rabbit:
## Change Summary

This exposes an option to bridge a container's network instead of masquerading

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3769

## Component(s) name

Containers

## Proposed changes

Change interfaces for containers to include a bridge option.  Add in the bridge option in cod

## Checklist:

- [ X] My commit headlines contain a valid Task id
- [ X] My change requires a change to the documentation
- [ X] I have updated the documentation accordingly I will

### DIFF
--- a/interface-definitions/containers.xml.in
+++ b/interface-definitions/containers.xml.in
@@ -165,6 +165,12 @@
               <multi/>
             </properties>
           </leafNode>
+          <leafNode name="enable-bridging">
+            <properties>
+              <help>Treat this network as a bridge instead of automatically NATing it (cni-podmanX)</help>
+                <valueless/>
+            </properties>
+          </leafNode>
         </children>
       </tagNode>
       <leafNode name="registry">
--- a/src/conf_mode/containers.py
+++ b/src/conf_mode/containers.py
@@ -16,6 +16,7 @@
 
 import os
 import json
+import re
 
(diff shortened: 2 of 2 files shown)


c-po:
Hmm isn't this the same as `allow-host-networks`?

kroy-the-rabbit:
> Hmm isn't this the same as `allow-host-networks`?

Using the host network namespace is nothing short of a security blight on a good day, and in my opinion, should honestly be removed.  And this is doubly-so on a router/firewall device. Especially when you are pulling random containers from popular, but untrusted docker registries.  In many scenarios and deployments, it would be flat out disallowed due to the security implications.  

https://medium.com/nttlabs/dont-use-host-network-namespace-f548aeeef575

In fact, most people mainly use it when you have a container that has a large number of ports that need forwarding as a convenience.  

This PR treats container networking as a completely separate network.  It can be firewalled, NAT'd, etc all independently of the host using normal VyOS functionality.

c-po:
Implemented via 209ce3d9b6fb09626a7abe3540b888566b739de8. It is a bad idea to mangle the JSON configuration file by string replacement. Rather load the file `json.loads()`, edit the content and the dump it back `json.dumps()`.

PR #991, opened 2021-09-02
[DRAFT, do not merge!] T3697: fix the deadlock in IPsec options script

dmbaturin:
## Change Summary

An argument is passed to ipsec-settings.py when it's run from `vpn ipsec options` and IPsec is not fully configured yet, so that it doesn't try to restart StrongSWAN when it's not even supposed to be running yet.

## Types of changes


dmbaturin:
Sorry, wrong branch.

PR #992, opened 2021-09-02
T3697: fix the deadlock in IPsec options script

dmbaturin:
## Change Summary

An argument is passed to ipsec-settings.py when it's run from `vpn ipsec options` and IPsec is not fully configured yet, so that it doesn't try to restart StrongSWAN when it's not even supposed to be running yet.

## Types of changes

### DIFF
--- a/interface-definitions/ipsec-settings.xml.in
+++ b/interface-definitions/ipsec-settings.xml.in
@@ -4,7 +4,7 @@
     <children>
       <node name="ipsec">
         <children>
-          <node name="options" owner="${vyos_conf_scripts_dir}/ipsec-settings.py">
+          <node name="options" owner="${vyos_conf_scripts_dir}/ipsec-settings.py from-options">
             <properties>
               <help>Global IPsec settings</help>
             </properties>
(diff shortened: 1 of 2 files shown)


jestabro:
Discussed with Daniil; testing the version of this solution that was considered.

jestabro:
The PR (with import fixes, below) avoids the deadlock on boot when 'vpn ipsec options ..' is set; the 'version' referred to means that no other modifications are needed, for example, to vyatta-cfg-vpn.

Import corrections:
```
> @@ -18,7 +18,7 @@ import re
>  import os
>  
>  from time import sleep
> -from sys import exit
> +from sys import argv, exit
>  
>  from vyos.config import Config
>  from vyos import ConfigError
40c38
< +    res = system("ipsec stroke memusage >&/dev/null")
---
> +    res = os.system("ipsec stroke memusage >&/dev/null")
```

c-po:
@jestabro recently I found out that there is no need to import exit, its a keyword always available like if/for

jestabro:
@c-po that's good to know --- the exit was there previously, but useful that it can be avoided; thanks !
@dmbaturin I'll run a sanity check on final version, and merge assuming all is fine

jestabro:
This revision looks good to me.

jestabro (approved):

PR #1066, opened 2021-11-04
pki: T3970: Allow op-mode PKI commands in a config session to install directly

sarthurdev:
## Change Summary

Allows existing PKI op-mode `install <name>` syntax to install directly when called from configure mode

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3970

## Component(s) name

pki

## Proposed changes

- Adds function `vyos.util.install_into_config` to show or install config values depending if a config session is available.
  - Has user prompt to ensure no existing config values are overwritten without user consent
- Adds all PKI `install <name>` op-mode commands to use `install_into_config` allowing for directly installing PKI values into pending config of active config session.

## How to test

Op-mode:
```
vyos@vyos:~$ generate pki ca install test
Enter private key type: [rsa, dsa, ec] (Default: rsa) ec
...
You are not in configure mode, commands to install manually from configure mode:
set pki ca test certificate '...'
set pki ca test private key '...'
```

Conf-mode:
```
vyos@vyos:~$ conf
WARNING: You are currently configuring a live-ISO environment, changes will not persist until installed
[edit]
vyos@vyos# run generate pki ca install test
Enter private key type: [rsa, dsa, ec] (Default: rsa) ec
...
2 value(s) installed. Use "compare" to see the pending changes, and "commit" to apply.
[edit]
vyos@vyos# comp
+pki {
+    ca test {
+        certificate ...
+        private {
+            key ...
+        }
+    }
+}
```

## Checklist:

### DIFF
--- a/python/vyos/util.py
+++ b/python/vyos/util.py
@@ -664,7 +664,6 @@ def ask_yes_no(question, default=False) -> bool:
         except EOFError:
             stdout.write("\nPlease respond with yes/y or no/n\n")
 
-
 def is_admin() -> bool:
     """Look if current user is in sudo group"""
     from getpass import getuser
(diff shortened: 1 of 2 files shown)


python/vyos/util.py
```diff
@@ -903,3 +902,34 @@ def check_port_availability(ipaddress, port, protocol):
...
+    # override_prompt: if True, user will be prompted before existing nodes are overwritten
+
+    if not config_paths:
+        return None
+
+    from vyos.config import Config
+
+    conf = Config()
+
+    if not conf.in_session():
```
c-po: Can we use `Config().in_session()` here?

python/vyos/util.py
```diff
@@ -903,3 +902,34 @@ def check_port_availability(ipaddress, port, protocol):
...
+        print('You are not in configure mode, commands to install manually from configure mode:')
+        for path in config_paths:
+            print(f'set {path}')
+        return None
+
+    count = 0
+
+    for path in config_paths:
+        if override_prompt and conf.exists(path) and not conf.is_multi(path):
```
c-po, line 925: Can we use an instance of `ConfigTreeQuery` for `conf`?

c-po (changes requested):

python/vyos/util.py
sarthurdev: As the underlying code essentially uses the `Config` class and we needed an instance for the session check, it seemed more efficient to use just the single instance for both checks.

c-po: Please still use ConfigTreeQuery whenever possible from op mode.

PR #1068, opened 2021-11-09
atop: T3774: Atop log file rotation fix

zdc:
## Change Summary

Atop service by default rotate files only daily and keeps the latest 28 files. The rotation logic is too complex. This commit changes service in a way that allows us to control rotation strategy.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3774

## Component(s) name

atop, logrotate

## Proposed changes

The systemd unit for atop service is changed, so the log file name and location will be always the same. It also adds the logrotate configuration to conditionally rotate a log file.
Hardcoded values:
- maximum log file size: 10 MB
- maximum count of files: 10
These values can be easily changed within the `/etc/logrotate.d/vyos-atop`, no additional configuration is required.
Rotation will be done hourly, if necessary, according to `/etc/cron.hourly/vyos-logrotate-hourly`.

This change has two benefits:
- rotation strategy control can be done via logrotate, and can be exposed to CLI now;
- the total size of all logs is now controlled more aggressively, so the chance to get a situation when atop logs took all the space on a drive is significantly lower. Also, if this will be necessary, rotation may be done even each minute what reduces risks related to logs size even more.

## Checklist:

### DIFF
--- a/src/etc/systemd/system/atop.service.d/10-override.conf
+++ b/src/etc/systemd/system/atop.service.d/10-override.conf
@@ -1,0 +1,6 @@
+[Service]
+ExecStartPre=
+ExecStart=
+ExecStart=/bin/sh -c 'exec /usr/bin/atop ${LOGOPTS} -w "${LOGPATH}/atop.log" ${LOGINTERVAL}'
+ExecStartPost=
+
--- a/debian/vyos-1x.install
+++ b/debian/vyos-1x.install
@@ -1,6 +1,7 @@
 etc/cron.hourly
 etc/dhcp
 etc/ipsec.d
+etc/logrotate.d
 etc/netplug
 etc/opennhrp
 etc/ppp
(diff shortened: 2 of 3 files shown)


src/etc/systemd/system/atop.service.d/10-override.conf
```diff
@@ -0,0 +1,6 @@
+[Service]
+ExecStartPre=
+ExecStart=
```
DmitriyEshenko, line 3: Why do we need empty and duplicate ExecStart* entries?

zdc: To overwrite it "from the scratch".
> if one wants to remove entries from a setting that is parsed as a list (and is not a dependency), such as AssertPathExists= (or e.g. ExecStart= in service units), one needs to first clear the list before re-adding all entries except the one that is to be removed

DmitriyEshenko: Looks good, thank you for the explanation

c-po (approved):

AlexanderTheGr8-gr:
@zdc Thank you bro! Saved me hours! <3

PR #1076, opened 2021-11-22
tftp: T4012: Add TFTP VRF support

DmitriyEshenko:
## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4012

## Checklist:

- [X] I have updated the documentation accordingly


src/conf_mode/tftp_server.py
```diff
@@ -91,6 +93,9 @@ def generate(tftpd):
         else:
             config['listen_address'] = f'[{address}]:{port} -6'
 
+        if 'vrf' in tftpd['listen_address'][address]:
+            config['vrf'] = tftpd['listen_address'][address]['vrf']
```
c-po: Please change iterator to iterate over the dictionary using key/value pairs - `for address, address_config in tftp['listen_address'].items()`  to make reading the code easier and not iterate over multiple dictionaries.

data/templates/tftp-server/default.tmpl
```diff
@@ -1,2 +1,7 @@
 ### Autogenerated by tftp_server.py ###
 DAEMON_ARGS="--listen --user tftp --address {{ listen_address }} {{ "--create --umask 000" if allow_upload is defined }} --secure {{ directory }}"
+{% if vrf is defined  %}
```
c-po: Please remove additional whitespace.

interface-definitions/include/listen-address-vrf.xml.i
```diff
@@ -0,0 +1,25 @@
+<!-- include start from listen-address.xml.i -->
```
c-po: Please fix include comment. References the wrong file.

src/conf_mode/tftp_server.py
```diff
@@ -69,6 +70,7 @@ def verify(tftpd):
         if not is_addr_assigned(address):
             print(f'WARNING: TFTP server listen address "{address}" not ' \
                   'assigned to any interface!')
+        verify_vrf(tftpd['listen_address'][address])
```
c-po: Please change iterator to iterate over the dictionary using key/value pairs - `for address, address_config in tftp['listen_address'].items()` to make reading the code easier and not iterate over multiple dictionaries.

c-po (changes requested):

src/conf_mode/tftp_server.py
DmitriyEshenko: Done

data/templates/tftp-server/default.tmpl
DmitriyEshenko: Done

interface-definitions/include/listen-address-vrf.xml.i
DmitriyEshenko: Done

src/conf_mode/tftp_server.py
DmitriyEshenko: Done

src/systemd/tftpd@.service
```diff
@@ -7,7 +7,7 @@ RequiresMountsFor=/run
 Type=forking
 #NotifyAccess=main
 EnvironmentFile=-/etc/default/tftpd%I
-ExecStart=/usr/sbin/in.tftpd "$DAEMON_ARGS"
+ExecStart=/bin/sh -c "${VRF_ARGS} /usr/sbin/in.tftpd ${DAEMON_ARGS}"
```
c-po, line 10: Is there a particular reason this uses `/bin/sh` as executor and not like the other patterns? Just trying to understand.

* https://github.com/vyos/vyos-1x/blob/current/data/templates/ssh/override.conf.tmpl#L9
* https://github.com/vyos/vyos-1x/blob/current/data/templates/ntp/override.conf.tmpl#L9

DmitriyEshenko: Yes, we can not use ENV arg as the first param if it is not defined exactly in service. In this case, it is not possible to define VRF_ARGS here and we are defining it in /etc/default/tftpN files.

c-po:
Can you please also extend https://github.com/vyos/vyos-1x/blob/current/smoketest/scripts/cli/test_service_tftp-server.py to start some instances in different VRFs and then verify if the VRF runs a tftpd process (`ip vrf pids red`).

I see no documentation PR, but the template mentions documentation needs update?

PR #1128, opened 2021-12-29
dhclient: T4121: Fixed resolv.conf generation at early boot stage

zdc:
## Change Summary

Fixed resolv.conf generation at early boot stage

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4121

## Component(s) name

dhclient, vyos-hostsd

## Proposed changes

In case if a CLI configuration is not available, dhclient cannot add nameservers to a `resolv.conf` file, because `vyos-hostsd` requires that an interface be listed in the `set system name-server` option.
This commit introduces two changes:
- `vyos-hostsd` service will not be started before Cloud-Init fetch all remote data. This is required because all meta-data should be available for Cloud-Init before any of VyOS-related services start since it is used for configuration generation.
- the `vyos-hostsd-client` in the `dhclient-script` will be used only if the `vyos-hostsd` is running. In other words - if VyOS services already started, dhclient changes `resolv.conf` using `vyos-hostsd`; in other cases - does this directly.

These changes should protect us from problems with DHCP during system boot if DHCP is required by third-party utils.

## How to test

To test this change, you need to deploy VyOS in an environment that uses Cloud-Init with a DNS-based URL for meta-data service. Without the fix, Cloud-Init will not be able to resolve the service and meta-data will not be fetched.

## Checklist:

### DIFF
--- a/src/etc/dhcp/dhclient-enter-hooks.d/04-vyos-resolvconf
+++ b/src/etc/dhcp/dhclient-enter-hooks.d/04-vyos-resolvconf
@@ -1,44 +1,48 @@
-# modified make_resolv_conf () for VyOS
-make_resolv_conf() {
-    hostsd_client="/usr/bin/vyos-hostsd-client"
-    hostsd_changes=
+# modified make_resolv_conf() for VyOS
+# should be used only if vyos-hostsd is running
 
-    if [ -n "$new_domain_name" ]; then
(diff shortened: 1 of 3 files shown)


dmbaturin (approved):
Looks good to me! @UnicronNL, do you also think it's the fix we need for cloud-init environments problem that you've been having?

PR #1167, opened 2022-01-13
firewall: T4178: Use lowercase for TCP flags and add an validator

sarthurdev:
## Change Summary

* Fix for case sensitivity in nftables config
* Add tcp flags validator

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4178

## Component(s) name

firewall, policy

## Checklist:

### DIFF
--- a/interface-definitions/include/firewall/common-rule.xml.i
+++ b/interface-definitions/include/firewall/common-rule.xml.i
@@ -274,12 +274,42 @@
         <help>TCP flags to match</help>
         <valueHelp>
           <format>txt</format>
-          <description>TCP flags to match</description>
+          <description>Multiple comma-separated flags</description>
+        </valueHelp>
+        <valueHelp>
+          <format>syn</format>
+          <description>Syncronise flag</description>
+        </valueHelp>
+        <valueHelp>
+          <format>ack</format>
+          <description>Acknowledge flag</description>
+        </valueHelp>
+        <valueHelp>
+          <format>fin</format>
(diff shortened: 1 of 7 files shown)


c-po:
I wonder if we can not use the new firewall systems migration script whichvisbin place to also migrate the flags from upper to lowercase in the CLI?

interface-definitions/include/firewall/common-rule.xml.i
```diff
@@ -280,6 +280,9 @@
           <format> </format>
           <description>\n\n  Allowed values for TCP flags : SYN ACK FIN RST URG PSH ALL\n  When specifying more than one flag, flags should be comma-separated.\n For example : value of 'SYN,!ACK,!FIN,!RST' will only match packets with\n  the SYN flag set, and the ACK, FIN and RST flags unset</description>
         </valueHelp>
```
c-po, line 278: Please also add valueHelp for each individual flag and also a completionHelper list so tab complete will work :)

c-po (changes requested):

sarthurdev:
Don't think migration is necessary for text case, I think we should handle both and just output to lowercase for nftables.

c-po:
It might work but would be bad practice if CLI accepts both upper and loser case versions ;) - prone to error

c-po:
this needs to be refactored in a later sped, as multiple flags on a comma separated line just feels wrong

PR #1279, opened 2022-04-08
Firewall: T990: Add snat and dnat connection status on firewall

nicolas-fort:
## Change Summary

source nat and destination nat connection status available in firewall rules configuration.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T990

## Component(s) name

firewall
## Proposed changes

On this first approach, we are adding only snat and dnat options for ct status.
In future PR, all options should be covered.

Netfilter references: 

- https://wiki.nftables.org/wiki-nftables/index.php/Matching_connection_tracking_stateful_metainformation#ct_status_-_conntrack_status
- https://wiki.nftables.org/wiki-nftables/index.php/Quick_reference-nftables_in_10_minutes#Ct

## How to test

cli options
```
vyos@vyos# set firewall name FOO rule 10 
Possible completions:
   action       Rule action [REQUIRED]
 > connection-status
                Connection status
   description  Description
 > destination  Destination parameters
   disable      Option to disable firewall rule
 > fragment     IP fragment match
 > icmp         ICMP type and code information
 > ipsec        Inbound IPsec packets
 > limit        Rate limit using a token bucket filter
   log          Option to log packets matching rule
   protocol     Protocol to match (protocol name, number, or "all")
 > recent       Parameters for matching recently seen sources
 > source       Source parameters
 > state        Session state
 > tcp          TCP flags to match
 > time         Time to match rule

      
[edit]
vyos@vyos# set firewall name FOO rule 10 connection-status 
Possible completions:
   nat          NAT connection status

      
[edit]
vyos@vyos# set firewall name FOO rule 10 connection-status nat 
Possible completions:
   destination  Match connections that are subject to destination NAT
   source       Match connections that are subject to source NAT
                

```

Configurations and validations
```
set firewall ipv6-name FOO rule 10 action 'accept'
set firewall ipv6-name FOO rule 10 connection-status nat destination
set firewall ipv6-name FOO rule 10 state new 'enable'
set firewall name FOO rule 10 connection-status nat source
set firewall name FOO rule 10 state invalid 'enable'

vyos@vyos# sudo nft list chain ip filter NAME_FOO
table ip filter {
        chain NAME_FOO {
                ct state { invalid } ct status { snat } counter packets 0 bytes 0 drop comment "FOO-10"
                counter packets 0 bytes 0 return comment "FOO default-action accept"
        }
}
[edit]
vyos@vyos# sudo nft list chain ip6 filter NAME6_FOO
table ip6 filter {
        chain NAME6_FOO {
                ct state { new } ct status { dnat } counter packets 0 bytes 0 return comment "FOO-10"
                counter packets 0 bytes 0 return comment "FOO default-action accept"
        }
}
```
## Checklist:

### DIFF
--- a/interface-definitions/include/firewall/common-rule.xml.i
+++ b/interface-definitions/include/firewall/common-rule.xml.i
@@ -95,51 +95,25 @@
     </constraint>
   </properties>
 </leafNode>
-<node name="ct-status">
+<leafNode name="connection-status">
   <properties>
-    <help>Connection status in conntrack</help>
+    <help>Connection status</help>
+    <completionHelp>
+      <list>dnat snat</list>
+    </completionHelp>
+    <valueHelp>
+      <format>dnat</format>
+      <description>Match connections that are subject to destination NAT</description>
+    </valueHelp>
+    <valueHelp>
+      <format>snat</format>
+      <description>Match connections that are subject to source NAT</description>
+    </valueHelp>
+    <constraint>
+      <regex>^(dnat|snat)$</regex>
+    </constraint>
   </properties>
-  <children>
-    <leafNode name="dnat">
-      <properties>
-        <help>Set when connection needs DNAT in original direction</help>
-        <completionHelp>
-          <list>enable disable</list>
-        </completionHelp>
-        <valueHelp>
-          <format>enable</format>
-          <description>Enable</description>
-        </valueHelp>
-        <valueHelp>
-          <format>disable</format>
-          <description>Disable</description>
-        </valueHelp>
-        <constraint>
-          <regex>^(enable|disable)$</regex>
-        </constraint>
-      </properties>
-    </leafNode>
-    <leafNode name="snat">
-      <properties>
-        <help>Set when connection needs SNAT in original direction</help>
-        <completionHelp>
-          <list>enable disable</list>
-        </completionHelp>
-        <valueHelp>
-          <format>enable</format>
-          <description>Enable</description>
-        </valueHelp>
-        <valueHelp>
-          <format>disable</format>
-          <description>Disable</description>
-        </valueHelp>
-        <constraint>
-          <regex>^(enable|disable)$</regex>
-        </constraint>
-      </properties>
-    </leafNode>
-  </children>
-</node>
+</leafNode>
 <leafNode name="protocol">
   <properties>
     <help>Protocol to match (protocol name, number, or "all")</help>
--- a/interface-definitions/include/firewall/common-rule.xml.i
+++ b/interface-definitions/include/firewall/common-rule.xml.i
@@ -95,6 +95,51 @@
     </constraint>
   </properties>
 </leafNode>
+<node name="ct-status">
+  <properties>
+    <help>Connection status in conntrack</help>
+  </properties>
+  <children>
+    <leafNode name="dnat">
+      <properties>
+        <help>Set when connection needs DNAT in original direction</help>
+        <completionHelp>
+          <list>enable disable</list>
+        </completionHelp>
+        <valueHelp>
+          <format>enable</format>
+          <description>Enable</description>
+        </valueHelp>
+        <valueHelp>
+          <format>disable</format>
+          <description>Disable</description>
+        </valueHelp>
+        <constraint>
+          <regex>^(enable|disable)$</regex>
+        </constraint>
+      </properties>
+    </leafNode>
+    <leafNode name="snat">
+      <properties>
+        <help>Set when connection needs SNAT in original direction</help>
+        <completionHelp>
+          <list>enable disable</list>
+        </completionHelp>
+        <valueHelp>
+          <format>enable</format>
+          <description>Enable</description>
+        </valueHelp>
+        <valueHelp>
+          <format>disable</format>
+          <description>Disable</description>
+        </valueHelp>
+        <constraint>
+          <regex>^(enable|disable)$</regex>
+        </constraint>
+      </properties>
+    </leafNode>
+  </children>
+</node>
 <leafNode name="protocol">
   <properties>
     <help>Protocol to match (protocol name, number, or "all")</help>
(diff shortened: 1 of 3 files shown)


sarthurdev (approved):

interface-definitions/include/firewall/common-rule.xml.i
```diff
@@ -95,6 +95,51 @@
...
+<node name="ct-status">
+  <properties>
+    <help>Connection status in conntrack</help>
+  </properties>
+  <children>
+    <leafNode name="dnat">
+      <properties>
+        <help>Set when connection needs DNAT in original direction</help>
+        <completionHelp>
+          <list>enable disable</list>
```
dmbaturin: I understand that you added "enable/disable" for uniformity with the `state` node, so I can't blame you. But our general agreement was that boolean flags should be made valueless (i.e. just `set ct-status dnat`), since values like "enable/disable" or "true/false" don't add any information to the config. They also make code more complicated since you need to handle the case when a node is absent and also the case when it's set to "disable".

I believe we should apply that policy here and make it valueless, since eventually the `state` node will also be converted to valueless. A little temporary non-uniformity is an acceptable cost for less migration in the future.

interface-definitions/include/firewall/common-rule.xml.i
```diff
@@ -95,6 +95,51 @@
...
   </properties>
 </leafNode>
+<node name="ct-status">
+  <properties>
+    <help>Connection status in conntrack</help>
+  </properties>
+  <children>
+    <leafNode name="dnat">
+      <properties>
+        <help>Set when connection needs DNAT in original direction</help>
```
dmbaturin: I think a better wording would be `Match connections that are subject to destination NAT`. The word "set" is clear in the original context in the nftables flag descriptions, but in our CLI I feel it would be ambiguous and may be interpreted as if this option _sets_ a flag, rather than make the rule match connections where that flag is _already set_.

interface-definitions/include/firewall/common-rule.xml.i
```diff
@@ -95,6 +95,51 @@
     </constraint>
   </properties>
 </leafNode>
+<node name="ct-status">
```
dmbaturin: I believe `connection-status` would be a better name. Experienced Linux network admins will associate `ct` with conntrack immediately, but many VyOS users actually come from non-Linux background and that name will be opaque to them.

interface-definitions/include/firewall/common-rule.xml.i
```diff
@@ -95,6 +95,51 @@
...
+  <properties>
+    <help>Connection status in conntrack</help>
```
dmbaturin: Maybe just "Connection status". Conntrack is an implementation detail that many users may not really care about, or even know what conntrack is.

dmbaturin (changes requested):
I think the idea and implementation are good, but the CLI command names and help strings can be improved, I left some suggestions there.

c-po:
snat and dnat are common terms - VyOS refers to it as destination nat and source nat on the CLI:

`set nat source | destination` maybe we can create an analogy here?

`set firewall name FOO rule 10 conntrack nat source`
`set firewall name FOO rule 10 conntrack nat destination`

?

dmbaturin:
@c-po I'm not fond of exposing internal names in the user-visible CLI. We have changed underlying components quite a few times already, and had components change their names in the upstream (through direct renaming or a fork).

If we named the firewall "set iptables",  it would have become an anachronism and a misnomer after migration to nftables. There's a small but non-zero chance that kernel developers will replace conntrack with something else (hypothetically, "nftrack") and we'll have to either keep a name that became wrong or make a syntax change.

I think `connection-status` is both neutral and clear, but if you have better ideas, please share!

nicolas-fort:
Daniil.
Yes, as you said, I added "enable/disable" for uniformity with the `state` node. I'll will remove `enable/disable` for this new feature.
So, if everyone agrees, new command should be `set firewall name <name> rule <rule> connection-status [snat|dnat]`

Of course, with corrected info on helpers/descriptions.

c-po:
> @c-po I'm not fond of exposing internal names in the user-visible CLI. We have changed underlying components quite a few times already, and had components change their names in the upstream (through direct renaming or a fork).
> 
> If we named the firewall "set iptables", it would have become an anachronism and a misnomer after migration to nftables. There's a small but non-zero chance that kernel developers will replace conntrack with something else (hypothetically, "nftrack") and we'll have to either keep a name that became wrong or make a syntax change.
> 
> I think `connection-status` is both neutral and clear, but if you have better ideas, please share!

I took a very bad example here. My main issue is the snat and dnat terminus which we should name nat source and nat destination to match the NAT subsystem.

`set firewall name <name> rule <rule> connection-status nat <source | destination >` would work for me.

nicolas-fort:
Christians proposal sound good to me.
@dmbaturin? I already made the changes you requested. So, we leave it as it is, or move to 
`set firewall name <name> rule <rule> connection-status nat <source | destination >`??

nicolas-fort:
> > @c-po I'm not fond of exposing internal names in the user-visible CLI. We have changed underlying components quite a few times already, and had components change their names in the upstream (through direct renaming or a fork).
> > If we named the firewall "set iptables", it would have become an anachronism and a misnomer after migration to nftables. There's a small but non-zero chance that kernel developers will replace conntrack with something else (hypothetically, "nftrack") and we'll have to either keep a name that became wrong or make a syntax change.
> > I think `connection-status` is both neutral and clear, but if you have better ideas, please share!
> 
> I took a very bad example here. My main issue is the snat and dnat terminus which we should name nat source and nat destination to match the NAT subsystem.
> 
> `set firewall name <name> rule <rule> connection-status nat <source | destination >` would work for me.

Sound good to me too. Already changed and updated.

c-po (approved):

c-po:
This implementation supports either source or destination nat. A rule can not have both source and destination, as the leafNode is not a multi node.

Is this intentional?

dmbaturin (approved):

PR #1282, opened 2022-04-11
T4327: ignore PermissionError caused by ethtool spee/duplex/autoneg setting

dmbaturin:
Certain NICs seem to fail to report that they don't support speed/duplex setting,
so they look as if it's supported, but the command fails in practice.

The Hyper-V NIC is one example.

It's probably better to preserve a working config in that case, since the script will try to set autoneg by default (arguably, should, unless the NIC says it can't). If a NIC says it can set it but actually can't, there's nothing we can do to keep its config functional, other than ignore the error.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4327

## Component(s) name

* ethernet

## Checklist:

### DIFF
--- a/python/vyos/ifconfig/ethernet.py
+++ b/python/vyos/ifconfig/ethernet.py
@@ -170,11 +170,18 @@ class EthernetIf(Interface):
                 return
 
         cmd = f'ethtool --change {ifname}'
-        if speed == 'auto' or duplex == 'auto':
-            cmd += ' autoneg on'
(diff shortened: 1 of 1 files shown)


c-po:
Please add the driver to the "Known issue" list of drivers not capable of setting speed/duplex:

https://github.com/vyos/vyos-1x/blob/current/python/vyos/ethtool.py#L22

PR #1322, opened 2022-05-11
Firewall: T3907: add log-level options in firewall

nicolas-fort:
## Change Summary

Add possibility for choosing log-level on firewall rules.
## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T3907

## Component(s) name

firewall
## Proposed changes

Add possibility for choosing log-level on firewall rules. All options supported for nftables are included.
[Nftables reference](https://wiki.nftables.org/wiki-nftables/index.php/Quick_reference-nftables_in_10_minutes#Log)
## How to test

cli
```
vyos@vyos# set firewall name FOO rule 10 
Possible completions:
   action       Rule action [REQUIRED]
 > connection-status
                Connection status
   description  Description
 > destination  Destination parameters
   disable      Option to disable firewall rule
 > fragment     IP fragment match
 > icmp         ICMP type and code information
 > ipsec        Inbound IPsec packets
 > limit        Rate limit using a token bucket filter
   log          Option to log packets matching rule
   log-level    Set log-level. Log must be enable.
   protocol     Protocol to match (protocol name, number, or "all")
 > recent       Parameters for matching recently seen sources
 > source       Source parameters
 > state        Session state
 > tcp          TCP flags to match
 > time         Time to match rule                
      
[edit]
vyos@vyos#set firewall name FOO rule 10 log-level 
Possible completions:
   emerg        Emerg log level
   alert        Alert log level
   crit         Critical log level
   err          Error log level
   warn         Warning log level
   notice       Notice log level
   info         Info log level
   debug        Debug log level

```
Example:
```
set firewall ipv6-name TEST default-action 'accept'
set firewall ipv6-name TEST enable-default-log 
set firewall ipv6-name TEST rule 10 action 'drop'
set firewall ipv6-name TEST rule 10 log 'enable'
set firewall ipv6-name TEST rule 10 log-level 'err'
set firewall ipv6-name TEST rule 10 protocol 'udp'

set firewall name FOO default-action 'drop'
set firewall name FOO enable-default-log
set firewall name FOO rule 1 action 'accept'
set firewall name FOO rule 1 log 'enable'
set firewall name FOO rule 1 log-level 'notice'
set firewall name FOO rule 1 protocol 'tcp'


## In nftables:
vyos@vyos# sudo nft list chain ip filter NAME_FOO
table ip filter {
	chain NAME_FOO {
		meta l4proto tcp log prefix "[FOO-1-A]" level notice counter packets 0 bytes 0 return comment "FOO-1"
		counter packets 0 bytes 0 log prefix "[FOO-default-D]" drop comment "FOO default-action drop"
	}
}

vyos@vyos# sudo nft list chain ip6 filter NAME6_TEST
table ip6 filter {
	chain NAME6_TEST {
		meta l4proto udp log prefix "[TEST-10-D]" level err counter packets 0 bytes 0 drop comment "TEST-10"
		counter packets 0 bytes 0 log prefix "[TEST-default-A]" return comment "TEST default-action accept"
	}
}

```
## Checklist:

### DIFF
--- a/interface-definitions/firewall.xml.in
+++ b/interface-definitions/firewall.xml.in
@@ -599,7 +599,7 @@
             </properties>
             <children>
               #include <include/firewall/action-accept-drop-reject.xml.i>
-              #include <include/firewall/log.xml.i>
+              #include <include/firewall/rule-log-level.xml.i>
             </children>
           </node>
           <node name="invalid">
@@ -608,7 +608,7 @@
             </properties>
             <children>
               #include <include/firewall/action-accept-drop-reject.xml.i>
-              #include <include/firewall/log.xml.i>
+              #include <include/firewall/rule-log-level.xml.i>
             </children>
           </node>
           <node name="related">
@@ -617,7 +617,7 @@
             </properties>
             <children>
               #include <include/firewall/action-accept-drop-reject.xml.i>
-              #include <include/firewall/log.xml.i>
+              #include <include/firewall/rule-log-level.xml.i>
             </children>
           </node>
         </children>
--- a/interface-definitions/include/firewall/common-rule.xml.i
+++ b/interface-definitions/include/firewall/common-rule.xml.i
@@ -76,25 +76,7 @@
     </leafNode>
   </children>
 </node>
-<leafNode name="log">
-  <properties>
-    <help>Option to log packets matching rule</help>
-    <completionHelp>
-      <list>enable disable</list>
-    </completionHelp>
-    <valueHelp>
-      <format>enable</format>
-      <description>Enable log</description>
-    </valueHelp>
-    <valueHelp>
-      <format>disable</format>
-      <description>Disable log</description>
-    </valueHelp>
-    <constraint>
-      <regex>(enable|disable)</regex>
-    </constraint>
-  </properties>
-</leafNode>
+#include <include/firewall/rule-log-level.xml.i>
 <node name="connection-status">
   <properties>
     <help>Connection status</help>
--- a/interface-definitions/include/firewall/common-rule.xml.i
+++ b/interface-definitions/include/firewall/common-rule.xml.i
@@ -76,6 +76,25 @@
     </leafNode>
   </children>
 </node>
+<leafNode name="log">
+  <properties>
+    <help>Option to log packets matching rule</help>
+    <completionHelp>
+      <list>enable disable</list>
+    </completionHelp>
+    <valueHelp>
+      <format>enable</format>
+      <description>Enable log</description>
+    </valueHelp>
+    <valueHelp>
+      <format>disable</format>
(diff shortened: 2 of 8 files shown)


nicolas-fort:
I'll change migration scripts: revert changes on 6-to-7 and create a new 7-to-8

dmbaturin (changes requested):
I like the idea, but I'm not sure if this is the best UI for it. I have a few arguments in favor of a separate option (like `default-log-level` and per-rule `log-level`) and against changing the behavior of the current options.

* Changing current options requires a migration script, new options don't.
* Separate log level options will be easily discoverable in tab completion, while arguments will not (as in, the user will not know about log-level support until they see the help text inside `log`/`enable-default-log`).
* This is somewhat subjective, but forcing the user to think about the log level doesn't feel right—more of the time people only enable logging for a short time to grep for it, since its performance impact is very noticeable.

I believe if we are ready to go the migration script route, we may overhaul the log syntax for the better. For example, it could be a `rule 10 log` node that enables logging when created amd may offer multiple options inside it, e.g. `rule 10 level ...` and `rule 10 log prefix ...` for overriding the default prefix to something distinctive.

nicolas-fort:
I agree that forcing user to choose for a log-level might not be the best option, when he wants and he is used to simply enable log.
So if I understand you correctly, you are suggesting to leave log option as it is:
```
vyos@Receiver-side# set firewall name FOO rule 1 log 
Possible completions:
   <text>       Option to log packets matching rule
   disable      
   enable       
```
And also, add new functionality as:
```
vyos@vyos# set firewall name FOO rule 1 log-level
Possible completions:
   emerg        Emerg log level
   alert        Alert log level
   crit         Critical log level
   err          Error log level
   warn         Warning log level (default)
   notice       Notice log level
   info         Info log level
   debug        Debug log level
```
If this is correct, I only want to remark that command `set firewall name FOO rule 1 log disable` is useless. It does nothing. Shall we keep it just for backwards compatibility?

dmbaturin:
@nicolas-fort Sorry for late reply. Let me clarify.

First, I think _this_ PR may indeed better be limited to _just_ adding a new `log-level` option.

Second, I think we should also rework the way logging it enabled: switch from a boolean `log <enable|disable>` node to a valueless node `log`. My reasoning is exactly like yours: `log disable` is a useless state that duplicates simply not setting or removing the `log enable` option. A valueless node `log` wouldn't have that problem.

I also think that changing the syntax for enabling logging is a separate concern. We should make a separate task and a separate PR for it.

In other words, after this PR, the user should be able to change the log level for a rule with `set firewall name Foo rule 10 log-level $level`, but the way to enable or disable logging will remain unchanged. We will improve the way to enable logging later.

nicolas-fort:
Fixed

PR #1341, opened 2022-05-29
T3214: Fixed ovpn config incompatibilities for IPv4+IPv6 and IPv6-onl…

leonkramer:
## Change Summary
T3214: Fixed ovpn config incompatibilities for IPv4+IPv6 and IPv6-only use. "nopool" option not necessary. net30 "ifconfig-pool" statement handled by OpenVPN if "nopool" is omitted. IPv6 route push was nested inside IPv4 if block, now in IPv6 block. Removed mandatory IPv4 subnet for IPv6-only use.

## Types of changes

## Related Task(s)
* https://phabricator.vyos.net/T3214

## Component(s) name
openvpn

## Proposed changes
1) Fixed OpenVPN config incompatibilities for IPv4+IPv6 and IPv6-only use. "nopool" option not necessary in generated openvpn Configs because net30 "ifconfig-pool" statements are handled by OpenVPN if "nopool" is omitted. 
2) The generated config would allow a IPv6 only usage of the OpenVPN, but a commit error appears which prevent that saying that an IPv4 subnet is required. I have removed that check.
3) Another incompatibility after change of 2) is that if only IPv6 "push-route" statement is defined, the IPv6 route is not being pushed because it is nested inside an IPv4 if block. I have moved the code into the IPv6 block. 

## How to test
* Configure OpenVPN interface with mode=server. 
* Configure IPv4-only Subnet @ \[interfaces openvpn vtunN server] = PrePatch: OK - PostPatch: OK
* Configure IPv6-only Subnet @ \[interfaces openvpn vtunN server\] = PrePatch: FAIL - PostPatch: OK
* * Commit Error: IPv6 server requires an IPv4 server subnet
* Configure IPv6+IPv4 Subnet @ \[interfaces openvpn vtunN server] = PrePatch: FAIL - PostPatch: OK
* * OpenVPN Startup Error1: Options error: --server-ipv6 is incompatible with 'nopool' option
* * OpenVPN Startup Error2: Options error: --server already defines an ifconfig-pool, so you can't also specify --ifconfig-pool explicitly

## Checklist:

### DIFF
--- a/data/templates/openvpn/server.conf.j2
+++ b/data/templates/openvpn/server.conf.j2
@@ -71,28 +71,24 @@ topology {{ server.topology }}
 {%             endif %}
 {%             for subnet in server.subnet %}
 {%                 if subnet | is_ipv4 %}
-server {{ subnet | address_from_cidr }} {{ subnet | netmask_from_cidr }} nopool
+server {{ subnet | address_from_cidr }} {{ subnet | netmask_from_cidr }}
 {# First ip address is used as gateway. It's allows to use metrics #}
 {%                     if server.push_route is vyos_defined %}
 {%                         for route, route_config in server.push_route.items() %}
 {%                             if route | is_ipv4 %}
 push "route {{ route | address_from_cidr }} {{ route | netmask_from_cidr }} {{ subnet | first_host_address ~ ' ' ~ route_config.metric if route_config.metric is vyos_defined }}"
-{%                             elif route | is_ipv6 %}
+{%                             endif %}
+{%                         endfor %}
+{%                     endif %}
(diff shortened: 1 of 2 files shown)


c-po:
HI @leonkramer,

sorry for the late reply, can you share a VyOS configuration that you used to get it up and running? I am courious in also adding it to our documentation.

dmbaturin:
We'll merge it once 1.4/sagitta is branched off.

dmbaturin:
It seems to be the same functionality as in https://github.com/vyos/vyos-1x/pull/1637 which is merged now.

PR #1388, opened 2022-07-04
event-handler: T4508: Fixed environment variables

zdc:
## Change Summary

Fixed usage of environment variables - made an individual environment variable for an event a true copy, instead of a reference to a single environ dictionary.
Also, reorganized imports according to PEP8.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4508

## Component(s) name

event-handler

## Proposed changes

The `environment = environ` assignment is wrong because it creates a reference to `environ`, so each next event config will overwrite environment items. There must be a `deepcopy()` to do the assignment properly.

## How to test

A simple configuration to trigger the problem:
```
set service event-handler event event1 filter pattern 'testp1'
set service event-handler event event1 script environment vari1 value 'value11'
set service event-handler event event1 script environment vari2 value 'value21'
set service event-handler event event1 script path '/bin/true'
set service event-handler event event2 filter pattern 'testp2'
set service event-handler event event2 script environment vari1 value 'value12'
set service event-handler event event2 script environment vari2 value 'value22'
set service event-handler event event2 script path '/bin/true'
```
Without the fix, both events have the same environments.

## Checklist:

### DIFF
--- a/src/system/vyos-event-handler.py
+++ b/src/system/vyos-event-handler.py
@@ -15,14 +15,16 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import argparse
-import select
-import re
 import json
+import re
+import select
+from copy import deepcopy
 from os import getpid, environ
 from pathlib import Path
 from signal import signal, SIGTERM, SIGINT
-from systemd import journal
 from sys import exit
+from systemd import journal
+
 from vyos.util import run, dict_search
 
 # Identify this script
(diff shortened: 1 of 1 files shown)


dmbaturin (approved):
I agree with the real change, but I think it shouldn't have been in the same commit with the import section restyle.

c-po:
> but I think it shouldn't have been in the same commit with the import section restyle

I agree here, but as we do not have bullet proof rules/commit guidelines I am happy to merge it.

PR #1402, opened 2022-07-07
T4517: ip: Add option to enable directed broadcast forwarding

vfreex:
Directed broadcast is described in rfc1812#section-5.3.5.2
and rfc2644.

By default Linux kernel doesn't forward directed broadcast
packets unless both of `/proc/sys/net/ipv4/conf/all/bc_forwarding`
and `/proc/sys/net/ipv4/conf/$iface/bc_forwarding` are set to 1.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4517

## Checklist:

### DIFF
--- a/interface-definitions/include/interface/enable-directed-broadcast.xml.i
+++ b/interface-definitions/include/interface/enable-directed-broadcast.xml.i
@@ -1,0 +1,8 @@
+<!-- include start from interface/enable-directed-broadcast.xml.i -->
+<leafNode name="enable-directed-broadcast">
+  <properties>
+    <help>Enable directed broadcast forwarding on this interface</help>
+    <valueless/>
+  </properties>
+</leafNode>
+<!-- include end -->
--- a/interface-definitions/include/interface/ipv4-options.xml.i
+++ b/interface-definitions/include/interface/ipv4-options.xml.i
@@ -8,6 +8,7 @@
     #include <include/interface/arp-cache-timeout.xml.i>
     #include <include/interface/disable-arp-filter.xml.i>
     #include <include/interface/disable-forwarding.xml.i>
+    #include <include/interface/enable-directed-broadcast.xml.i>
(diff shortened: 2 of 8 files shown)


c-po:
Thanks for this PR. Unfortunately it changes VyOS default behaviour which is something that we do not take easy.

Adding `net.ipv4.conf.all.bc_forwarding=1` to `src/etc/sysctl.d/30-vyos-router.conf` will enable directed broadcast forwarding on ALL interfaces, which is currently not the case in VyOS 1.2 - 1.4, and we should keep it that way. So please drop this from your commit. 

The enabling of this feature on a per interface basis is good and I would like to merge it once the other part is resolved.

vfreex:
@c-po Hi, setting `net.ipv4.conf.all.bc_forwarding=1` on sysctl doesn't enable directed broadcast forwarding automatically. To enable directed broadcast on an interface, both of `net.ipv4.conf.all.bc_forwarding` and per interface `net.ipv4.conf.$iface.bc_forwarding=1` need to be set to `1`. This is described at https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/networking/ip-sysctl.rst?h=linux-5.18.y#n270. I also verified this behavior in my local environment: The logic is `AND` instead of `OR`.

Even with this PR, per interface value is still default to `0`. This is because `/proc/sys/net/ipv4/conf/default/bc_forwarding` is default to `0` on new interfaces and https://github.com/vyos/vyos-1x/pull/1402/files#diff-e0ccbc8a48513bf9440d14497c51fd2ec1583a051bf6c4b3f74d03632958d101R1518 explicitly sets managed interfaces to `0`.

vfreex:
I added a new commit to explicitly set `net.ipv4.conf.default.bc_forwarding = 0` and describe the behavior in the comment.

PR #1498, opened 2022-08-26
pppoe: T4648: fix incorrect installation of IPv6 default route even when default-route is set to none 

initramfs:
## Change Summary
Currently an IPv6 default route is installed as a response to the received router advertisement on the PPPoE interface regardless of the `default-route` setting in the PPPoE interface.

This change suppresses the automated route installation by tuning the kernel sysctl `accept_ra_defrtr` to ignore default routers advertised in RAs on the ppp interface coming up in `/etc/ppp/ipv6-up.d/`

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4648

## Component(s) name
pppoe

## Proposed changes

The template `ipv6-up.script.tmpl` is adjusted to add a line:
```sh
echo 0 > /proc/sys/net/ipv6/conf/{{ ifname }}/accept_ra_defrtr
```
to suppress automatic default route installation by the kernel, only when the `default-route` parameter is set to `none`.

## How to test

1. Create a basic PPPoE interface with `none` the `default-route` with IPv6 configured.  e.g.:
```
interfaces {
    pppoe pppoe0 {
        ...

        default-route none

        ipv6 {
            address {
                autoconf
            }
        }
    }
}
```
2. Once the interface is connected, run `show ipv6 route` and ensure there are no default routes installed.

## Checklist:

### DIFF
--- a/data/templates/pppoe/ipv6-up.script.tmpl
+++ b/data/templates/pppoe/ipv6-up.script.tmpl
@@ -36,6 +36,14 @@ echo 1 > /proc/sys/net/ipv6/conf/{{ ifname }}/forwarding
 #
 echo 2 > /proc/sys/net/ipv6/conf/{{ ifname }}/accept_ra
 
+{% if default_route == 'none' %}
+# Prevent learning of default router from router advertisements
+echo 0 > /proc/sys/net/ipv6/conf/{{ ifname }}/accept_ra_defrtr
+{% else %}
+# Enable learning of default router from router advertisements
+echo 1 > /proc/sys/net/ipv6/conf/{{ ifname }}/accept_ra_defrtr
+{% endif %}
+
 # Autoconfigure addresses using Prefix Information in Router Advertisements.
 echo 1 > /proc/sys/net/ipv6/conf/{{ ifname }}/autoconf
 {% endif %}


data/templates/pppoe/ipv6-up.script.tmpl
```diff
@@ -36,6 +36,11 @@ echo 1 > /proc/sys/net/ipv6/conf/{{ ifname }}/forwarding
 #
 echo 2 > /proc/sys/net/ipv6/conf/{{ ifname }}/accept_ra
 
+{% if default_route == 'none' %}
+# Prevent learning of default router from router advertisements
+echo 0 > /proc/sys/net/ipv6/conf/{{ ifname }}/accept_ra_defrtr
```
c-po, line 41: If I switch back to `default_route auto`, where is this case handled?

initramfs: If `accept_ra_defrtr` is not set, the functional default is to follow what `accept_ra` is set to (in the sense that if `accept_ra` is >= 1, `accept_ra_defrtr` will also be 1). This is also why we have never needed to manually set this parameter.

Since the PPPoE interface disappears when it goes down, we don't have to worry about any 'old' value staying around. I have tested this with setting 'default-route' to none, committing, and the followed by `default-route` to auto, and the routing table correctly gets the default route.

If this is not satisfactory, I can explicitly enable `accept_ra_defrtr` when `default-route` is not none.

c-po: Please explicitly enable it so it's easier to read/understand. Other then that it looks good to me. THanks.

initramfs: I've updated the template to explicitly enable `accept_ra_defrtr` when `default-route` is not none.

c-po:
Forward ported this to the `current` branch in 9126170f.

c-po (approved):

sever-sever (approved):

PR #1517, opened 2022-09-02
bonding: T4668: Fix bond members not adding/interface state incorrect

initramfs:
## Change Summary
Fixes bugs around bonding member interfaces not being added to the bond or being added/removed with incorrect interface states (as defined by if the interface is meant to be disabled).

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4668

## Component(s) name
bonding

## Proposed changes
There were a few bugs ranging from simple typos to incorrect behavioral assumptions being made in the bonding code that have been corrected by cleaning the code and being more explicit about interface states. In particular:

In `interfaces-bonding.py`:
* There was a typo of `interfaces` which has been fixed
* A bug with assuming the wrong config level has been addressed

Additionally a new disabled configuration state has been added to track if a member interface is disabled such that we set correct interface states when enslaving interfaces.

## How to test
Create a simple bond using config similar to this:
```
interfaces {
    bonding bond0 {
        mode 802.3ad
        hash-policy layer2+3
        lacp-rate fast
        member {
            interface eth0
        }
    }
}
```
 * Attempt to add bonding members, one at a time (and committing), to the bond, observe that the current behavior eventually stops adding interfaces after the first two.
 * Mark any of the bond members as disabled whilst adding/removing the interface, observe the interface does not stay down.

After the patch is applied, validate both cases no longer hold true.
## Checklist:

### DIFF
--- a/src/conf_mode/interfaces-bonding.py
+++ b/src/conf_mode/interfaces-bonding.py
@@ -89,6 +89,11 @@ def get_config(config=None):
 
     # determine which members have been removed
     interfaces_removed = leaf_node_changed(conf, base + [ifname, 'member', 'interface'])
+
+    # Reset config level to interfaces
+    old_level = conf.get_level()
+    conf.set_level(['interfaces'])
+
     if interfaces_removed:
         bond['shutdown_required'] = {}
         if 'member' not in bond:
@@ -97,7 +102,7 @@ def get_config(config=None):
         tmp = {}
         for interface in interfaces_removed:
             section = Section.section(interface) # this will be 'ethernet' for 'eth0'
-            if conf.exists(['insterfaces', section, interface, 'disable']):
+            if conf.exists([section, interface, 'disable']):
                 tmp[interface] = {'disable': ''}
             else:
                 tmp[interface] = {}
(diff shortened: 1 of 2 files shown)


src/conf_mode/interfaces-bonding.py
```diff
@@ -73,63 +73,83 @@ def get_config(config=None):
...
+            section = Section.section(interface) # this will be 'ethernet' for 'eth0'
+            if conf.exists([section, interface, 'disable']):
+                interface_config['disable'] = ''
+
+            conf.set_level(old_level)
+
             # Check if member interface is already member of another bridge
             tmp = is_member(conf, interface, 'bridge')
-            if tmp: bond['member']['interface'][interface].update({'is_bridge_member' : tmp})
+            if tmp: interface_config['is_bridge_member'] = tmp
```
c-po, line 133: That somehow never worked for me - operating on the reference. Are you sure it works now? That's why I stick with `bond['member']['interface'][interface]`

initramfs: As long as the reference itself is not replaced (i.e. `interface_config = ...`) this code should function the same (as they operate on the same reference). Just as a sanity check, I did print out `bond['member']['interface'][interface] is interface_config` just now in the for loop as a debugging measure and it returns `True`.

While I haven't checked individually if each key (e.g. 'is_bridge_member', 'is_source_address', etc) works, dumping the generated config during testing did show the config updating correctly.

If the reference is replaced though (i.e. `interface_config = ...`) then there is a difference as `bond['member']['interface'][interface] = ...` internally calls `__setitem__()` on the `bond['member']['interface']` dict, whereas assigning to `interface_config` merely mutates the copy of the reference that was made (in essence making it reference something else). ~Perhaps you were running issues here?~

**Edit:** Nevermind I think you were experiencing a bug I fixed on line 79 (where each interface's dict was the same reference), mutating one would mutate them all (and cause all manner of bugs such as wrong interfaces failing the validation step).

c-po: You can output the dict and see what happens by running „touch /tmp/vyos.ifconfig.debug“ and then commit

initramfs: I have already done something similar to that during testing (via insertion of print statements to file) but that's a good shortcut to know.

PR #1551, opened 2022-09-21
ospf: T4707: Add OSPF segment routing for FRR

Cheeze-It:
This change is to add OSPF segment routing for FRR. 

## Change Summary

With this change, we will be introducing the code to configure segment routing for OSPF.

We will be adding to the ospfd jinja2 template, adding tie breakers for the protocol handler,
adding smoketests, adding the CLI hierarchy. This configuration should make OSPF be able
to enable segment routing for prefixes, as well as enable it for interfaces. 

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4707

## Component(s) name

OSPF

## Proposed changes

We added the XML hierarchy. This involved adding the actual segment routing nodes, as well as
adding the high and low label file for segment routing. We added smoketests so that if for some
reason the config hierarchy fails between builds then we should catch it. We added the jinja2
FRR template changes for ospfd that way we can configure the actual lines in FRR. 

## How to test

Here is the baseline VyOS config here:

```
vyos@vyos:~$ vtysh -c "show run"
Building configuration...

Current configuration:
!
frr version 8.3.1
frr defaults traditional
hostname vyos
log syslog
log facility local7
service integrated-vtysh-config
!
ip route 0.0.0.0/0 10.0.0.65
!
end
```

Here is what we are adding:

```
set protocols ospf interface lo passive
set protocols ospf segment-routing enable
set protocols ospf segment-routing global-block high-label-value '100'
set protocols ospf segment-routing global-block low-label-value '50'
set protocols ospf segment-routing prefix 192.168.0.1/32 index explicit-null
set protocols ospf segment-routing prefix 192.168.0.1/32 index value '1'
set protocols ospf segment-routing prefix 192.168.0.2/32 index no-php-flag
set protocols ospf segment-routing prefix 192.168.0.2/32 index value '2'
set protocols ospf segment-routing prefix 192.168.0.3/32 index explicit-null
set protocols ospf segment-routing prefix 192.168.0.3/32 index value '3'
set protocols ospf segment-routing prefix 192.168.0.4/32 index no-php-flag
set protocols ospf segment-routing prefix 192.168.0.4/32 index value '4'
```

Here is how it looks after commit:

```
vyos@vyos# compare
[edit protocols]
+ospf {
+    interface lo {
+        passive {
+        }
+    }
+    segment-routing {
+        enable
+        global-block {
+            high-label-value 100
+            low-label-value 50
+        }
+        prefix 192.168.0.1/32 {
+            index {
+                explicit-null
+                value 1
+            }
+        }
+        prefix 192.168.0.2/32 {
+            index {
+                no-php-flag
+                value 2
+            }
+        }
+        prefix 192.168.0.3/32 {
+            index {
+                explicit-null
+                value 3
+            }
+        }
+        prefix 192.168.0.4/32 {
+            index {
+                no-php-flag
+                value 4
+            }
+        }
+    }
+}
[edit]
vyos@vyos# commit

vyos@vyos# save
Saving configuration to '/config/config.boot'...
Done
[edit]
vyos@vyos# exit
exit
vyos@vyos:~$ vtysh -c "show run"
Building configuration...

Current configuration:
!
frr version 8.3.1
frr defaults traditional
hostname vyos
log syslog
log facility local7
service integrated-vtysh-config
!
ip route 0.0.0.0/0 10.0.0.65
!
interface eth0
 bandwidth 10000
exit
!
interface eth1
 bandwidth 10000
exit
!
interface lo
 ip ospf dead-interval 40
 ip ospf passive
exit
!
router ospf
 auto-cost reference-bandwidth 100
 timers throttle spf 200 1000 10000
 segment-routing global-block 50 100
 segment-routing prefix 192.168.0.1/32 index 1 explicit-null
 segment-routing prefix 192.168.0.2/32 index 2 no-php-flag
 segment-routing prefix 192.168.0.3/32 index 3 explicit-null
 segment-routing prefix 192.168.0.4/32 index 4 no-php-flag
exit
!
```

When we try to configure both explicit-null and no-php-flag we get a proper failure:

```
vyos@vyos# set protocols ospf segment-routing prefix 192.168.0.1/32 index no-php-flag
[edit]
vyos@vyos# compare
[edit protocols ospf segment-routing prefix 192.168.0.1/32 index]
+no-php-flag
[edit]
vyos@vyos# commit

Segment routing prefix 192.168.0.1/32 cannot have both explicit-null and
no-php-flag configured at the same time.

[[protocols ospf]] failed
Commit failed
```

If we remove the "value" for the segment ID number we get the proper failure:

```
vyos@vyos# delete protocols ospf segment-routing prefix 192.168.0.1/32 index value
[edit]
vyos@vyos# compare
[edit protocols ospf segment-routing prefix 192.168.0.1/32 index]
-value 1
[edit]
vyos@vyos# commit

Segment routing prefix 192.168.0.1/32 index value cannot be blank.

[[protocols ospf]] failed
Commit failed
```

## Checklist:

- [ ] I have updated the documentation accordingly  <--- I plan on making ISIS and OSPF changes to documentation once merged

### DIFF
--- a/interface-definitions/include/accel-ppp/radius-additions-rate-limit.xml.i
+++ b/interface-definitions/include/accel-ppp/radius-additions-rate-limit.xml.i
@@ -6,7 +6,7 @@
   <children>
     <leafNode name="attribute">
       <properties>
-        <help>Specifies which RADIUS attribute contains rate information. (default is Filter-Id)</help>
+        <help>Specifies which RADIUS attribute contains rate information</help>
       </properties>
       <defaultValue>Filter-Id</defaultValue>
     </leafNode>


Cheeze-It:
I did some digging, and it *seems* that inside of FRR, OSPF segment routing may be missing the command that is there in ISIS. The command specifically is "segment-routing on" which is kind of odd.

I tried to add it and it did *NOT* show up like it should have:

```
vyos# show running-config
Building configuration...

Current configuration:
!
frr version 8.3.1
frr defaults traditional
hostname vyos
log syslog
log facility local7
service integrated-vtysh-config
!
ip route 0.0.0.0/0 10.0.0.65
!
interface lo
 ip ospf dead-interval 40
 ip ospf passive
exit
!
router ospf
 auto-cost reference-bandwidth 100
 timers throttle spf 200 1000 10000
 segment-routing global-block 50 100
 segment-routing prefix 192.168.0.1/32 index 1 explicit-null
 segment-routing prefix 192.168.0.2/32 index 2 no-php-flag
 segment-routing prefix 192.168.0.3/32 index 3 explicit-null
 segment-routing prefix 192.168.0.4/32 index 4 no-php-flag
exit
!
```

Adding it here:


```
vyos# configure terminal
vyos(config)# router ospf
vyos(config-router)# segment-routing
  global-block  Segment Routing Global Block label range
  node-msd      Maximum Stack Depth for this router
  on            Enable Segment Routing
  prefix        Prefix SID
vyos(config-router)# segment-routing on
vyos(config-router)# end
```


But it still didn't show up:


```
vyos# show running-config
Building configuration...

Current configuration:
!
frr version 8.3.1
frr defaults traditional
hostname vyos
log syslog
log facility local7
service integrated-vtysh-config
!
ip route 0.0.0.0/0 10.0.0.65
!
interface lo
 ip ospf dead-interval 40
 ip ospf passive
exit
!
router ospf
 auto-cost reference-bandwidth 100
 timers throttle spf 200 1000 10000
 segment-routing global-block 50 100
 segment-routing prefix 192.168.0.1/32 index 1 explicit-null
 segment-routing prefix 192.168.0.2/32 index 2 no-php-flag
 segment-routing prefix 192.168.0.3/32 index 3 explicit-null
 segment-routing prefix 192.168.0.4/32 index 4 no-php-flag
exit
!
```


We *may* have to bring this up to the FRR team to see if it was just omitted. This seems to cause my smoketests to fail as it's not there, but the command *is* executed.

interface-definitions/include/ospf/protocol-common-config.xml.i
```diff
@@ -621,6 +621,97 @@
     </constraint>
   </properties>
 </leafNode>
+<node name="segment-routing">
+  <properties>
+    <help>Segment-Routing (SPRING) settings</help>
+  </properties>
+  <children>
+    <leafNode name="enable">
```
c-po: We do not require a dedicated `enable` node. If SR is configured on the CLI we assume the user want's to enable it.

interface-definitions/include/ospf/protocol-common-config.xml.i
```diff
@@ -621,6 +621,97 @@
...
+    </leafNode>
+    <tagNode name="prefix">
+      <properties>
+        <help>Static IPv4/IPv6 prefix segment/label mapping</help>
+        <valueHelp>
+          <format>ipv4net</format>
+          <description>IPv4 prefix segment</description>
+        </valueHelp>
+        <valueHelp>
+          <format>ipv6net</format>
```
c-po: OSPFv2 supports OSPFv3 IPv6 prefixes?

c-po (changes requested):

interface-definitions/include/ospf/protocol-common-config.xml.i
Cheeze-It: Right, so in this case I believe it is needed as a toggle as a distinct on/off for really one reason only. For segment routing normally there are defaults that are there. If someone doesn't want to mess with the defaults they should be able to just enable segment routing. Furthermore, if one doesn't want to change the SRGB/SRLB and wants to use the router as a transit LSR and doesn't want to inject prefixes then they can. That's why I believe in FRR there is a toggle for "on." 

In this case, ISIS and OSPF in FRR both have "segment-routing on" as can be seen:

```
vyos(config)# router isis test
vyos(config-router)# segment-routing
  global-block  Segment Routing Global Block label range
  node-msd      Maximum Stack Depth for this router
  on            Enable Segment Routing
  prefix        Prefix SID
vyos(config-router)# exit

vyos(config)# router ospf
vyos(config-router)# segment-routing
  global-block  Segment Routing Global Block label range
  node-msd      Maximum Stack Depth for this router
  on            Enable Segment Routing
  prefix        Prefix SID
```

That's the only reason why I generally use the word "enable" in there. It offers a functionality that doesn't have any other command for.

I had another idea. Instead of having "enable" in ISIS and OSPF for segment routing, maybe we can match on the word "segment-routing" and if that is present to just switch it on instead of having a specific "enable" command. I honestly would rather not do that though. An explicit enable I think is the right choice.

interface-definitions/include/ospf/protocol-common-config.xml.i
Cheeze-It: Ahh yes, this I'll remove. Thank you.

Cheeze-It: It is complete.

interface-definitions/include/ospf/protocol-common-config.xml.i
Cheeze-It: I have made it to now that the "segment-routing" node just needs to be there and it will enable segment-routing within ISIS and OSPF.

runborg:
Hi!

It looks like we are hitting some FRR buge while adding this.. 
1. Enabling segment-routing results in a new `router ospf\n router-info area` command beeing added to the configuration. this is a default and i dont think it should be vissible to the user.  not adding this to our templates results in frr-reload trying to remove it from the configuration and that results in an error:
```
vyos(config-router)# no router-info area
% Unknown command: no router-info area
```
2. Enabling `segment-routing on` inside vtysh results in it beeing enabled, but when doing anything else segment-routing related after that point results in the `segment-routing on` command beeing removed from the configuration. that also makes frr-reload fail as it failes to add the frr commands because it beleves segment-rouing isnt enabled.
```
vyos(config-router)# do sh run
Building configuration...

Current configuration:
!
frr version 8.3.1
frr defaults traditional
hostname debian
log syslog
log facility local7
hostname vyos
service integrated-vtysh-config
!
router ospf
 segment-routing on
 router-info area
exit
!
end
```
then adding more segment-routing commands: 
```
vyos(config-router)# segment-routing global-block 50 100
vyos(config-router)# do sh run
Building configuration...

Current configuration:
!
frr version 8.3.1
frr defaults traditional
hostname debian
log syslog
log facility local7
hostname vyos
service integrated-vtysh-config
!
router ospf
 segment-routing global-block 50 100
 router-info area
exit
!
end
```
Here we see that `segment-routing on` is removed from the configuration which is no good.

3.  while trying to remove a ` segment-routing prefix 192.168.0.1/32 index 1 explicit-null` command, `explicit-null` needs to be skipped, and the whole index needs to be removed for the command to succeed.. because of the earlier noted issues i've not gotten to check if frr-reload handles this the correct way or not.. but i'm wondering if this is the correct way to handle this, or if it should be possible to get it removed by applying the whole command.

the most serious issue on this is the missing `segment-routing on`, as this makes frr-reload error out by the missing command
```
line 33: Failure to communicate[13] to ospfd, line:  segment-routing global-block 50 100

% OSPF SR is not turned on
line 36: Failure to communicate[13] to ospfd, line:  segment-routing prefix 192.168.0.1/32 index 1

% OSPF SR is not turned on
```

here is a full config loadable by frr-reload:
Loading this into frr-reload the first time makes frr-reload choke on execution.. by executing it multiple times it looks like its successful, and will choke on not be able to remove the `router-info area` command. 
```
vyos@vyos:~$ cat test.vtysh

!
frr version 8.3.1
frr defaults traditional
hostname vyos
log syslog
log facility local7
service integrated-vtysh-config
!
!
interface lo
 ip ospf priority 1
 ip ospf hello-interval 10
 ip ospf retransmit-interval 5
 ip ospf transmit-delay 1
 ip ospf dead-interval 40
 ip ospf passive
exit
!
!
router ospf
 auto-cost reference-bandwidth 100
 ospf abr-type cisco
 segment-routing on
 segment-routing global-block 50 100
 segment-routing prefix 192.168.0.1/32 index 1
 segment-routing prefix 192.168.0.1/32 index 1 explicit-null
 segment-routing prefix 192.168.0.2/32 index 2
 segment-routing prefix 192.168.0.2/32 index 2 no-php-flag
 segment-routing prefix 192.168.0.3/32 index 3
 segment-routing prefix 192.168.0.3/32 index 3 explicit-null
 segment-routing prefix 192.168.0.4/32 index 4
 segment-routing prefix 192.168.0.4/32 index 4 no-php-flag
 timers throttle spf 200 1000 10000
exit
!
end
```
which is loadable by:
```
sudo /usr/lib/frr/frr-reload.py --daemon ospfd --stdout --reload --log-level debug test.vtysh
```

Cheeze-It:
Opened an FRR issue. I'm sure they'll fix it in very little time as it's probably something simple.

c-po:
https://github.com/FRRouting/frr/issues/12007

Cheeze-It:
I made changes to the minimum label value for segment routing. I raised it from 16 to 100. It seems that in Zebra we have certain reserved minimum ranges when enabling SR. I don't believe this should cause much of a problem as we can go up to 1,048,576. This is a completely fine compromise for minimum label value. It also requires no changes to FRR either.

dmbaturin (approved):

data/templates/frr/isisd.frr.j2
```diff
@@ -107,9 +107,7 @@ router isis VyOS {{ 'vrf ' + vrf if vrf is vyos_defined }}
  mpls-te inter-as{{ level }}
 {% endif %}
 {% if segment_routing is vyos_defined %}
-{%     if segment_routing.enable is vyos_defined %}
```
c-po: Please move this change to a dedicated commit - it also requires a CLI migrator. Also the XML node must be altered to not have the enable node anymore. If you're not familiar with the migration script process just leave this out and we will take care of it.

interface-definitions/include/ospf/protocol-common-config.xml.i
c-po: I still see the `ipv6net` completion helper

c-po:
```
FAIL: test_ospf_14_segment_routing_configuration (__main__.TestProtocolsOSPF)
  File "/usr/libexec/vyos/tests/smoke/cli/test_protocols_ospf.py", line 429, in test_ospf_14_segment_routing_configuration
    self.assertIn(f' segment-routing on', frrconfig)
AssertionError: ' segment-routing on' not found in 'router ospf\n auto-cost reference-bandwidth 100\n timers throttle spf 200 1000 10000\n segment-routing global-block 100 199 local-block 200 299\n segment-routing node-msd 5\n segment-routing prefix 192.168.0.1/32 index 1\n segment-routing prefix 192.168.0.2/32 index 2\nexit\n!'

Ran 14 tests in 265.365s

FAILED (failures=1)
```

Unfortunately smoketests fail to run as `segment-routing on` is not shown when requesting the FRR configuration. It feels like an ordering issue. When moving the `segment-routing on` statement to the end of the Jinja2 template it does work.

```Jinja2
{% if segment_routing is vyos_defined %}
{%     if segment_routing.maximum_label_depth is vyos_defined %}
 segment-routing node-msd {{ segment_routing.maximum_label_depth }}
{%     endif %}
{%     if segment_routing.global_block is vyos_defined %}
{%         if segment_routing.local_block is vyos_defined %}
 segment-routing global-block {{ segment_routing.global_block.low_label_value }} {{ segment_routing.global_block.high_label_value }} local-block {{ segment_>
{%         else %}
 segment-routing global-block {{ segment_routing.global_block.low_label_value }} {{ segment_routing.global_block.high_label_value }}
{%         endif %}
{%     endif %}
{%     if segment_routing.prefix is vyos_defined %}
{%         for prefix, prefix_config in segment_routing.prefix.items() %}
{%             if prefix_config.index is vyos_defined %}
{%                 if prefix_config.index.value is vyos_defined %}
 segment-routing prefix {{ prefix }} index {{ prefix_config.index.value }}
{%                     if prefix_config.index.explicit_null is vyos_defined %}
 segment-routing prefix {{ prefix }} index {{ prefix_config.index.value }} explicit-null
{%                     elif prefix_config.index.no_php_flag is vyos_defined %}
 segment-routing prefix {{ prefix }} index {{ prefix_config.index.value }} no-php-flag
{%                     endif %}
{%                 endif %}
{%             endif %}
{%         endfor %}
{%     endif %}
 segment-routing on
{% endif %}
```

c-po (changes requested):

data/templates/frr/isisd.frr.j2
Cheeze-It: Should I remove all of my changes for ISIS on this and then just do a separate PR to address it? 

I was unsure if I should do that as I figured to make both OSPF and ISIS the same configs wise in this one commit.

Also, yeah, I can add a migrator.

PR #1557, opened 2022-09-24
firewall: T4709: fix firewall MSS clamping issues

initramfs:
## Change Summary

* ~Fixes TCP MSS clamping by converting the relevant rules to native nftables.~
* Fiixes the MSS clamping range for both IPv4 and IPv6.
* Backports the `clamp-mss-to-pmtu` option to equuleus

## Types of changes

## Related Task(s)
* https://phabricator.vyos.net/T4709

## Component(s) name
firewall

## Proposed changes

1. ~We refactor the MSS clamping rule from iptables to nftables to fix the conversion issue.~
2. We fix the MSS clamping range by raising the maximum to 65535 octets. We also move the arbitrary minimum of 500 octets to 536 for IPv4 to conform with RFC9293 and RFC791. We reduce the minimum MSS for IPv6 from 1280 to 1220 in accordance to RFC9293 and RFC8200.
3. Since we've already already converted the relevant iptables to nftables we can easily implement the `clamp-mss-to-pmtu` option, as such we add the ability to specify that option to use the PMTU instead of a hard-coded MSS value.

## How to test
Create an example firewall configuration such as:
```
firewall {
    options {
        interface eth0 {
            adjust-mss6 1220
        }
    }
}
```
and validate that commit succeeds.

Additionally when using the `clamp-mss-to-pmtu` option, validate the generated rule looks of the form:

> oifname "eth0" tcp flags & (syn | rst) == syn counter tcp option maxseg size set rt mtu

## Checklist:

### DIFF
--- a/interface-definitions/firewall-options.xml.in
+++ b/interface-definitions/firewall-options.xml.in
@@ -20,24 +20,40 @@
               <leafNode name="adjust-mss">
                 <properties>
                   <help>Adjust MSS for IPv4 transit packets</help>
+                  <completionHelp>
+                      <list>clamp-mss-to-pmtu</list>
+                  </completionHelp>
                   <valueHelp>
-                    <format>500-1460</format>
+                      <format>clamp-mss-to-pmtu</format>
+                      <description>Automatically sets the MSS to the proper value</description>
+                 </valueHelp>
+                  <valueHelp>
+                    <format>536-65535</format>
                     <description>TCP Maximum segment size in bytes</description>
                   </valueHelp>
                   <constraint>
-                    <validator name="numeric" argument="--range 500-1460"/>
+                    <validator name="numeric" argument="--range 536-65535"/>
+                    <regex>(clamp-mss-to-pmtu)</regex>
                   </constraint>
                 </properties>
               </leafNode>
               <leafNode name="adjust-mss6">
                 <properties>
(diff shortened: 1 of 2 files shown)


c-po:
Please kote, VyOS 1.3 does kot use nftables, it is all based on iptables

initramfs:
Maybe I'm wrong but it seems like the NAT components are using native nftables, in fact I followed the code in `src/conf_mode/nat.py` and template in `data/templates/firewall/nftables-nat.tmpl` when writing changes for this PR.

Also stated in the phabricator task, it's all nftables under the hood anyway (from debian's use of `iptables-nft`).

c-po:
nat.py is in the repos but not used in production in 1.3. The XML/node.def files are deleted during make so noone calls nat.py - VyOS 1.3 uses vyatta-cfg-nat package. Only VyOS 1.4 uses the new NAT implementation.

initramfs:
@c-po It seems I was wrong about the iptables version of the MSS clamp not working (I accidentally broke it whilst fixing something else and attributed it to iptables/nftables). I've updated the phabricator task to indicate as such.

As such I've removed the conversion to nftables and only left the MSS range fixes and the dynamic PMTU clamping options in this PR.

c-po (approved):

sever-sever (approved):

PR #1590, opened 2022-10-13
system login: T874: add 2FA support for local and ssh auth. Bugfix_2

goodNETnick:
## Change Summary

Bugfix for PR:
https://github.com/vyos/vyos-1x/pull/1555
Bugfix for error in parsing default parameters

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T874

## Component(s) name

system login, ssh

## Proposed changes

Bugfix for error in parsing default parameters

## Checklist:

### DIFF
--- a/src/conf_mode/system-login.py
+++ b/src/conf_mode/system-login.py
@@ -64,6 +64,13 @@ def get_config(config=None):
     login = conf.get_config_dict(base, key_mangling=('-', '_'),
                                  no_tag_node_value_mangle=True, get_first_key=True)
 
+    # Get global OTP parameters
+    default_values = defaults(base + ['authentication'])
+    if 'authentication' in login:
+        login['authentication'] = dict_merge(default_values, login['authentication'])
+    else:
+        login['authentication'] = default_values
+
     # users no longer existing in the running configuration need to be deleted
     local_users = get_local_users()
     cli_users = []


sever-sever (approved):

src/conf_mode/system-login.py
```diff
@@ -64,6 +64,11 @@ def get_config(config=None):
     login = conf.get_config_dict(base, key_mangling=('-', '_'),
                                  no_tag_node_value_mangle=True, get_first_key=True)
 
+    # Get global OTP parameters
+    default_values_otp = defaults(base + ['authentication'])
+    login['authentication'] = {}
```
sarthurdev: Why override this part of the config with an empty dict before merging default values?

sarthurdev (changes requested):

src/conf_mode/system-login.py
goodNETnick: Without it the script will display an error that this key does not exist:
```
  File "/usr/libexec/vyos/conf_mode/system-login.py", line 339, in <module>
    c = get_config()
  File "/usr/libexec/vyos/conf_mode/system-login.py", line 70, in get_config
    login['authentication'] = dict_merge(default_values_otp, login['authentication'])
KeyError: 'authentication'
```

sarthurdev: Should add a check if it doesn't exist, If authentication node does exist in the config it would be cleared leaving only default values.
```
if 'authentication' not in login:
    login['authentication'] = {}
```

c-po: Actually @goodNETnick and @sarthurdev all is already there and should follow the style already used. In this case:

```python
if 'authentication' in login:
    default_values = defaults(base + ['authentication'])
    login['authentication'] = dict_merge(default_values, login['authentication'])
```

c-po: The actual common style used in all implementations is:

```python
# retrieve default values
default_values = defaults(base)

# XXX: T2665: When there is no FOO configuration given, we can safely
# remove the default values from the dict.
if 'user' not in login:
    if 'user' in default_values:
        del default_values['user']

# ... same goes for radius ...
if RADIUS-FOO not in login ....
```

goodNETnick: Thanks! Fixed

c-po: Actually you did not fix it. You create a key if it's not there - you should rather use the well known good approach from every other implementation we have.

goodNETnick: Is it fixed now?

c-po: Why not use the solution I provided? Is there something wrong with it?

goodNETnick: I think your solution is great.
But in this case the default values are required not only for the **login authentication** section, but also for the **login user name authentication** section.
In that case, my new solution does not look bad either.
Or is it just as bad as the original one?

c-po:
Fixed via https://github.com/vyos/vyos-1x/commit/da535ef5697f6ce87a7f34ff185e4df239e6af63

goodNETnick:
There is a dangerous bug in your fix:
https://github.com/vyos/vyos-1x/blob/da535ef5697f6ce87a7f34ff185e4df239e6af63/debian/vyos-1x.postinst#L25-L30
**grep** checks one line but **sed** inserts another
This will potentially lead to a lot of unnecessary authentication checks and not being able to log into the device.

c-po:
It's not "dangerous" as it will only install multiple lines if the package is installed multiple times (development only, not during ISO build)

It's still an error that I will fix shortly!

PR #1763, opened 2023-01-18
remote: T4454: Add an option to check free space for install-image

erkin:
## Change Summary

Calls to `install-image` now check `/usr/lib/live/mount/persistence` for free storage.
Please let me know if this path can be changed in some way (and if so, how to query it) because it's hardcoded right now.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4454

## Component(s) name

op-mode, install-image, remote

## Proposed changes

Before attempting to download a file, the destination directory is checked for free storage, so that the download can be aborted in advance if there isn't enough space. For image installs, we need to check not only the destination of the .iso file (ie `/var/tmp`) but also the directory where the image is installed (ie `/usr/lib/live/mount/persistence`). This commit adds a special case to `check_storage()` and `friendly_download()` to be used by `install-image`.

## How to test

```
$ add system image ...
```

## Checklist:

### DIFF
--- a/python/vyos/remote.py
+++ b/python/vyos/remote.py
@@ -42,6 +42,8 @@ from vyos.version import get_version
 
 
 CHUNK_SIZE = 8192
+CHECK_SPACE_FOR_IMAGE = False
+IMAGE_DIRECTORY = '/usr/lib/live/mount/persistence'
 
 class InteractivePolicy(MissingHostKeyPolicy):
     """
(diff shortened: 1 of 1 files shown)


dmbaturin (changes requested):
I think we should move actual size check logic to a small function that only does that so that it's easy to call from non-interactive scripts, and use it in the larger, interactive function.

When we get to adding ability to install images via the API, we will need a size check there, and I believe the API version should always just refuse to proceed if there may not be enough space.

erkin:
In that case, the free storage check logic is to become a helper function in `vyos.util` and it should be called on these two paths during installation. But I'm not entirely certain on how to modify this and `install-image` to accommodate for both interactive and API-scripted calls.

erkin:
Perhaps it's the best if we deal with this matter by rewriting `install-image` as a fresh op-mode script in Python. That's what I'm going to do.

PR #1766, opened 2023-01-18
igmp-proxy: T4912: Rewrite show IGMP proxy commands in the new op-mode format

erkin:
## Change Summary

A straightforward rewrite, but the `/proc/net/ip_mr_{vif,mfc}` parsers were kept as they are. The `show_*` routines also support JSON output through calls with `--raw` (not accessible through op-mode).

**Note:** I couldn't test `show ip multicast mfc` much due to the absence of documentation but there shouldn't be any change in behaviour since the parser itself is untouched.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4912

## Component(s) name

op-mode, IGMP proxy, multicast

## Proposed changes

Other than the `--raw` output support, the behaviour is identical to the previous version besides cosmetic changes (eg no more redundant linebreaks in output). Also moved the process check into the script from the hackish one-liner in the template and pruned a superfluous helper routine in favour of the in-house one.

## How to test

```
$ show ip multicast interface
$ show ip multicast mfc
```

## Checklist:


src/op_mode/igmp-proxy.py
```diff
@@ -194,48 +146,73 @@ def do_mr_mfc():
...
+# tun1       0.0 B        0            0.0 B        0            xxx.xxx.xxx.2
+def show_interface(raw: bool):
+    vif_out_template = """{%- for r in data -%}
+{{ "%-10s"|format(r.interface) }} {{ "%-12s"|format(r.bytes_in) }} {{ "%-12s"|format(r.pkts_in) }} {{ "%-12s"|format(r.bytes_out) }} {{ "%-12s"|format(r.pkts_out) }} {{ "%-15s"|format(r.loc) }}
+{% endfor %}"""
+    if data := _process_mr_vif():
+        if raw:
+            return json.loads(json.dumps(data))
+        else:
+            return jinja2.Template(vif_out_template).render(data)
```
c-po: If I remember correctly the idea was to use `Tabulate` here over Jinja2 if possible to get nice looking tables.

erkin: I wasn't sure whether I should hack on that part but I guess it's a good opportunity.

erkin:
Now it uses `tabulate` and the JSON output for the API is much clearer. I went ahead and diked out the hairy parser for the `show ip multicast mfc` command as a scream test. If anyone actually demands it, I'll rewrite that part and document it this time.

dmbaturin (approved):

c-po (approved):

PR #1767, opened 2023-01-19
config-mgmt: T4942: rewrite vyatta-config-mgmt to Python/XML

jestabro:
## Change Summary

Note that this PR is dependent on the two PR's:
https://github.com/vyos/vyatta-cfg/pull/57
https://github.com/vyos/vyos-world/pull/7
and must be merged with the above.

This is a rewrite of the legacy package vyatta-config-mgmt to Python/XML.

Behavioral changes:
the legacy behavior of 'commit-confirm' adds a log entry and archives config files even when the commit is 'canceled', in the sense that 'confirm' is not issued and the system reboots to the saved config. That is no longer the case: the log entry is cached and only added if 'confirm' is issued, in which case the config file is archived.

Secondly, a warning is now given at a prompt if 'commit-confirm' is issued with unsaved commits, so one can save commits first before proceeding; these changes were silently discarded in the legacy package.

Finally, the Python library difflib's 'unified_diff' is used for diff output instead of the legacy vyatta-cfg hand-crafted diff formatting; I contend this is cleaner, and allows us to avoid further calls to cli-shell-api functions.

## Types of changes

- [X] Migration from an old Vyatta component to vyos-1x, please link to related PR inside obsoleted component

## Related Task(s)

* https://phabricator.vyos.net/T4942

## Checklist:

### DIFF
--- a/python/vyos/config_mgmt.py
+++ b/python/vyos/config_mgmt.py
@@ -1,0 +1,674 @@
+# Copyright 2023 VyOS maintainers and contributors <maintainers@vyos.io>
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import re
+import sys
+import gzip
+import logging
+from typing import Optional, Tuple, Union
+from filecmp import cmp
+from datetime import datetime
+from tabulate import tabulate
+
+from vyos.config import Config
+from vyos.configtree import ConfigTree
+from vyos.defaults import directories
+from vyos.util import is_systemd_service_active, ask_yes_no, rc_cmd
+
+SAVE_CONFIG = '/opt/vyatta/sbin/vyatta-save-config.pl'
+
+# created by vyatta-cfg-postinst
+commit_post_hook_dir = '/etc/commit/post-hooks.d'
+
+commit_hooks = {'commit_revision': '01vyos-commit-revision',
+                'commit_archive': '02vyos-commit-archive'}
+
+DEFAULT_TIME_MINUTES = 10
+timer_name = 'commit-confirm'
+
+config_file = os.path.join(directories['config'], 'config.boot')
+archive_dir = os.path.join(directories['config'], 'archive')
+archive_config_file = os.path.join(archive_dir, 'config.boot')
+commit_log_file = os.path.join(archive_dir, 'commits')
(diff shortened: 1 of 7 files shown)


c-po (approved):

python/vyos/config_mgmt.py
```diff
@@ -0,0 +1,664 @@
...
+from shutil import copy
+from filecmp import cmp
+from time import time
+from datetime import datetime
+from difflib import unified_diff
+from tabulate import tabulate
+
+from vyos.config import Config
+from vyos.configtree import ConfigTree
+from vyos.remote import upload
```
dmbaturin: Since it's only used in one place, maybe make it a local import there?

python/vyos/config_mgmt.py
```diff
@@ -0,0 +1,664 @@
...
+
+import os
+import re
+import sys
+import pwd
+import gzip
+import logging
+from typing import Optional, Tuple, Union
+from argparse import ArgumentParser, REMAINDER
+from shutil import copy
```
dmbaturin: I have a feeling that quite a few of these may be better off as local imports inside functions that use them, in general.

python/vyos/config_mgmt.py
```diff
@@ -0,0 +1,664 @@
...
+        """Upload config to remote archive.
+        """
+        hostname = self.hostname
+        t = datetime.now()
+        timestamp = t.strftime('%Y%m%d_%H%M%S')
+        remote_file = f'config.boot-{hostname}.{timestamp}'
+        source_address = self.source_address
+
+        for location in self.locations:
+            upload(archive_config_file, f'{location}/{remote_file}',
```
dmbaturin, line 350: One occasionally-requested feature is saving config revisions in a remote git repository. I think we may want to add an abstract "archive storage" interface right away and keep it to a single "URL upload" implementation for now, so that later these functions will be easy to adjust to instantiating a storage object:

Pseudocode:
```
if location:
    storage = vyos.commit_archive.URL(location)
    storage.save(remote_file, archive_config_file)
```

dmbaturin (changes requested):
I think a few things can be cleaned up and generalized, let me know what you think.

python/vyos/config_mgmt.py
jestabro: upload is now a local import

python/vyos/config_mgmt.py
jestabro: Several other imports made local, where appropriate; thanks !

jestabro:
Force-push to correct raw output for op-mode 'show_commit_diff' and 'show_commit_file' in light of:
https://github.com/vyos/vyos-1x/pull/1770

python/vyos/config_mgmt.py
jestabro: I think I will open this as a separate task, now: note that support for those who archive config files under revision control depends on the update to vyatta-save-config.pl:
https://github.com/jestabro/vyos-1x/blob/save-config/src/helpers/vyos-save-config.py
as mentioned for example here:
https://forum.vyos.io/t/config-written-with-quotes-after-upgrade-but-no-quotes-after-change/8161
One  blocker for the rewrite of vyatta-save-config.pl is precisely this PR: namely, backwards comparability of diffs with the commit archive will be preserved, as it is now mediated by configtree.

dmbaturin (approved):

PR #1801, opened 2023-02-05
smoketest: tftp: T4012: add busy waiting loop when validating service availability

c-po:
## Change Summary

TFTP daemon is started as "fire and forget" and systemctl can return (thus commit will return) but the daemon itself is not yet running.

This adds a loop checking if the service runs and will fail after 10 seconds.

This will make 1.3 ISO images build again on Jenking

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4012

## Component(s) name

TFTP smoketest

## Proposed changes

Add busy waiting loop

## How to test

```
vyos@vyos:~$ /usr/libexec/vyos/tests/smoke/cli/test_service_tftp-server.py
test_01_tftpd_single (__main__.TestServiceTFTPD) ... ok
test_02_tftpd_multi (__main__.TestServiceTFTPD) ... ok
test_03_tftpd_vrf (__main__.TestServiceTFTPD) ... ok

Ran 3 tests in 43.754s

OK
```

## Checklist:

### DIFF
--- a/smoketest/scripts/cli/test_service_tftp-server.py
+++ b/smoketest/scripts/cli/test_service_tftp-server.py
@@ -18,6 +18,7 @@ import unittest
 
 from psutil import process_iter
 from base_vyostest_shim import VyOSUnitTestSHIM
+from time import sleep
 
 from vyos.configsession import ConfigSessionError
 from vyos.util import cmd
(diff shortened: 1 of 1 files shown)


sever-sever (approved):

dmbaturin (approved):
Minor note: I think a loop that has a `sleep()` call in it is not a busy loop in the strict sense. A true busy loop is one that keeps doing its check on every CPU cycle it gets, and their only good use is on CPUs that don't have programmable interrupt timers. Apart from that, it's usually shows up in "absurdly tricky 'lock-free'  algorithm vs a busy loop" discussions that completely ignore that adding `sleep(1)` to the loop solves 99% of the problems for 0.1% effort. :)

To the point, I agree with the change.

c-po:
> Minor note: I think a loop that has a `sleep()` call in it is not a busy loop in the strict sense. A true busy loop is one that keeps doing its check on every CPU cycle it gets, and their only good use is on CPUs that don't have programmable interrupt timers. Apart from that, it's usually shows up in "absurdly tricky 'lock-free' algorithm vs a busy loop" discussions that completely ignore that adding `sleep(1)` to the loop solves 99% of the problems for 0.1% effort. :)
> 
> To the point, I agree with the change.

You are right it‘s not a real busy loop keeping the cpu busy all the time.

PR #1862, opened 2023-03-01
graphql: T5040: generate schema on installation, rather than dynamically

jestabro:
## Change Summary

GraphQL schema are generated dynamically, on configuration, as a convenience in order to distinguish schema for key authentication (non-nullable input field 'key') and JWT token (no such field).

It is preferable to generate schema statically, during installation of the debian package vyos-1x, as one can then apply code generation tools for client types and operations.

Consequently, use nullable field 'key' and generate schema in vyos-1x.postinst.

## Types of changes

        Reorganize schema generation.

## Related Task(s)

* https://vyos.dev/T5040

## Component(s) name

http-api

## Checklist:

### DIFF
--- a/src/services/api/graphql/generate/schema_from_composite.py
+++ b/src/services/api/graphql/generate/schema_from_composite.py
@@ -26,7 +26,7 @@ from jinja2 import Template
 
 from vyos.defaults import directories
 if __package__ is None or __package__ == '':
-    sys.path.append("/usr/libexec/vyos/services/api")
+    sys.path.append(os.path.join(directories['services'], 'api'))
     from graphql.libs.op_mode import snake_to_pascal_case, map_type_name
     from composite_function import queries, mutations
 else:
--- a/src/services/api/graphql/generate/schema_from_composite.py
+++ b/src/services/api/graphql/generate/schema_from_composite.py
@@ -29,9 +29,6 @@ if __package__ is None or __package__ == '':
     sys.path.append("/usr/libexec/vyos/services/api")
     from graphql.libs.op_mode import snake_to_pascal_case, map_type_name
     from composite_function import queries, mutations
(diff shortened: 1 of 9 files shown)


sever-sever (approved):

sarthurdev (approved):

src/services/api/graphql/generate/schema_from_composite.py
```diff
@@ -29,50 +29,23 @@
     sys.path.append("/usr/libexec/vyos/services/api")
```
dmbaturin: I think we should use global constants rather than duplicate hardcoded paths in files.

src/services/api/graphql/generate/schema_from_config_session.py
```diff
@@ -29,50 +29,23 @@
     sys.path.append("/usr/libexec/vyos/services/api")
```
dmbaturin: Another instance of a hardcoded path. I don't expect that we change it, but if switching distros will make us move it to elsewhere, storing all such constants in one place will make it a lot easier.

dmbaturin (changes requested):
My only very minor objection is against the use of hardcoded directory paths duplicated in multiple files. Everything else seems fine.

src/services/api/graphql/generate/schema_from_composite.py
jestabro: Quite right, thanks. Path from defaults.py now used in all instances.

src/services/api/graphql/generate/schema_from_config_session.py
jestabro: As above.

dmbaturin (approved):

c-po (approved):

PR #1904, opened 2023-03-22
T5081: ISIS and OSPF syncronization with IGP-LDP sync

Cheeze-It:
## Change Summary

This is the initial implementation for LDP-IGP synchronization that FRR has implemented for ISIS and OSPF.
This has the operational commands, the configuration commands, the smoketests implemented for the way
that this feature works within FRR.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5081

## Component(s) name

ldp, isis, ospf

## Proposed changes

The changes that are added are fairly straightforward. The operational commands were taken straight
from the way FRR handles the commands. This is to follow convention in said operational commands.
I formatted the configuration commands to follow the operational commands so that they are named
the same. This should make command association easier for configuration and checking. I added the 
smoketests for the configuration of these features as well with them being about as in depth as the
CLI allows. I will admit though, this is the initial implementation and there may be some handlers that
are missing. But I will investigate that in the future. They should be fairly easy to figure out.

## How to test

Here is a rudimentary ISIS configuration that turns on LDP-IGP sync, and changes the hold down for an interface:
```
vyos@vyos:~$ configure
[edit]
vyos@vyos# set protocols isis net 49.0001.1921.6800.1002.00
[edit]
vyos@vyos# set protocols isis ldp-sync
[edit]
vyos@vyos# set protocols isis interface eth1 ldp-sync holddown 100
[edit]
vyos@vyos# compare
[protocols]
+ isis {
+     interface eth1 {
+         ldp-sync {
+             holddown "100"
+         }
+     }
+     ldp-sync { }
+     net "49.0001.1921.6800.1002.00"
+ }

[edit]
vyos@vyos# commit
[edit]
vyos@vyos# exit
Warning: configuration changes have not been saved.
exit
vyos@vyos:~$ vtysh -c "show run"
Building configuration...

Current configuration:
!
frr version 8.4.2
frr defaults traditional
hostname vyos
log syslog
log facility local7
service integrated-vtysh-config
!
ip route 0.0.0.0/0 10.0.0.65
!
interface eth0
 bandwidth 10000
exit
!
interface eth1
 bandwidth 10000
 ip router isis VyOS
 ipv6 router isis VyOS
 isis mpls ldp-sync holddown 100  <--- the interface command here
exit
!
interface lo
exit
!
router isis VyOS
 net 49.0001.1921.6800.1002.00
 mpls ldp-sync    <---- the main ISIS command here
exit
!
rpki
exit
!
end
```

Here it is in OSPF:
```
vyos@vyos:~$ configure
[edit]
vyos@vyos# set protocols ospf ldp-sync
[edit]
vyos@vyos# set protocols ospf interface eth1 ldp-sync holddown 100
[edit]
vyos@vyos# commit
[edit]
vyos@vyos# exit

vyos@vyos:~$ vtysh -c "show run"
Building configuration...

Current configuration:
!
frr version 8.4.2
frr defaults traditional
hostname vyos
log syslog
log facility local7
service integrated-vtysh-config
!
ip route 0.0.0.0/0 10.0.0.65
!
interface eth1
 ip ospf dead-interval 40
 ip ospf mpls ldp-sync       <--- This gets added to enable, but it's odd as it doesn't go away afterwards. Might be a bug.
 ip ospf mpls ldp-sync holddown 100  <--- This properly gets added and removed. 
exit
!
router ospf
 auto-cost reference-bandwidth 100
 timers throttle spf 200 1000 10000
 redistribute babel      <--- I noticed this being enabled for some reason. Might be an FRR jinja2 template problem. But I didn't make any changes to enable this.
 mpls ldp-sync            <--- This also shows that it's enabled for the actual protocol. 
exit
!
rpki
exit
!
end
```

## Checklist:

I have not made a change to the documentation yet. If the PR is approved, then I will add do the documentation.


interface-definitions/include/ldp-sync.xml.i
```diff
@@ -0,0 +1,27 @@
+<!-- include start from ldp-igp-sync.xml.i -->
+<node name="ldp-sync">
+  <properties>
+    <help>LDP-IGP synchronization configuration for interface</help>
+  </properties>
+  <children>
+    <leafNode name="disable">
```
c-po: Please use `#include <include/generic-disable-node.xml.i>`

interface-definitions/include/ldp-sync.xml.i
```diff
@@ -0,0 +1,27 @@
...
+      <properties>
+        <help>Disable LDP-IGP synchronization for interface</help>
+        <valueless/>
+      </properties>
+    </leafNode>
+    <leafNode name="holddown">
```
c-po: The `holddown` CLI node is defined multiple times - please move it to an include file and reference it.

op-mode-definitions/include/isis-common.xml.i
```diff
@@ -54,6 +54,30 @@
...
+  <children>
+    <node name="ldp-sync">
+      <properties>
+        <help>Show IS-IS LDP-IGP synchronization information</help>
+      </properties>
+      <command>${vyos_op_scripts_dir}/vtysh_wrapper.sh $@</command>
+      <children>
+        <tagNode name="interface">
+          <properties>
+            <help>Show specific IS-IS LDP-IGP synchronization for an interface</help>
```
c-po: Can we probably remove IS-IS and OSPF from the help string so we can make this an includable XML snipped to become more and more single-source?

c-po (changes requested):

interface-definitions/include/ldp-sync.xml.i
Cheeze-It: Sure ok, will change it over to that.

op-mode-definitions/include/isis-common.xml.i
Cheeze-It: Sure of course. Will make those changes.

interface-definitions/include/ldp-sync.xml.i
Cheeze-It: So this is something that I believe we *can* do but in FRR it's kind of split.

So the reason I did it the way I did it was because the way LDP-IGP sync is done is it's done in two places. Under the routing protocol (for global enablement) and then under the interface. By default from what I read/understood is that we are to enable/disable it under the protocol and it will propagate it down to the interfaces. Now the interfaces in *theory* have 2 (and it seems 3, but this is where it seems to be buggy) options. One option is to disable LDP-IGP sync for that interface (the disable) and one is to change the `holddown` value for just that interface. I figured to make a template of the interfaces and include that, but to make the protocol portion be independent from the interfaces.

If you're thinking we can consolidate them then can I do a protocol LDP-IGP sync include file, and an interfaces include file? That way if there's changes in the global it doesn't affect the interfaces. Vice versa with the interfaces affecting the global.

Cheeze-It:
I was able to do some changes and I validated that they work. Which is good :) 

Let me know if things look ok your end my good sirs.

interface-definitions/include/ldp-sync-protocol.xml.i
```diff
@@ -0,0 +1,21 @@
+<!-- include start from ldp-igp-sync.xml.i -->
+<node name="ldp-sync">
+  <properties>
+    <help>Protocol wide LDP-IGP synchronization configuration</help>
+  </properties>
+  <children>
+    <leafNode name="holddown">
```
c-po, line 7: Please make `holddown` and includable building block. The only difference I see is the help string. I think one can find a common help that fits both needs.

c-po (approved):

PR #1935, opened 2023-04-03
dns: T5143: Apply constraint for domain name in DNS forwarding

indrajitr:
## Change Summary

Apply contraint for domain name in DNS forwarding, so that arbitrary strings aren't allowed. Also, move the common XML fragment to a new include file.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5143

## Component(s) name

dns forwarding

## How to test

Something like this should not be allowed.
```
set service dns forwarding domain 👋 name-server 1.2.3.4
```

## Checklist:

### DIFF
--- a/interface-definitions/dns-forwarding.xml.in
+++ b/interface-definitions/dns-forwarding.xml.in
@@ -83,6 +83,13 @@
               <tagNode name="domain">
                 <properties>
                   <help>Domain to forward to a custom DNS server</help>
+                  <valueHelp>
+                    <format>text</format>
+                    <description>An absolute DNS domain name</description>
+                  </valueHelp>
+                  <constraint>
+                    <validator name="fqdn"/>
+                  </constraint>
                 </properties>
                 <children>
                   #include <include/name-server-ipv4-ipv6-port.xml.i>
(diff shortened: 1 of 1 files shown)


c-po:
That is a good idea, but we have never just had the body/multiple children of XML nodes as an include.
So far we either had a full XML definition or constraints only (which lived in the `interface-definitions/include/constraint` folder.

In my opinion when we now start only inlcuding fragments of XML nodes we might get lost as it will be simply to many. I usually tried to migrate CLI nodes to make use of a common include instead where applicable. Nevertheless dns domain name were identified in https://vyos.dev/T5109 to get their discrete OCaml validator. So in the meantime we can still make half use of this PR by creating only the `interface-definitions/include/constraint/dns-domain-name.xml.i` validator for common reference.

@jestabro @dmbaturin what you think?

indrajitr:
> In my opinion when we now start only inlcuding fragments of XML nodes we might get lost as it will be simply to many.

Indeed, we don't want to end up overdoing these fragments!

> Nevertheless dns domain name were identified in https://vyos.dev/T5109 to get their discrete OCaml validator.

Great! As I stated in T5143, our current validator isn't RFC5321 compliant anyway. OCaml will handle the compliance better (and faster) than regex.

>  only the `interface-definitions/include/constraint/dns-domain-name.xml.i` validator for common reference

~Should we do `constraint/dns-domain-name.xml.i` or should we just stick to `constraint/domain-name.xml.i` to stay consistent with the other constraints in there (like `constraint/host-name.xml.i`, `constraint/interface-name.xml.i` etc.)?~

Edit: we can reuse `fqdn` validator in all cases instead.

@c-po, please let me know if this makes sense. Else, will add back the regex constraint.

c-po (approved):

PR #1955, opened 2023-04-13
updating network name due to change to aardvark networking

trilom:
## Change Summary

When following the [documentation for container's example configuration](https://docs.vyos.io/en/latest/configuration/container/index.html#example-configuration) it is clear that something isn't working with the network name 'zabbix-net'.

After reducing the **network_interface** of the network in the file located at _/etc/containers/networks/zabbix-net.json_ to a length of 15 or less I am able to commit my configuration.  With a network name greater than 15 I receive a failure in systemd.

>Error: netavark: get bridge interface: Netlink error: Numerical result out of range (os error 34)

## Types of changes

## Related Task(s)

* https://vyos.dev/Txxxx

## Proposed changes

Either use the change as suggested in the commit to further limit network name from 11 to 9.

Or rewrite the python code that generates the template to a different **network_interface** scheme.  It used to be "cni" which if we continued this it is a misnomer and could generate misleading information (interface called cni using ardvark).  I am for leaving the interfaces named "podman-" as a transition from "cni-" but it restricts the user names from 11 to 9.

### some bad config
```bash
set container network zabbix-net prefix 172.20.0.0/16
```
creates a file that looks similar to this, where **len(network_interface) == 17**
```json
{
  "name": "zabbix_net",
  "id": "f39303dff1194a6c063abcb21c5a6ba858268dbcdcb5ce1812274efac54820bc",
  "driver": "bridge",
  "network_interface": "podman-zabbix_net",
  "subnets": [
    {
      "subnet": "172.20.0.0/16",
      "gateway": "172.20.0.1"
    }
  ],
  "ipv6_enabled": false,
  "internal": false,
  "dns_enabled": false,
  "ipam_options": {
    "driver": "host-local"
  }
}
```
### some good config
```bash
set container network zabbix-n prefix 172.20.0.0/16
```
creates a file that looks similar to this, where **len(network_interface) == 15**
```json
{
  "name": "zabbix_net",
  "id": "f39303dff1194a6c063abcb21c5a6ba858268dbcdcb5ce1812274efac54820bc",
  "driver": "bridge",
  "network_interface": "podman-zabbix_n",
  "subnets": [
    {
      "subnet": "172.20.0.0/16",
      "gateway": "172.20.0.1"
    }
  ],
  "ipv6_enabled": false,
  "internal": false,
  "dns_enabled": false,
  "ipam_options": {
    "driver": "host-local"
  }
}
```

## Checklist:

  - **I am waiting on an account to submit a bug report** 

### DIFF
--- a/interface-definitions/container.xml.in
+++ b/interface-definitions/container.xml.in
@@ -343,9 +343,9 @@
         <properties>
           <help>Network name</help>
           <constraint>
-            <regex>[-_a-zA-Z0-9]{1,11}</regex>
+            <regex>[-_a-zA-Z0-9]{1,9}</regex>
           </constraint>
-          <constraintErrorMessage>Network name cannot be longer than 11 characters</constraintErrorMessage>
+          <constraintErrorMessage>Network name cannot be longer than 9 characters</constraintErrorMessage>
         </properties>
         <children>
           #include <include/generic-description.xml.i>


sarthurdev:
When you get a task ID, please update the PR and commit titles to the format in [CONTRIBUTING](https://github.com/vyos/vyos-1x/blob/current/CONTRIBUTING.md).

Also if this line is the problem, would it not be better to instead shorten the prefix so we don't need to further restrict the name length?
https://github.com/vyos/vyos-1x/blob/current/src/conf_mode/container.py#L379 `podman-` to `pod-` for example?

trilom:
I am in favor of shortening to `pod-`, i'm not married to either approach.
Personally it's better to change it to `pod-` in order to not make the name length be a consistently moving target since `pod-` and `cni-` are equal it facilitates easily swapping back in the CNI if desired and maintain consistency in `network name` length.

trilom:
Another thing I noticed that might be a side effect of switching from `cni` recently in relation to the example provided it seems that resolving container hostnames isn't functional.

The config file explicitly sets DNS to false for the networks created.  If you manually enable it and attempt to run the container (with `dns_enabled set to True`) it says:
>WARN[0000] aardvark-dns binary not found, container dns will not be enabled

Within the container, for example "mysql-server" the `/etc/hosts` file is similar to the host's with the addition of a host entry pointing the bridge address to "mysql-server" but no other way to resolve the other zabbix server's by name.

c-po:
Closing this as of the following changes:
* [container: T5082: enable aardvark-dns support](https://github.com/vyos/vyos-1x/commit/42775f287)
* [container: T5082: shorten container network prefix to allow longer names](https://github.com/vyos/vyos-1x/commit/2a8760598)

PR #1979, opened 2023-05-04
cloud-init: T5190: Added Cloud-init pre-configurator

zdc:
## Change Summary

Added Cloud-init pre-configurator

## Types of changes

## Related Task(s)

* https://vyos.dev/T5190

## Component(s) name

cloud-init

## Proposed changes

Added a new service that starts before Cloud-init, waits for all network interfaces initialization, and if requested by config, checks which interfaces can get configuration via DHCP server and creates a corresponding Cloud-init network configuration.

This protects from two situations:

* when Cloud-init tries to get meta-data via eth0 (default and fallback variant for any data source which depends on network), but the real network is connected to another interface
* when Cloud-init starts simultaneously with udev and initializes the first interface to get meta-data before it is renamed to eth0 by udev

## How to test

To test this, you need to deploy VyOS in an environment with Cloud-init, where meta-data should be fetched from the network and connect the VM to the network via any interface, except `eth0`. A good test environment is Equinix Metal, where usually `eth2` is the main interface.

Without this change, the system will fail to get any meta-data and possibly will have an `e0` interface except `eth0`.

## Checklist:

### DIFF
--- a/debian/vyos-1x.postinst
+++ b/debian/vyos-1x.postinst
@@ -42,3 +42,6 @@ for file in $DELETE; do
         rm -f ${file}
     fi
 done
+
+# Enable Cloud-init pre-configuration service
+systemctl enable vyos-config-cloud-init.service
--- a/data/templates/system/cloud_init_networking.tmpl
+++ b/data/templates/system/cloud_init_networking.tmpl
@@ -1,0 +1,9 @@
+network:
+  version: 2
+  ethernets:
+{% for iface in ifaces_list %}
+    {{ iface['name'] }}:
+      dhcp4: true
+      match:
+        macaddress: "{{ iface['mac'] }}"
+{% endfor %}
--- a/src/system/vyos-config-cloud-init.py
+++ b/src/system/vyos-config-cloud-init.py
@@ -1,0 +1,170 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) 2023 VyOS maintainers and contributors
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or later as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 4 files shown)


sever-sever (approved):

debian/vyos-1x.postinst
```diff
@@ -42,3 +42,6 @@ for file in $DELETE; do
         rm -f ${file}
     fi
 done
+
+# Enable Cloud-init pre-configuration service
```
c-po, line 46: Is it a good idea to enable cloud-init service for every system by default?

dmbaturin (approved):
If cloud-init by itself will not allow a rogue DHCP server trick VyOS into loading a maliciously crafted config, I suppose it's safe enough.

c-po:
> If cloud-init by itself will not allow a rogue DHCP server trick VyOS into loading a maliciously crafted config, I suppose it's safe enough.

How can I disable cloudinit from the VyOS CLI once my machine is fully deployed/configured? Or does it simply run in the background and eat RAM?

zdc:
> > If cloud-init by itself will not allow a rogue DHCP server trick VyOS into loading a maliciously crafted config, I suppose it's safe enough.
> 
> How can I disable cloudinit from the VyOS CLI once my machine is fully deployed/configured? Or does it simply run in the background and eat RAM?

There are no extra security exposures provided by this PR. It adds only extra service which can be triggered by Cloud-init as a dependency and help it to use proper interfaces for retrieving meta-data and protect the system from broken network interfaces, which is the case now and happens when Cloud-init initializes eth0 before udev finish renaming.

On system resources and service state, everything is more complex and the behavior depends on multiple conditions. Highlights:

1. The main Cloud-init service status (enabled/disabled) is controlled by the systemd generator, which runs every boot and checks if Cloud-init needs to be used.
2. If Cloud-init is required by a specific platform, it normally should be run on each system boot to allow using modules that must be started per boot.
3. Cloud-init does not have any persistent processes by itself, it only starts services requested by modules and provided config. In our case, we have no such processes, therefore after Cloud-init service stops its work, no extra processes keep running in the system.

c-po (approved):

PR #2031, opened 2023-06-07
op-mode: T5262: move 'unsaved_commits' to module scope and correct false positive

jestabro:
## Change Summary

This provides an addendum to PR:
https://github.com/vyos/vyos-1x/pull/2028

The function 'unsaved_commits' was added in config_mgmt to warn a user of unsaved commits before commit-confirm, as that entails a possible reboot. As it has other uses and no dependence on the object itself, move to module scope. For general use, add simple check for live image to avoid false positive, due to config migration reformatting.

## Types of changes

- [X] Refactoring (no functional changes)

## Related Task(s)

* https://vyos.dev/T5262

## Checklist:

### DIFF
--- a/python/vyos/config_mgmt.py
+++ b/python/vyos/config_mgmt.py
@@ -26,6 +26,7 @@ from tabulate import tabulate
 from vyos.config import Config
 from vyos.configtree import ConfigTree, ConfigTreeError, show_diff
 from vyos.defaults import directories
+from vyos.version import get_full_version_data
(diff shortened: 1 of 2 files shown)


sever-sever (approved):

sever-sever (approved):

dmbaturin (approved):
My only very minor objection is that `unsaved_commits()` sounds as if it would return a list of unsaved commits... of course, the way it's used here would still work if it returned a list of unsaved commits rather than a boolean value.

PR #2060, opened 2023-06-28
T5320: warn on entering config mode if boot config errors present

jestabro:
## Change Summary

In the unlikely case of a boot config error of a verified config (for example, T5300), the user may miss the error message on boot, and then go on to config/commit/save, overwriting the failed config section. Add a warning on entering config mode in case of boot config error.

Also, a boot config error, if not overwritten, will present a 'unsaved commit' warning on reboot: this is misleading (cf. comment in T5262), so modify the check (since the reboot will restore/re-error the section, as in the case above for an intermittent error due to race condition).

## Types of changes

        Warn user in config mode if necessary; modify warning on reboot.

## Related Task(s)

* https://vyos.dev/T5320

## Checklist:

### DIFF
--- a/op-mode-definitions/configure.xml.in
+++ b/op-mode-definitions/configure.xml.in
@@ -11,7 +11,12 @@
         echo "Please do it as an administrator level VyOS user instead."
     else
         if grep -q -e '^overlay.*/filesystem.squashfs' /proc/mounts; then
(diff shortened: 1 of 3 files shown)


c-po (approved):

dmbaturin (approved):
I love the idea!

PR #2075, opened 2023-07-05
T5336: Add Swedish keyboard-layout

Apachez-:
## Types of changes

## Related Task(s)

* https://vyos.dev/T5336

## Component(s) name

set system option keyboard-layout

## Proposed changes

Adding swedish keymap as system option keyboard-layout

## How to test

Switched between keymaps using "set system option keyboard-layout" and verified through console that se-latin1 is fully functional using a swedish keyboard.

Note that this is only valid when connected through local console. Connecting through SSH is usually a workaround for this issue.

## Checklist:

### DIFF
--- a/interface-definitions/system-option.xml.in
+++ b/interface-definitions/system-option.xml.in
@@ -36,7 +36,7 @@
              <properties>
                <help>System keyboard layout, type ISO2</help>
                <completionHelp>
-                 <list>us uk fr de es fi jp106 no dk dvorak</list>
+                 <list>us uk fr de es fi jp106 no dk se-latin1 dvorak</list>
                </completionHelp>
                <valueHelp>
                  <format>us</format>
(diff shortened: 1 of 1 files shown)


dmbaturin (approved):
This works, but I wonder: can we list all layouts available in the system programmatically somehow? Are there reasons not to expose them all, or it hard to do?

sever-sever (approved):

c-po (approved):

c-po:
> This works, but I wonder: can we list all layouts available in the system programmatically somehow? Are there reasons not to expose them all, or it hard to do?

I think that is a much better approach then we currently have

Apachez-:
> This works, but I wonder: can we list all layouts available in the system programmatically somehow? Are there reasons not to expose them all, or it hard to do?

That would be prefered but there are ALOT of them in /usr/share/keymaps/i386/ so tab completing all of them in configuration mode would be hilarious.

One way might be to add the most popular languages along with a final option named "custom" or similar where one then can define whatever one like from the /usr/share/keymaps/i386/ directories.

dmbaturin:
I think a custom layout option is a good idea.

PR #2116, opened 2023-07-27
xml: T5403: add support for supplemental xml cache

jestabro:
## Change Summary

Add support for adding supplemental entries to the vyos.xml_ref cache.py outside of the native vyos-1x build.

Addon packages to supplement CLI interface definitions should call `generate_cache` during build, and `update_cache` in `some_package.postinst`, for example:

(In debian/rules, after transcluding interface definitions:)

`./vyos-1x-current/python/vyos/xml_ref/generate_cache.py --xml-dir $(TMP_DIR)/${CLI_SRC_CONF} --package-name some_package --output-path ${TMP_DIR}
`

(and then:)

`cp $(TMP_DIR)/some_package_cache.py $(DESTDIR)/${PYTHON_CACHE_DIR}`

(In debian/some_package.postinst:)

`python3 /usr/lib/python3/dist-packages/vyos/xml_ref/update_cache.py`

----

In the above,` ${PYTHON_CACHE_DIR}` is `/usr/lib/python3/dist-packages/vyos/xml_ref/pkg_cache`

Similarly, in `some_package.postrm`, `update_cache` should be called after removing the package cache from  ${PYTHON_CACHE_DIR}, thus the need to maintain package specific dict caches in the subdirectory, and update the central cache.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5403

## Checklist:

### DIFF
--- a/debian/vyos-1x.postinst
+++ b/debian/vyos-1x.postinst
@@ -180,6 +180,9 @@ systemctl enable vyos-config-cloud-init.service
 # Generate API GraphQL schema
 /usr/libexec/vyos/services/api/graphql/generate/generate_schema.py
 
+# Update XML cache
+python3 /usr/lib/python3/dist-packages/vyos/xml_ref/update_cache.py
+
 # T1797: disable VPP support for rolling release, should be used by developers
 # only (in the initial phase). If you wan't to enable VPP use the below command
 # on your VyOS installation:
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,4 @@ debian/*.substvars
 data/component-versions.json
 # vyos-1x XML cache
 python/vyos/xml_ref/cache.py
+python/vyos/xml_ref/pkg_cache/*_cache.py
--- a/python/vyos/xml_ref/generate_cache.py
+++ b/python/vyos/xml_ref/generate_cache.py
@@ -18,10 +18,14 @@
 
 import sys
 import json
-import argparse
+from argparse import ArgumentParser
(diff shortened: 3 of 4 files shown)


dmbaturin (approved):

sever-sever (approved):

debian/vyos-1x.postinst
```diff
@@ -180,6 +180,9 @@ systemctl enable vyos-config-cloud-init.service
 # Generate API GraphQL schema
 /usr/libexec/vyos/services/api/graphql/generate/generate_schema.py
 
+# Update XML cache
+python3 /usr/lib/python3/dist-packages/vyos/xml_ref/update_cache.py
```
c-po, line 184: As there is a #shebang inside `update_cache.py` why call it with the python3 interpreter?

c-po (approved):

debian/vyos-1x.postinst
jestabro: Simply a permissions issue: otherwise one needs to set +x, as python3 setup install unsets executable permissions, and in *.postinst it is needed (though not in Makefile); good point the shebang is not then needed in update_cache.py.

jestabro: ^unsets^ignores ... note that `generate_schema.py` is not installed via setuptools, so the executable permissions are retained. I'd prefer to leave it for now, as there may be further reorganization depending on the needs of the vpp package.

c-po: It was just a note - please continue.

PR #2119, opened 2023-07-28
T5014: nat: add source and destination nat options for configuring lo…

nicolas-fort:
…ad balance within a single rule.

## Change Summary

Add balancing options for source and destination NAT.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5014

## Component(s) name

nat

## Proposed changes

Add options for applyin nat load balancing, with multiple options, all configre withing a single rule,.
## How to test

```
set nat destination rule 10 balance hash 'source-address'
set nat destination rule 10 balance hash 'source-port'
set nat destination rule 10 balance member 198.51.100.11 weight '25'
set nat destination rule 10 balance member 198.51.100.21 weight '25'
set nat destination rule 10 balance member 198.51.100.31 weight '50'
set nat destination rule 10 description 'DNAT balance HTTPS'
set nat destination rule 10 destination port '443'
set nat destination rule 10 inbound-interface 'eth0'
set nat destination rule 10 protocol 'tcp'
set nat destination rule 20 balance hash 'random'
set nat destination rule 20 balance member 8.8.4.4 weight '44'
set nat destination rule 20 balance member 8.8.8.8 weight '56'
set nat destination rule 20 description 'DNAT balance DNS'
set nat destination rule 20 destination port '53'
set nat destination rule 20 inbound-interface 'eth0'
set nat destination rule 20 protocol 'udp'
set nat destination rule 999 balance hash 'source-port'
set nat destination rule 999 balance member 1.1.1.1 weight '50'
set nat destination rule 999 inbound-interface 'pppoe0'
set nat destination rule 999 protocol 'udp'
set nat source rule 10 balance hash 'source-address'
set nat source rule 10 balance hash 'destination-address'
set nat source rule 10 balance member 203.0.113.59 weight '50'
set nat source rule 10 balance member 203.0.113.224 weight '50'
set nat source rule 10 description 'Snat - Balance'
set nat source rule 10 outbound-interface 'eth1'
vyos@HASH# sudo nft list table ip vyos_nat
table ip vyos_nat {
        chain PREROUTING {
                type nat hook prerouting priority dstnat; policy accept;
                counter packets 152 bytes 42237 jump VYOS_PRE_DNAT_HOOK
                iifname "eth0" tcp dport 443 counter packets 0 bytes 0 dnat to jhash ip saddr . tcp sport mod 100 map { 0-24 : 198.51.100.11, 25-49 : 198.51.100.21, 50-99 : 198.51.100.31 } comment "DST-NAT-10"
                iifname "eth0" udp dport 53 counter packets 0 bytes 0 dnat to numgen random mod 100 map { 0-43 : 8.8.4.4, 44-99 : 8.8.8.8 } comment "DST-NAT-20"
                iifname "pppoe0" meta l4proto udp counter packets 0 bytes 0 dnat to jhash udp sport mod 100 map { 0-49 : 1.1.1.1 } comment "DST-NAT-999"
        }

        chain POSTROUTING {
                type nat hook postrouting priority srcnat; policy accept;
                counter packets 46 bytes 3496 jump VYOS_PRE_SNAT_HOOK
                oifname "eth1" counter packets 0 bytes 0 snat to jhash ip saddr . ip daddr mod 100 map { 0-49 : 203.0.113.59, 50-99 : 203.0.113.224 } comment "SRC-NAT-10"
        }

        chain VYOS_PRE_DNAT_HOOK {
                return
        }

        chain VYOS_PRE_SNAT_HOOK {
                return
        }
}
[edit]
vyos@HASH# 

```

## Checklist:


sarthurdev (approved):

sever-sever (approved):

interface-definitions/include/firewall/firewall-hashing-parameters.xml.i
```diff
@@ -0,0 +1,35 @@
+<!-- include start from firewall/firewall-hashing-parameters.xml.i -->
+<leafNode name="hash">
+  <properties>
+    <help>Define the parameters of the packet header to apply the hashing, or choose random</help>
```
c-po: Please remove `, or choose random` as this node has a `defaultValue`. The following string will be automatically appended when generating CLI nodes: `(default: random)`.

interface-definitions/include/firewall/firewall-hashing-parameters.xml.i
```diff
@@ -0,0 +1,35 @@
...
+      <description>Use destination port for hashing</description>
+    </valueHelp>
+    <valueHelp>
+      <format>random</format>
+      <description>Do not use information from ip header. Use random value.</description>
+    </valueHelp>
+    <constraint>
+      <regex>(source-address|destination-address|source-port|destination-port|random)</regex>
+    </constraint>
+    <multi/>
```
c-po, line 31: Does it make sense to have a hash value based on multiple configurable sources?

interface-definitions/include/firewall/nat-balance.xml.i
```diff
@@ -0,0 +1,28 @@
+<!-- include start from firewall/nat-balance.xml.i -->
+<tagNode name="member">
```
c-po: What does `member` stand for as the help states "translated" - maybe there is a better node name available?

src/conf_mode/nat.py
```diff
@@ -223,7 +235,7 @@ def verify(nat):
                 Warning(f'rule "{rule}" interface "{config["inbound_interface"]}" does not exist on this system')
 
             if not dict_search('translation.address', config) and not dict_search('translation.port', config) and not dict_search('translation.redirect.port', config):
-                if 'exclude' not in config:
+                if 'exclude' not in config  and 'member' not in config['balance']:
```
c-po: Please remove the second whitespace here.

src/conf_mode/nat.py
```diff
@@ -199,7 +211,7 @@ def verify(nat):
                 Warning(f'rule "{rule}" interface "{config["outbound_interface"]}" does not exist on this system')
 
             if not dict_search('translation.address', config) and not dict_search('translation.port', config):
-                if 'exclude' not in config:
+                if 'exclude' not in config and 'member' not in config['balance']:
```
c-po: What happens if `balance` is not specified? A KeyError will be reported.

Maybe use: `if 'exclude' not in config and dict_search('balance.member', config) == None:`

c-po (changes requested):

interface-definitions/include/firewall/firewall-hashing-parameters.xml.i
nicolas-fort: From my point of view, it does make sense.
- More flexibility
- If more parameters are used, it easier to get different values of hashing for every different connection. In consequence, better load balance distribution.

For example if hash is only calculated using source IP address (think on a LAN, applying source nat to reach internet through 2 uplinks), then all connection form host A will always use the same uplink... If more parameters are included in the hashing, then connections from same host will be balanced.

src/conf_mode/nat.py
nicolas-fort: `balance` is always defined, since in `hash` section we have a default value. So if firewall is defined, we always get the default value `balance hash random` (unless something different is defined).. I faced with that issue while preparing this PR

interface-definitions/include/firewall/nat-balance.xml.i
nicolas-fort: We had internal talks and cli proposals with @zdc and @sever-sever, and so far this was the best approach. Let me know if you have a better option

c-po: I propose a slightly different CLI. Why not have something like:

```
set nat destination rule 10 hash 'source-address'
set nat destination rule 10 hash 'source-port'
set nat destination rule 10 load-balance 198.51.100.11 weight '25'
set nat destination rule 10 load-balance 198.51.100.21 weight '25'
set nat destination rule 10 load-balance 198.51.100.31 weight '50'
```

I had problems to understand what balance is used for. Checking nftables revealed it's actually load-balancing, so we should name it accordingly. Also `hash` is per nftables entry/rule so move it up one level.

If you think the backend server and the has should be under one parent node this would work, too:

```
set nat destination rule 10 load-balance hash 'source-address'
set nat destination rule 10 load-balance hash 'source-port'
set nat destination rule 10 load-balance backend 198.51.100.11 weight '25'
set nat destination rule 10 load-balance backend 198.51.100.21 weight '25'
set nat destination rule 10 load-balance backend 198.51.100.31 weight '50'
```

as usually load-balancing implies a frontend and a backend service - never heard of member so far.

nicolas-fort: Ok, I'll change and go for second option:
```
set nat destination rule 10 load-balance hash 'source-address'
set nat destination rule 10 load-balance hash 'source-port'
set nat destination rule 10 load-balance backend 198.51.100.11 weight '25'
set nat destination rule 10 load-balance backend 198.51.100.21 weight '25'
set nat destination rule 10 load-balance backend 198.51.100.31 weight '50'
```

c-po (approved):

PR #2130, opened 2023-08-02
wireguard: T5409: Added 'set interfaces wireguard wgX threaded'

aapostoliuk:
## Change Summary

Added 'set interfaces wireguard wgX threaded' command. 
Process traffic from each peer in a dedicated thread.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5409

## Component(s) name

wireguard
## Proposed changes

Added 'set interfaces wireguard wgX threaded' command. 
Process traffic from each peer in a dedicated thread.
This command is allowed only if one enabled peer exists

## Checklist:

### DIFF
--- a/interface-definitions/include/interface/per-client-thread.xml.i
+++ b/interface-definitions/include/interface/per-client-thread.xml.i
@@ -1,0 +1,8 @@
+<!-- include start from interface/per-client-thread.xml.i -->
+<leafNode name="per-client-thread">
+  <properties>
+    <help>Process traffic from each client in a dedicated thread</help>
+    <valueless/>
+  </properties>
+</leafNode>
+<!-- include end -->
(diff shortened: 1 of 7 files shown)


c-po:
Any reason to not enable this by default?

zdc:
> Any reason to not enable this by default?

Zero, if not negative, benefit for p2p interfaces. Also, bigger overall CPU usage (but distributed between cores).

sever-sever (approved):

c-po (changes requested):
IMHO this should be enabled by default as it should improve overall system responsiveness if load is distributed among multiple cores.

dmbaturin (approved):
I think if WireGuard has it off by default, we should probably follow the same approach.

c-po:
> I think if WireGuard has it off by default, we should probably follow the same approach.

If it will be a config option, can we then try to streamline it, as it‘s kind of an offload option? As we use offloading for other interface types ethernet, openvpn, too?

`set interfaces wireguard wgN offload multicore`

c-po:
As discussed in todays maintainer meeting we agreed on:
`set interfaces wireguard wgN per-client-thread`

And also implement this for the WIFI subsystem

c-po (approved):

PR #2157, opened 2023-08-18
T5494: add SSSD IPA and Kerberos support

h-eberhardt:
## Types of changes

## Related Task(s)

* https://vyos.dev/T5494

## Component(s) name

sssd, krb5

## Proposed changes

I have prepared an initial implementation of FreeIPA and Kerberos support. It allows to define a global Kerberos configuration (Keytabs for host and service principals) and to use one or multiple FreeIPA domains through SSSD. The current implementation does not add SSSD to the common PAM configuration, as this would probably allow all directory users administrative access to the VyOS system.
The feature is intended to be used for Single-Sign-On functions of  services, that are capable of performing PAM authentication combined with Kerberos GSSAPI, e.g. ocserv. Another service that could be implemented is the certmonger daemon, to have CSRs signed by a FreeIPA PKI.

## How to test

With an already working FreeIPA setup, the configuration is fairly straightforward. First you need to generate a keytab for the host itself. This needs to be done on a machine joined to a FreeIPA domain or a FreeIPA server itself.
To add the VyOS host to the domain run the following commands on the domain-joined machine
```
# add a host entry
ipa host-add --ip-address=xxx.xxx.xxx.xxx vyos.idm.example.org

# allow VyOS host to be managed by the domain-joined machine
ipa host-add-managedby --hosts=ipa-server.idm.example.org vyos.idm.example.org

# get the host principal keytab
ipa-getkeytab -s ipa-server.idm.example.org -p host/vyos.idm.example.org -o /tmb/krb5.keytab

# BASE64 encode the keytab
base64 /tmp/krb5.keytab
```
The CA certificate can be fetched with the follwing command
```
curl http://ipa-server.idm.example.org/ipa/config/ca.crt
```
The follwing commands set-up Kerberos and SSSD for the domain.
```
# IP address of a FreeIPA server
set system name-server yyy.yyy.yyy.yyy
set pki ca ipa.example.org certificate 'BASE64 Encoded RootCA Public Key'
set system host-name 'vyos.ipa.example.org'
set system login kerberos default-keytab 'host/vyos.ipa.example.org'
set system login kerberos default-realm 'IPA.EXAMPLE.ORG'
set system login kerberos dns-lookup-kdc
set system login kerberos dns-lookup-realm
set system login kerberos forwardable
set system login kerberos keytab principal host/vyos.ipa.example.org@IPA.EXAMPLE.ORG krb-keytab 'BASE64 Encoded Keytab'
set system login kerberos rdns
set system login sssd ipa-domain ipa.example.org ca-certificate 'ipa.example.org'
set system login sssd ipa-domain ipa.example.org cache-credentials
set system login sssd ipa-domain ipa.example.org ipa-hostname 'vyos.ipa.example.org'
set system login sssd ipa-domain ipa.example.org keytab 'host/vyos.ipa.example.org@IPA.EXAMPLE.ORG'
set system login sssd ipa-domain ipa.example.org krb-realm 'IPA.EXAMPLE.ORG'
set system login sssd ipa-domain ipa.example.org krb5-store-password-if-offline
```
After successfull commit SSSD should be running and an id lookup of the _admin_ user should return the uid and gid of the FreeIPA admin user

## Checklist:


sever-sever:
@h-eberhardt Are you still working on it?

h-eberhardt:
@sever-sever As there was a discussion in the corresponding phabricator task if this feature fits the scope of vyos (especially the kerberos part) I am waiting for some sort of consensus to form. @c-po wanted to wait for opinions from other maintainers, but no one has raised opinions yet.

c-po:
Well as outlined before - Kerberos is too heavy for a Router. The maximum I'd would go is LDAP based authentication using SSSD

h-eberhardt:
I submitted this PR to try building the foundation for enabling the Kerberos SSO- and KKDCP-Feature of ocserv and to maybe get Certmonger implemented.

This requires to have VyOS and ocserv properly integrated into FreeIPAs authentication, authorization and PKI systems. In the end, this could have enabled VyOS to serve as SSO-capable Openconnect VPN-Gateway, to get automatically renewing internal certificates (which could also be used for other services) and enabling dynamic Updates of the DNS Zones managed by FreeIPAs internal DNS. All of those features are usable if the entity (VyOS in this case) is authenticating via Kerberos. It should also not be too hard to expand the authentication and authorization part to Active Directory, to have sort of the same feature-set (excluding PKI I think) for those environments.

If there is no chance to get the Kerberos part integrated, I see no sense for me to put additional work into this, as in the end the approach with LDAP binds is not going to help with the problem I wanted to deal with initially.

As it seems like the decision regarding Kerberos is final, you can as well reject or close the PR and phabricator task.

c-po:
But why is OS system login Kerberos a dependency for ocserv kerberos support?

Isn't this what a SPN is for so it can only be implemented for a discrete service?

h-eberhardt:
As far as I understand, when you authenticate to ocserv via Kerberos, ocserv does the authentication and authorization independently of each other. Kerberos is only used for authentication of the user, as there is no Group-Information stored in the standard tickets (Microsoft does this differently via their MS-PAC extension).
The authorization part uses the regular PAM infrastructure to get the group memberships for the authenticated user, which itself gets the required information by querying sssd. As sssd must authenticate itself to an IDM domain to have access to all the required directory entries, there is the need to have sssd configured for Kerberos too.

Dec 17, 2023 17:38:21 Christian Breunig ***@***.***>:

c-po:
Closed as there are standing conflicts and we're unlikely adding global Kerberos support.

PR #2196, opened 2023-09-03
T5543: IGMP: Fix source address handling in static joins

vfreex:
The following command expects to join source-specific multicast group 239.1.2.3 on interface eth0, where the source address is 192.0.2.1.
```
set protocols igmp interface eth0 join 239.1.2.3 source 192.0.2.1
```

This command should generate FRR config:

```
interface eth0
 ip igmp
 ip igmp join 239.1.2.3 192.0.2.1
exit

```

However, there is a bug in the Jinja template where `if ifaces[iface].gr_join[group]` is mostly evaluated as `false` because `iface` is a loop variable from another loop.

## Types of changes

## Related Task(s)

* https://vyos.dev/Txxxx

## Checklist:

### DIFF
--- a/data/templates/frr/igmp.frr.j2
+++ b/data/templates/frr/igmp.frr.j2
@@ -27,9 +27,9 @@ interface {{ interface }}
 {%     if interface_config.query_max_resp_time %}
  ip igmp query-max-response-time {{ interface_config.query_max_resp_time }}
 {%     endif %}
-{%     for group in interface_config.gr_join %}
-{%         if ifaces[iface].gr_join[group] %}
-{%             for source in ifaces[iface].gr_join[group] %}
+{%     for group, sources in interface_config.gr_join.items() %}
+{%         if sources is vyos_defined %}
+{%             for source in sources %}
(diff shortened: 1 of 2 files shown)


data/templates/frr/igmp.frr.j2
```diff
@@ -27,9 +27,9 @@ interface {{ interface }}
 {%     if interface_config.query_max_resp_time %}
  ip igmp query-max-response-time {{ interface_config.query_max_resp_time }}
 {%     endif %}
-{%     for group in interface_config.gr_join %}
-{%         if ifaces[iface].gr_join[group] %}
-{%             for source in ifaces[iface].gr_join[group] %}
+{%     for group, sources in interface_config.gr_join.items() %}
+{%         if sources %}
```
c-po: Please use: `if sources is vyos_defined`

c-po (changes requested):

data/templates/frr/igmp.frr.j2
vfreex: done

c-po (approved):

PR #2361, opened 2023-10-12
pmacct: T5232: Fixed pmacct service control via systemctl

zdc:
## Change Summary

Fixed pmacct service control via systemctl

## Types of changes

## Related Task(s)

* https://vyos.dev/T5232, https://vyos.dev/T5233

## Component(s) name

pmacct

## Proposed changes

pmacct daemons have one very important specific - they handle control signals in the same loop as packets. And packets waiting is blocking operation.

Because of this, when systemctl sends SIGTERM to uacctd, this signal has no effect until uacct receives at least one packet via nflog. In some cases, this leads to a 90-second timeout, sending SIGKILL, and improperly finished tasks. As a result, a working folder is not cleaned properly.

This commit contains several changes to fix service issues:

- add a new nftables table for pmacct with a single rule to get the ability to send a packet to nflog and unlock uacctd
- remove PID file options from the uacctd and a systemd service file. Systemd can detect proper PID, and PIDfile is created by uacctd too late, which leads to extra errors in systemd logs
- KillMode changed to mixed. Without this, SIGTERM is sent to all plugins and the core process exits with status 1 because it loses connection to plugins too early. As a result, we have errors in logs, and the systemd service is in a failed state.
- added logging to uacctd
- systemctl service modified to send packets to specific address during a service stop which unlocks uacctd and allows systemctl to finish its work properly

## How to test

Configure `system flow-accounting`:

```
set system flow-accounting interface eth0
```

Disconnect all interfaces from a router and try to reboot it or change flow-accounting parameters. Without this change, it will take +90 seconds.

## Smoketest result

```
vyos@vyos:~$ /usr/libexec/vyos/tests/smoke/cli/test_system_flow-accounting.py
test_basic (__main__.TestSystemFlowAccounting.test_basic) ... 
Flow accounting requires at least one interface to be configured!

ok
test_netflow (__main__.TestSystemFlowAccounting.test_netflow) ... 
You need to configure at least one NetFlow server!

ok
test_sflow (__main__.TestSystemFlowAccounting.test_sflow) ... 
You need to configure at least sFlow or NetFlow, or not set "disable-imt" for flow-accounting!


You need to configure at least sFlow or NetFlow, or not set "disable-imt" for flow-accounting!

ok
test_sflow_ipv6 (__main__.TestSystemFlowAccounting.test_sflow_ipv6) ... 
You need to configure at least sFlow or NetFlow, or not set "disable-imt" for flow-accounting!


You need to configure at least sFlow or NetFlow, or not set "disable-imt" for flow-accounting!

ok

Ran 4 tests in 25.976s

OK

```
## Checklist:


c-po:
Interesting PR. Any chance this can be upstreamed? Or the maintainers could be made aware of the issues?

c-po (approved):

sever-sever (approved):

c-po:
@MergifyIo backport sagitta

zdc:
> Interesting PR. Any chance this can be upstreamed? Or the maintainers could be made aware of the issues?

This is a very dirty trick from my point of view - it is acceptable only because we control all the firewall/networking stack and can be sure that it will work for VyOS specifically. Repeating the same in upstream (Debian) seems not to be possible - there are too many combinations that can break it.

P.S. As I understand, pmacct developers are aware of this nuance, but for some reason cannot change this right now.

c-po:
@zdc thanks for the clarification.

PR #2507, opened 2023-11-20
op-mode: T4038: Python rewrite of image tools

erkin:
## Change Summary

Rewrite of `vyatta-image-tools.pl`. Brings with it a minor change to the template schema to add a special case for `comptype`. Please see the corresponding PR on `vyatta-op` below.

## Types of changes

## Related Task(s)

* https://vyos.dev/T4038

## Related PR(s)

* vyos/vyatta-op#84

## Component(s) name

op-mode, image tools

## Proposed changes

It should be functionally identical with the Perl implementation, except it eschews external programs like `rsync` and `curl` in favour of in-house routines and there are a few user-facing cosmetic changes.
It adds the `<imagePath/>` empty tag to the XML schema as an option for `<completionType>`. It should theoretically work with others like `<list>` and `<script>` too, but I haven't tried it yet. (I'll try it with `add system image` later.)
A certain `is_interactive()` check for known hosts is replaced with `isatty()` since it can be gracefully answered by automated scripts and because it was preventing the SFTP operations of `copy file`.
I tested all routines, except for some cases of `copy file upload` but it would definitely benefit from more eyes if we're going to backport this to 1.4.

## How to test

```
$ copy file [<image>://<path>|<url>] to [<image>://<path>|<url>]
$ delete file <image>://<path>
$ show file [<image>://<path>|<url>]
$ clone system config <image>
$ clone system config <image> from <image>
```

## Checklist:

### DIFF
--- a/src/op_mode/file.py
+++ b/src/op_mode/file.py
@@ -1,0 +1,383 @@
+#!/usr/bin/python3
+
+# Copyright 2023 VyOS maintainers and contributors <maintainers@vyos.io>
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import contextlib
+import datetime
+import grp
+import os
+import pwd
+import shutil
+import sys
+import tempfile
+
+from vyos.remote import download
+from vyos.remote import upload
+from vyos.utils.io import ask_yes_no
+from vyos.utils.io import print_error
+from vyos.utils.process import cmd
+from vyos.utils.process import run
+
+
+parser = argparse.ArgumentParser(description='view, copy or remove files and directories',
+                                 formatter_class=argparse.RawDescriptionHelpFormatter)
+parser.epilog = """
+TYPE is one of 'remote', 'image' and 'local'.
+A local path is <path> or ~/<path>.
+A remote path is <scheme>://<urn>.
+An image path is <image>:<path>.
+
+Clone operation is between images only.
+Copy operation does not support directories from remote locations.
+Delete operation does not support remote paths.
+"""
+operations = parser.add_mutually_exclusive_group(required=True)
+operations.add_argument('--show', nargs=1, help='show the contents of file PATH of type TYPE',
+                        metavar=('PATH'))
+operations.add_argument('--copy', nargs=2, help='copy SRC to DEST',
+                        metavar=('SRC', 'DEST'))
+operations.add_argument('--delete', nargs=1, help='delete file PATH',
+                        metavar=('PATH'))
+operations.add_argument('--clone', help='clone config from running image to IMG',
+                        metavar='IMG')
+operations.add_argument('--clone-from', nargs=2, help='clone config from image SRC to image DEST',
+                        metavar=('SRC', 'DEST'))
+
+## Helper procedures
+def fix_terminal() -> None:
+    """
+    Reset terminal after potential breakage caused by abrupt exits.
+    """
+    run('stty sane')
+
+def get_types(arg: str) -> tuple[str, str]:
+    """
+    Determine whether the argument shows a local, image or remote path.
+    """
+    schemes = ['http', 'https', 'ftp', 'ftps', 'sftp', 'ssh', 'scp', 'tftp']
+    s = arg.split("://", 1)
+    if len(s) != 2:
+        return 'local', arg
+    elif s[0] in schemes:
+        return 'remote', arg
+    else:
+        return 'image', arg
+
+def zealous_copy(source: str, destination: str) -> None:
+    # Even shutil.copy2() doesn't preserve ownership across copies.
+    # So we need to resort to this.
+    stats = os.stat(source)
+    shutil.copy2(source, destination)
+    os.chown(destination, stats.st_uid, stats.st_gid)
+
+def get_file_type(path: str) -> str:
+    return cmd(['file', '-sb', path])
+
+def print_header(string: str) -> None:
+    print('#' * 10, string, '#' * 10)
+
+def octal_to_symbolic(octal: str) -> str:
+    perms = ['---', '--x', '-w-', '-wx', 'r--', 'r-x', 'rw-', 'rwx']
+    result = ""
+    # We discard all but the last three digits because we're only
+    # interested in the permission bits.
+    for i in octal[-3:]:
+        result += perms[int(i)]
+    return result
+
+def get_user_and_group(stats: os.stat_result) -> tuple[str, str]:
+    try:
+        user = pwd.getpwuid(stats.st_uid).pw_name
+    except (KeyError, PermissionError):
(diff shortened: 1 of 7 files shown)


erkin:
Before you approve/merge, I have a very important question to those familiar with `vyatta-op`'s workings: There are checks for admin permissions [in the deleted template definitions](https://github.com/vyos/vyatta-op/pull/84/files#diff-5a5a7f18cd9434f64b9c82b15783cf3527f2381886c9664f1fc33ba7024cc582L4). Is this still necessary? I omitted it because `$VYATTA_USER_LEVEL_DIR` seems to be set to `/opt/vyatta/etc/shell/level/admin` for all users by default.
There are no such checks in the current implementation and there are no calls to `sudo` in the new template definitions. If it is necessary, please let me know because its omission could be a huge security hole.

dmbaturin:
@erkin Since it's no longer possible to set the user level to operator (we removed that option), I believe it's safe to just set it to admin for all users.

src/op_mode/file.py
```diff
@@ -0,0 +1,378 @@
...
+        # No need to duplicate local copy operations for image copying.
+        elif destination_type == 'image':
+            copy('local', source, 'local', parse_image_path(destination_path))
+        # Try to preserve metadata when copying.
+        elif destination_type == 'local':
+            if os.path.isdir(source):
+                if os.path.isdir(destination_path):
+                    destination_path = os.path.join(destination_path, os.path.basename(source))
+                shutil.copytree(source, destination_path, copy_function=zealous_copy)
+            else:
```
jestabro, line 249: A missing check here causes problems, in applying 'chown' to the destination directory:
```
diff --git a/src/op_mode/file.py b/src/op_mode/file.py
index b09631ba2..b675afebf 100755
--- a/src/op_mode/file.py
+++ b/src/op_mode/file.py
@@ -244,6 +244,8 @@ def copy(source_type: str, source_path: str,
                     destination_path = os.path.join(destination_path, os.path.basename(source))
                 shutil.copytree(source, destination_path, copy_function=zealous_copy)
             else:
+                if os.path.isdir(destination_path):
+                    destination_path = os.path.join(destination_path, os.path.basename(source))
                 zealous_copy(source, destination_path)
         else:
             print_error(f'Unknown destination type: {source_type}')
```

jestabro:
This one is a simplification using some of the utils from the image-tools revision and, for consideration, dropping the 'disk-install' case --- my impression is that it is antiquated/obsolete, but I may be missing something; we should check with @dmbaturin 

```
diff --git a/src/completion/list_images.py b/src/completion/list_images.py
index 2f22c2aab..e5af8b7ae 100755
--- a/src/completion/list_images.py
+++ b/src/completion/list_images.py
@@ -16,25 +16,25 @@
 
 import argparse
 import os
-import re
 import sys
 
+from vyos.system.image import is_live_boot
+from vyos.system.image import get_running_image
+
 parser = argparse.ArgumentParser(description='list available system images')
 parser.add_argument('--no-running', action='store_true',
                     help='do not display the currently running image')
 
-def get_current_image() -> str:
-    with open('/proc/cmdline', 'r') as f:
-        return re.match(r'BOOT_IMAGE=/boot/([^/]+)/vmlinuz', f.readline()).group(1)
-
 def get_images(omit_running: bool = False) -> list[str]:
+    if is_live_boot():
+        return []
     images = os.listdir("/lib/live/mount/persistence/boot")
-    # Check if we're running on a live disk
-    if os.path.isdir("/lib/live/mount/persistence/config") and not omit_running:
-        images += 'disk-install'
-    elif omit_running:
-        images.remove(get_current_image())
-    images.remove('grub')
+    if omit_running:
+        images.remove(get_running_image())
+    if 'grub' in images:
+        images.remove('grub')
+    if 'efi' in images:
+        images.remove('efi')
     return sorted(images)
 
 if __name__ == '__main__':
```

src/op_mode/file.py
erkin: I moved the shared check outside the conditional for simplicity.

src/op_mode/file.py
```diff
@@ -0,0 +1,378 @@
...
+        # The stream to the temporary file could break for any reason.
+        # It's much less fragile than if we streamed directly to the process stdin.
+        # But anything could still happen and we don't want to scare the user.
+        except (BrokenPipeError, EOFError, KeyboardInterrupt, OSError):
+            fix_terminal()
+            sys.exit(1)
+
+def show(type: str, path: str) -> None:
+    if type == 'remote':
+        with tempfile.NamedTemporaryFile(delete=False) as temp:
```
jestabro: Since using as a context manager here, one does not want `delete=False`

jestabro:
Comment added here with reference to issue discussed with Erkin; patch is for the companion PR https://github.com/vyos/vyatta-op/pull/84 and cited below.

Unless op-mode `file.py` is excluded from piping through VYATTA_PAGER, prompts to `input`, when called from `remote.download(...)`, are not printed to stdout --- input is still accepted, but the user is at a loss.
Current behavior:
```
vyos@vyos:~$ show file scp://vyos:dummy@192.168.1.5/home/vyos/files/test-file
Host '192.168.1.5' not found in known hosts.
Fingerprint: b4a8c0f8a8777137a03966fcf25ba58d
```
... and that's it: 'y'+'enter' or 'enter' is accepted, but prompt is missing.
After excluding from paging:
```
vyos@vyos:~$ show file scp://vyos:dummy@192.168.1.5/home/vyos/files/test-file
Host '192.168.1.5' not found in known hosts.
Fingerprint: b4a8c0f8a8777137a03966fcf25ba58d
Do you wish to continue? [y/N] y
########## FILE INFO ##########
Path:           /tmp/tmphlsbic_0
Type:           ASCII text
Owner:          root:root
Permissions:    rw-------
Modified:       2023-12-16 02:21:20

########## FILE DATA ##########

slug test-file: vyos on sirius; some data ...
```

patch for vyatta-op:
```
diff --git a/functions/interpreter/vyatta-op-run b/functions/interpreter/vyatta-op-run
index ee4cb1c..3260781 100644
--- a/functions/interpreter/vyatta-op-run
+++ b/functions/interpreter/vyatta-op-run
@@ -218,7 +218,7 @@ _vyatta_op_run ()
     local run_cmd=$(_vyatta_op_get_node_def_field $tpath/node.def run)
     run_cmd=$(_vyatta_op_conv_run_cmd "$run_cmd") # convert the positional parameters
     local ret=0
-    local cmd_regex="^(LESSOPEN=|less|pager|tail|/opt/vyatta/bin/vyatta-tshark-interface-port.pl).*"
+    local cmd_regex="^(.*/file.py.*|LESSOPEN=|less|pager|tail|/opt/vyatta/bin/vyatta-tshark-interface-port.pl).*"
     if [ -n "$run_cmd" ]; then
       eval $restore_shopts
       if [[ -t 1 &&  "${args[1]}" == "show" && ! $run_cmd =~ $cmd_regex ]] ; then
```

There are two things to note:
(1) as op-mode `file.py`  `show file` does its own paging, piping through `VYATTA_PAGER` is not recommended nor needed.
(2) The previous fact is _not_ the cause of the bug, as isolating to a simple download is sufficient to trigger; rather this has to do with paramiko's `MissingHostKeyPolicy`, though details not yet clear..

jestabro:
cf. comment in
https://github.com/vyos/vyatta-op/pull/84

jestabro:
A typo in `parse_image_path` prevents show/copy across images from working (as it leaves a leading slash in the second arg, so `os.path.join` is not what one wants):
https://github.com/vyos/vyos-1x/commit/38044479b34cbadc8271dad00a17ee2b6a87b486
Also, one can consider the following for manually setting perms on a file from remote download: it's a bit crude, and up to you whether/how to implement ... just a thought.
https://github.com/vyos/vyos-1x/commit/212f5d66a144ece81f76455d7bd4c167c1098023

dmbaturin (approved):

jestabro (approved):

PR #2541, opened 2023-11-25
remote: T5773: Fix for broken config download

erkin:
## Change Summary

Fix for a regression that broke conf-mode `load` (and probably a few other calls to `download()` and `upload()`).

## Types of changes

## Related Task(s)

* https://vyos.dev/T5773
* https://vyos.dev/T5774

## Component(s) name

remote, conf-mode

## Proposed changes

The original behaviour for failed transfer operations was printing out the error (after clipping the stack trace) but that was deemed too scary for the end-user, so the current version that simply prints the transfer error and exits came to be. The addition of a new positional argument breaks calls that `download()` and `upload()` that rely on positional arguments, although they should still probably be replaced with keyword arguments for clarity and most of them are from legacy code. If there is a need for detailed error logging, that can be arranged based on a feature request.

The `@umask()` context manager would benefit from being moved to a place like `vyos.utils.io` for broader use but it would have to be rewritten in class syntax because of top-level `import contextlib`.

## How to test

`# load <remote URL>`

## Checklist:

### DIFF
--- a/python/vyos/remote.py
+++ b/python/vyos/remote.py
@@ -46,17 +46,6 @@ from vyos.version import get_version
 
 CHUNK_SIZE = 8192
 
-@contextmanager
-def umask(mask: int):
-    """
-    Context manager that temporarily sets the process umask.
-    """
-    oldmask = os.umask(mask)
-    try:
-        yield
-    finally:
-        os.umask(oldmask)
-
 class InteractivePolicy(MissingHostKeyPolicy):
     """
     Paramiko policy for interactively querying the user on whether to proceed
@@ -88,6 +77,17 @@ class SourceAdapter(HTTPAdapter):
             num_pools=connections, maxsize=maxsize,
             block=block, source_address=self._source_pair)
 
+@contextmanager
+def umask(mask: int):
+    """
+    Context manager that temporarily sets the process umask.
+    """
+    import os
+    oldmask = os.umask(mask)
+    try:
+        yield
+    finally:
+        os.umask(oldmask)
 
 def check_storage(path, size):
     """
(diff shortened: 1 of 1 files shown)


sever-sever:
Looks good
```

vyos@r4# load http://192.168.122.14:8080/config.txt
Loading configuration from 'http://192.168.122.14:8080/config.txt'
[edit]
vyos@r4# 
[edit]
vyos@r4# compare 
[interfaces]
+ dummy dum0 {
+     address "203.0.113.1/32"
+ }

[edit]
vyos@r4# 
vyos@r4# 
vyos@r4# curl -k --location --request POST 'https://localhost/config-file' \
> --form data='{"op": "load", "file": "http://192.168.122.14:8080/config.txt"}' \
> --form key='foo'

{"success": true, "data": null, "error": null}[edit]
vyos@r4# 

```

sever-sever (approved):

jestabro (approved):

c-po:
Off topic question, can I POST an entire config as JSON to the API?

jestabro:
> Off topic question, can I POST an entire config as JSON to the API?

Not in practice, since the argument 'path' to the config section to be replaced must be non-empty, as validated by the http-api, so one can only replace a section of the config at a specified path.

It would be simple to extend the /configure or /config-file endpoint to allow loading a config from JSON, but this raises issues with hw-id --- @sever-sever 's solution was to restrict to reasonable sub-sections of the config.

sever-sever:
There is community member writes ansible module based on JSON load https://www.reddit.com/r/vyos/comments/152aurr/comment/jsj40ke/

c-po (approved):

sever-sever:
@MergifyIo backport sagitta

PR #2576, opened 2023-12-06
git: T5803: Adjust git configuration for baseline defaults

indrajitr:
## Change Summary

Apply baseline defaults for `.gitattributes` and `.vscode/settings.json` for improved developer experience.

The `.gitattrbutes` settings are based on:
- [Git documentation](https://git-scm.com/docs/gitattributes#_effects)
- [GitHub documentation](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings)
- [Community templates](https://github.com/gitattributes/gitattributes)

Since editor-agnostic line-ending specific settings are applied to `.gitattributes`, they can be removed from `.vscode/settings.json`. The global VSCode defaults have also been removed to avoid duplication.

## Types of changes

- [x] Other (please describe): git attributes, VSCode settings

## Related Task(s)

* https://vyos.dev/T5803

## Checklist:


.vscode/settings.json
```diff
@@ -1,18 +1,26 @@
 {
-    "files.trimTrailingWhitespace": true,
-    "editor.tabSize": 4,
```
c-po, line 3: Why remove remove the whitespace indention setting?

.gitattributes
```diff
@@ -1,4 +1,73 @@
...
+# *.svg    binary
+
+# Scripts
+*.bash     text eol=lf diff=bash
+*.sh       text eol=lf diff=bash
+
+# Serialisation
+*.json     text
+*.toml     text
+*.tmpl     text linguist-language=Jinja
```
c-po, line 44: .j2 is missing for Jinja2

c-po (changes requested):

.vscode/settings.json
indrajitr: `"editor.insertSpaces": true` and `"editor.tabSize": 4` are VSCode defaults.
Besides, with `"editor.detectIndentation": true` (also VSCode default), `"editor.tabSize"` and `"editor.insertSpaces"` are automatically detected.

So they would be redundant.

Of course, one can make a case for the situation where a developer might have changed these settings in their local editor config. But in that case, we would have to apply project overrides for far too many settings to compensate for the developer's "creativity" with local settings 🙂. Such situations are usually rare though.

.gitattributes
indrajitr: Right you are, I missed `.j2` and some other extensions. Have added them and tweaked some other ones. Generally, line endings for most text files should be handled properly by git but good to have them explicitly listed.

c-po (approved):

indrajitr:
_Note from https://vyos.dev/T5803 for somebody arriving here in the future from their custom fork:_

Since this PR affects the line-endings of files, once this PR is applied, one might need to reset the local git working tree from the local git index (after necessary care to `stash` or temporarily `commit` the local changes) by performing the following operations in the working directory:

```
git rm -rf --cached .
git reset --hard HEAD
```

indrajitr:
@c-po can this be considered for sagitta so that Mergify doesn't fail for future backports on the affected files?

PR #2584, opened 2023-12-07
login: T4943: use pam-auth-update to enable/disable Google authenticator

c-po:
## Change Summary

The initial version always enabled Google authenticator (2FA/MFA) support by hardcoding the PAM module for sshd and login.

This change only enables the PAM module on demand if any use has 2FA/MFA configured. Enabling the module is done system wide via pam-auth-update by using a predefined template.

Can be tested using:

```console
set system login user vyos authentication plaintext-password vyos
set system login user vyos authentication otp key 'QY735IG5HDHBFHS5W7Y2A4EM274SMT3O'
```

See https://docs.vyos.io/en/latest/configuration/system/login.html for additional details.

## Types of changes

## Related Task(s)

* https://vyos.dev/T4943

## Component(s) name

login

## Smoketest result

image

## Checklist:


src/pam-configs/mfa-google-authenticator
```diff
@@ -0,0 +1,13 @@
+Name: Google Authenticator PAM module (2FA/MFA)
+Default: no
+Priority: 9999
```
zdc: I think the priority should be `384` (based on https://wiki.ubuntu.com/PAMConfigFrameworkSpec)

`384` also will put it between `unix` and `radius/tacplus`, thus successful authentication using network PAM modules will skip MFA.

src/pam-configs/mfa-google-authenticator
```diff
@@ -0,0 +1,13 @@
...
+
+Auth-Type: Primary
+Auth:
+    required      pam_google_authenticator.so nullok forward_pass
+
+Account-Type: Primary
+Account:
+
+Session-Type: Additional
+Session:
```
zdc: Missing types can be omitted.

src/pam-configs/mfa-google-authenticator
```diff
@@ -0,0 +1,13 @@
+Name: Google Authenticator PAM module (2FA/MFA)
+Default: no
+Priority: 9999
+
+Auth-Type: Primary
+Auth:
+    required      pam_google_authenticator.so nullok forward_pass
```
zdc: This is equal to:

```
[default=bad success=ok new_authtok_reqd=ok ignore=ignore] pam_google_authenticator.so nullok forward_pass
```

But I would prefer:

```
[default=ignore success=ok auth_err=die] pam_google_authenticator.so nullok forward_pass
```

The module returns only `PAM_SUCCESS`, `PAM_IGNORE`, and `PAM_AUTH_ERR`. The main difference is that we will not call `unix` or other modules if the MFA code is wrong.

zdc (changes requested):

src/pam-configs/mfa-google-authenticator
c-po: In the original implementation it was the first entry, that's why I re-added it at first position.
Happy to use a sane and reasonable position in the config

c-po:
Incorporated requested changes by @zdc

GurliGebis:
Would it make sense to rename it from "Google Authenticator" to just "Authenticator"?
Since you can use several different apps for it (Like Microsoft Authenticator, which a lot of people who has to use Office 365 already have installed) - it would make sense to remove the Google branding.

zdc (approved):

sever-sever (approved):

c-po:
> Would it make sense to rename it from "Google Authenticator" to just "Authenticator"? Since you can use several different apps for it (Like Microsoft Authenticator, which a lot of people who has to use Office 365 already have installed) - it would make sense to remove the Google branding.

Well it was invented by Google that should be honored. I use it with Microsoft authenticator, too.

c-po:
@Mergifyio backport sagitta

GurliGebis:
@c-po that makes sense.
Just as long as people aren't misled into thinking they only need to use the Google app 🙂

PR #2617, opened 2023-12-11
ddclient: T5144,T5791: Fix migration to avoid config name conflict

indrajitr:
## Change Summary

This PR has fix for two potential issues:

* As part of T5144, we merged all services that used web-based address lookup under `service dns dynamic address web ...`. When migrating from `service dns dynamic interface <interface> ...` to `service dns dynamic address <address> ...`, the config name can potentially have a conflict when `address == 'web'`. Although the `/run/ddclient/ddclient.conf` that was generated earlier was incorrect, one could still potentially have misconfigured VyOS config without realizing it.
We now append the old <interface> name to the config name to avoid conflict.

* As part of T5791, since `service dns dynamic address <address> service <service> ...` changed to `service dns dynamic name <service> address <address> ...`, the resulting service and address config flip can result in conflicting `service` name.
We now migrate the service name to `<service>-<address>` to avoid the conflict.
So `service dns dynamic address <address> service <service> ...` is migrated to `service dns dynamic name <service>-<address> address <address> ...`

* Additionally, based on the recommendation in the PR review, service names now conform to alphanumeric constraints. The migration script normalizes the old names to conform to the constraints.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5144
* https://vyos.dev/T5791
* https://vyos.dev/T5721

## Component(s) name

dns dynamic

## Checklist:

### DIFF
--- a/interface-definitions/dns-dynamic.xml.in
+++ b/interface-definitions/dns-dynamic.xml.in
@@ -19,6 +19,10 @@
                     <format>txt</format>
                     <description>Dynamic DNS service name</description>
                   </valueHelp>
+                  <constraint>
+                    #include <include/constraint/alpha-numeric-hyphen-underscore.xml.i>
+                  </constraint>
+                  <constraintErrorMessage>Dynamic DNS service name must be alphanumeric and can contain hyphens and underscores</constraintErrorMessage>
                 </properties>
                 <children>
                   #include <include/generic-description.xml.i>
--- a/src/migration-scripts/dns-dynamic/0-to-1
+++ b/src/migration-scripts/dns-dynamic/0-to-1
@@ -81,20 +81,33 @@ for address in config.list_nodes(new_base_path):
                 config.rename(new_base_path + [address, 'service', svc_cfg, 'login'], 'username')
             # Apply global 'ipv6-enable' to per <config> 'ip-version: ipv6'
             if config.exists(new_base_path + [address, 'ipv6-enable']):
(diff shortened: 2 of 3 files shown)


c-po (changes requested):
Can dou please use zyphens in the tag Node service-address representation? Then its identical to the IPSec migrator and we can have a proper constraint on the tagNode

indrajitr:
> Can dou please use zyphens in the tag Node service-address representation? Then its identical to the IPSec migrator and we can have a proper constraint on the tagNode

`service dns dynamic address <address> service <service> ...` is now migrated to `service dns dynamic name <service>-<address> address <address> ...`.

c-po:
Oh, sorry - this is not what I ment. The service name should be (after migration) `dyndns-172-16-1-1` instead of `dyndns-172.16.1.1`

c-po:
Please add a contraint here https://github.com/vyos/vyos-1x/blob/current/interface-definitions/dns-dynamic.xml.in#L15-L22 so only valid names can be used and not something like `set service dns dynamic name fooÄÖÜBARß?=/(` 

You are most likely well suited using this statement:
```xml
<constraint>
  #include <include/constraint/alpha-numeric-hyphen-underscore.xml.i>
</constraint>
```

indrajitr:
My apologies too, I didn't catch what you meant. PR is now updated based on your recommendation.

c-po (approved):

PR #2651, opened 2023-12-18
firewall: T5834: Rename 'enable-default-log' to 'default-log'

indrajitr:
## Change Summary

Rename chain level defaults log option from `enable-default-log` to `default-log` for consistency.

Additionally, implement `default-log` in `policy route` template so that a configuration like `set policy route foo default-log` actually adds `log` in the respective chain.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5834

## Component(s) name

firewall

## Proposed changes

After the recent conversion of log leafNodes to valueless, a typical firewall rule entry that enables log looks like so:
```
set firewall ipv4 forward filter rule 101 log
```
However, the chain level defaults log looks like so (with redundant `enable` prefix):
```
set firewall ipv4 forward filter enable-default-log
```
We can rename `enable-default-log` for consistency with rule-level entries like so:
```
set firewall ipv4 forward filter default-log
```

### Firewall chain 

Configuration:
```
vyos@v15-c26# show firewall ipv4 name testname 
 default-action drop
 default-log
 rule 1 {
     action accept
     destination {
         address 172.16.10.10
     }
     log
     log-options {
         level debug
     }
     source {
         address 172.16.20.10
     }
     ttl {
         eq 15
     }
 }
```
Resulting output:
```
vyos@v15-c26$ sudo nft list chain ip vyos_filter NAME_testname
table ip vyos_filter {
	chain NAME_testname {
		ip daddr 172.16.10.10 ip saddr 172.16.20.10 ip ttl 15 log prefix "[ipv4-NAM-testname-1-A]" log level debug counter packets 0 bytes 0 accept comment "ipv4-NAM-testname-1"
		counter packets 0 bytes 0 log prefix "[ipv4-testname-default-D]" drop comment "testname default-action drop"
	}
}
```

### Policy chain

Configuration:
```
vyos@v15-c26# show policy route testroute 
 default-log
 rule 1 {
     action drop
     mark 2020
     protocol udp
 }
```
Resulting output:
```
vyos@v15-c26$ sudo nft list chain ip vyos_mangle VYOS_PBR_UD_testroute
table ip vyos_mangle {
	chain VYOS_PBR_UD_testroute {
		meta l4proto udp meta mark 0x000007e4 counter packets 0 bytes 0 drop comment "ipv4-route-testroute-1"
		counter packets 0 bytes 0 log prefix "[ipv4-testroute-default]"
	}
}
```

## Smoketest result

```
vyos@v15-c26:~$ /usr/libexec/vyos/tests/smoke/cli/test_firewall.py 
test_bridge_basic_rules (__main__.TestFirewall.test_bridge_basic_rules) ... ok
test_flow_offload (__main__.TestFirewall.test_flow_offload) ... 
Interface "eth0" requires "offload hw-tc-offload"

ok
test_geoip (__main__.TestFirewall.test_geoip) ... Updating GeoIP. Please wait...
Updating GeoIP. Please wait...
ok
test_ipv4_synproxy (__main__.TestFirewall.test_ipv4_synproxy) ... 
"synproxy" option allowed only for action synproxy

ok
test_nested_groups (__main__.TestFirewall.test_nested_groups) ... 
Group "smoketest_network1" has a circular reference

ok
test_zone_flow_offload (__main__.TestFirewall.test_zone_flow_offload) ... 
Interface "eth0" requires "offload hw-tc-offload"

ok

Ran 17 tests in 98.948s

OK
vyos@v15-c26:~$ /usr/libexec/vyos/tests/smoke/cli/test_policy_route.py 

Ran 5 tests in 36.096s

OK

```

## Checklist:


sever-sever:
As I remember `enable-default-log` logs only entries that are not defined in the firewall as regular rules (i.e, only for the default rule/action)

Renaming to `default log` can confuse users who expect to log all defined rules by default.

On the other hand, it would be a good idea if we had a proper description of it.

dmbaturin (commented):
I wonder if we should rethink the option name completely indeed. `log-all-traffic`? `log-all-rules`?

c-po:
> I wonder if we should rethink the option name completely indeed. `log-all-traffic`? `log-all-rules`?

We do not log traffic, we log "hitting" a rule - this should be taken into account, so I think `log-all-traffic` might be missleading

> As I remember enable-default-log logs only entries that are not defined in the firewall as regular rules (i.e, only for the default rule/action)

Correct, this is/was used to add a log entry to the Kernel, if a firewall rule "walked" to the end and the default rule (most likely deny) was hit, resulting in a message like this (generated from `VyOS 1.4-rolling-202312170908`):

```
Dec 17 19:08:11 kernel: [ipv6-ALLOW-ESTABLISHED-6-default-D]IN=pppoe0 OUT=eth0.10 MAC= SRC=2600:9000:237d:ca00:0017:d593:2800:93a1 DST=2003:00ec:0748:c40a:ad69:9377:77b9:b0bf LEN=111 TC=2 HOPLIMIT=60 FLOWLBL=208271 PROTO=TCP SPT=443 DPT=49771 WINDOW=161 RES=0x00 ACK PSH URGP=0
Dec 17 19:08:15 kernel: [ipv4-LOCAL-WAN-default-D]IN= OUT=pppoe0 SRC=193.159.53.91 DST=80.151.69.205 LEN=176 TOS=0x08 PREC=0x80 TTL=64 ID=64668 PROTO=UDP SPT=10200 DPT=10200 LEN=156
```

So I think the best is a mix of both worlds:
* `set firewall ipv4 name DMZ-GUEST rule 10 log` current 1.4 CLI which feels fine - the help can be improved to: "Create log entry once rule is hit" over the current "Enable log"
* `set firewall ipv4 name DMZ-GUEST enable-default-log` my idea would be: `set firewall ipv4 name DMZ-GUEST log-default-action` as It really logs, when the default action of the rule is hit.

indrajitr:
`enable-default-log` is probably more like a _default_ fallback clause in a _switch-case_ statement. This means individual actions (for example a _jump target_) can have their own log as well.

Here is the thought process I had:

If we see the options for a chain, it looks like this right now:
```
vyos@v15-0b12# set firewall ipv4 name IPV4-WAN-LAB
Possible completions:
   default-action       Default-action for rule-set (default: drop)
   default-jump-target  Set jump target. Action jump must be defined in default-action
                        to use this setting
   description          Description
   enable-default-log   Log packets hitting default-action
+> rule                 IPv4 Firewall custom rule number
```
So we already have `default-action` and `default-jump-target`. By renaming `enable-default-log` to `default-log` it would have two effects:
- Make it obvious that `default-log` is associated with `default-action` (just like `default-jump-target` is associated with `default-action`).
- Remove the redundant prefix `enable-` to align with the valueless `log` in rules. With this:
  - chains have: `default-action`, `default-log` (default/last fallback, implicit action)
  - rules have: `action`, `log` (rule-specific action)

indrajitr:
> * `set firewall ipv4 name DMZ-GUEST enable-default-log` my idea would be: `set firewall ipv4 name DMZ-GUEST log-default-action` as It really logs, when the default action of the rule is hit.

Indeed, had this thought as well, but then it occurred to me that if at the chain level we end up having `default-action` and `log-default-action`, it would end up being inconsistent at rule level -- where we have `action` and `log` (instead of `action`, `log-action`).

c-po:
`default-log` or `default-action-log` would both work for me at the definition level `set firewall ipv4 name IPV4-WAN-LAB` - whereas I'd prefer the first one because it's shorter.

So if this is "set" and we go down that path, it should logically be at a rule level:
* `set firewall ipv4 name IPV4-WAN-LAB action ...`
* `set firewall ipv4 name IPV4-WAN-LAB action-log` - somehow `log-action` more explains what we do and I think we should not overcomplicate it and rather use CLI nodes defining what really happens, thus `log-action` should be preferred of `action-log`

The more parties, the more opinions 😄

indrajitr:
Okay, summarizing the combinations here:

* Option 1:
```
// chain level
set firewall ipv4 name IPV4-WAN-LAB default-action 'jump'
set firewall ipv4 name IPV4-WAN-LAB default-jump-target ...
set firewall ipv4 name IPV4-WAN-LAB default-log

// rule level
set firewall ipv4 name IPV4-WAN-LAB rule 101 action 'drop'
set firewall ipv4 name IPV4-WAN-LAB rule 101 log
```

* Option 2:
```
// chain level
set firewall ipv4 name IPV4-WAN-LAB default-action 'jump'
set firewall ipv4 name IPV4-WAN-LAB default-jump-target ...
set firewall ipv4 name IPV4-WAN-LAB default-log-action

// rule level
set firewall ipv4 name IPV4-WAN-LAB rule 101 action 'drop'
set firewall ipv4 name IPV4-WAN-LAB rule 101 log-action
```

* Option 3:
```
// chain level
set firewall ipv4 name IPV4-WAN-LAB action 'jump'
set firewall ipv4 name IPV4-WAN-LAB jump-target ...
set firewall ipv4 name IPV4-WAN-LAB log-action

// rule level
set firewall ipv4 name IPV4-WAN-LAB rule 101 action 'drop'
set firewall ipv4 name IPV4-WAN-LAB rule 101 log-action
```

* Option 4:
```
set firewall ipv4 name IPV4-WAN-LAB action 'jump'
set firewall ipv4 name IPV4-WAN-LAB jump-target ...
set firewall ipv4 name IPV4-WAN-LAB log

set firewall ipv4 name IPV4-WAN-LAB rule 101 action 'drop'
set firewall ipv4 name IPV4-WAN-LAB rule 101 log
```

NB: Of course, the help messages should be improved regardless.

dmbaturin:
@indrajitr I'm leaning towards option number 1.

c-po:
@indrajitr I'm leaning towards option number 1.

indrajitr:
Thank you for weighing in @dmbaturin @c-po! This PR is ready in that case.

c-po (approved):

c-po:
@nicolas-fort any objections?

nicolas-fort:
I' like this PR, but are you sure, as it is, it doesn't break policy route? I saw only one file changed regarding policy route, so maybe now smoketest for policy route and route6 breaks.
Same as migrations.
If possible, please add results of those smoketests.

indrajitr:
> I saw only one file changed regarding policy route, so maybe now smoketest for policy route and route6 breaks

Thanks, will push an update shortly.

indrajitr:
1. Updated migrations
2. @nicolas-fort, looks like `test_policy_route.py` doesn't have tests applying `enable-default-log` and thus it succeeds even though `enable-default-log` is renamed to `default-log`. I wanted to add tests for this but looks like `enable-default-log` isn't used at all in `nftables-policy.j2`. Am I missing something?

c-po:
I can confirm builds an ISO from scratch with this PR all smoketests passed.

Question is, when doing policy based routing - do we wan't to have a log entry for every packet/flow traversing?

nicolas-fort:
Maybe is not a big request logging such entries, but if it's supported in the cli, then it should definitely do something. I don't know if it should be fixed in this PR, or create a patch in separate PR.

nicolas-fort:
And I can see a new migration script 7-to-8 for policy route, please also bump version number in https://github.com/vyos/vyos-1x/blob/current/interface-definitions/include/version/policy-version.xml.i (as you did for firewall).

c-po:
> Maybe is not a big request logging such entries, but if it's supported in the cli, then it should definitely do something. I don't know if it should be fixed in this PR, or create a patch in separate PR.

If it's broken it should be fixed in this PR, too as it deals with the log situation.

c-po:
> And I can see a new migration script 7-to-8 for policy route, please also bump version number in https://github.com/vyos/vyos-1x/blob/current/interface-definitions/include/version/policy-version.xml.i (as you did for firewall).

Bumped version to 8 - also running a full build with smoketests now

c-po:
image

indrajitr:
Updated policy route template and smoke tests. The description has been updated to include sample configuration and the generated chain.

indrajitr:
Since this has log adjustments for the policy route, would this also be a backport candidate?

c-po:
@MergifyIo backport sagitta

PR #2708, opened 2023-12-29
T5871: ipsec remote access VPN: specify "cacerts" for client auth

lucasec:
## Change Summary

For authentication methods that depend on validating a client certificate against a CA (e.g. EAP-TLS), we currently do not explicitly tell strongswan which CA to use. This PR specifies the `cacerts` option explicitly for every connection to ensure the connection only accepts certificates signed by its specific CA.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5871

## Related PR(s)

N/A

## Component(s) name

ipsec remote-access

## Proposed changes

The configured CA certificate fro the connection is added to the swanctl.conf connection definition using the `cacerts` option.

## How to test

Example configuration:
```
[edit vpn ipsec remote-access connection ClientVPN]
lucas@lcn-router# show
 authentication {
     client-mode eap-tls
     local-id <local id>
     server-mode x509
     x509 {
         ca-certificate <ca certificate name>
         certificate <server certificate name>
     }
 }
 local-address <router IP address>
 ...
```

Validate that `/etc/swanctl/swanctl.conf` specifies `cacerts = <ca certificate name>_1.pem` under `remote`.

## Checklist:

### DIFF
--- a/data/templates/ipsec/swanctl/remote_access.j2
+++ b/data/templates/ipsec/swanctl/remote_access.j2
@@ -35,6 +35,11 @@
             auth = {{ rw_conf.authentication.client_mode }}
             eap_id = %any
 {% endif %}
+{% if rw_conf.authentication.client_mode is vyos_defined('eap-tls') or rw_conf.authentication.client_mode is vyos_defined('x509') %}
+{#          pass all configured CAs as filenames, separated by commas #}
+{#          this will produce a string like "MyCA1.pem,MyCA2.pem" #}
+            cacerts = {{ '.pem,'.join(rw_conf.authentication.x509.ca_certificate) ~ '.pem' }}
+{% endif %}
         }
         children {
             ikev2-vpn  {
--- a/smoketest/scripts/cli/test_vpn_ipsec.py
+++ b/smoketest/scripts/cli/test_vpn_ipsec.py
@@ -757,6 +757,7 @@ class TestVPNIPsec(VyOSUnitTestSHIM.TestCase):
             f'id = "{local_id}"',
             f'auth = pubkey',
             f'certs = peer1.pem',
+            f'cacerts = MyVyOS-CA.pem',
             f'auth = eap-tls',
             f'eap_id = %any',
             f'esp_proposals = aes256-sha512,aes256-sha384,aes256-sha256,aes256-sha1,aes128gcm128-sha256',
@@ -840,6 +841,7 @@ class TestVPNIPsec(VyOSUnitTestSHIM.TestCase):
         with self.assertRaises(ConfigSessionError):
             self.cli_commit()
         self.cli_set(base_path + ['remote-access', 'connection', conn_name, 'authentication', 'x509', 'ca-certificate', ca_name])
+        self.cli_set(base_path + ['remote-access', 'connection', conn_name, 'authentication', 'x509', 'ca-certificate', int_ca_name])
 
         self.cli_set(base_path + ['remote-access', 'connection', conn_name, 'esp-group', esp_group])
         self.cli_set(base_path + ['remote-access', 'connection', conn_name, 'ike-group', ike_group])
@@ -867,6 +869,7 @@ class TestVPNIPsec(VyOSUnitTestSHIM.TestCase):
             f'id = "{local_id}"',
             f'auth = pubkey',
             f'certs = peer1.pem',
+            f'cacerts = MyVyOS-CA.pem,MyVyOS-IntCA.pem',
             f'esp_proposals = aes256-sha512,aes256-sha384,aes256-sha256,aes256-sha1,aes128gcm128-sha256',
             f'rekey_time = {eap_lifetime}s',
             f'rand_time = 540s',
@@ -894,6 +897,7 @@ class TestVPNIPsec(VyOSUnitTestSHIM.TestCase):
 
         # Check Root CA, Intermediate CA and Peer cert/key pair is present
         self.assertTrue(os.path.exists(os.path.join(CA_PATH, f'{ca_name}.pem')))
+        self.assertTrue(os.path.exists(os.path.join(CA_PATH, f'{int_ca_name}.pem')))
         self.assertTrue(os.path.exists(os.path.join(CERT_PATH, f'{peer_name}.pem')))
 
         self.tearDownPKI()


data/templates/ipsec/swanctl/remote_access.j2
```diff
@@ -32,6 +32,9 @@
             auth = {{ rw_conf.authentication.client_mode }}
 {% if rw_conf.authentication.client_mode.startswith("eap") %}
             eap_id = %any
+{% endif %}
+{% if rw_conf.authentication.client_mode == 'eap_tls' or rw_conf.authentication.client_mode == 'x509' %}
```
c-po: You should use `eap-tls` over `eap_tls`. Please rephrase it to:

`{% if rw_conf.authentication.client_mode is vyos_defined('eap-tls') or rw_conf.authentication.client_mode is vyos_defined('x509') %}`

Else the code path is not hit for eap-tls client auth.

c-po (changes requested):

data/templates/ipsec/swanctl/remote_access.j2
c-po: Please also update the IPSec remote-access smoketest (test_07) in https://github.com/vyos/vyos-1x/blob/current/smoketest/scripts/cli/test_vpn_ipsec.py#L551 to validate this option.

lucasec:
@c-po Updated with smoketests and rebase onto latest.

@sarthurdev I did a little more investigation into the CA chain behavior you introduced in https://github.com/vyos/vyos-1x/commit/1ac230548c86d3308ff5b479b79b0e64b75a0e8a. From my understanding, the user would specify the intermediate CA name under `authentication x509 ca-certificate`, and then the config logic would render the intermediate as `CA_name_1.pem`, next parent as `CA_name_2.pem`, up to the root being `CA_name_N.pem`.

I would think in most cases the user would want strongswan to be "strict", as in only accept client certs signed by the intermediate. In that case, the current behavior in this PR, which renders `cacerts = CA_name_1.pem`, would match the user's expectation. But I could see valid cases where the user would want more broad acceptance, e.g. accept client certs signed by other intermediates under the same root (closer to the current behavior without this PR).

Definitely open to more discussion on this, e.g. are we overloading too much behavior on the `x509 ca-certificate` option.

sarthurdev:
> @sarthurdev I did a little more investigation into the CA chain behavior you introduced in [1ac2305](https://github.com/vyos/vyos-1x/commit/1ac230548c86d3308ff5b479b79b0e64b75a0e8a). From my understanding, the user would specify the intermediate CA name under `authentication x509 ca-certificate`, and then the config logic would render the intermediate as `CA_name_1.pem`, next parent as `CA_name_2.pem`, up to the root being `CA_name_N.pem`.
> 
> I would think in most cases the user would want strongswan to be "strict", as in only accept client certs signed by the intermediate. In that case, the current behavior in this PR, which renders `cacerts = CA_name_1.pem`, would match the user's expectation. But I could see valid cases where the user would want more broad acceptance, e.g. accept client certs signed by other intermediates under the same root (closer to the current behavior without this PR).

Thanks for raising this, it's a very good point. Thinking my commit should be reverted and a multi value `ca-certificate` node is made available instead.

lucasec:
If I understand correctly, there are sort of two behaviors an admin might want to accomplish by supplying an intermediate to strongswan:

1. Allow strongswan to "complete the chain" if the client supplies a cert signed by the intermediate but does not supply the intermediates along with it (I am familiar with transmitting multiple certs over the wire in the context of HTTP, but not as much with if/how that is implemented in IKE).
2. The admin might want to only have certs issued by a specific intermediate trusted.

In case 1, if there were other intermediates below the root and the admin also wanted those trusted, they would need to be able to load all of the different intermediates at the same level into strongswan. If there are multiple levels of intermediates it seems like the number of certs that need to be loaded could explode out quite a bit.

And in case 2, if you only want the intermediate trusted, would strongswan need to have access to the root at all?

data/templates/ipsec/swanctl/remote_access.j2
```diff
@@ -34,6 +34,11 @@
 {% elif rw_conf.authentication.client_mode.startswith("eap") %}
             auth = {{ rw_conf.authentication.client_mode }}
             eap_id = %any
+{% endif %}
+{% if rw_conf.authentication.client_mode is vyos_defined('eap-tls') or rw_conf.authentication.client_mode is vyos_defined('x509') %}
+{#          in the case of an intermediate/root chain, assume user wants us to valiadate against #}
+{#          the lowest specified cert in the chain (e.g. the intermediate) #}
+            cacerts = {{ rw_conf.authentication.x509.ca_certificate }}_1.pem
```
sarthurdev: Need to change to:
`cacerts = {{ '.pem,'.join(rw_conf.authentication.x509.ca_certificate) ~ '.pem' }}`

(Should be correct, though will need testing)

smoketest/scripts/cli/test_vpn_ipsec.py
```diff
@@ -754,6 +754,7 @@ def test_08_ikev2_road_warrior_client_auth_eap_tls(self):
             f'id = "{local_id}"',
             f'auth = pubkey',
             f'certs = peer1.pem',
+            f'cacerts = MyVyOS-CA_1.pem',
```
sarthurdev: `_1` needs to be dropped

smoketest/scripts/cli/test_vpn_ipsec.py
```diff
@@ -864,6 +865,7 @@ def test_09_ikev2_road_warrior_client_auth_x509(self):
...
```
sarthurdev: `_1` needs to be dropped

sarthurdev (changes requested):

c-po:
@lucasec can you rebase your work to the latest `current` now that #3205 is merged?

lucasec:
Rebased accordingly. Verified the updated smoketests pass but I am holding testing this on my live system until I finish up some other changes tomorrow (feel free to merge or hold at your discretion).

data/templates/ipsec/swanctl/remote_access.j2
lucasec: That syntax should work but I ended up taking what I thought was a "cleaner" approach of adding a filter. You are welcome to veto that though.

lucasec:
Installed on my live system and looks fine.

I am still trying to chase some other issue causing system crashes seemingly related to ipsec, but those predate this PR (I am now investigating whether they are possibly related to https://github.com/vyos/vyos-1x/pull/3157).

sarthurdev (approved):

sever-sever:
@lucasec is it ready to merge?

lucasec:
Yes—this one is clear to merge.

sever-sever (approved):

python/vyos/template.py
```diff
@@ -172,6 +172,14 @@ def force_to_list(value):
     else:
         return [value]
 
+@register_filter('prepend_str')
+def prepend_str(value, prefix):
```
c-po: Why not use the Jinja2 ~ operator to concat strings?

c-po (changes requested):

python/vyos/template.py
lucasec: We could.

I thought this was cleaner and would be easier to maintain than @sarthurdev’s suggestion here:
> Need to change to: `cacerts = {{ '.pem,'.join(rw_conf.authentication.x509.ca_certificate) ~ '.pem' }}`

(in particular, the use of `join` with `.pem` as the separator is clever, but IMHO not immediately obvious what it’s doing)

c-po: Please proceed with join and pem separator and add a proper comment. Then I'm happy to merge this one!

lucasec: Updated and confirmed smoke tests still pass.

c-po (approved):

c-po:
@MegifyIo backport sagitta

GurliGebis:
@c-po it looks like it didn't create the backport pr

sarthurdev:
@mergifyio backport sagitta

PR #2806, opened 2024-01-11
image-tools: T5910: explicitly set transmission speed of serial console

jestabro:
## Change Summary

The image-tools install scripts do not explicitly set the serial console speed, resulting in a GRUB default of 9600. In the case of a serial-only installation, this is a confounding, as the live image sets transmission speed to 115200, whereas the resulting installed image does not. Explicitly set for both fresh installs and for compatibility mode when legacy images are subsequently removed.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5910

## Checklist:

### DIFF
--- a/data/templates/grub/grub_common.j2
+++ b/data/templates/grub/grub_common.j2
@@ -8,9 +8,9 @@ fi
 function setup_serial {
     # initialize the first serial port by default
     if [ "${console_type}" == "ttyS" ]; then
-        serial --unit=${console_num}
+        serial --unit=${console_num} --speed=${console_speed}
     else
-        serial --unit=0
+        serial --unit=0 --speed=${console_speed}
     fi
     terminal_output --append serial console
     terminal_input --append serial console
--- a/data/templates/grub/grub_vyos_version.j2
+++ b/data/templates/grub/grub_vyos_version.j2
@@ -6,16 +6,21 @@
 {% endif %}
 menuentry "{{ version_name }}" --id {{ version_uuid }} {
     set boot_opts="{{ boot_opts_rendered }}"
+    if [ "${console_type}" == "ttyS" ]; then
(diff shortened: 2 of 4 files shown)


c-po:
Why not reference the serial console speed from CLI? Looks like now we have two places defining the speed.

Updating serial speed on the CLI should also be reflected in the GRUB configuration

jestabro:
Yes, good point, and in fact this reveals a second bug: config-mode `system_console.py` is only appropriate for legacy/compatibility grub.cfg. The good news is that the new image tools allow an easy utility for setting this from the config-mode script; this will need some care for the compatibility mode, to be addressed in a separate task.

jestabro:
(addendum:) A fix for https://vyos.dev/T5923, pending, will correctly update the GRUB settings; the default values in PR above are only for initial install.

fghorow:
> Why not reference the serial console speed from CLI? Looks like now we have two places defining the speed.
> 
> Updating serial speed on the CLI should also be reflected in the GRUB configuration

Just because I did this search yesterday when I was seeing if there was some way for me to get the codebase into my head enough to contribute a fix, [this search](https://github.com/search?q=repo%3Avyos%2Fvyos-1x+115200&type=code) shows 42 hits in the repo.

It's significantly more than just two places in the codebase where 115200 is a "magic number"...

jestabro:
@fghorow yes, but most of those hits are in the example configs that are used in one category of smoketests, and have no instrumental effect.

c-po (approved):

c-po:
Approving this as hotfix, but we should properly resolve the issue as mentioned above with the console speed on the CLI config.

jestabro:
A fix for the follow-on https://vyos.dev/T5923 should be ready shortly.

c-po:
@MergifyIo backport sagitta

PR #2839, opened 2024-01-17
dhcp: T5952: Fix validate duplicate MAC Address on same subnet

josephillips85:
## Types of changes

## Related Task(s)

* https://vyos.dev/T5952

## Checklist:

### DIFF
--- a/src/conf_mode/service_dhcp-server.py
+++ b/src/conf_mode/service_dhcp-server.py
@@ -233,6 +233,7 @@ def verify(dhcp):
             if 'static_mapping' in subnet_config:
                 # Static mappings require just a MAC address (will use an IP from the dynamic pool if IP is not set)
                 used_ips = []
+                used_mac = []
                 for mapping, mapping_config in subnet_config['static_mapping'].items():
                     if 'ip_address' in mapping_config:
                         if ip_address(mapping_config['ip_address']) not in ip_network(subnet):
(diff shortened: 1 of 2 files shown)


c-po (approved):

src/conf_mode/service_dhcp-server.py
```diff
@@ -247,7 +248,11 @@ def verify(dhcp):
                         if mapping_config['ip_address'] in used_ips:
                             raise ConfigError(f'Configured IP address for static mapping "{mapping}" exists on another static mapping')
 
+                        if mapping_config['mac'] in used_mac:
+                            raise ConfigError(f'Configured MAC address for static mapping "{mapping}" exists on another static mapping')
+
                         used_ips.append(mapping_config['ip_address'])
+                        used_mac.append(mapping_config['mac'])
```
sarthurdev, line 256: This needs to be inside a check that the mac field is present. The code can reach this point with only a DUID and IP set.

sarthurdev (changes requested):

src/conf_mode/service_dhcp-server.py
c-po: Good catch! We should have an identical if statement for the `mac` address key

c-po (changes requested):

src/conf_mode/service_dhcp-server.py
josephillips85: Ok i perform the corrections the IF statement should't be identical . acording the documentation you are able to set a mac address only and will assigned by the dynamic pool .

c-po: I wonder if we should add a check for duplicate duids as well @sarthurdev?

c-po (approved):

PR #2843, opened 2024-01-17
dhcp: T5948: Strip trailing dot from hostnames

sarthurdev:
## Change Summary

Strip trailing dot from received DHCP hostname

## Types of changes

## Related Task(s)

* https://vyos.dev/T5948

## Component(s) name

dhcp-server

## Proposed changes

Strips '.' from hostname variable, hostnames should only contain alphanumeric and hyphen characters.

## Checklist:

### DIFF
--- a/src/system/on-dhcp-event.sh
+++ b/src/system/on-dhcp-event.sh
@@ -63,7 +63,7 @@ case "$action" in
 
       client_ip=${!client_ip_var}
       client_mac=${!client_mac_var}
-      client_name=${!client_name_var}
+      client_name=${!client_name_var//./}
       client_subnet_id=${!client_subnet_id_var}
 
       if [ -z "$client_name" ]; then


c-po (approved):

c-po:
@MergifyIo backport sagitta

sarthurdev:
No backport necessary here, should only be issue with Kea.

chenxiaolong:
Thanks for the quick fix! Would it make sense to use `${!client_name_var%.}` instead? The current solution also eliminates dots in the middle.

If the client runs eg. `sudo dhcpcd -h foo.bar -F both -4B enp2s0f0np0`, then VyOS will treat it as `foobar.<domain>` instead of `foo.bar.<domain>`. (Corporate Windows machines tend to do this sort of thing in my experience.)

EDIT: That said, I have no objection at all if VyOS prefers to just keep the solution of getting rid of all dots.

sarthurdev:
The hostname should follow the domain segment character limitations (alphanumeric/hyphen only). Though the RFC does seem to indicate the field may instead contain the whole FQDN.

I'm not sure however If Kea hooks send the FQDN or strip the suffix, I suppose if it does send a FQDN, the script would mangle the hostname (e.g. recv'd hostname = testserver.network.lan, resulting client_name = testservernetworklan.network.lan).

I'll experiment in my Kea lab.

PR #2960, opened 2024-02-07
xml: T302: replace references to Quagga with FRRouting

c-po:
Just replace all help string references


c-po:
@mergifyIo backport sagitta

PR #2965, opened 2024-02-08
T5872: ipsec remote access VPN: support dhcp-interface.

lucasec:
## Change Summary
Enables the `dhcp-interface` option to be used with ipsec remote-access VPNs.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5872

## Component(s) name

ipsec remote-access

## Proposed changes

This adds the ability to use `dhcp-interface` with the ipsec remote-access VPN. It works essentially the same way as the `dhcp-interface` behavior for the ipsec site-to-site VPN.

While most IKEv2 Remote Access ("Road Warrior") VPN setups will likely be at sites with a fixed IP, the DHCP functionality may be useful for isolated installs, say, at a small business or remote branch office that may be on an ISP package that only offers DHCP addressing. Combined with dynamic DNS, VyOS can still provide a one-stop routing and remote access solution.

## How to test

Example configuration:
```
[edit vpn ipsec remote-access connection ClientVPN]
lucas@lcn-router# show
 authentication {
     client-mode eap-tls
     local-id <local id>
     server-mode x509
     x509 {
         ca-certificate <ca certificate name>
         certificate <server certificate name>
     }
 }
 dhcp-interface eth0
 ...
```

The ipsec daemon should start normally and accept connections.
If you inspect `/etc/swanctl/swanctl.conf` on the running router, you should see a peer configuration that begins with the following (replace `1.2.3.4` with the router's WAN interface IP on eth0):

```
    ra-ClientVPN {
        remote_addrs = %any
        local_addrs = 1.2.3.4 # dhcp:eth1
```

You can simulate the DHCP address change behavior by manually invoking the dhclient exit hook:
```
sudo interface=eth1 old_ip_address=1.2.3.4 new_ip_address=5.6.7.8 /etc/dhcp/dhclient-exit-hooks.d/99-ipsec-dhclient-hook
```

## Smoketest result

```
DEBUG - Running Testcase: /usr/libexec/vyos/tests/smoke/cli/test_vpn_ipsec.py
DEBUG - test_01_dhcp_fail_handling (__main__.TestVPNIPsec.test_01_dhcp_fail_handling) ... ok
DEBUG - test_02_site_to_site (__main__.TestVPNIPsec.test_02_site_to_site) ... ok
DEBUG - test_03_site_to_site_vti (__main__.TestVPNIPsec.test_03_site_to_site_vti) ... ok
DEBUG - test_04_dmvpn (__main__.TestVPNIPsec.test_04_dmvpn) ... ok
DEBUG - test_05_x509_site2site (__main__.TestVPNIPsec.test_05_x509_site2site) ... ok
DEBUG - test_06_flex_vpn_vips (__main__.TestVPNIPsec.test_06_flex_vpn_vips) ... ok
DEBUG - test_07_ikev2_road_warrior (__main__.TestVPNIPsec.test_07_ikev2_road_warrior) ... ok
DEBUG - test_08_ikev2_road_warrior_client_auth_eap_tls (__main__.TestVPNIPsec.test_08_ikev2_road_warrior_client_auth_eap_tls) ... ok
DEBUG - test_09_ikev2_road_warrior_client_auth_x509 (__main__.TestVPNIPsec.test_09_ikev2_road_warrior_client_auth_x509) ... ok
DEBUG - test_10_ikev2_road_warrior_dhcp_fail_handling (__main__.TestVPNIPsec.test_10_ikev2_road_warrior_dhcp_fail_handling) ... ok
```

## Checklist:

### DIFF
--- a/data/templates/ipsec/swanctl/peer.j2
+++ b/data/templates/ipsec/swanctl/peer.j2
@@ -8,7 +8,7 @@
 {% if peer_conf.virtual_address is vyos_defined %}
         vips = {{ peer_conf.virtual_address | join(', ') }}
 {% endif %}
-        local_addrs = {{ peer_conf.local_address if peer_conf.local_address != 'any' else '%any' }} # dhcp:{{ peer_conf.dhcp_interface if 'dhcp_interface' in peer_conf else 'no' }}
+        local_addrs = {{ peer_conf.local_address if peer_conf.local_address != 'any' else '%any' }} # dhcp:{{ peer_conf.dhcp_interface if 'dhcp_interface' in peer_conf else 'no' }} reset:{{ name }}
         remote_addrs = {{ peer_conf.remote_address | join(",") if peer_conf.remote_address is vyos_defined and 'any' not in peer_conf.remote_address else '%any' }}
 {% if peer_conf.authentication.mode is vyos_defined('x509') %}
         send_cert = always
--- a/data/templates/ipsec/swanctl/peer.j2
+++ b/data/templates/ipsec/swanctl/peer.j2
@@ -8,7 +8,7 @@
 {% if peer_conf.virtual_address is vyos_defined %}
         vips = {{ peer_conf.virtual_address | join(', ') }}
 {% endif %}
-        local_addrs = {{ peer_conf.local_address if peer_conf.local_address != 'any' else '%any' }} # dhcp:{{ peer_conf.dhcp_interface if 'dhcp_interface' in peer_conf else 'no' }} reset:{{ name }}
+        local_addrs = {{ peer_conf.local_address if peer_conf.local_address != 'any' else '%any' }} # dhcp:{{ peer_conf.dhcp_interface if 'dhcp_interface' in peer_conf else 'no' }}
         remote_addrs = {{ peer_conf.remote_address | join(",") if peer_conf.remote_address is vyos_defined and 'any' not in peer_conf.remote_address else '%any' }}
 {% if peer_conf.authentication.mode is vyos_defined('x509') %}
         send_cert = always
--- a/src/etc/dhcp/dhclient-exit-hooks.d/99-ipsec-dhclient-hook
+++ b/src/etc/dhcp/dhclient-exit-hooks.d/99-ipsec-dhclient-hook
@@ -14,60 +14,32 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-DHCP_HOOK_IFLIST="/tmp/ipsec_dhcp_waiting"
+DHCP_HOOK_IFLIST="/tmp/ipsec_dhcp_interfaces"
 
-if [ -f $DHCP_HOOK_IFLIST ] && [ "$reason" == "BOUND" ]; then
-    if grep -qw $interface $DHCP_HOOK_IFLIST; then
-        sudo rm $DHCP_HOOK_IFLIST
-        sudo /usr/libexec/vyos/conf_mode/vpn_ipsec.py
-        exit 0
-    fi
+if ! { [ -f $DHCP_HOOK_IFLIST ] && grep -qw $interface $DHCP_HOOK_IFLIST; }; then
+    exit 0
 fi
 
+# Re-generate the config on the following events:
+# - BOUND: always re-generate
+# - RENEW: re-generate if the IP address changed
+# - REBIND: re-generate if the IP address changed
 if [ "$reason" == "RENEW" ] || [ "$reason" == "REBIND" ]; then
     if [ "$old_ip_address" == "$new_ip_address" ]; then
         exit 0
     fi
-else
+elif [ "$reason" != "BOUND" ]; then
     exit 0
 fi
 
-python3 - <<PYEND
-import os
-import re
-
-from vyos.utils.process import call
-from vyos.utils.process import cmd
-from vyos.utils.file import read_file
-from vyos.utils.file import write_file
-
-SWANCTL_CONF="/etc/swanctl/swanctl.conf"
+# Best effort wait for any active commit to finish
+sudo python3 - <<PYEND
+from vyos.utils.commit import wait_for_commit_lock
 
 if __name__ == '__main__':
-    interface = os.getenv('interface')
-    new_ip = os.getenv('new_ip_address')
-    old_ip = os.getenv('old_ip_address')
-
-    if os.path.exists(SWANCTL_CONF):
-        conf_lines = read_file(SWANCTL_CONF).split("\n")
-        found = False
-        reset_conns = set()
-        to_match = f'# dhcp:{interface}'
-
-        for i, line in enumerate(conf_lines):
-            if line.find(to_match) > 0:
-                conf_lines[i] = line.replace(old_ip, new_ip)
-                found = True
-                regex_match = re.search(r'#.* reset:([-_a-zA-Z0-9|@]+)', line)
-                if regex_match:
-                   connection_name = regex_match[1]
-                   reset_conns.add(connection_name)
-
-        if found:
-            write_file(SWANCTL_CONF, "\n".join(conf_lines))
-            for connection_name in reset_conns:
-                call(f'sudo swanctl -t -i {connection_name}')
-            call('sudo swanctl -q')
-
+    wait_for_commit_lock()
     exit(0)
 PYEND
+
+# Now re-generate the config
+sudo /usr/libexec/vyos/conf_mode/vpn_ipsec.py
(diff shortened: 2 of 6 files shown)


c-po:
What happens on dhcp renew when you receive a NEW IP address?

Can you test this please?

lucasec:
It's like you know where all the skeletons are buried ;)

I did run this on my lab system, and it seemed to persist an IP change, but after poking around with the hook script a bit more directly I must have just gotten lucky that some other change I was experimenting with caused the swanctl config to re-render.

That script is definitely quite broken. It appears it may even be broken all the way back to 1.3 too (due to `ipsec statusall` command outputting `peer-` instead of `peer_`, at least on 1.3.5 where I tested).

I pushed a fix here that should be valid for 1.4 and 1.5 under the current ticket (T5872). If you have another one open for this issue already let me know and I can re-word the second commit.

c-po:
Thanks for the proposed change in `src/etc/dhcp/dhclient-exit-hooks.d/99-ipsec-dhclient-hook`

I think it's not a good idea to mangle the service definition files underneath - @sarthurdev please correct me.

I would rather see if the dhclient lease changed, and if so - call `/usr/libexec/vyos/conf_mode/vpn_ipsec.py` so the entire cofig is regenerated from a single point to not leaf space for any arbitrary errors in scripts.

lucasec:
I agree the mangling feels like a bit of a hack left over from the 1.2 days. It would also remove the need for error-prone regular expressions (including the new "hack" I added to put the connection name on the same line as the DHCP interface comment to make it easy to find the right connection name to reset).

Would there be any race conditions if someone was mutating the config in config-mode at the time the hook executes?

edit to add: we already invoke the conf_mode script in the case of when the DHCP IP is not available at configure time, if we can do this for IP changes can also remove the need for the `/tmp/ipsec_dhcp_waiting` file.

c-po:
You can safe from such races by checking if a commit is in progress

https://github.com/vyos/vyos-1x/blob/current/python/vyos/utils/commit.py

sarthurdev:
> I think it's not a good idea to mangle the service definition files underneath - @sarthurdev please correct me.

Altering the generated files from the hook would likely be more performant than calling the conf script again, but does add complexity (linking interface/connection via comments in-file).

I agree that simplifying the hook to just call the conf script is likely an overall safer solution (with the commit in progress check).

lucasec:
The exit hook I pushed that does the mangling is proving to still be pretty broken, so added a few more fixes. In particular, the logic that tires to clear old connections/SAs is proving to be quite dangerous. If the IP actually changes, a new connection will come up after clearing the connection and reloading swanctl. But if the IP does not change, swanctl recognizes that the configuration is still the same when it is asked to reload so the connection stays in the "terminated" state.

It's not really clear to me that logic is still needed. At least for site-to-site peers, swanctl seems to reliably clear old SAs when it reloads and the configuration is changed (at least when using `connection-type: initiate`). For road warrior, existing child SAs stay up until they time out (unless the client can actually re-resolve and re-negotiate after the IP change, I'm still trying to test if any real-world clients have that behavior).

c-po:
To be honest - IKEv2 road warriors terminating at a dynamic endpoint with changing DHCP IPs and a ddns record needing an update can handle a bit of downtime, as also the DNS update requires some time.

IMHO this should not be overengineered.

Such a setup is nice for a lab but broken by design for production/enterprise use

lucasec:
@c-po Thanks for your patience while time to focus on this has been hard to come by over here. I just pushed up an update to the hook that dramatically simplifies it. Do take a look at my attempt at race condition handling though—I don't think it's 100% robust but hopefully better than nothing.

I deployed this on my system and am letting it soak for a bit. In general `swanctl`'s reload mechanism seems to work fine for site-to-site connections. If it detects a change it will tear down the old connection and re-create. For road-warrior configurations, it does not seem to tear down the old connections. Like you said though, that probably doesn't need to be over-engineered.

data/templates/ipsec/swanctl/peer.j2
```diff
@@ -8,7 +8,7 @@
 {% if peer_conf.virtual_address is vyos_defined %}
         vips = {{ peer_conf.virtual_address | join(', ') }}
 {% endif %}
-        local_addrs = {{ peer_conf.local_address if peer_conf.local_address != 'any' else '%any' }} # dhcp:{{ peer_conf.dhcp_interface if 'dhcp_interface' in peer_conf else 'no' }}
```
c-po: Any reason why removing this comment? It could make things easier while debugging swanctl file

src/etc/dhcp/dhclient-exit-hooks.d/99-ipsec-dhclient-hook
```diff
@@ -14,69 +14,32 @@
...
-            write_file(SWANCTL_CONF, conf_lines)
-            ipsec_down(old_ip)
-            call('sudo ipsec rereadall')
-            call('sudo ipsec reload')
-            call('sudo swanctl -q')
-
+    wait_for_commit_lock()
     exit(0)
 PYEND
+
```
c-po, line 43: I think you should `source /opt/vyatta/etc/functions/script-template` here.

https://docs.vyos.io/en/latest/automation/command-scripting.html

c-po (changes requested):

data/templates/ipsec/swanctl/peer.j2
lucasec: Removed because it no longer serves a functional purpose, but I agree it's helpful for debugging. Will restore.

src/etc/dhcp/dhclient-exit-hooks.d/99-ipsec-dhclient-hook
lucasec: I tested the hook by manually invoking it and it seemed to happily run without needing any of the aliases/functions in there. Unless there is something else in that file I am missing that is triggered solely by sourcing it.

c-po: Yeah, I've ran down that rabbit hole some months ago. Executing such a hook manually worked from your tty, but as soon it was launched in background by a daemon it was missing environment variables - that's why you should source that script and also test it by manually renewing your DHCP lease to trigger the execution of the script: `renew dhcp interface eth0`

lucasec: I was able to run a few tests on my machine and at least from what I can tell, as-is, the script executes fine even when called from dhclient.
 
I performed the following manual test:
- Modify the swanctl.conf file by replacing the current IP with a fake IP
- Modify the dhclient hook to not short-circuit `exit 0` when old_ip == new_ip.
- `renew dhcp interface eth0`
- Confirmed `swanctl.conf` was re-rendered, had acceptable owner `root:vyattacfg`, and that logs showed swanctl performing a reload

lucasec: Are you sure we're talking about the same file to source? The `/opt/vyatta/etc/functions/script-template` on my build only contains bash functions/aliases, which wouldn't survive a `sudo` regardless

c-po: If all your tests passed thats very good - then ignore my paranoia.

c-po (approved):

lucasec:
`ignore my paranoia`
famous last words ;)

That being said, I think we've shaken out all the issues here. Also going to add the "bug fix" checkbox in the description since that hook re-write also fixes the functionality for site-to-site users, which I guess none of us (myself included) realized was broken before...

lucasec:
@c-po Is this clear to merge or are we waiting on any further input?
Also, thoughts on backporting to 1.4?

c-po:
> @c-po Is this clear to merge or are we waiting on any further input? Also, thoughts on backporting to 1.4?

I‘m waiting on a review from @sarthurdev, after that is done we can backport this of course

lucasec:
While we wait for review on this, may be worth taking a look at https://vyos.dev/T6177 which I just filed.
I am going to try temporarily changing the DHCP hook to a no-op to see if that increases the time between crashes (I do not believe they are related as the hook only reloads strongswan if the IP actually changes, which I do not believe has happened). Currently the issue seems to reproduce roughly once every 7 days.

sarthurdev (approved):
LGTM

dmbaturin:
@mergifyio backport sagitta

lucasec:
> I am going to try temporarily changing the DHCP hook to a no-op to see if that increases the time between crashes (I do not believe they are related as the hook only reloads strongswan if the IP actually changes, which I do not believe has happened). Currently the issue seems to reproduce roughly once every 7 days.

Tested this and still experienced a crash consistent with https://vyos.dev/T6177. So that regression was not introduced here. Continuing to investigate and will add further findings to that ticket.

PR #2689, opened 2023-12-25
ethernet: T5566: disable energy efficient ethernet (EEE) for interfaces

c-po:
## Change Summary

VyOS is a routing (packet pushing) platform, thus supporting EEE which potentially causes issues is not a good idea. Some recent Intel drivers enable EEE by default, thus we will disable this for every NIC supporting EEE.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes)
- [ ] Migration from an old Vyatta component to vyos-1x, please link to related PR inside obsoleted component
- [ ] Other (please describe):

## Related Task(s)

* https://vyos.dev/T5311
* https://vyos.dev/T5566

## Component(s) name

Ethernet

## Proposed changes

Always disable EEE via ethtool

## How to test

This has been developed by injecting EEE output provided by @zdc to ethtool.py in lack of hardware.

```console
$ sudo ethtool --show-eee eno1 
EEE settings for eno1:
	EEE status: enabled - active
	Tx LPI: 17 (us)
	Supported EEE link modes:  100baseT/Full
	                           1000baseT/Full
	Advertised EEE link modes:  100baseT/Full
	                            1000baseT/Full
	Link partner advertised EEE link modes:  100baseT/Full
	                                         1000baseT/Full
$ sudo ethtool --set-eee eno1 eee off
$ sudo ethtool --show-eee eno1 
EEE settings for eno1:
	EEE status: disabled
	Tx LPI: 17 (us)
	Supported EEE link modes:  100baseT/Full
	                           1000baseT/Full
	Advertised EEE link modes:  100baseT/Full
	                            1000baseT/Full
	Link partner advertised EEE link modes:  100baseT/Full
	                                         1000baseT/Full
$ sudo ethtool --set-eee eno1 eee on
$ sudo ethtool --show-eee eno1 
EEE settings for eno1:
	EEE status: enabled - inactive
	Tx LPI: 17 (us)
	Supported EEE link modes:  100baseT/Full
	                           1000baseT/Full
	Advertised EEE link modes:  100baseT/Full
	                            1000baseT/Full
	Link partner advertised EEE link modes:  Not reported
(some time later...)
$ sudo ethtool --show-eee eno1 
EEE settings for eno1:
	EEE status: enabled - active
	Tx LPI: 17 (us)
	Supported EEE link modes:  100baseT/Full
[...]

PR #3052, opened 2024-02-26
vyos.ethtool: T6070: fix EEE reading the incorrect status line

up-n-atom:
EEE enabled status is on the 2nd line of ethtool output and not the 3rd. Subsequently, reading the 3rd line was causing an out-of-bounds access for the bnx2x driver as well.

## Types of changes

## Related Task(s)

* https://vyos.dev/T6070

## Related PR(s)

N/A

## Component(s) name

ethernet

## Proposed changes

Same as commit message

## How to test

```
ethtool --show-eee eth0
```

## Smoketest result

N/A

## Checklist:

### DIFF
--- a/python/vyos/ethtool.py
+++ b/python/vyos/ethtool.py
@@ -150,7 +150,7 @@ class Ethtool:
             self._eee = True
             # read current EEE setting, this returns:
             # EEE status: disabled || EEE status: enabled - inactive || EEE status: enabled - active
-            self._eee_enabled = bool('enabled' in out.splitlines()[2])
+            self._eee_enabled = bool('enabled' in out.splitlines()[1])
 
     def check_auto_negotiation_supported(self):
         """ Check if the NIC supports changing auto-negotiation """


c-po:
I wonder why ghe artifical data used while testing worked out?

https://github.com/vyos/vyos-1x/pull/2689

Any idea? I fed back the ethtool and hardcoded the result for validation testing, as I do not have an EEE capable NIC

up-n-atom:
It certainly wouldn't error out with those tests cases because it would be reading the `Tx LPI: ...` line but should have always returned False or disabled.

Were you possibly including the command into the output? ie. `$ sudo ethtool --show-eee eno1`

dmbaturin (changes requested):
I have two (plus one) concerns about all such code.

1. Without an example of the output in question, it's completely impossible to judge the code when reading it unless you remember the output by heart (unlikely) or have a chance to easily test the command (also not always possible).
I believe we must always include examples.

2. Do you think we can generalize that code to avoid hardcoded line numbers? That would be perfect because we never know how it may change.

3. With `ethtool` specifically: is there a specific reason why we still parse its human-readable output instead of using `ethtool --json`? Can we start doing it from this case?

up-n-atom:
As a first time contributor, I agree fully but I'm definitely not the person to go about that. From my perspective, I was actually quite surprised to witness output parsing rather than netlink being utilized for such purposes but that is neither here nor there.

dmbaturin:
I don't mind merging the emergency fix now, but I wanted to draw attention to the issue: I think we all should work together to switch to `--json` here, since it will greatly simplify the script.

c-po:
> I don't mind merging the emergency fix now, but I wanted to draw attention to the issue: I think we all should work together to switch to `--json` here, since it will greatly simplify the script.

Unfortunately `ethtool` is not feature on par when using JSON output. It‘s only implemented for a subset of commands.

https://git.kernel.org/pub/scm/network/ethtool/ethtool.git/tree/ethtool.c and search for `.json` - maybe we should spend time in adding missing JSON support at least for the features we use.

c-po (approved):

c-po:
Reviewed the original PR and output used for testing, seems like an off-by-one error in the initial implementation.

We should come up with a JSON patch for EEE to ethtool and until it‘s upstreamed can use our own build.

c-po:
@MergifyIo backport sagitta

PR #3218, opened 2024-03-30
accel-ppp: T6187: use correct CPU counts adjusted for SMT

dmbaturin:
## Change Summary

Use `vyos.cpu.get_core_count()` to get the correct number of physical cores instead of halving the number of apparent logical cores — halving `os.cpu_count()` makes the system appear to have fewer CPUs than it really has if SMT is disabled.

## Types of changes

- [x] Other (please describe): optimization.

## Component(s) name

Mainly accel-ppp

## Checklist:

### DIFF
--- a/python/vyos/configdict.py
+++ b/python/vyos/configdict.py
@@ -633,7 +633,7 @@ def get_accel_dict(config, base, chap_secrets, with_pki=False):
 
     Return a dictionary with the necessary interface config keys.
     """
-    from vyos.utils.system import get_half_cpus
+    from vyos.cpu import get_core_count
     from vyos.template import is_ipv4
 
     dict = config.get_config_dict(base, key_mangling=('-', '_'),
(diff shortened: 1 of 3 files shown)


Apachez-:
Something that not just affect accel-ppp is the use of E- and P-cores (mainly for Intel CPU's but I guess its a matter of time before AMD will also do that).

Can E-cores be avoided (by configuration) or should they be counted as a 0.5 core or such?

c-po (approved):

c-po:
> Something that not just affect accel-ppp is the use of E- and P-cores (mainly for Intel CPU's but I guess its a matter of time before AMD will also do that).
> 
> Can E-cores be avoided (by configuration) or should they be counted as a 0.5 core or such?

It's the users choice and thus responsibility if you use a system with E and P cores or not. Introducing a factor feels wrong - how should it be? Is 0.5 the right value? What would happen e.g. with 20th Gen processors, will the factor be 0.66 then? We should count them all as one and hand this one off to the user.

Apachez-:
The factor could be part of the configuration to make it forward compatible.

Like having an "e-core_weight" to default to lets say 0.5 and a "p-core_weight" to default to 1.0.

dmbaturin:
@mergifyio backport sagitta

PR #3340, opened 2024-04-21
T6255: static-routing: don't render whitespace from static table descriptions

Embezzle:
## Change Summary

When rendering static-routing tables in `static.conf.j2` replace whitespace from table descriptions with underscores.

## Types of changes

## Related Task(s)

[https://vyos.dev/T6255](https://vyos.dev/T6255)

## Component(s) name
protocol -> static-routing

## How to test

1. Set a static table description containing whitespace:
```
[edit]
vyos@vyos# set protocols static table 100 description "table name with spaces"
[edit]
vyos@vyos# set protocols static table 100 route 192.168.1.1/32 blackhole
[edit]
vyos@vyos# commit ; save
[edit]
vyos@vyos# exit
```

2. Validate in rendered configuration file  whitespace has been replaced with underscores:
```
vyos@vyos:~$ cat /etc/iproute2/rt_tables.d/vyos-static.conf
# Generated by VyOS (protocols_static.py), do not edit by hand
100    table_name_with_spaces
```

3. Validate table is accessible via alias/description:
```
vyos@vyos:~$ sudo ip route show table table_name_with_spaces
blackhole 192.168.1.1 proto static metric 20
```

## Checklist:

### DIFF
--- a/data/templates/iproute2/static.conf.j2
+++ b/data/templates/iproute2/static.conf.j2
@@ -2,7 +2,7 @@
 {% if table is vyos_defined %}
 {%     for t, t_options in table.items() %}
 {%         if t_options.description is vyos_defined %}
-{{ "%-6s" | format(t) }} {{ "%-40s" | format(t_options.description) }}
+{{ "%-6s" | format(t) }} {{ "%-40s" | format(t_options.description | replace(" ", "_")) }}
 {%         endif %}
 {%     endfor %}
 {% endif %}


dmbaturin (changes requested):
Since whitespace used to be allowed and didn't cause commit failures, this will require a migration script if we are to merge it.

Embezzle:
> Since whitespace used to be allowed and didn't cause commit failures, this will require a migration script if we are to merge it.

I have added a migration script to replace whitespace in existing table descriptions with underscores.

interface-definitions/include/version/static-routing-version.xml.i
```diff
@@ -0,0 +1,3 @@
+<!-- include start from include/version/static-routing-version.xml.i -->
+<syntaxVersion component='static-routing' version='1'></syntaxVersion>
```
c-po: I wonder if we should introduce a new component here or rely on the good old fashioned `quagga-version` here.

c-po:
I see the main reason for the description use-case at the moment is to alias a routing table by name instead of it's table ID. There is no VyOS CLI for retrieving the routing table with a given alias.

Using the alias can only by done by `sudo ip route show table <description alias>`. Why not keep it simple: When rendering the description into `/etc/iproute2/rt_tables.d/vyos-static.conf` replace whitespace with `_`. If we really wan't to make use of mnemonics we should probably add a proper `alias` CLI node making explicit use of this feature.

Embezzle:
> I see the main reason for the description use-case at the moment is to alias a routing table by name instead of it's table ID. There is no VyOS CLI for retrieving the routing table with a given alias.
> 
> Using the alias can only by done by `sudo ip route show table <description alias>`. Why not keep it simple: When rendering the description into `/etc/iproute2/rt_tables.d/vyos-static.conf` replace whitespace with `_`. If we really wan't to make use of mnemonics we should probably add a proper `alias` CLI node making explicit use of this feature.

Thank you @c-po for the feedback, looks like I have been approaching this from the wrong angle!
As requested, I have editing my commit to just replace whitespace in rendering & have updated the PR description to reflect.

c-po (approved):

c-po:
@MergifyIo backport sagitta

PR #3371, opened 2024-04-28
openconnect: T4982: Support defining minimum TLS version in openconnect VPN

Embezzle:
## Change Summary

Allow configuration of minimum acceptable TLS version for openconnect VPN. 
Default is set at TLSv1.2 to ensure out-of-box/unconfigured option is not insecure.

## Types of changes

## Related Task(s)

[https://vyos.dev/T4982](https://vyos.dev/T4982)

## Component(s) name
vpn -> openconnect

## How to test

1. Create an openconnect VPN configuration:
```
set vpn openconnect authentication local-users username example password 'test'
set vpn openconnect authentication mode local 'password'
set vpn openconnect network-settings client-ip-settings subnet '192.168.1.1/30'
set vpn openconnect network-settings name-server '192.168.1.254'
set vpn openconnect ssl ca-certificate 'test-ca'
set vpn openconnect ssl certificate 'test-certificate'
set vpn openconnect tls-version-min 1.2
```

2. Validate in the configuration file rendered that minimum TLS version has been set correctly: 
```
vyos@vyos:~$ cat /var/run/ocserv/ocserv.conf | grep "tls-priorities"
tls-priorities = "NORMAL:%SERVER_PRECEDENCE:%COMPAT:-RSA:-VERS-SSL3.0:-ARCFOUR-128:-VERS-TLS1.0:-VERS-TLS1.1"
```

## Smoketest result

```
vyos@vyos:~$ /usr/libexec/vyos/tests/smoke/cli/test_vpn_openconnect.py
test_ocserv (__main__.TestVPNOpenConnect.test_ocserv) ...
SSL missing on OpenConnect config!

ok

Ran 1 test in 8.705s

OK
```

## Checklist:

### DIFF
--- a/src/migration-scripts/openconnect/2-to-3
+++ b/src/migration-scripts/openconnect/2-to-3
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) 2024 VyOS maintainers and contributors
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or later as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# T4982: Retain prior default TLS version (v1.0) when upgrading installations with existing openconnect configurations
+
+import sys
+
+from vyos.configtree import ConfigTree
+
+if len(sys.argv) < 2:
+    print("Must specify file name!")
+    sys.exit(1)
+
+file_name = sys.argv[1]
+
+with open(file_name, 'r') as f:
+    config_file = f.read()
+
+
+config = ConfigTree(config_file)
+cfg_base = ['vpn', 'openconnect']
+
+# bail out early if service is unconfigured
+if not config.exists(cfg_base):
+    sys.exit(0)
+
+# new default is TLS 1.2 - set explicit old default value of TLS 1.0 for upgraded configurations to keep compatibility
+tls_min_path = cfg_base + ['tls-version-min']
+if not config.exists(tls_min_path):
+    config.set(tls_min_path, value='1.0')
+
+try:
+    with open(file_name, 'w') as f:
+        f.write(config.to_string())
+except OSError as e:
+    print("Failed to save the modified config: {}".format(e))
+    sys.exit(1)
--- a/data/templates/ocserv/ocserv_config.j2
+++ b/data/templates/ocserv/ocserv_config.j2
@@ -61,7 +61,15 @@ keepalive = 300
 dpd = 60
 mobile-dpd = 300
 switch-to-tcp-timeout = 30
+{% if tls_version_min == '1.0' %}
 tls-priorities = "NORMAL:%SERVER_PRECEDENCE:%COMPAT:-RSA:-VERS-SSL3.0:-ARCFOUR-128"
+{% elif tls_version_min == '1.1' %}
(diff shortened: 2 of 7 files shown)


c-po (changes requested):
Hi @Embezzle,

thank you for the contribution. I have some minor comments.

`tls-version-min` is already defined for OpenVPN interfaces at https://github.com/vyos/vyos-1x/blob/current/interface-definitions/interfaces_openvpn.xml.in#L742 - please create a new file interface-definitions/include/tls-version-min.xml.i which can then be included by both OpenVPN and OpenConnect. This saves redundant lines and makes future maintenance easier.

In addition you define that the default TLS version will be 1.2 - this breaks backward compatibility as the current default TLS version is 1.0. You will either need to downgrade the default to 1.0 or add a migration script which explicitly defined tls-version-min to be set at 1.0 if it‘s not defined in the old configurations. As VyOS 1.4.0-epa3 is yet not released we can add this change to lift the default TLS version.

Embezzle:
Thank you for the feedback @c-po, I have amended the PR with your requested changes.

The OpenVPN interface implementation of **tls-version-min** did not previously have a `<defaultValue>` tag, so I have had to make a minor tweak in `interfaces_openvpn.py` to account for this.

I do not think this requires a migration script for OpenVPN, as the default minimum version supported if not explicitly configured by VyOS, as was the case before this PR, is also TLSv1.2.

https://openvpn.net/community-resources/reference-manual-for-openvpn-2-6 _(see section for: --tls-version-min)_

c-po:
Hi @Embezzle,

no need for any changes on OpenVPN code. Please remove this again as it should happen in a different PR if needed.

Our `defaultValue` system works as follows:

OpenVPN will only have:
```
#include <include/tls-version-min.xml.i>
```

Whereas OpenConnect will have:

```
#include <include/tls-version-min.xml.i>
<leafNode name="tls-version-min">
    <defaultValue>1.2</defaultValue>
</leafNode>
```

Meaning that one can overload the default of a previous XML node definition

src/migration-scripts/openconnect/2-to-3
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) 2022 VyOS maintainers and contributors
```
c-po: 2024 ;)

src/migration-scripts/openconnect/2-to-3
```diff
@@ -0,0 +1,43 @@
...
+
+file_name = sys.argv[1]
+
+with open(file_name, 'r') as f:
+    config_file = f.read()
+
+config = ConfigTree(config_file)
+cfg_base = ['vpn', 'openconnect']
+
+if config.exists(cfg_base):
```
c-po: You're logic should rather be:

```python3
config = ConfigTree(config_file)
cfg_base = ['vpn', 'openconnect']

# bail out early if service is unconfigured
if not config.exists(cfg_base):
    sys.exit(0)

# new default is TLS 1.2 - set explicit old default value of TLS 1.0 for upgraded configurations to keep compatibility
tls_min_path = cfg_base + ['tls-version-min']
if not config.exists(tls_min_path):
    config.set(tls_min_path, value='1.0')

try:
    with open(file_name, 'w') as f:
        f.write(config.to_string())
except OSError as e:
    print("Failed to save the modified config: {}".format(e))
    sys.exit(1)
```

c-po (changes requested):

Embezzle:
Hi @c-po,

I did not realise you could append tags to includes in the XML definitions like shown, that makes much more sense!
Thank you for your help, I have amended my commit to correct issues you raised & have re-run smoke tests.

src/migration-scripts/openconnect/2-to-3
Embezzle:
```python
if not config.exists(tls_min_path):
```
No idea how I managed to forget this check! 😲

c-po (approved):

c-po:
@MergifyIo backport sagitta

sever-sever (approved):

PR #3616, opened 2024-06-10
firewall: T4694: Adding rt ipsec exists/missing match to firewall configs

talmakion:
## Change Summary

Current firewall syntax only permits inbound matching of IPsec, using "meta ipsec", which indicates a packet was decrypted from IPsec encap.

This extends firewall capability to also be able to match outbound traffic using "rt ipsec", which indicates a packet will be encrypted into an IPsec encap. 

"meta ipsec" is not valid (nft will refuse to load config) inside output hooks so additional care has been taken to ensure it doesn't end up in output or any inappropriate jump targets from an output chain. 

In doing these checks, I've also added an infinite jump-target loop check to the validation pass. 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes)
- [ ] Migration from an old Vyatta component to vyos-1x, please link to related PR inside obsoleted component
- [x] Other (please describe): Changes to existing firewall ipsec match syntax

## Related Task(s)

* https://vyos.dev/T4694
* https://vyos.dev/T4667

## Related PR(s)

* https://github.com/vyos/vyos-1x/pull/3637

## Component(s) name

* firewall

## Proposed changes

I've covered some concerns I have about my implementation in the forum: https://forum.vyos.io/t/outbound-ipsec-filtering-by-firewall-would-like-some-dev-opinions/14710.

 * Change ipsec match-ipsec/none to match-ipsec-in and match-none-in for fw rules
 * Add ipsec match-ipsec-out and match-none-out
   * *-in means "decrypted" and *-out means "encrypting", the names may benefit from tweaking to make intent clearer
 * Change all the points where the match-ipsec.xml.i include was used before, making sure the new includes (match-ipsec-in/out.xml.i) are used appropriately. There were a handful of spots where match-ipsec.xml.i had snuck back in for output hooked chains already (the common-rule-* includes)
 * Add the -out generators to rendered templates
 * Heavy modification to firewall config validators:
   * I needed to check for ipsec-in matches no matter how deeply nested under an output-hook chain(via jump-target) - this always generates an error.
   * Ended up retrofitting the jump-targets validator from root chains and for named custom chains. It checks for recursive loops and bad IPsec. However, it could be improved - it will not detect a cycle between 2 otherwise un-referenced named chains.
[...]

PR #3637, opened 2024-06-11
firewall: T4694: Adding GRE flags & fields matches to firewall rules

talmakion:
## Change Summary

Allow firewall rules to match flags and fields in GRE protocol headers, primarily, the "key" field. 

Adding this will allow firewall rules to distinguish individual GRE sessions and apply appropriate policy, such as blocking sending or forwarding certain tunnels in the clear if IPsec is down. 

## Types of changes

## Related Task(s)

* https://vyos.dev/T4694
* https://vyos.dev/T4667

## Related PR(s)

* https://github.com/vyos/vyos-1x/pull/3616

## Component(s) name

* firewall

## Proposed changes

I've covered some concerns I have about my implementation in the forum: https://forum.vyos.io/t/outbound-ipsec-filtering-by-firewall-would-like-some-dev-opinions/14710.

* Only matching flags and fields used by modern RFC2890 "extended GRE" - this is backwards-compatible, but does not match all possible flags. 
* There are no nftables helpers for the GRE key field, which is critical to match individual tunnel sessions (more detail in the forum post)
  * nft expression syntax is not flexible enough for multiple field matches in a single rule and the key offset changes depending on flags.
  * Thus, clumsy compromise in requiring an explicit match on the "checksum" flag if a key is present, so we know where key will be. In most cases, nobody uses the checksum, but assuming it to be off or automatically adding a "not checksum" match unless told otherwise would be confusing
  * The automatic "flags key" check when specifying a key doesn't have similar validation, I added it first and it makes sense. I would still like to find a workaround to the "checksum" offset problem.
  * If we could add 2 rules from 1 config definition, we could match both cases with appropriate offsets, but this would break existing FW generation logic, logging, etc.
* Added a "gre-protocol" validator for the fields we can pass to nft's gre matches.
  * The protocol names are out of synch with other parts of the firewall def, but for eg, I can't call a completion+valueHelp "ipv6" without VyOS deciding to show an IPv6 pattern instead of my help text.
  * I've allowed arbitrary radix numbers for the ethertype, for eg, it's common to use 0x8100 for .1q instead of 33024. nft should accept these as well.

## How to test

* Configured VyOS testing pair in a "triangle" (LEFT -> FAKEWWW <- RIGHT) 
* 2 additional VMs behind LEFT and RIGHT for forward chain testing
* Applied numerous combinations of gre/ip6gre matches in logging/nflogging rules across all filter chains
* Created GRE/GRETAP point to point and thru-tunnels between LEFT and RIGHT and the VMs behind them
* Checking dmesg output and tcpdump on FAKEWWW for expected operation in hitting all the right rules

Starting firewall config (note ethtype 0x6558 is gretap):
```
firewall {
    ipv4 {
        forward {
            filter {
                rule 100 {
                    action jump
                    jump-target TEST-LOG-UNKEYED
                }
                rule 110 {
                    action jump
                    jump-target TEST-LOG-KEYED
                }
                rule 120 {
                    action jump
                    jump-target TEST-LOG-CSUMMED
                }
                rule 130 {
                    action jump
                    jump-target TEST-LOG-ETHTYPE
                }
            }
        }
        input {
            filter {
                rule 100 {
                    action jump
                    jump-target TEST-LOG-UNKEYED
                }
                rule 110 {
                    action jump
                    jump-target TEST-LOG-KEYED
                }
                rule 120 {
                    action jump
                    jump-target TEST-LOG-CSUMMED
                }
                rule 130 {
                    action jump
                    jump-target TEST-LOG-ETHTYPE
                }
            }
        }
        name TEST-LOG-CSUMMED {
            default-action return
            rule 10 {
                action continue
                gre {
                    flags {
                        checksum
                    }
                }
                log
                protocol gre
            }
        }
        name TEST-LOG-ETHTYPE {
            default-action return
            rule 10 {
                action continue
                gre {
                    inner-proto ip
                }
                log
                protocol gre
            }
            rule 20 {
                action continue
                gre {
                    inner-proto 0x6558
                }
                log
                protocol gre
            }
        }
        name TEST-LOG-KEYED {
            default-action return
            rule 10 {
                action continue
                gre {
                    flags {
                        key
                        not {
                            checksum
                        }
                    }
                    key 1234
                }
                log
                protocol gre
            }
            rule 20 {
                action continue
                gre {
                    flags {
                        key
                        not {
                            checksum
                        }
                    }
                    key 4321
                }
                log
                protocol gre
            }
            rule 30 {
                action continue
                gre {
                    flags {
                        key
                        not {
                            checksum
                        }
                    }
                }
                log
                protocol gre
            }
        }
        name TEST-LOG-UNKEYED {
            default-action return
            rule 10 {
                action continue
                gre {
                    flags {
                        not {
                            key
                        }
                    }
                }
                log
                protocol gre
            }
        }
        output {
            filter {
                rule 100 {
                    action jump
                    jump-target TEST-LOG-UNKEYED
                }
                rule 110 {
                    action jump
                    jump-target TEST-LOG-KEYED
                }
                rule 120 {
                    action jump
                    jump-target TEST-LOG-CSUMMED
                }
                rule 130 {
                    action jump
                    jump-target TEST-LOG-ETHTYPE
                }
            }
        }
        prerouting {
            raw {
                rule 100 {
                    action jump
                    jump-target TEST-LOG-UNKEYED
                }
                rule 110 {
                    action jump
                    jump-target TEST-LOG-KEYED
                }
                rule 120 {
                    action jump
                    jump-target TEST-LOG-CSUMMED
                }
                rule 130 {
                    action jump
                    jump-target TEST-LOG-ETHTYPE
                }
            }
        }
    }
    ipv6 {
        forward {
            filter {
                rule 100 {
                    action jump
                    jump-target TEST6-LOG-UNKEYED
                }
                rule 110 {
                    action jump
                    jump-target TEST6-LOG-KEYED
                }
                rule 120 {
                    action jump
                    jump-target TEST6-LOG-CSUMMED
                }
                rule 130 {
                    action jump
                    jump-target TEST6-LOG-ETHTYPE
                }
            }
        }
        input {
            filter {
                rule 100 {
                    action jump
                    jump-target TEST6-LOG-UNKEYED
                }
                rule 110 {
                    action jump
                    jump-target TEST6-LOG-KEYED
                }
                rule 120 {
                    action jump
                    jump-target TEST6-LOG-CSUMMED
                }
                rule 130 {
                    action jump
                    jump-target TEST6-LOG-ETHTYPE
                }
            }
        }
        name TEST6-LOG-CSUMMED {
            default-action return
            rule 10 {
                action continue
                gre {
                    flags {
                        checksum
                    }
                }
                log
                protocol gre
            }
        }
        name TEST6-LOG-ETHTYPE {
            default-action return
            rule 10 {
                action continue
                gre {
                    inner-proto ip
                }
                log
                protocol gre
            }
            rule 20 {
                action continue
                gre {
                    inner-proto 0x6558
                }
                log
                protocol gre
            }
        }
        name TEST6-LOG-KEYED {
            default-action return
            rule 10 {
                action continue
                gre {
                    flags {
                        key
                        not {
                            checksum
                        }
                    }
                    key 1234
                }
                log
                protocol gre
            }
            rule 20 {
                action continue
                gre {
                    flags {
                        key
                        not {
                            checksum
                        }
                    }
                    key 4321
                }
                log
                protocol gre
            }
            rule 30 {
                action continue
                gre {
                    flags {
                        key
                        not {
                            checksum
                        }
                    }
                }
                log
                protocol gre
            }
        }
        name TEST6-LOG-UNKEYED {
            default-action return
            rule 10 {
                action continue
                gre {
                    flags {
                        not {
                            key
                        }
                    }
                }
                log
                protocol gre
            }
        }
        output {
            filter {
                rule 100 {
                    action jump
                    jump-target TEST6-LOG-UNKEYED
                }
                rule 110 {
                    action jump
                    jump-target TEST6-LOG-KEYED
                }
                rule 120 {
                    action jump
                    jump-target TEST6-LOG-CSUMMED
                }
                rule 130 {
                    action jump
                    jump-target TEST6-LOG-ETHTYPE
                }
            }
        }
        prerouting {
            raw {
                rule 100 {
                    action jump
                    jump-target TEST6-LOG-UNKEYED
                }
                rule 110 {
                    action jump
                    jump-target TEST6-LOG-KEYED
                }
                rule 120 {
                    action jump
                    jump-target TEST6-LOG-CSUMMED
                }
                rule 130 {
                    action jump
                    jump-target TEST6-LOG-ETHTYPE
                }
            }
        }
    }
}
```

This was a starting point, specific rules were fed directly into top level chains for testing outside of custom named chains and more yes/no logging firewall pairs were added to cover as many combinations of attributes & flags as possible. 

The validators were also given exercise to make sure invalid flag/attribute combos weren't possible. 

## Smoketest result

Note, the groups failure below is also encountered in a clean rolling image:
```
# python3 /usr/libexec/vyos/tests/smoke/cli/test_firewall.py 
test_bridge_basic_rules (__main__.TestFirewall.test_bridge_basic_rules) ... ok
test_groups (__main__.TestFirewall.test_groups) ... FAIL

FAIL: test_groups (__main__.TestFirewall.test_groups)
  File "/usr/libexec/vyos/tests/smoke/cli/test_firewall.py", line 152, in test_groups
    self.verify_nftables(nftables_search, 'ip vyos_filter')
  File "/usr/libexec/vyos/tests/smoke/cli/base_vyostest_shim.py", line 122, in verify_nftables
    self.assertTrue(not matched if inverse else matched, msg=search)
AssertionError: False is not true : ['elements = { 192.0.2.5, 192.0.2.8,']

Ran 21 tests in 184.628s

FAILED (failures=1)
```

## Checklist:

### DIFF
--- a/interface-definitions/include/firewall/gre.xml.i
+++ b/interface-definitions/include/firewall/gre.xml.i
@@ -1,0 +1,116 @@
+<!-- include start from firewall/gre.xml.i -->
+<node name="gre">
+  <properties>
+    <help>GRE fields to match</help>
+  </properties>
+  <children>
+    <node name="flags">
+      <properties>
+        <help>GRE flag bits to match</help>
+      </properties>
+      <children>
+        <node name="key">
+          <properties>
+            <help>Header includes optional key field</help>
+          </properties>
+          <children>
+            <leafNode name="unset">
+              <properties>
+                <help>Header does not include optional key field</help>
+                <valueless/>
+              </properties>
+            </leafNode>
+          </children>
+        </node>
+        <node name="checksum">
+          <properties>
+            <help>Header includes optional checksum</help>
+          </properties>
+          <children>
+            <leafNode name="unset">
+              <properties>
+                <help>Header does not include optional checksum</help>
+                <valueless/>
+              </properties>
+            </leafNode>
+          </children>            
+        </node>
+        <node name="sequence">
+          <properties>
+            <help>Header includes a sequence number field</help>
+          </properties>
+          <children>
+            <leafNode name="unset">
+              <properties>
+                <help>Header does not include a sequence number field</help>
+                <valueless/>
+              </properties>
+            </leafNode>
+          </children>
+        </node>
+      </children>
+    </node>
+    <leafNode name="inner-proto">
+      <properties>
+        <help>EtherType of encapsulated packet</help>
+        <completionHelp>
+          <list>ip ip6 arp 802.1q 802.1ad</list>
+        </completionHelp>
+        <valueHelp>
+          <format>u32:0-65535</format>
+          <description>Ethernet protocol number</description>
+        </valueHelp>
+        <valueHelp>
+          <format>u32:0x0-0xffff</format>
+          <description>Ethernet protocol number (hex)</description>
+        </valueHelp>
+        <valueHelp>
+          <format>ip</format>
+          <description>IPv4</description>
+        </valueHelp>
+        <valueHelp>
+          <format>ip6</format>
+          <description>IPv6</description>
+        </valueHelp>
+        <valueHelp>
+          <format>arp</format>
+          <description>Address Resolution Protocol</description>
+        </valueHelp>
+        <valueHelp>
+          <format>802.1q</format>
+          <description>VLAN-tagged frames (IEEE 802.1q)</description>
+        </valueHelp>
+        <valueHelp>
+          <format>802.1ad</format>
+          <description>Provider Bridging (IEEE 802.1ad, Q-in-Q)</description>
+        </valueHelp>
+        <valueHelp>
+          <format>gretap</format>
+          <description>Transparent Ethernet Bridging (L2 Ethernet over GRE, gretap)</description>
+        </valueHelp>
+        <constraint>
+          <regex>(ip|ip6|arp|802.1q|802.1ad|gretap|0x[0-9a-fA-F]{1,4})</regex>
+          <validator name="numeric" argument="--range 0-65535"/>
+        </constraint>
+      </properties>
+    </leafNode>
+    #include <interface/parameters-key.xml.i>
+    <leafNode name="version">
+      <properties>
+        <help>GRE Version</help>
+        <valueHelp>
+          <format>gre</format>
+          <description>Standard GRE</description>
+        </valueHelp>
+        <valueHelp>
+          <format>pptp</format>
+          <description>Point to Point Tunnelling Protocol</description>
+        </valueHelp>
+        <constraint>
+          <regex>(gre|pptp)</regex>
+        </constraint>
+      </properties>
+    </leafNode>
+  </children>
+</node>
+<!-- include end -->
--- a/src/conf_mode/firewall.py
+++ b/src/conf_mode/firewall.py
@@ -231,6 +231,36 @@ def verify_rule(firewall, family, hook, priority, rule_id, rule_conf):
         if not {'count', 'time'} <= set(rule_conf['recent']):
             raise ConfigError('Recent "count" and "time" values must be defined')
 
+    if 'gre' in rule_conf:
+        if dict_search_args(rule_conf, 'protocol') != 'gre':
+            raise ConfigError('Protocol must be gre when matching GRE flags and fields')
+
+        if dict_search_args(rule_conf, 'gre', 'key'):
+            if dict_search_args(rule_conf, 'gre', 'version') == 'pptp':
+                raise ConfigError('GRE tunnel keys are not present in PPTP')
+
+            if dict_search_args(rule_conf, 'gre', 'flags', 'checksum') is None:
+                # There is no builtin match in nftables for the GRE key, so we need to do a raw lookup.
+                # The offset of the key within the packet shifts depending on the C-flag. 
+                # 99% of the time, nobody will have checksums enabled - it's usually a manual config option. 
+                # We can either assume it is unset unless otherwise directed 
+                # (confusing, requires doco to explain why it doesn't work sometimes)
+                # or, demand an explicit selection to be made for this specific match rule. 
+                # This check enforces the latter. The user is free to create rules for both cases. 
+                raise ConfigError('Matching GRE tunnel key requires an explicit checksum flag match. For most cases, use "gre flags checksum unset"')
+
+            if dict_search_args(rule_conf, 'gre', 'flags', 'key', 'unset') is not None:
+                raise ConfigError('Matching GRE tunnel key implies "flags key", cannot specify "flags key unset"')
+
+        gre_inner_proto = dict_search_args(rule_conf, 'gre', 'inner_proto')
+        if gre_inner_proto is not None:
+            try:
+                gre_inner_value = int(gre_inner_proto, 0)
+                if gre_inner_value < 0 or gre_inner_value > 65535:
+                    raise ConfigError('inner-proto outside valid ethertype range 0-65535')
+            except ValueError:
+                pass # Symbolic constant, pre-validated before reaching here. 
+
     tcp_flags = dict_search_args(rule_conf, 'tcp', 'flags')
     if tcp_flags:
         if dict_search_args(rule_conf, 'protocol') != 'tcp':
--- a/interface-definitions/include/firewall/common-rule-inet.xml.i
+++ b/interface-definitions/include/firewall/common-rule-inet.xml.i
@@ -19,5 +19,6 @@
 #include <include/firewall/synproxy.xml.i>
 #include <include/firewall/tcp-flags.xml.i>
 #include <include/firewall/tcp-mss.xml.i>
+#include <include/firewall/gre.xml.i>
 #include <include/firewall/time.xml.i>
 <!-- include end -->
--- a/python/vyos/firewall.py
+++ b/python/vyos/firewall.py
@@ -409,6 +409,41 @@ def parse_rule(rule_conf, hook, fw_name, rule_id, ip_name):
         time = rule_conf['recent']['time']
         output.append(f'add @RECENT{def_suffix}_{hook}_{fw_name}_{rule_id} {{ {ip_name} saddr limit rate over {count}/{time} burst {count} packets }}')
 
+    if 'gre' in rule_conf:
+        gre_key = dict_search_args(rule_conf, 'gre', 'key')
+
+        gre_flags = dict_search_args(rule_conf, 'gre', 'flags')
+        output.append(parse_gre_flags(gre_flags or {}, force_keyed=gre_key is not None))
+
+        gre_proto_alias_map = {
+            '802.1q': '8021q',
+            '802.1ad': '8021ad',
+            'gretap': '0x6558',
+        }
+
+        gre_proto = dict_search_args(rule_conf, 'gre', 'inner_proto')
+        if gre_proto is not None:
+            gre_proto = gre_proto_alias_map.get(gre_proto, gre_proto)
+            output.append(f'gre protocol {gre_proto}')
+
+        gre_ver = dict_search_args(rule_conf, 'gre', 'version')
+        if gre_ver == 'gre':
+            output.append('gre version 0')
+        elif gre_ver == 'pptp':
+            output.append('gre version 1')
+
+        if gre_key:
+            # The offset of the key within the packet shifts depending on the C-flag. 
+            # nftables cannot handle complex enough expressions to match multiple 
+            # offsets based on bitfields elsewhere.
+            # We enforce a specific match for the checksum flag in validation, so the 
+            # gre_flags dict will always have a 'checksum' key when gre_key is populated. 
+            if not gre_flags['checksum']: 
+                # No "unset" child node means C is set, we offset key lookup +32 bits
+                output.append(f'@th,64,32 == {gre_key}')                
+            else:
+                output.append(f'@th,32,32 == {gre_key}')
+
     if 'time' in rule_conf:
         output.append(parse_time(rule_conf['time']))
 
@@ -543,6 +578,32 @@ def parse_rule(rule_conf, hook, fw_name, rule_id, ip_name):
 
     output.append(f'comment "{family}-{hook}-{fw_name}-{rule_id}"')
     return " ".join(output)
+
+def parse_gre_flags(flags, force_keyed=False):
+    flag_map = { # nft does not have symbolic names for these. 
+        'checksum': 1<<0,
+        'routing':  1<<1,
+        'key':      1<<2,
+        'sequence': 1<<3,
+        'strict_routing': 1<<4,
+    }
+
+    include = 0
+    exclude = 0
+    for fl_name, fl_state in flags.items():
+        if not fl_state: 
+            include |= flag_map[fl_name]
+        else: # 'unset' child tag
+            exclude |= flag_map[fl_name]
+
+    if force_keyed:
+        # Implied by a key-match.
+        include |= flag_map['key']
+
+    if include == 0 and exclude == 0:
+        return '' # Don't bother extracting and matching no bits
+
+    return f'gre flags & {include + exclude} == {include}'
 
 def parse_tcp_flags(flags):
     include = [flag for flag in flags if flag != 'not']
--- a/smoketest/scripts/cli/test_firewall.py
+++ b/smoketest/scripts/cli/test_firewall.py
@@ -1100,5 +1100,60 @@ class TestFirewall(VyOSUnitTestSHIM.TestCase):
         with self.assertRaises(ConfigSessionError):
             self.cli_commit()
 
+    def test_gre_match(self):
+        name = 'smoketest-gre'
+
+        self.cli_set(['firewall', 'ipv4', 'name', name, 'default-action', 'return'])
+        self.cli_set(['firewall', 'ipv4', 'name', name, 'rule', '1', 'action', 'accept'])
+        self.cli_set(['firewall', 'ipv4', 'name', name, 'rule', '1', 'protocol', 'gre'])
+        self.cli_set(['firewall', 'ipv4', 'name', name, 'rule', '1', 'gre', 'flags', 'key'])
+        self.cli_set(['firewall', 'ipv4', 'name', name, 'rule', '1', 'gre', 'flags', 'checksum', 'unset'])
+        self.cli_set(['firewall', 'ipv4', 'name', name, 'rule', '1', 'gre', 'key', '1234'])
+        self.cli_set(['firewall', 'ipv4', 'name', name, 'rule', '1', 'log'])
+
+        self.cli_set(['firewall', 'ipv4', 'forward', 'filter', 'rule', '2', 'action', 'continue'])
+        self.cli_set(['firewall', 'ipv4', 'forward', 'filter', 'rule', '2', 'protocol', 'gre'])
+        self.cli_set(['firewall', 'ipv4', 'forward', 'filter', 'rule', '2', 'gre', 'inner-proto', '0x6558'])
+        self.cli_set(['firewall', 'ipv4', 'forward', 'filter', 'rule', '2', 'log'])
+
+        self.cli_set(['firewall', 'ipv6', 'input', 'filter', 'rule', '3', 'action', 'drop'])
+        self.cli_set(['firewall', 'ipv6', 'input', 'filter', 'rule', '3', 'protocol', 'gre'])
+        self.cli_set(['firewall', 'ipv6', 'input', 'filter', 'rule', '3', 'gre', 'flags', 'checksum'])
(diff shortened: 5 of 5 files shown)


talmakion:
Please excuse the badly formatted commit message, this isn't intended as a final PR.

talmakion:
Fixed mistake in validator & rebased on current

talmakion:
Testing is completed, clumsy design but operational.

nicolas-fort:
Can you extend [firewall smoketest](https://github.com/vyos/vyos-1x/blob/current/smoketest/scripts/cli/test_firewall.py), and include a test case for this setup?

talmakion:
@nicolas-fort Added smoketests in a [separate PR](https://github.com/vyos/vyos-1x/pull/3800) for both IPsec & GRE match changes - they're all in the same file, so there's merge conflicts in combined testing. 

Let me know if you'd like them split up into the separate PR branches instead.

talmakion:
Split smoketests back into their respective PRs.

interface-definitions/include/firewall/gre.xml.i
```diff
@@ -0,0 +1,126 @@
...
+            </leafNode>
+          </children>
+        </node>
+      </children>
+    </node>
+    <leafNode name="inner-proto">
+      <properties>
+        <help>EtherType of encapsulated packet</help>
+        <completionHelp>
+          <list>ip ip6 arp vlan 8021q 8021ad</list>
```
dmbaturin: In VLAN interface settings we use `802.1q` and `802.1ad` with dots (see https://github.com/vyos/vyos-1x/blob/current/interface-definitions/include/interface/vlan-protocol.xml.i)

I think it's more readable with dots and there's no real reason to spell it `8021ad`, since dots are allowed in leaf node values, without quoting, and it's very easy to convert it to NFT options in the conf script.

interface-definitions/include/firewall/gre.xml.i
```diff
@@ -0,0 +1,126 @@
...
+        <valueHelp>
+          <format>ip6</format>
+          <description>IPv6</description>
+        </valueHelp>
+        <valueHelp>
+          <format>arp</format>
+          <description>Address Resolution Protocol</description>
+        </valueHelp>
+        <valueHelp>
+          <format>vlan</format>
```
dmbaturin: What's the reason to have both `vlan` and `802.1q` options that are actually the same ethertype? I don't think we should.

interface-definitions/include/firewall/gre.xml.i
```diff
@@ -0,0 +1,126 @@
...
+          <description>VLAN-tagged Ethernet</description>
+        </valueHelp>
+        <valueHelp>
+          <format>8021q</format>
+          <description>VLAN-tagged Ethernet</description>
+        </valueHelp>
+        <valueHelp>
+          <format>8021ad</format>
+          <description>Bridged Ethernet</description>
```
dmbaturin: I think this description should be "802.1ad QinQ stacked VLAN" or similar. Saying "bridged" without explicitly mentioning "provider bridge" is just misleading, and even "provider bridge" isn't really a common term. "VLAN stacking" is intuitive and quite common.

interface-definitions/include/firewall/gre.xml.i
```diff
@@ -0,0 +1,126 @@
...
+        </valueHelp>
+        <constraint>
+          <validator name="gre-protocol"/>
+        </constraint>
+      </properties>
+    </leafNode>
+    <leafNode name="key">
+      <properties>
+        <help>Tunnel Key</help>
```
dmbaturin: We don't capitalize every word, unless it's an acronym/abbreviation. "Tunnel key" is just a normal phrase.

interface-definitions/include/firewall/gre.xml.i
```diff
@@ -0,0 +1,126 @@
...
+    <leafNode name="version">
+      <properties>
+        <help>GRE Version</help>
+        <valueHelp>
+          <format>gre</format>
+          <description>Standard GRE</description>
+        </valueHelp>
+        <valueHelp>
+          <format>pptp</format>
+          <description>Point to Point Protocol</description>
```
dmbaturin: Point to Point _Tunnelling_ Protocol

src/conf_mode/firewall.py
```diff
@@ -192,6 +192,36 @@ def verify_rule(firewall, rule_conf, ipv6):
...
+                # This check enforces the latter. The user is free to create rules for both cases. 
+                raise ConfigError('Matching GRE tunnel key requires an explicit checksum flag match. For most cases, use "gre flags not checksum"')
+
+        gre_flags = dict_search_args(rule_conf, 'gre', 'flags')
+        if gre_flags:
+            gre_not_flags = dict_search_args(rule_conf, 'gre', 'flags', 'not')
+            if gre_not_flags:
+                duplicates = [flag for flag in gre_flags if flag in gre_not_flags]
+                if duplicates:
+                    raise ConfigError(f'Cannot match a GRE flag as set and not set')
```
dmbaturin: I believe in cases like this we should always look for ways to make the illegal state unrepresentable. One idea is to make the commands like `set ... rule 1 gre flags checksum [unset]`.

If the user does `set ... rule 1 gre flags checksum`, the rule matches packets where GRE checksum is present. If the user adds `set ... rule 1 gre flags checksum unset`, then it matches packets without a checksum. That way the config doesn't have mutually exclusive options.

src/conf_mode/firewall.py
```diff
@@ -192,6 +192,36 @@ def verify_rule(firewall, rule_conf, ipv6):
...
+                if 'key' in gre_not_flags and gre_checking_key:
+                    raise ConfigError('Matching GRE tunnel key implies "flags key", cannot specify "flags not key"')
```
dmbaturin: Here mutually exclusive options are inevitable, but I still think these commands might be better as `set ... rule 1 gre flags key value <...>` and `set ... rule 1 gre flags key unset`.

src/validators/gre-protocol
```diff
@@ -0,0 +1,38 @@
...
+    input = argv[1]
+    try:
+        # It's common practice to pass hex for ethtype (eg, 0x8100), so we allow 
+        # other radix prefixes with the int conversion:
+        if int(input, 0) in range(0, 65535):
+            exit(0)
+    except ValueError:
+        pass
+
+    pattern = "!?\\b(ip|ip6|arp|vlan|8021q|8021ad)\\b"
```
dmbaturin: This script is unnecessary, since the use case is already covered by the `<regex>` validation mechanism, which is a part of the `validate-value` executable and doesn't require spawning a subprocess (and spawning a Python interpreter is a notoriously slow operation).

dmbaturin (changes requested):
I think the UI and the implementation can be improved.

interface-definitions/include/firewall/gre.xml.i
talmakion: This one was pure laziness, I'll sort it out.

interface-definitions/include/firewall/gre.xml.i
talmakion: Again, just following the symbolic names available in nft. Sorting it.

interface-definitions/include/firewall/gre.xml.i
talmakion: I'm updating it to track the same terminology as include/vlan-protocol.xml.i (slightly less verbose): "Provider Bridging (IEEE 802.1ad, Q-in-Q)". 

If you think it's useful to mention VLAN stacking as well when stuffing QinQ into GRE, it's no big effort for me to tweak again. 

Since it's being tweaked, I'll add an alias for gretap 0x6558 as well, as it's an explicitly supported GRE interface type in VyOS.

src/validators/gre-protocol
talmakion: Main reason I did this was for the multi-radix validation (0xNNNN syntax is pretty common for protocols and ethertypes).

The OCaml numeric validator doesn't like radix prefixes, I'll pack this all into a single regex for now - is that something that might be useful for the validator?

interface-definitions/include/firewall/gre.xml.i
talmakion: Whoops.

src/validators/gre-protocol
talmakion: Also, Is there a neat or "proper" way to offer hex integers in valueHelp, or is it something that is best not done at all?

Right now I'm using:
```
        <valueHelp>
          <format>u32:0x0-0xFFFF</format>
          <description>Ethernet protocol number (hex)</description>
        </valueHelp>
```

This formats correctly in the completion under vbash. I picked through the vyatta-cfg node parser + completion scripts and couldn't immediately spot anywhere that definitely expects a base-10 integer, but I'm not familiar with the API and other internals that may treat those tags differently.

talmakion:
Updated with requested changes, smoketests & generated rules look OK and re-ran through a short underlying test set with real traffic.

I cannot figure out a way to do a CLI-level range validation on integers with a hex prefix (0xNNNN, for ethertype) without a custom validator, this has been moved into conf_mode. I can drop this convenience altogether if this doesn't match preferred style, but it's common to use these for ethertypes and protocol numbers. 

I've added 'gretap' as an ethertype alias in GRE matches, there doesn't seem to be a nice standard name for this (Cisco likes 'eogre', there's also 'gre-teb', 'l2gre' or 'teb'). I've stuck with the tunnel type name already used by VyOS. 

Code comments have been tweaked around a bit as well for clarity.

talmakion:
I'm not entirely happy with my help text that follows the `gre flags <FLAG> [unset]` schema either, I may tweak this again if I think of a better way to explain that node structure. 

Potentially, can also switch to `gre flags <FLAG> set|unset`, but my future RSI likes keeping things short.

dmbaturin (approved):

src/validators/gre-protocol
dmbaturin: >The OCaml numeric validator doesn't like radix prefixes

The only reason for that is that I never thought it would be useful anywhere. It would be quite simple to add, although the UI for it is an interesting question. I made a task about it (https://vyos.dev/T6622), if you have any ideas, please leave a comment there.

c-po: > This formats correctly in the completion under vbash

This is defined in `/etc/bash_completion.d/vyatta-cfg`

```
  case "$vtype" in
    _*)
      echo -n "${vtype#_}"
      ;;
    txt)
      echo -n '<text>'
      ;;
    u32)
      echo -n '<0-4294967295>'
      ;;
    u32:*)
      echo -n "<${vtype##u32:}>"
      ;;
   ...
```


Using 

```
    <valueHelp>
      <format>u32:0x0-0xffff</format>
      <description>Numeric IP port</description>
    </valueHelp>
```

will yield `<0x0-0xffff>         Numeric IP port`

and 

```
    <valueHelp>
      <format>_0x0-0xffff</format>
      <description>Numeric IP port</description>
    </valueHelp>
```

will yield `0x0-0xffff         Numeric IP port`

c-po:
> I'm not entirely happy with my help text that follows the `gre flags <FLAG> [unset]` schema either, I may tweak this again if I think of a better way to explain that node structure.
> 
> Potentially, can also switch to `gre flags <FLAG> set|unset`, but my future RSI likes keeping things short.

I understand it this way: `gre flags key` checks if the optional key is present and `gre flags key unset` does not check if the optional key is present? This feels weird.

It should simply be: `gre flags has-key` check if optional key is present - if CLI node is unset - optional key checking is ignored.

c-po:
Can you please rebase this PR to the latest `current`?

talmakion:
> I understand it this way: `gre flags key` checks if the optional key is present and `gre flags key unset` does not check if the optional key is present? This feels weird.
> 
> It should simply be: `gre flags has-key` check if optional key is present - if CLI node is unset - optional key checking is ignored.

If I understand you - not quite, those are explicit checks for whether a flag bit is set or not set in the GRE header structure. That means there are 3 match states: "Don't care" (no node at all), "Is Set" or `gre flags key`, "Not Set" or `gre flags key unset`. 

There are 2 flags we need to care about if we're checking for an optional key, because the packet structure changes depending on how they are set and there's no builtin lookups, so I'm doing raw transport header bitfields lookups to match the key value. 

The other flags are exposed more for completeness than anything else - somebody may find them useful. 

I don't really like the way this works, the protocol guts shouldn't need to be this exposed to the user. The alternatives would be using something as capable as BPF, since nftables can't handle it in a single rule (iptables with BPF expressions could do these multiple flagged-bitwise matches in a single rule), or re-engineering the nft config generation (and anything depending on it) to allow multiple rules in nft output from a single firewall config rule. 

The alternative I went with was simplicity of implementation and guiding the user if they're not sure - where it requires a setting for the "checksum" flag it gives a recommendation, and automatically assumes the "key" flag to be set if we want to match a key, disallowing an "unset" match.

dmbaturin:
>If I understand you - not quite, those are explicit checks for whether a flag bit is set or not set in the GRE header structure. 

I do believe we should have a way to match only packets where the GRE key is not set, and the CLI for it will, unfortunately, be somewhat awkward because the issue is awkward. I think the current syntax is acceptable.

c-po:
> If I understand you - not quite, those are explicit checks for whether a flag bit is set or not set in the GRE header structure. That means there are 3 match states: "Don't care" (no node at all), "Is Set" or gre flags key, "Not Set" or gre flags key unset.

Thanks for the clarification.

interface-definitions/include/firewall/gre.xml.i
```diff
@@ -0,0 +1,126 @@
...
+        <help>EtherType of encapsulated packet</help>
+        <completionHelp>
+          <list>ip ip6 arp 802.1q 802.1ad</list>
+        </completionHelp>
+        <valueHelp>
+          <format>u32:0-65535</format>
+          <description>Ethernet protocol number</description>
+        </valueHelp>
+        <valueHelp>
+          <format>u32:0x0-0xFFFF</format>
```
c-po: Can we please use lowercase hex digits?

interface-definitions/include/firewall/gre.xml.i
```diff
@@ -0,0 +1,126 @@
...
+        <valueHelp>
+          <format>802.1ad</format>
+          <description>Provider Bridging (IEEE 802.1ad, Q-in-Q)</description>
+        </valueHelp>
+        <valueHelp>
+          <format>gretap</format>
+          <description>Transparent Ethernet Bridging (L2 Ethernet over GRE, gretap)</description>
+        </valueHelp>
+        <constraint>
+          <regex>(ip|ip6|arp|802.1q|802.1ad|gretap|\d+|0x[0-9a-fA-F]+)</regex>
```
c-po: Please drop `\d+` from the regex and rather use this instead, as validators are logically OR'ed:

```

         <constraint>
-          <regex>(ip|ip6|arp|802.1q|802.1ad|gretap|\d+|0x[0-9a-fA-F]+)</regex>
+          <regex>(ip|ip6|arp|802.1q|802.1ad|gretap|0x[0-9a-fA-F]+)</regex>
+          <validator name="numeric" argument="--range 0-65535"/>
         </constraint>
```

Please also not that the hex validation logic is NOT precise enough:

```
cpo@LR2.wue3# set firewall ipv4 name foo rule 10 gre inner-proto 0x20000
```

This works and is an illegal value. I think it should be: `0x[0-9a-fA-F]{1,4}` instead.

@dmbaturin we probably should also make the numeric validator HEX aware.

c-po (changes requested):

interface-definitions/include/firewall/gre.xml.i
```diff
@@ -0,0 +1,126 @@
...
+        </constraint>
+      </properties>
+    </leafNode>
+    <leafNode name="key">
```
c-po: Please use: `#include <interface/parameters-key.xml.i>` instead of defining your own `key` CLI node

c-po (changes requested):

interface-definitions/include/firewall/gre.xml.i
talmakion: > as validators are logically OR'ed

Thanks. I was trying to figure out a way to do that neatly. There is a range check in verify_rule() but it's nicer to have it driven from the CLI with immediate feedback to the user. 

> make the numeric validator HEX aware

dmbaturin picked that up already from earlier comments: https://vyos.dev/T6622. It would be the ideal way to do this - the regex stuff is just to get something workable until a radix-tolerant validator is available and can be retrofitted in.

interface-definitions/include/firewall/gre.xml.i
talmakion: The " (only GRE tunnels)" part of the help text is a bit odd but it doesn't hurt anything.

talmakion:
Applied requested changes & rebased

c-po (approved):

PR #3684, opened 2024-06-19
op mode: T6498: move uptime helpers to vyos.utils.system

dmbaturin:
## Change Summary

Move the functions for getting system uptime and load averages from `uptime.py` to `vyos.utils.system`, mainly to be able to call them from the new tech-support script, but that's not all.

There's also a "while we are at it" change that makes the performance of those functions better.

1. `get_uptime_seconds()` no longer reads the uptime file twice.
2. `get_load_averages()` now reads `/proc/loadavg` directly instead of spawning `uptime` as a subprocess.

## Types of changes

## Component(s) name

Op mode.

## How to test

`run show system uptime` should work as before.

## Checklist:

### DIFF
--- a/python/vyos/utils/system.py
+++ b/python/vyos/utils/system.py
@@ -1,4 +1,4 @@
-# Copyright 2023 VyOS maintainers and contributors <maintainers@vyos.io>
+# Copyright 2023-2024 VyOS maintainers and contributors <maintainers@vyos.io>
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


c-po (approved):

jestabro (commented):
One minor point: pylint complains about '`\s`' and '`\.`' if the regex is not a raw string; annoyingly, it can produce a vague warning message when pylint called on an unrelated file:
`<unknown>:108: SyntaxWarning: invalid escape sequence '\.'`


To make pylint happy:
```
diff --git a/python/vyos/utils/system.py b/python/vyos/utils/system.py
index dc8c9aa89..b3b1dbb7f 100644
--- a/python/vyos/utils/system.py
+++ b/python/vyos/utils/system.py
@@ -105,7 +105,7 @@ def get_uptime_seconds():
     from vyos.utils.file import read_file
 
     data = read_file("/proc/uptime")
-    seconds = search("([0-9\.]+)\s", data).group(1)
+    seconds = search(r'([0-9\.]+)\s', data).group(1)
     res  = int(float(seconds))
 
     return res
```

I have encountered this recently and have in a working repo (part of a larger PR):
```
diff --git a/python/vyos/utils/dict.py b/python/vyos/utils/dict.py
index 062ab9c81..1eb6abcd5 100644
--- a/python/vyos/utils/dict.py
+++ b/python/vyos/utils/dict.py
@@ -34,7 +34,7 @@ def colon_separated_to_dict(data_string, uniquekeys=False):
         otherwise they are always lists of strings.
     """
     import re
-    key_value_re = re.compile('([^:]+)\s*\:\s*(.*)')
+    key_value_re = re.compile(r'([^:]+)\s*\:\s*(.*)')
 
     data_raw = re.split('\n', data_string)
```

jestabro (approved):

PR #3745, opened 2024-06-30
T6527: add legacy Vyatta interpreter files still in use

c-po:
## Change Summary

Remove legacy packages from ISO build.
* vyatta-cfg-system
* vyatta-op

Are no longer required and migrated to vyos-1y. vyatta-wanloadbalance, vyatta-bash and vyatta-cfg are now a direct dependency of vyos-1x, so everything in one place.

## Types of changes

## Related Task(s)

* https://vyos.dev/T6527

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/680
* https://github.com/vyos/vyatta-cfg/pull/83

## Smoketest result

image

## Checklist:


src/etc/default/vyatta
```diff
@@ -0,0 +1,203 @@
+# declare configured Vyatta shell environment variables
```
dmbaturin, line 15: I think we should likely add a license header with attribution to Vyatta, to avoid creating an impression that we wrote us, and to ensure legal compliance.

src/opt/vyatta/share/vyatta-op/functions/interpreter/vyatta-unpriv
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
```
dmbaturin, line 1: This is also certainly a highly non-trivial file that requires license info and attribution.

src/opt/vyatta/share/vyatta-op/templates/clear/interfaces/connection/node.tag/node.def
```diff
@@ -0,0 +1,53 @@
+help:  Bring datalink layer connection down and back up
+
+allowed: local -a array ;
+         array=( /etc/ppp/peers/pppoe* /etc/ppp/peers/pppoa* /etc/ppp/peers/wan* /etc/ppp/peers/wlm* ) ;
```
dmbaturin: `/etc/ppp/peers/wan*` is irrelevant since we don't have serial WAN support now.

Can `wlm*` interfaces really be restarted this way?

src/opt/vyatta/share/vyatta-op/templates/clear/interfaces/connection/node.tag/node.def
```diff
@@ -0,0 +1,53 @@
...
+
+run: (	
+	IFNAME=${4}
+	LOGFILE=/var/log/vyatta/ppp_${IFNAME}.log
+        if [ ! -e /etc/ppp/peers/$IFNAME ]; then
+            echo "Invalid interface: $IFNAME"
+	    builtin exit 1
+	fi
+
+	# Serial PPP interfaces are started with "pon", which
```
dmbaturin: We can remove this block since serial WAN is not supported in VyOS anyway.

src/opt/vyatta/share/vyatta-op/templates/clear/interfaces/connection/node.tag/node.def
```diff
@@ -0,0 +1,53 @@
...
+	
+	# now PPP layer should be down
+        echo "Bringing interface $IFNAME back up..."
+	echo "`date`: User $USER starting PPP daemon for $IFNAME by clear command" >> $LOGFILE
+        if [ "${IFNAME::3}" = "wan" ]; then
+	    # Serial interfaces are started with "pon"
+            (umask 0; sudo /usr/sbin/pon $IFNAME > \
+                /dev/null 2>&1 & )
+        else
+            ( umask 0; sudo /usr/sbin/pppd call $IFNAME > \
```
dmbaturin: We only really need this branch.

dmbaturin (changes requested):
I think we should add license headers with attribution to Vyatta at least to all non-trivial files. Maybe even just add them to all files, even allowed-op and similar, just to make it clear where they came from and who the real author is — it's unlikely that we'll get into legal trouble for not attributing trivial files, but I think it's a good thing to do.

We can also clean files from legacy stuff like serial WAN, since it was never supported in VyOS anyway.

src/opt/vyatta/share/vyatta-op/templates/clear/interfaces/connection/node.tag/node.def
c-po: Added by accident - rewritten by proper XML logic. Thanks!

src/opt/vyatta/share/vyatta-op/templates/clear/interfaces/connection/node.tag/node.def
c-po: Added by accident - rewritten by proper XML logic. Thanks!

src/opt/vyatta/share/vyatta-op/templates/clear/interfaces/connection/node.tag/node.def
c-po: Added by accident - rewritten by proper XML logic. Thanks!

src/etc/default/vyatta
c-po: Of course - attribution added

src/opt/vyatta/share/vyatta-op/functions/interpreter/vyatta-unpriv
c-po: Of course - attribution added

c-po:
> I think we should add license headers with attribution to Vyatta at least to all non-trivial files. Maybe even just add them to all files, even allowed-op and similar, just to make it clear where they came from and who the real author is — it's unlikely that we'll get into legal trouble for not attributing trivial files, but I think it's a good thing to do.
> 
> We can also clean files from legacy stuff like serial WAN, since it was never supported in VyOS anyway.

All requested changes addressed.

sever-sever (approved):

dmbaturin (approved):

c-po:
@MergifyIo backport circinus

PR #3753, opened 2024-07-02
T6539: add logging options to load-balancer reverse-proxy

jvoss:
## Change Summary

Add logging configuration options to `load-balancer reverse-proxy`

## Types of changes

## Related Task(s)

* https://vyos.dev/T6539

## Component(s) name

* load-balancer reverse-proxy (haproxy)

## Proposed changes

The current configuration of haproxy (load-balancing reverse-proxy) includes two statically configured global options in `haproxy.cfg`:

```
global
  log /dev/log local0
  log /dev/log local1 notice
```

This does not provide flexibility to the user. Additionally, the current configuration logs all connection attempts across all services. This can be problematic for busy systems by producing large logs of unnecessary information.

This task aims to provide configuration options to allow the user to configure the syslog facility and severity of messages received. New options are provided as global-parameters, as well as within each service and backend:

```
[edit load-balancing reverse-proxy]
vyos@vyos# show
 backend BACKEND {
     logging {
         facility local1 {
             level err
         }
     }
 }
 global-parameters {
     logging {
         facility local0 {
         }
     }
 }
 service TEST {
     logging {
         facility local7 {
             level debug
         }
     }
 }
```

Multiple logging facilities may be specified for each of the attachment points and this hierarchy allows for additional logging configuration options in the future (i.e. format, etc).

Reference haproxy documentation: https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#8

When none of these options are specified, the original behavior is preserved and the original global logging configuration will be present in `haproxy.cfg`.

## How to test

Configure logging options

```
set load-balancer reverse-proxy global-parameters logging facility local0 level notice
```

Ensure the configuration has been rendered:

```
vyos@vyos# cat /run/haproxy/haproxy.cfg | grep log
    log /dev/log local0 notice
```

Ensure the service has been reloaded and is running (valid configuration):

```
vyos@vyos# systemctl status haproxy
● haproxy.service - HAProxy Load Balancer
     Loaded: loaded (/lib/systemd/system/haproxy.service; disabled; preset: enabled)
    Drop-In: /run/systemd/system/haproxy.service.d
             └─10-override.conf
     Active: active (running) since Tue 2024-07-02 18:21:55 UTC; 3s ago
```

## Smoketest result

New test ensures that original logging behavior is preserved if not configured. Also tests multiple facilities configured for `global-parameters`, `service` and `backend` items:

```
vyos@vyos:~$ /usr/libexec/vyos/tests/smoke/cli/test_load-balancing_reverse-proxy.py
[...]
test_09_lb_reverse_proxy_logging (__main__.TestLoadBalancingReverseProxy.test_09_lb_reverse_proxy_logging) ... ok

Ran 9 tests in 131.619s

OK
```

## Checklist:


jvoss:
I will also update the documentation if this change is accepted.

interface-definitions/include/haproxy/logging.xml.i
```diff
@@ -0,0 +1,146 @@
+<!-- include start from haproxy/logging.xml.i -->
+<node name="logging">
+  <properties>
+    <help>Logging parameters</help>
+  </properties>
+  <children>
+    <tagNode name="facility">
```
dmbaturin: This looks like an exact copy of https://github.com/vyos/vyos-1x/blob/current/interface-definitions/include/syslog-facility.xml.i

I think it's better to include that file rather than copy its contents, since syslog facilities and levels are standardized and won't need custom options here.

dmbaturin (approved):
Looks good other than the part about includes — I think let's just use the existing syslog facility and level file.

interface-definitions/include/haproxy/logging.xml.i
jvoss: Thanks, I wanted to see a maintainer's opinion on this. 

I chose this direction as it would isolate system logging configuration changes from haproxy. There were also configuration options that are invalid for haproxy.

jvoss: Updated as suggested. However some of these options will be invalid configuration such as `facility all` and cause haproxy startup/reload to fail. Do you have suggestions on how to handle these situations?

I could add checks to the configuration verifier (https://github.com/vyos/vyos-1x/blob/current/src/conf_mode/load-balancing_reverse-proxy.py) but it seems inappropriate to even list it as a configuration option if it is invalid.

data/templates/load-balancing/haproxy.cfg.j2
```diff
@@ -11,6 +9,14 @@ global
     daemon
 
 {% if global_parameters is vyos_defined %}
+{%     if global_parameters.logging is vyos_defined %}
+{%         for facility, facility_config in global_parameters.logging.facility.items() %}
+    log /dev/log {{ facility }} {{ facility_config.level }}
+{%         endfor %}
+{%     else %}
+    log /dev/log local0
```
c-po: This keeps the current behavior:

> This does not provide flexibility to the user. Additionally, the current configuration logs all connection attempts across all services. This can be problematic for busy systems by producing large logs of unnecessary information.

I wonder if this should either be changed to something less verbose or be dropped at all to require the user to enable explicit logging.

c-po (approved):

data/templates/load-balancing/haproxy.cfg.j2
jvoss: I can do that too. I would personally probably err on the side of requiring explicit configuration myself.

I was not sure what you all would prefer as far as maintaining how it functions prior to this change.

jvoss: @c-po updated, removed the default logging behavior in favor of explicit logging.

sever-sever (approved):

c-po:
@MergifyIo backport sagitta

sever-sever:
@mergifyio backport circinus

PR #3811, opened 2024-07-14
op-mode: T6580: read active nodes directly from the config tree without calling cli-shell-api

dmbaturin:
## Change Summary

For op mode completion purposes, we only need to care about nodes in the active config, so none of the complexities of config sessions apply. In the current implementation that uses a directory tree for the config tree, that's equivalent to `ls`.

The task is made easier by the fact that op mode template interpreter doesn't care about exit codes, so even if `ls` exits with a non-zero code on non-existent dirs, that has no negative effects, we can just supress its error output with `ls 2>/dev/null`. And node names cannot contain whitespace, so we can always use them in UNIX path strings without escaping. 

## Types of changes

- [x] Other (please describe): performance optimization.

## Component(s) name

Op mode.

## How to test

Every command that uses a node path in completion, like `run show interfaces`, should work as before.

## Checklist:

### DIFF
--- a/scripts/build-command-op-templates
+++ b/scripts/build-command-op-templates
@@ -3,7 +3,7 @@
 #    build-command-template: converts new style command definitions in XML
 #      to the old style (bunch of dirs and node.def's) command templates
 #
-#    Copyright (C) 2017 VyOS maintainers <maintainers@vyos.net>
+#    Copyright (C) 2017-2024 VyOS maintainers <maintainers@vyos.net>
 #
 #    This library is free software; you can redistribute it and/or
 #    modify it under the terms of the GNU Lesser General Public
@@ -20,6 +20,7 @@
 #    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301
 #    USA
 
+import re
 import sys
 import os
 import argparse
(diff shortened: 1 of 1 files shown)


scripts/build-command-op-templates
```diff
@@ -108,7 +109,8 @@ def get_properties(p):
             for i in lists:
                 comp_exprs.append("echo \"{0}\"".format(i.text))
             for i in paths:
-                comp_exprs.append("/bin/cli-shell-api listActiveNodes {0} | sed -e \"s/'//g\" && echo".format(i.text))
+                path = re.sub(r'\s+', '/', i.text)
+                comp_exprs.append("ls /opt/vyatta/config/active/{0} 2>/dev/null".format(path))
```
c-po, line 113: Please de-hardcode the path to `vyos.defaults.directories`

https://github.com/vyos/vyos-1x/blob/current/python/vyos/defaults.py#L38

c-po (changes requested):

scripts/build-command-op-templates
dmbaturin: This script is only executed at build time and doesn't have `vyos.*` modules in its environment, so we'll need to either mess with Python path in the Makefile or import by absolute path. I don't think it's worth it in this case.

c-po: It's not worth the trouble.

c-po (approved):

jestabro (approved):

PR #3817, opened 2024-07-16
op-mode: T6577: create generic service restart helper to work with the API

c-po:
## Change Summary

 Right now we have multiple restart helpers (e.g. dhcp server, ssh, ntp) that  all do the same (more or less):

* Check if service is configured on CLI
* Restart if configured
* Error out if unconfigured

This is not available via the op-mode API. Create a new restart.py op-mode helper that takes the service name and possible VRF as argument so it's also exposed via API.

## Types of changes

## Related Task(s)

* https://vyos.dev/T6577

## Component(s) name

op-mode restart

## How to test

* `restart dhcp server`
* `restart dhcpv6 server`
* `restart ssh`
* ...

## Checklist:

### DIFF
--- a/op-mode-definitions/dhcp.xml.in
+++ b/op-mode-definitions/dhcp.xml.in
@@ -245,7 +245,7 @@
             <properties>
               <help>Restart DHCP server</help>
             </properties>
-            <command>if cli-shell-api existsActive service dhcp-server; then sudo systemctl restart kea-dhcp4-server.service; else echo "DHCP server not configured"; fi</command>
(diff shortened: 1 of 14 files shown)


c-po:
@MergifyIo backport circinus

dmbaturin (approved):
This goes against the current "subsystem first" approach of the op mode API. 

*However*, my recent ideas include a plan to move away from it and use a "command first approach" to unify the CLI and the API, so I like this change.

sever-sever (approved):

PR #3825, opened 2024-07-17
openvpn: T6591: deprecate OpenVPN server net30 topology

dmbaturin:
## Change Summary

Set the default topology to `subnet` for new configs, add a deprecation warning, and insert `net30` in old configs explicitly for compatibility.

## Types of changes

- [x] Other (please describe): default change, deprecation.

## Component(s) name

OpenVPN

## Proposed changes

The `net30` topology is:

* only there because it was the default in OpenVPN, but even then it was just for compatibility reasons, not because it's good;
* bad for Windows clients, they don't work with it;
* deprecated in OpenVPN and may eventually be removed.

We should set the new default to `subnet` and mark `net30` as deprecated. But some people may rely on the `net30` behavior, so we'll use a migration script to inject the old default into old configs explicitly.

## Checklist:


c-po (approved):

sever-sever (approved):

dmbaturin:
@mergifyio backport circinus

PR #3927, opened 2024-08-01
nat64: T6627: call check_kmod within standard config function

jestabro:
## Change Summary

The config mode script nat64.py calls the modprobe utility `check_kmod` outside of the standard functions, leading to an `OSError` when run under configd. Current behavior of configd will re-run the script within the CLI context, leading to config and smoketest success; as that behavior will be changed in the future to accommodate all scripts running within the configd context (T6608), correct now.

There are several other config scripts that use this pattern, which will be all changed in a separate PR. The nat64.py script deserved special attention as it provides an example of state configuration informing the syntax verification stage, which should be kept in mind as we move to distinguishing syntax verification from system state verification.

## Types of changes

## How to test

Confirmation provided in task. Success in smoketest and journalctl output, below.

## Smoketest result

```
vyos@vyos:/usr/libexec/vyos/tests/smoke/cli$ ./test_nat64.py
test_snat64 (__main__.TestNAT64.test_snat64) ... ok

Ran 1 test in 3.788s

OK
vyos@vyos:/usr/libexec/vyos/tests/smoke/cli$ journalctl -u vyos-configd --no-pager

...
Aug 01 16:43:35 vyos vyos-configd[4299]: Received message: {"type": "node", "last": true, "data": "/usr/libexec/vyos/conf_mode/nat64.py"}
...
Aug 01 16:43:35 vyos vyos-configd[4299]: Sending response 1
...
```

## Checklist:

### DIFF
--- a/src/conf_mode/nat64.py
+++ b/src/conf_mode/nat64.py
@@ -46,7 +46,12 @@ def get_config(config: Config | None = None) -> None:
     base = ["nat64"]
     nat64 = config.get_config_dict(base, key_mangling=("-", "_"), get_first_key=True)
 
-    base_src = base + ["source", "rule"]
+    return nat64
+
+
+def verify(nat64) -> None:
+    check_kmod(["jool"])
+    base_src = ["nat64", "source", "rule"]
 
     # Load in existing instances so we can destroy any unknown
     lines = cmd("jool instance display --csv").splitlines()
@@ -76,12 +81,8 @@ def get_config(config: Config | None = None) -> None:
         ):
             rules[num]["recreate"] = True
 
-    return nat64
-
-
-def verify(nat64) -> None:
     if not nat64:
-        # no need to verify the CLI as nat64 is going to be deactivated
+        # nothing left to do
         return
 
     if dict_search("source.rule", nat64):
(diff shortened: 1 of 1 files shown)


sever-sever (approved):

src/conf_mode/nat64.py
```diff
@@ -76,12 +81,9 @@ def get_config(config: Config | None = None) -> None:
         ):
             rules[num]["recreate"] = True
 
-    return nat64
-
-
-def verify(nat64) -> None:
     if not nat64:
-        # no need to verify the CLI as nat64 is going to be deactivated
+        unload_kmod(['jool'])
```
dmbaturin: I'm quite uneasy about system state changes in `verify()`. I'm inclined to think that `verify()` should never include any logic except config verification. Do you think it's feasible to move this to `apply()`?

dmbaturin (changes requested):

src/conf_mode/nat64.py
jestabro: Agreed: though I consider the real problem with the logic of this script to be the need to modprobe in the `verify()` stage, no need to make it worse. Moved to `apply()`.

dmbaturin (approved):

jestabro:
@Mergifyio backport circinus

PR #3955, opened 2024-08-07
configd: T6640: enforce in_session returns False under configd

jestabro:
## Change Summary

Injecting the environment variables needed for avoiding the use of my_set under configd (T6633) causes in_session to return True; we should enforce legacy checks to be consistent with the separation of concerns between the config session and conf_mode script execution.

For this (T6489/T6633) and other extensions as we move towards dropping CStore, in_session should return False if running under configd.

The complex of interactions is described in T6639.

## Types of changes

        Enforce distinction between configd and config session to allow extensions/alternatives to CStore.

## Related Task(s)

* https://vyos.dev/T6633
* https://vyos.dev/T6639

## Checklist:

### DIFF
--- a/src/services/vyos-configd
+++ b/src/services/vyos-configd
@@ -267,6 +267,8 @@ if __name__ == '__main__':
     cfg_group = grp.getgrnam(CFG_GROUP)
     os.setgid(cfg_group.gr_gid)
 
+    os.environ['VYOS_CONFIGD'] = 't'
+
     def sig_handler(signum, frame):
         shutdown()
 
--- a/python/vyos/configsource.py
+++ b/python/vyos/configsource.py
@@ -204,6 +204,8 @@ class ConfigSourceSession(ConfigSource):
         Returns:
             True if called from a configuration session, False otherwise.
         """
+        if os.getenv('VYOS_CONFIGD', ''):
+            return False
         try:
             self._run(self._make_command('inSession', ''))
             return True


src/services/vyos-configd
```diff
@@ -267,6 +267,8 @@ if __name__ == '__main__':
     cfg_group = grp.getgrnam(CFG_GROUP)
     os.setgid(cfg_group.gr_gid)
 
+    os.environ['VYOS_CONFIGD'] = 't'
```
c-po, line 270: What does `t` stand for? Or is it only a virtual marker so the `VYOS_CONFIGD` environment variable is populated?

Why not check if vyos-configd service is running?

jestabro: A virtual/dummy marker: 't' for true, whereas 'true' would be misleading, as it is not a boolean value.

Whether or not vyos-configd is running, the script and its imported modules may not be called under vyos-configd --- though this is the design now in 1.5 after your work, it may not be the case if a daemon timeout or non-fatal error (not seen since early development) or an uncaught error before T6608 is merged.

To untangle the interactions in T6639, an explicit check is preferred.

c-po: Thanks for the detailed explanation.

c-po (approved):

sever-sever (approved):

dmbaturin (approved):

jestabro:
@Mergifyio backport circinus

PR #4022, opened 2024-08-27
T6681: Add option for SLAAC to support suppress Interval Advertisement in RA Packets

rgwan:
## Types of changes

## Related Task(s)

https://vyos.dev/T6681

## Component(s) name

SLAAC,  RADVD, IPv6
## Proposed changes

MikroTik's IPv6 SLAAC client has a bug, when it received a RA packet with Interval advertisement field, it can't obtain an address correctly from the router.

The log on MikroTik shows like that:

```
 22:01:50 radvd,debug received Router Advertisement on ether1 from <some mac>
 22:01:50 radvd,debug   prefix: <some prefix> valid: 1800 preferred: 900
 22:01:50 radvd,debug   unsupported option 7

```
So I'm submitted a patch which allows VyOS don't send Interval advertisement in RA packet.

e.g. the command line will be:

```
set service router-advert interface <some int> no-send-interval
```

## How to test

Run the command I described above, and use Wireshark to capture packet from router, and the 'Interval Advertisement' field in RA packet is gone.

## Checklist:

### DIFF
--- a/data/templates/router-advert/radvd.conf.j2
+++ b/data/templates/router-advert/radvd.conf.j2
@@ -19,7 +19,7 @@ interface {{ iface }} {
 {%         if iface_config.reachable_time is vyos_defined %}
     AdvReachableTime {{ iface_config.reachable_time }};
 {%         endif %}
-    AdvIntervalOpt {{ 'off' if iface_config.no_send_advert is vyos_defined else 'on' }};
+    AdvIntervalOpt {{ 'off' if iface_config.no_send_interval is vyos_defined else 'on' }};
     AdvSendAdvert {{ 'off' if iface_config.no_send_advert is vyos_defined else 'on' }};
 {%         if iface_config.default_lifetime is vyos_defined %}
     AdvDefaultLifetime {{ iface_config.default_lifetime }};
--- a/interface-definitions/service_router-advert.xml.in
+++ b/interface-definitions/service_router-advert.xml.in
@@ -390,6 +390,12 @@
                   <valueless/>
                 </properties>
               </leafNode>
+              <leafNode name="no-send-interval">
+                <properties>
+                  <help>Do not send Advertisement Interval option in RAs</help>
+                  <valueless/>
+                </properties>
+              </leafNode>
             </children>
           </tagNode>
         </children>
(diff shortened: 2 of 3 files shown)


c-po (changes requested):
Thank you for the contribution. This is a behavioral change to VyOS, as in the past both options got set with a single command `set service router-advert interface <some int> no-send-advert`.

To preserve existing behavior a migration script is mandatory which will set the new CLI node once it detects that an interface has `no-send-advert` defined. This will preserve old behavior and add support for users to individually toggle the fields.

rgwan:
> Thank you for the contribution. This is a behavioral change to VyOS, as in the past both options got set with a single command `set service router-advert interface <some int> no-send-advert`.
> 
> To preserve existing behavior a migration script is mandatory which will set the new CLI node once it detects that an interface has `no-send-advert` defined. This will preserve old behavior and add support for users to individually toggle the fields.

In fact, when `no-send-advert` is set, the RADVD daemon will not send any router-advertise to the interface, so `AdvIntervalOpt` option will be omitted by RADVD daemon in this situation.

I don't think we need a migration script for it, it didn't changes old behavior of `no-send-advert`.

rgwan:
@c-po  So, do you have any further question about this PR? I'm glad to hear from you.

c-po:
> In fact, when `no-send-advert` is set, the RADVD daemon will not send any router-advertise to the interface, so `AdvIntervalOpt` option will be omitted by RADVD daemon in this situation.
> 
> I don't think we need a migration script for it, it didn't changes old behavior of `no-send-advert`.

Confirmed:

https://github.com/radvd-project/radvd/blob/f2de4764559e102fe8492520be328858af0f964d/send.c#L880-L885

RADVD code bails out early if  `no-send-advert` is set, thos omitting any other option and not adding a need for a migrator.

c-po (approved):

c-po:
@MergifyIo backport sagitta circinus

c-po:
Added smoketests for the new option


```
cpo@LR2.wue3:~$ /usr/libexec/vyos/tests/smoke/cli/test_service_router-advert.py
test_advsendadvert_advintervalopt (__main__.TestServiceRADVD.test_advsendadvert_advintervalopt) ... ok
test_common (__main__.TestServiceRADVD.test_common) ... ok
test_deprecate_prefix (__main__.TestServiceRADVD.test_deprecate_prefix) ... ok
test_dns (__main__.TestServiceRADVD.test_dns) ... ok
test_nat64prefix (__main__.TestServiceRADVD.test_nat64prefix) ... ok
test_rasrcaddress (__main__.TestServiceRADVD.test_rasrcaddress) ... ok
test_route (__main__.TestServiceRADVD.test_route) ... ok

Ran 7 tests in 19.518s

OK
```

dmbaturin (approved):

PR #4077, opened 2024-09-16
T6716: don't automatically set ethernet offload

dvlogic:
## Change Summary
Remove the lines of code that checked if the kernel had offloading enabled and was then forcing the config to set it to "on." The behavior now mirrors the config and offloading will only be enabled if the config is explicitly set to enabled.

Note: the code is still present to disable the offloading, in the config, if the kernel doesn't support it.

## Types of changes

## Related Task(s)

 https://vyos.dev/T6716

## Component(s) name

Ethernet activation script (20-ethernet_offload.py)

## Proposed changes

Delete lines of code that set the ethernet offload setting based on the Kernel offload setting.
## How to test

On a system where the kernel enables offloading at boot and/or the offload is currently set in the config:

```
configure
delete interfaces ethernet eth0 offload 
commit
load
show interfaces ethernet eth0 offload 
```
No offloading settings should be shown. On VyOS 1.5, without this PR, the offload settings will automatically be restored.
 
## Checklist:

### DIFF
--- a/src/activation-scripts/20-ethernet_offload.py
+++ b/src/activation-scripts/20-ethernet_offload.py
@@ -17,9 +17,12 @@
 #        CLI. See https://vyos.dev/T3619#102254 for all the details.
 # T3787: Remove deprecated UDP fragmentation offloading option
 # T6006: add to activation-scripts: migration-scripts/interfaces/20-to-21
+# T6716: Honor the configured offload settings and don't automatically add
+#        them to the config if the kernel has them set (unless its a live boot)
 
 from vyos.ethtool import Ethtool
 from vyos.configtree import ConfigTree
+from vyos.system.image import is_live_boot
 
 def activate(config: ConfigTree):
     base = ['interfaces', 'ethernet']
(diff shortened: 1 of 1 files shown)


jestabro (changes requested):
One modification that we may want to consider: only set in config during live boot, before installation:
```
diff --git a/src/activation-scripts/20-ethernet_offload.py b/src/activation-scripts/20-ethernet_offload.py
index 33b0ea469..4c7d8b521 100755
--- a/src/activation-scripts/20-ethernet_offload.py
+++ b/src/activation-scripts/20-ethernet_offload.py
@@ -20,6 +20,7 @@

 from vyos.ethtool import Ethtool
 from vyos.configtree import ConfigTree
+from vyos.system.image import is_live_boot

 def activate(config: ConfigTree):
     base = ['interfaces', 'ethernet']
@@ -36,7 +37,7 @@ def activate(config: ConfigTree):
         enabled, fixed = eth.get_generic_receive_offload()
         if configured and fixed:
             config.delete(base + [ifname, 'offload', 'gro'])
-        elif enabled and not fixed:
+        elif is_live_boot() and enabled and not fixed:
             config.set(base + [ifname, 'offload', 'gro'])

         # If GSO is enabled by the Kernel - we reflect this on the CLI. If GSO is
```
... etc. It will still hold that after installation, activation will not enable against user settings.

dvlogic:
I like @jestabro suggestion and updated the PR commit to reflect.

jestabro (approved):

jestabro:
@Mergifyio backport circinus

c-po (approved):

PR #4124, opened 2024-10-03
cli: T6740: add a converter from set commands to config

dmbaturin:
## Change Summary

Add new utility `vyos-commands-to-config` that converts `set` commands to configs suitable for loading. 

Whether we want a CLI completion for it is an open question.

## Types of changes

## Checklist:

### DIFF
--- a/debian/vyos-1x.install
+++ b/debian/vyos-1x.install
@@ -28,6 +28,7 @@ usr/bin/initial-setup
 usr/bin/vyos-config-file-query
 usr/bin/vyos-config-to-commands
 usr/bin/vyos-config-to-json
+usr/bin/vyos-commands-to-config
 usr/bin/vyos-hostsd-client
 usr/lib
 usr/libexec/vyos/activate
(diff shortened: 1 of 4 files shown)


sever-sever (approved):

jestabro (changes requested):
The following concerns may be purely cosmetic, but it is a timely reminder that we are still relying on the legacy loader and parser to provide:
(1) setting tag data
(2) ordering of nodes

We have explicit mechanisms for both, and if one wants the result of the converter to provide a stand alone config rendered with respect to those constraints, the following patch illustrates the idea:
https://github.com/vyos/vyos-1x/commit/088fa9a37037a66f986d66993a171354446d9344

The approach to (1) there is reasonable and provides a utility that may be otherwise useful to set all tag data long a path; the approach to (2) is a hack by simply reading and re-rendering the result --- we currently order nodes when parsing, not rendering, as we haven't had need of the latter (other than an optional ordering of values for unittests).

It is the author's discretion as to whether we should bother with cosmetic parity for the output.

jestabro (approved):

PR #4221, opened 2024-12-04
dhcpv6-server: T5992: Fix op-mode Kea DHCP lease output

nvandamme:
## Change Summary
Fixing for op commands:
* `show dhcpv6 server static-mappings` to include `ipv6-address` from actual dhcpv6 static mappings in config.
* `show dhcpv6 server leases` to include `hostname` field (and removing trailing dot if any).

## Types of changes

## Related Task(s)

* https://vyos.dev/T5992
* https://vyos.dev/T3316

## Component(s) name

* dhcpv6-server

## Proposed changes
* Fix reading of `ipv6-address` field in config section `service dhcpv6-server [...] static-mapping` to report them in op command `show dhcpv6 server static-mappings`
* Fix reading of `hostname` field in leases for either `show dhcp server leases` / `show dhcpv6 server leases` op commands
* Fix `sort` arg in `show dhcpv6 server [...]` op commands

## Checklist:

### DIFF
--- a/src/op_mode/dhcp.py
+++ b/src/op_mode/dhcp.py
@@ -110,23 +110,19 @@ def _get_raw_server_leases(family='inet', pool=None, sorted=None, state=[], orig
         data_lease['pool'] = kea_get_pool_from_subnet_id(active_config, inet_suffix, lease['subnet-id']) if active_config else '-'
         data_lease['end'] = lease['expire_timestamp'].timestamp() if lease['expire_timestamp'] else None
         data_lease['origin'] = 'local' # TODO: Determine remote in HA
+        data_lease['hostname'] = lease.get('hostname', '-')
+        # remove trailing dot to ensure consistency for `vyos-hostsd-client`
+        if data_lease['hostname'][-1] == '.':
+            data_lease['hostname'] = data_lease['hostname'][:-1]
 
         if family == 'inet':
             data_lease['mac'] = lease['hw-address']
             data_lease['start'] = lease['start_timestamp'].timestamp()
-            data_lease['hostname'] = lease.get('hostname', "-")
-            # remove trailing dot to ensure consistency for `vyos-hostsd-client`
-            if data_lease['hostname'][-1] == '.':
-                data_lease['hostname'] = data_lease['hostname'][:-1]
 
         if family == 'inet6':
             data_lease['last_communication'] = lease['start_timestamp'].timestamp()
             data_lease['duid'] = _format_hex_string(lease['duid'])
             data_lease['type'] = lease['type']
-            data_lease['hostname'] = lease.get('hostname', "-")
-            # remove trailing dot to ensure consistency for `vyos-hostsd-client`
-            if data_lease['hostname'][-1] == '.':
-                data_lease['hostname'] = data_lease['hostname'][:-1]
 
             if lease['type'] == 'IA_PD':
                 prefix_len = lease['prefix-len']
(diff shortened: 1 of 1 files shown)


sarthurdev (changes requested):
Thanks for the PR.

While small fixes relating to the same area of code as the PR targets are usually no problem, each fix should be in separate commits with an identifying commit message.

The commits in this PR all have the same commit message, despite changing different things. Can you please reword the messages for what each commit changes?

src/op_mode/dhcp.py
```diff
@@ -279,22 +284,36 @@ def _get_raw_server_static_mappings(family='inet', pool=None, sorted=None):
...
+        for entry in raw_data:
+            pool = entry.get('pool')
+            subnet = entry.get('subnet')
+            name = entry.get('name')
+            ip_addr = entry.get('ip-address', 'N/A')
+            mac_addr = entry.get('mac', 'N/A')
+            duid = entry.get('duid', 'N/A')
+            description = entry.get('description', 'N/A')
+            data_entries.append([pool, subnet, name, ip_addr, mac_addr, duid, description])
+    elif family == 'inet6':
```
sarthurdev, line 307: Instead of duplicating all these lines for the one `ipv6-address` key difference. Could this instead be handled by:
`ip_addr = entry.get('ipv6-address' if family == 'inet6' else 'ip-address', 'N/A')`

src/op_mode/dhcp.py
```diff
@@ -128,12 +132,12 @@ def _get_raw_server_leases(family='inet', pool=None, sorted=None, state=[], orig
         data_lease['remaining'] = '-'
 
         if lease['valid-lft'] > 0:
-            data_lease['remaining'] = lease['expire_timestamp'] - datetime.utcnow()
+            data_lease['remaining'] = lease['expire_timestamp'] - datetime.now(timezone.utc)
```
sarthurdev, line 135: Does this not just produce the same output?

sarthurdev (dismissed):

src/op_mode/dhcp.py
nvandamme: datetime.utcnow() is deprecated since 3.12
https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

src/op_mode/dhcp.py
nvandamme: Of course, normally i'll be doing so, but most part of handling code inside the same file has been split for the two inet families, given their formatting headers have not the same labels:
```python
# IPv4 header for `show dhcp server leases`
headers = ['IP Address', 'MAC address', 'State', 'Lease start', 'Lease expiration', 'Remaining', 'Pool',
                   'Hostname', 'Origin']

# IPv6 header for `show dhcpv6 server leases`
headers = ['IPv6 address', 'State', 'Last communication', 'Lease expiration', 'Remaining', 'Type', 'Pool',
                   'DUID']
```

nvandamme:
> Thanks for the PR.
> 
> While small fixes relating to the same area of code as the PR targets are usually no problem, each fix should be in separate commits with an identifying commit message.
> 
> The commits in this PR all have the same commit message, despite changing different things. Can you please reword the messages for what each commit changes?

Not used to amend commit message, does these renaming works ?

dmbaturin (approved):

c-po:
Commits will be squashed and merged

c-po (approved):

src/op_mode/dhcp.py
c-po: > Could this instead be handled by: ip_addr = entry.get('ipv6-address' if family == 'inet6' else 'ip-address', 'N/A')

We need room for future improvements, too ;)

PR #4230, opened 2024-12-09
T6792: Autocomplete for "show arp interface" interfaces

HollyGurza:
## Types of changes

## Related Task(s)

* https://vyos.dev/T6792

## How to test

```
vyos@vyos:~$ show arp interface 
Possible completions:
  dum1                  Show Address Resolution Protocol (ARP) cache for specified
                        interface
  eth0
  eth1
  eth2
  eth3
  lo

      
vyos@vyos:~$ show arp interface lo 
Address    Interface    Link layer address    State
---------  -----------  --------------------  -------
```

## Checklist:

### DIFF
--- a/op-mode-definitions/show-arp.xml.in
+++ b/op-mode-definitions/show-arp.xml.in
@@ -12,7 +12,7 @@
             <properties>
               <help>Show Address Resolution Protocol (ARP) cache for specified interface</help>
               <completionHelp>
-                <script>${vyos_completion_dir}/list_interfaces --broadcast</script>
+                <script>${vyos_completion_dir}/list_interfaces</script>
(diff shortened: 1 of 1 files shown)


dmbaturin (changes requested):
I don't think we should show _all_ interfaces there. There's no ARP on IPIP, IP6IP, loopback, dummies, VTI... if `list_interfaces --broadcast` doesn't show all interfaces with a broadcast flag, we need to fix the `list_interfaces` script.

If that's an intractable problem (which I doubt), we should limit this list to interfaces that we know can send and receive ARP.

HollyGurza:
implemented here: https://github.com/vyos/vyos-utils/pull/28

PR #4237, opened 2024-12-13
T6948: Keep DHCP server leases in sync with hostd records

indrajitr:
## Change Summary

Keep DHCP server leases in sync with vyos-hostd records via helper script invoked with `ExecStartPost` directive in kea-dhcp4-server.service.
 
The helper script updates VyOS hostd records from DHCP server leases. This ensures that hostd records with the DHCP server leases are kept in sync with VyOS hostd records right after DHCP server is started.

This is similar to the capability exposed via kea hook `libdhcp_run_script.so` which is invoked internally by kea when a single lease changes state.
 
Since the kea hook is currently implemented for DHCPv4 only, this helper script is implemented for DHCPv4 only as well.

## Types of changes

## Related Task(s)

* https://vyos.dev/T6948

## Related PR(s)

https://github.com/vyos/vyos-1x/pull/4307

## Component(s) name

dhcp server

## How to test

1. Enable DHCP server and have active server leases
2. Ensure that relevant host entries exist in `/etc/hosts` and/or `/run/vyos-hostsd/vyos-hostsd.state`
4. Restart VyOS
5. Verify DHCP server leases are still active (`show dhcp server leases`)
6. Verify that the leases have NOT been applied to `/etc/hosts`
7. Restart dhcp server (`restart dhcp server`)
8. Verify that the leases have been applied to `/etc/hosts`

## Smoketest result

```
vyos@test-1216:~$ /usr/libexec/vyos/tests/smoke/cli/test_service_dhcp-server.py
test_dhcp_exclude_in_range (__main__.TestServiceDHCPServer.test_dhcp_exclude_in_range) ... ok
test_dhcp_exclude_not_in_range (__main__.TestServiceDHCPServer.test_dhcp_exclude_not_in_range) ... ok
test_dhcp_high_availability (__main__.TestServiceDHCPServer.test_dhcp_high_availability) ... ok
test_dhcp_high_availability_standby (__main__.TestServiceDHCPServer.test_dhcp_high_availability_standby) ... ok
test_dhcp_hostsd_lease_sync (__main__.TestServiceDHCPServer.test_dhcp_hostsd_lease_sync) ... ok
test_dhcp_multiple_pools (__main__.TestServiceDHCPServer.test_dhcp_multiple_pools) ... ok
test_dhcp_on_interface_with_vrf (__main__.TestServiceDHCPServer.test_dhcp_on_interface_with_vrf) ... ok
test_dhcp_relay_server (__main__.TestServiceDHCPServer.test_dhcp_relay_server) ... ok
test_dhcp_single_pool_options (__main__.TestServiceDHCPServer.test_dhcp_single_pool_options) ... ok
test_dhcp_single_pool_options_scoped (__main__.TestServiceDHCPServer.test_dhcp_single_pool_options_scoped) ... ok
test_dhcp_single_pool_range (__main__.TestServiceDHCPServer.test_dhcp_single_pool_range) ... ok
test_dhcp_single_pool_static_mapping (__main__.TestServiceDHCPServer.test_dhcp_single_pool_static_mapping) ... ok

```

## Checklist:

### DIFF
--- a/smoketest/scripts/cli/test_service_dhcp-server.py
+++ b/smoketest/scripts/cli/test_service_dhcp-server.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 #
-# Copyright (C) 2020-2024 VyOS maintainers and contributors
+# Copyright (C) 2020-2025 VyOS maintainers and contributors
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 2 or later as
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import re
 import unittest
 
 from json import loads
@@ -22,6 +23,9 @@ from json import loads
 from base_vyostest_shim import VyOSUnitTestSHIM
 
 from vyos.configsession import ConfigSessionError
+from vyos.kea import kea_add_lease
+from vyos.kea import kea_delete_lease
+from vyos.utils.process import cmd
 from vyos.utils.process import process_named_running
 from vyos.utils.file import read_file
 from vyos.template import inc_ip
@@ -31,6 +35,7 @@ PROCESS_NAME = 'kea-dhcp4'
 CTRL_PROCESS_NAME = 'kea-ctrl-agent'
 KEA4_CONF = '/run/kea/kea-dhcp4.conf'
 KEA4_CTRL = '/run/kea/dhcp4-ctrl-socket'
+HOSTSD_CLIENT = '/usr/bin/vyos-hostsd-client'
 base_path = ['service', 'dhcp-server']
 interface = 'dum8765'
 subnet = '192.0.2.0/25'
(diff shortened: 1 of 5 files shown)


indrajitr:
Keeping it as draft till I have a working test environment.

indrajitr:
Update: Ready for review.
This has been working fine for me in the local environment for the last few days.

sarthurdev (approved):
LGTM. Thanks for adding the smoketest.

src/etc/systemd/system/kea-dhcp4-server.service.d/override.conf
```diff
@@ -5,3 +5,5 @@ After=vyos-router.service
 [Service]
 ExecStart=
 ExecStart=/usr/sbin/kea-dhcp4 -c /run/kea/kea-dhcp4.conf
+ExecStartPost=!/usr/bin/python3 /usr/libexec/vyos/op_mode/dhcp.py update_dhcp_server_lease_to_hostd_state --family inet
```
indrajitr: Thanks for the review @sarthurdev! I need some feedback on the convention of invoking this one.
So far I haven't seen any precedence of invoking `op_mode/*.py` directly in the codebase. Is this okay, or do you want me to revisit this?

sarthurdev: it's not the ideal solution, as I don't see a use case where it'd be needed in op-mode, nor is it defined in XML.

Probably best done in a script in `src/system` like the Kea hook scripts.

indrajitr: Yes, that's why I asked :)

As you can see in the implementation of `update_dhcp_server_lease_to_hostd_state`, the method crucially depends on `_get_dhcp_pools` and `_get_raw_server_leases` which are implemented in `op_mode/dhcp.py` which is why I added the method in `update_dhcp_server_lease_to_hostd_state`. Could they be refactored to kea lib (`python/vyos/kea.py`) perhaps so that they can be reused?

Or what if we actually add an op-mode command - something like `update hostd state` and define it in XML as well?

NB: I realize that we are getting into design discussion here. We can discuss the options on Slack if necessary.

sarthurdev: Ideally there shouldn't be a scenario where the user needs to manually re-sync hostd state.

I agree that the utility functions could move to the `vyos.kea` module and be imported in op-mode and a new src/system script. Though for the module function, it should probably be just the logic that needs to be re-used (no sorting, filtering etc. Just returning the lease data as-is).

indrajitr: @sarthurdev, raised #4307 to refactor utility functions to `vyos.kea`. Would appreciate feedback on that.

indrajitr: Now that #4307 has been applied, I have made the necessary changes to better comply with the recommended convention.

sever-sever (approved):

c-po (approved):
This looks good to me. I asked @sarthurdev as our VyOS most experienced KEA user for his opinion, too.

src/system/sync-dhcp-lease-to-hosts.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+
+    # Load configuration
+    config = ConfigTreeQuery()
+
+    # Check if DHCP server is configured
+    # Using warning instead of error since this check may fail during first-time
+    # DHCP server setup when the service is not yet configured in the config tree.
+    # This happens when called from systemd's ExecStartPost the first time.
+    if not config.exists('service dhcp-server'):
```
sarthurdev: This should check v4/v6 node depending on the argument

src/system/sync-dhcp-lease-to-hosts.py
```diff
@@ -0,0 +1,111 @@
...
+        logger.warning('DHCP server is not configured')
+
+    # Check if hostfile-update is enabled
+    if not config.exists('service dhcp-server hostfile-update'):
```
sarthurdev: This should check v4/v6 node depending on the argument

sarthurdev (changes requested):
LGTM, just this last issue

src/system/sync-dhcp-lease-to-hosts.py
indrajitr: DHCPv6 doesn't support hostfile update and the script will hard fail before reaching here. But you are right, since there is a cli arg for v4/v6, we might as well have this in place.

src/system/sync-dhcp-lease-to-hosts.py
indrajitr: This config path doesn't exist anyway since DHCPv6 doesn't support hostfile update. But you are right, since there is a cli arg for v4/v6, we might as well have this in place.

sarthurdev (approved):

smoketest/scripts/cli/test_service_dhcp-server.py
```diff
@@ -821,27 +1093,187 @@ def test_dhcp_high_availability_standby(self):
...
+            ip = inc_ip(subnet, seq)
+            kea_add_lease(4, ip, host_name=client, mac_address=mac)
+
+        # 2. Verify that leases are not available in vyos-hostsd
+        tag_regex = re.escape(f'dhcp-server-{subnet.rsplit(".", 1)[0]}')
+        host_json = cmd(f'{HOSTSD_CLIENT} --get-hosts {tag_regex}')
+        self.assertFalse(host_json.strip('{}'))
+
+        # 3. Restart the service to trigger vyos-hostsd sync and wait for it to start
+        self.assertTrue(process_named_running(PROCESS_NAME, timeout=30))
```
sarthurdev, line 1270: Noticed this while working on something else, Is there meant to be a service restart here?

PR #4253, opened 2024-12-23
utils: T6975: Add 'vrf' and 'netns' arguments to functions in 'vyos.utils.process'

natali-rs1985:
## Types of changes

## Related Task(s)

* https://vyos.dev/T6975
## Component(s) name

vyos.utils
## Checklist:

### DIFF
--- a/python/vyos/utils/process.py
+++ b/python/vyos/utils/process.py
@@ -20,10 +20,23 @@ from subprocess import PIPE
 from subprocess import STDOUT
 from subprocess import DEVNULL
 
+
+def get_wrapper(vrf, netns, auth):
+    wrapper = ''
+    if vrf:
+        wrapper = f'ip vrf exec {vrf} '
+    elif netns:
+        wrapper = f'ip netns exec {netns} '
+    if auth:
+        wrapper = f'{auth} {wrapper}'
+    return wrapper
+
+
 def popen(command, flag='', shell=None, input=None, timeout=None, env=None,
-          stdout=PIPE, stderr=PIPE, decode='utf-8'):
+          stdout=PIPE, stderr=PIPE, decode='utf-8', auth='', vrf=None,
+          netns=None):
     """
-    popen is a wrapper helper aound subprocess.Popen
+    popen is a wrapper helper around subprocess.Popen
     with it default setting it will return a tuple (out, err)
     out: the output of the program run
     err: the error code returned by the program
@@ -45,6 +58,8 @@ def popen(command, flag='', shell=None, input=None, timeout=None, env=None,
               - DEVNULL, discard the output
     decode:  specify the expected text encoding (utf-8, ascii, ...)
              the default is explicitely utf-8 which is python's own default
+    vrf:     run command in a VRF context
+    netns:   run command in the named network namespace
 
     usage:
     get both stdout and stderr: popen('command', stdout=PIPE, stderr=STDOUT)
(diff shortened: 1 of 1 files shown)


python/vyos/utils/process.py
```diff
@@ -60,6 +75,9 @@ def popen(command, flag='', shell=None, input=None, timeout=None, env=None,
     if not debug.enabled(flag):
         flag = 'command'
 
+    wrapper = get_wrapper(vrf, netns, auth)
+    command = wrapper + command
```
dmbaturin:
```suggestion
    command = f'{wrapper} {command}'
```

I think it's better to use an approach that rules out any accidentally missing whitespace. If the whitespace is embedded  in the format string here, it's impossible to create a bad command by accidentally missing whitespace in different places.

dmbaturin (changes requested):
I left a minor suggestion.

Another question is that executing commands in VRFs and network namespace is only available with `sudo` at the moment. For now, we should probably add an effective UID check (there's a function for that in utils) and raise a "permission denied" error if it's not 0 (root), rather than expose users to odd error messages.

python/vyos/utils/process.py
```diff
@@ -20,10 +20,23 @@
...
 from subprocess import DEVNULL
 
+
+def get_wrapper(vrf, netns, auth):
+    wrapper = ''
+    if vrf:
+        wrapper = f'ip vrf exec {vrf} '
+    elif netns:
+        wrapper = f'ip netns exec {netns} '
+    if auth:
```
c-po, line 30: What does auth do? Where do we use it?

natali-rs1985: auth was added in this  [PR](https://github.com/vyos/vyos-1x/pull/4225#pullrequestreview-2494419628) to prevent showing username and password in error message

python/vyos/utils/process.py
```diff
@@ -139,13 +164,18 @@ def cmd(command, flag='', shell=None, input=None, timeout=None, env=None,
...
         stdout=stdout, stderr=stderr,
         input=input, timeout=timeout,
         env=env, shell=shell,
         decode=decode,
+        auth=auth,
+        vrf=vrf,
+        netns=netns,
     )
     if code not in expect:
+        wrapper = get_wrapper(vrf, netns, auth='')
```
c-po, line 179:
```suggestion
        wrapper = get_wrapper(vrf, netns, auth)
```

You overwrite `auth` in this statememnt. If one passes in `auth` to `cmd()` you will reset it to `''` in this line.

c-po (changes requested):

python/vyos/utils/process.py
natali-rs1985: In this line we generate wrapper for command without auth just to show it on the CLI on error

python/vyos/utils/process.py
```diff
@@ -60,6 +75,14 @@ def popen(command, flag='', shell=None, input=None, timeout=None, env=None,
     if not debug.enabled(flag):
         flag = 'command'
 
+    # Must be run as root to execute command in VRF or network namespace
+    if vrf or netns:
+        if os.getuid() != 0:
+            raise OSError('Permission denied')
```
dmbaturin:
```suggestion
            raise OSError('Permission denied: cannot execute commands in VRF and netns contexts as an unprivileged user')
```

Maybe make the error more specific?

dmbaturin (changes requested):
Looks good now, I left a small suggestion.

python/vyos/utils/process.py
natali-rs1985: Done

dmbaturin (approved):

PR #4259, opened 2024-12-26
frr: T6746: additional improvements after 10.2 upgrade

c-po:
## Change Summary

Some daemons (e.g. `bgpd`) have several nested substrings/sections like

```
router bgp 100
  address-family ipv4 unicast
  ..
  exit-address-family
exit
```

We can now use getFRRconfig() with the substring option to extract only
```
 address-family ipv4 unicast
 ..
 exit-address-family
```
Making config validation more granular

## Types of changes

## Related Task(s)

* https://vyos.dev/T6746

## Checklist:

### DIFF
--- a/python/vyos/frrender.py
+++ b/python/vyos/frrender.py
@@ -218,11 +218,11 @@ def get_frrender_dict(conf, argv=None) -> dict:
     # values present on the CLI - that's why we have if conf.exists()
     eigrp_cli_path = ['protocols', 'eigrp']
     if conf.exists(eigrp_cli_path):
-        isis = conf.get_config_dict(eigrp_cli_path, key_mangling=('-', '_'),
-                                    get_first_key=True,
-                                    no_tag_node_value_mangle=True,
-                                    with_recursive_defaults=True)
-        dict.update({'eigrp' : isis})
+        eigrp = conf.get_config_dict(eigrp_cli_path, key_mangling=('-', '_'),
+                                     get_first_key=True,
+                                     no_tag_node_value_mangle=True,
+                                     with_recursive_defaults=True)
+        dict.update({'eigrp' : eigrp})
     elif conf.exists_effective(eigrp_cli_path):
         dict.update({'eigrp' : {'deleted' : ''}})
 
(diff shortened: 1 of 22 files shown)


python/vyos/frrender.py
```diff
@@ -537,14 +537,25 @@ def dict_helper_pim_defaults(pim, path):
...
 
 class FRRender:
+    cached_config_dict = {}
     def __init__(self):
         self._frr_conf = '/run/frr/config/vyos.frr.conf'
 
     def generate(self, config_dict) -> None:
+        """
+        Generate FRR configuration file
+        Returns Fals if no changes to configuration were made, otherwise True
```
dmbaturin:
```suggestion
        Returns False if no changes to configuration were made, otherwise True
```

smoketest/scripts/cli/base_vyostest_shim.py
```diff
@@ -89,11 +86,7 @@ def cli_commit(self):
             # during a commit there is a process opening commit_lock, and run() returns 0
             while run(f'sudo lsof -nP {commit_lock}') == 0:
                 sleep(0.250)
-            # wait for FRR reload to be complete
-            while process_named_running('frr-reload.py'):
-                sleep(0.250)
-            # reset getFRRconfig() guard timer
-            self.commit_guard = time()
+            sleep(5)
```
dmbaturin: Can we make it a named constant and provide some considerations about the value? My immediate thought is "can we make it shorter?" or, in general, "can we reduce the wait time by doing a more precise condition check?". Let's document that knowledge.

dmbaturin (changes requested):
Just a few small suggestions.

smoketest/scripts/cli/base_vyostest_shim.py
c-po: Good one!

Added proper code documentation:
```
    The commit process is not finished until all pending files from
    VYATTA_CHANGES_ONLY_DIR are copied to VYATTA_ACTIVE_CONFIGURATION_DIR. This is
    done inside libvyatta-cfg1 and the FUSE UnionFS part. On large non-interactive
    commits FUSE UnionFS might not replicate the real state in time, leading to
    errors when querying the working and effective configuration.

    TO BE DELETED AFTER SWITCH TO IN MEMORY CONFIG
```

sever-sever (approved):

PR #4279, opened 2025-01-05
telegraf: T7017: Telegraf should log at least info to a configured logging output

ryphon:
## Change summary

Adding *.info to Loki logging templating. In my personal setup I am also including local7.debug, but I'm not quite sure how to access that map from inside the j2 template of the syslog for Telegraf.

## Types of changes

## Related Task(s)

https://vyos.dev/T7017

Adjusted the config in line in my own vyos, and validated at least that the j2 dictionary would be parsed and the *.info would be added if loki or splunk were defined in the monitoring map. 

## Checklist:

### DIFF
--- a/data/templates/telegraf/syslog_telegraf.j2
+++ b/data/templates/telegraf/syslog_telegraf.j2
@@ -2,7 +2,7 @@
 
 $ModLoad omuxsock
 $OMUxSockSocket /run/telegraf/telegraf_syslog.sock
-{% if telegraf.loki or telegraf.splunk %}
+{% if telegraf.loki is vyos_defined or telegraf.splunk is vyos_defined %}
 *.info;*.notice :omuxsock:
 {% else %}
 *.notice :omuxsock:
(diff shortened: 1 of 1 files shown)


data/templates/telegraf/syslog_telegraf.j2
```diff
@@ -2,4 +2,8 @@
 
 $ModLoad omuxsock
 $OMUxSockSocket /run/telegraf/telegraf_syslog.sock
+{% if telegraf.loki or telegraf.splunk %}
```
c-po: Please use: 

```suggestion
{% if telegraf.loki is vyos_defined or telegraf.splunk is vyos_defined %}
```

c-po (changes requested):

data/templates/telegraf/syslog_telegraf.j2
ryphon: Appreciate the suggestion, I assumed the j2 would parse 'existence' at all. I also don't think this is very scalable, it's hard coded to just understand splunk and loki. Do you agree with the foundation of "syslog should log more if a log output location is configured"?

dmbaturin (approved):

ryphon:
Bringing the comment out to here as well.

This isn't a scalable solution really, but if the 'idea' is good, I can expand upon it, and would rather do that than have this merged.

sever-sever (approved):

PR #4290, opened 2025-01-08
xml: T7029: allow wildcard in include directive

jestabro:
## Change summary

It is silly to explicitly include all version information files in
https://github.com/vyos/vyos-1x/blob/current/interface-definitions/xml-component-version.xml.in
and complicates merging of definition files. Add glob to the `transclude-template` script.

The only difference in the resulting file is that `xml-component-version.xml` is sorted.

## Types of changes

         Remove annoyance.

## Checklist:


c-po (approved):
I wonder why we need `interface-definitions/xml-component-version.xml.in` at all. As this now globs all files present, why not generate it explicitly? I see no benefit in relying on a file that only has the glob path as input?

Fully autogenerating the result by adding the explicit glob path into the transclude-template could further simplify it.

jestabro:
Good point, @c-po --- this was the quick fix for @sever-sever 's work, but I will take another look. I'm inclined to leave it as is for the moment.

jestabro:
Yes, my hesitation was for good reason: the syntax version has always occupied a unique position --- we want to keep it with the XML definitions, but need to process it separately for the Python cache; we require a single file with all inclusions to parse by xmltodict, whereas all other files are handled by a general mechanism.

dmbaturin (approved):

PR #4305, opened 2025-01-15
syslog: T6989: convert old configuration format to "advanced"

c-po:
## Change summary

We are still using the legacy config format for rsyslog in https://github.com/vyos/vyos-1x/blob/bb4611544ff73a979f927a7a1c8ea3dd1c56fb48/data/templates/rsyslog/rsyslog.conf.j2

It has been superseded by a new format:

https://www.rsyslog.com/doc/configuration/converting_to_new_format.html

There are some associated issues. For example, we use `$markMessagePeriod` for `set system syslog global marker interval`, which sends mark messages unconditionally. New rsyslog versions support sending mark messages only if there are no real messages, the old directive is now listed as obsolete (https://www.rsyslog.com/doc/configuration/input_directives/rsconf1_markmessageperiod.html) and its replacement seems to be only available in the advanced format.

*  Drop `file` CLI option to log to dedicated files.
*  Drop `user` CLI option to always log to user terminal - should be an op-mode command instead
* Rename `host <ipv4|ipv6|fqdn>` -> `remote <ipv4|ipv6|fqdn>`
* Add new source-address option for outgoing connections under `remote source-address <ipv4|ipv6|fqdn>`
* VRF can be defined per remote as  `remote vrf <name>`
* Move `global preserve-fqdn` -> `preserve-fqdn` - move up one CLI level
* Move `global marker interval` -> `marker interval` - move up one CLI level
* Rename `global` -> `local` as every facility configured under global is actually logged locally to `/var/log/messages`

## Types of changes

## Related Task(s)

* https://vyos.dev/T6989

## Related PR(s)

* https://github.com/vyos/vyos-documentation/pull/1589
* https://github.com/vyos/vyos-build/pull/898

```
vyos@vyos:~$ /usr/libexec/vyos/tests/smoke/cli/test_system_syslog.py
test_console (__main__.TestRSYSLOGService.test_console) ... ok
test_global (__main__.TestRSYSLOGService.test_global) ... ok
test_remote (__main__.TestRSYSLOGService.test_remote) ... ok
test_vrf_source_address (__main__.TestRSYSLOGService.test_vrf_source_address) ... ok

Ran 4 tests in 9.964s

OK
```

## Checklist:


data/templates/rsyslog/rsyslog.conf.j2
```diff
@@ -1,78 +1,123 @@
...
+{%     set tmp = [] %}
+{%     if local.facility is vyos_defined %}
+{%         for facility, facility_options in local.facility.items() %}
+{%             set _ = tmp.append(facility.replace('all', '*') ~ "." ~ facility_options.level.replace('all', 'debug')) %}
+{%         endfor %}
+if prifilt("{{ tmp | join(',') }}") then {
+    action(
+        type="omfile"
+        file="/var/log/messages"
+        queue.size="262144"
```
dmbaturin: Can we include some comment about this number? If I see a non-round number, I always start wondering if it's either completely random (and can be arbitrary) or it was calculated in some complex and precise way (and dangerous to change).

dmbaturin (changes requested):

data/templates/rsyslog/rsyslog.conf.j2
c-po: It's a relict from our history:

https://github.com/vyos/vyos-1x/blob/crux/src/conf_mode/syslog.py#L132

We probably should rethink this behavior and decide for a new/the same number and update it with a proper comment.

c-po: Adjusted

dmbaturin (approved):

jestabro:
Note that the smoketest `test_console` raises errors, although it passes as unittest.

c-po:
> Note that the smoketest `test_console` raises errors, although it passes as unittest.

Thank you - issue fixed in commit 9212b401d33

jestabro (approved):

Apachez-:
It looks like this commit wasnt properly tested before pushed to the public tree according to:

https://github.com/vyos/vyos-nightly-build/actions/runs/13210103854/job/36882252374#step:5:2710

https://github.com/vyos/vyos-1x/blame/38dacfeb8fb66d135f8ad219ff3c2fc7504f3029/smoketest/scripts/cli/test_system_syslog.py#L62

Above seems the cause for why there have been no succeeded nightly builds since 3th feb (today is 8th feb):

https://github.com/vyos/vyos-nightly-build/releases

https://github.com/vyos/vyos-nightly-build/actions

PR #4380, opened 2025-03-05
T4406: Add public API endpoint to display information

oniko94:
## Change summary

Adds a public endpoint `/info` to the HTTP(S) API to display some basic information about the router, namely router's hostname, VyOS version and pre-login banner. 

The endpoint accepts `version` and `hostname` query parameters, set to valid boolean values (1, 0, 'yes', 'no' etc. are acceptable as well), which dictate whether to include respective data into the response. 
- If values are set to `False` or equivalent, a correspondent field in the response object is set to an empty string. 
- If no query parameters are supplied, all three fields are populated with actual data by default. 
- The query parameters' type is validated. Extra parameters besides `version` and `hostname` are not allowed. 
## Types of changes

## Related Task(s)

https://vyos.dev/T4406

- Build the debian package for vyos-1x
- Install it to vyos-1x instance
- Enable the HTTP API
- Send a GET request to the API endpoint, for example `curl --insecure --location --request GET 'https://<INSTANCE_ADDRESS>/info`
 - Try different combinations of query parameters, for example `?version=0&hostname=1` or `?hostname=false` or `?version=yes&hostname=0`
## Checklist:

### DIFF
--- a/src/services/vyos-http-api-server
+++ b/src/services/vyos-http-api-server
@@ -20,18 +20,22 @@ import grp
 import json
 import logging
 import signal
+import traceback
 from time import sleep
+from typing import Annotated
 
-from fastapi import FastAPI
+from fastapi import FastAPI, Query
 from fastapi.exceptions import RequestValidationError
 from uvicorn import Config as UvicornConfig
 from uvicorn import Server as UvicornServer
 
 from vyos.configsession import ConfigSession
 from vyos.defaults import api_config_state
+from vyos.utils.file import read_file
+from vyos.version import get_version
 
 from api.session import SessionState
-from api.rest.models import error
+from api.rest.models import error, InfoQueryParams, success
 
 CFG_GROUP = 'vyattacfg'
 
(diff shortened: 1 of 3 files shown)


src/services/vyos-http-api-server
```diff
@@ -57,11 +61,49 @@ app = FastAPI(debug=True,
...
+            res.update(version=get_version())
+
+        if show_hostname:
+            try:
+                hostname = read_file(hostname_file)
+            except Exception:
+                hostname = 'vyos'
+            res.update(hostname=hostname)
+
+        banner = read_file(prelogin_file, defaultonfailure=f'{default}!')
```
dmbaturin: Maybe put the exclamation mark in `default` rather than in the format string? If we need an exclamation mark at all, that is!!! ;)

dmbaturin (approved):
Other than the trivial question of the exclamation mark, I see nothing against this. Brute force attack operators typically blast exploits at everything without any checks, so exposing the version string to the public is not a big risk even in public networks.

jestabro (commented):
@oniko94 why not keep this endpoint with the others in api/rest/routers.py ? I can think of reasons to do it the way that you have, just curious as to the reasoning of your decision ...

oniko94:
@jestabro I followed mainly two reasons behind putting this endpoint in this file:

1. 'src/services/api/rest/routers' contains the routes that all require authentication by an API key. Meanwhile, if I understood the ticket correctly, we would want this endpoint to be publicly available for anonymous users. It would require to declare additional FastAPI router instance in this case, which could potentially cause confusion which router does what.
2. In my opinion it would also make more sense to have this endpoint active as soon as the HTTPS API service is activated, even if the user hasn't enabled the REST API specifically, or requires only the GraphQL HTTPS API. Thus, it seemed logical to me to keep it at the FastAPI app initialization level

jestabro:
I fully agree; thanks.

jestabro (approved):

src/services/vyos-http-api-server
oniko94: Fixed, thank you!

PR #4413, opened 2025-03-25
T7278: Remove cracklib hack from postconfig script template

oniko94:
## Change summary

Remove the cracklib file generation hack from postconfig script template - it should be copied now during the image generation/update.
## Types of changes

## Related Task(s)

https://vyos.dev/T7278
## Related PR(s)

Blocked by: https://github.com/vyos/vyos-build/pull/935

1. Get an older rolling image which does not contain changes introduced with T6353 implementation, for example [this](https://github.com/vyos/vyos-nightly-build/releases/download/1.5-rolling-202502220006/vyos-1.5-rolling-202502220006-generic-amd64.iso)
2. Ensure the linked PR to vyos-build is merged, or checkout to respective branch locally
3. Build a new ISO image
4. Boot into the older image - it should not contain cracklib. Update with the freshly built image (`add system image`)
5. Boot into the new image. Ensure `/var/cache/cracklib` directory exists.
6. Try creating a user in configuration mode, preferably with a weak password. You should have a password complexity check warning displayed as expected. It should not throw an exception.
## Checklist:

### DIFF
--- a/debian/vyos-1x.postinst
+++ b/debian/vyos-1x.postinst
@@ -195,10 +195,6 @@ if [ ! -x $PRECONFIG_SCRIPT ]; then
 EOF
 fi
 
-# cracklib-runtime default database location
-CRACKLIB_DIR=/var/cache/cracklib
-CRACKLIB_DB=cracklib_dict
-
 # create /opt/vyatta/etc/config/scripts/vyos-postconfig-bootup.script
 POSTCONFIG_SCRIPT=/opt/vyatta/etc/config/scripts/vyos-postconfig-bootup.script
 if [ ! -x $POSTCONFIG_SCRIPT ]; then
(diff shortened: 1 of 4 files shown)


jestabro (commented):
Note that some smoketests don't pass due to running without the companion PR for vyos-build (https://github.com/vyos/vyos-build/pull/935); with it smoketests look fine. However, this brings up a point: as absence or corruption of the /var/cache/cracklib directory causes critical failure in adding users, one may consider bailing out on internal failure (since the consequence is only a warning anyway), for example:

```
diff --git a/python/vyos/utils/auth.py b/python/vyos/utils/auth.py
index a27d8a28a..879dc5e79 100644
--- a/python/vyos/utils/auth.py
+++ b/python/vyos/utils/auth.py
@@ -32,6 +32,7 @@ class EPasswdStrength(StrEnum):
     WEAK = 'Weak'
     DECENT = 'Decent'
     STRONG = 'Strong'
+    ERROR = 'Error'
 
 
 def calculate_entropy(charset: str, passwd: str) -> float:
@@ -63,6 +64,9 @@ def evaluate_strength(passwd: str) -> dict[str, str]:
             msg = f'should not be {e}'
         result.update(strength=EPasswdStrength.WEAK)
         result.update(error=WEAK_PASSWORD_MSG.replace('@MSG@', msg))
+    except Exception as e:
+        result.update(strength=EPasswdStrength.ERROR)
+        result.update(error=f'passwd strength evaluation not available: {e}')
     else:
         # Now check the password's entropy
         # Cast to Decimal for more precise rounding
```
The example is only a suggestion, but the recommendation would be to gracefully catch the general exception in case of corruption.

oniko94:
@jestabro thanks for the tip - I've updated my PR with the requested change. I believe the most sensible approach would be displaying a warning for now (same as with a weak password).

dmbaturin (approved):
I would not combine changes that are not directly related, although I have nothing against the substance of either change.

jestabro (approved):

PR #4429, opened 2025-03-31
installer: T7301: remove support for GnuPG signatures

dmbaturin:
## Change summary

I removed support for GPG but kept the logic that allows multiple signature types, in case the elliptic curve algorithm used by minisign is somehow compromised or some new and cool tool appears and we want to add it (unlikely, but nothing is impossible).

vyos-1x Debian package has no explicit dependency on GnuPG, so there's no need to remove any dependencies here. 

## Types of changes

- [x] Other (please describe): obsolete functionality removal

## Checklist:


c-po (approved):
As minisign was introduced in VyOS 1.2 there is a safe upgrade path from 1.2 -> 1.3 -> 1.4 to always have verified image upgrade.

Logic seems correct, too

sever-sever (approved):
For removing GPG the code  changes looks correct

PR #4444, opened 2025-04-10
bridge: T7322:  fix slow performance of allowed vlan

l0crian1:
## Change summary

Ranges that are configured in allowed-vlans (e.g. 1-4094), are deconstructed into all of their individual VLANs, which is unnecessary for the execution of these:
`bridge vlan add dev {interface} vid {vlan} master`
`bridge vlan del dev {interface} vid {vlan} master`

This leads to a very long commit time when doing large ranges. This change eliminates that unnecessary deconstruction.

Example:
Execution time of `allowed-vlans 1-4094` in current state: 38.94 seconds
Execution time of `allowed-vlans 1-4094` without unnecessary deconstruction of range: 4.94 seconds
## Types of changes

- [x] Other (please describe): Performance improvement
## Related Task(s)

https://vyos.dev/T7322

```set interfaces bridge br0 member interface eth0 allowed-vlan 1-4094```
You will see a massive  improvement on commit times, especially on slower systems.
## Checklist:

### DIFF
--- a/python/vyos/configdict.py
+++ b/python/vyos/configdict.py
@@ -626,6 +626,23 @@ def get_vlan_ids(interface):
 
     return vlan_ids
 
+def get_vlans_ids_and_range(interface):
+	vlan_ids = set()
+
+	vlan_filter_status  = json.loads(cmd(f'bridge -j -d vlan show dev {interface}'))
+
+    if vlan_filter_status is not None:
+        for interface_status in vlan_filter_status:
+			for vlan_entry in iface.get("vlans", []):
+				start = vlan_entry["vlan"]
+				end = vlan_entry.get("vlanEnd")
+				if end:
+					vlan_ids.add(f"{start}-{end}")
+				else:
+					vlan_ids.add(str(start))
+
+	return vlan_ids
+
 def get_accel_dict(config, base, chap_secrets, with_pki=False):
     """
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: python/vyos/configdict.py [hunk 1]
```diff
@@ -626,6 +626,23 @@ def get_vlan_ids(interface):
 
     return vlan_ids
 
+def get_vlans_ids_and_range(interface):
```
c-po, line 629: The current only reference to this helper is in `python/vyos/ifconfig/bridge.py` - it probably should reside there for now. Other then that the change looks legit.

c-po (changes requested):

python/vyos/configdict.py
l0crian1: @c-po I was replacing the previously used function that resided within configdict. I saw that in `python/vyos/ifconfig/bridge.py` there are no "get" functions, so I didn't know if it was canonical to not put get functions in there.

Just let me know if you 100% want me to move that directly into `python/vyos/ifconfig/bridge.py`. I'm fine with either way.

c-po: Keep it

c-po (approved):
Improved logic feels better then before and there is a significant boost in performance

dmbaturin (approved):
It worked for me in a simple test and seems safe for backports.

PR #4460, opened 2025-04-18
syslog: T7367: ensure rsyslog is registered as default systemd syslog service

c-po:
## Change summary

The default syslog implementation should make syslog.service a symlink to itself, so that this socket activates the right actual syslog service.

## Types of changes

## Related Task(s)

* https://vyos.dev/T7367

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/951

```
cpo@LR1.wue3:~$ /usr/libexec/vyos/tests/smoke/cli/test_system_syslog.py
test_basic (__main__.TestRSYSLOGService.test_basic) ... ok
test_console (__main__.TestRSYSLOGService.test_console) ... ok
test_remote (__main__.TestRSYSLOGService.test_remote) ... ok
test_vrf_source_address (__main__.TestRSYSLOGService.test_vrf_source_address) ... ok

Ran 4 tests in 13.359s

OK
```

## Checklist:

### DIFF
--- a/debian/vyos-1x.links
+++ b/debian/vyos-1x.links
@@ -1,3 +1,2 @@
 /etc/netplug/linkup.d/vyos-python-helper /etc/netplug/linkdown.d/vyos-python-helper
 /usr/libexec/vyos/system/standalone_root_pw_reset /opt/vyatta/sbin/standalone_root_pw_reset
-/lib/systemd/system/rsyslog.service /etc/systemd/system/syslog.service
--- a/python/vyos/defaults.py
+++ b/python/vyos/defaults.py
@@ -43,7 +43,7 @@ directories = {
 }
 
 systemd_services = {
-    'rsyslog' : 'rsyslog.service',
+    'syslog' : 'syslog.service',
     'snmpd' : 'snmpd.service',
 }
 
--- a/smoketest/scripts/cli/test_system_syslog.py
+++ b/smoketest/scripts/cli/test_system_syslog.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
 import unittest
 
 from base_vyostest_shim import VyOSUnitTestSHIM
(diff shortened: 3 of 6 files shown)


zdc:
**Question:** After this change, will syslog messages (`/dev/log`) be redirected to `rsyslog` or `journald`?

One of the goals for the not-so-distant future is to be able to store logs in memory only, for systems without read-write storage. This is easily achievable if *all* logs are processed by `journald` first, and other services like `rsyslog` are used only to redirect some of them to remote hosts. Additionally, `journald` is an established standard in many modern Debian-based systems as the main log processing system. Also, this behavior was exactly what we had until the recent `rsyslog` config refactoring.

Therefore, the question is: will this change restore the expected behavior?

```
lrwxrwxrwx 1 root root 28 Apr 20 10:34 /dev/log -> /run/systemd/journal/dev-log
```

nvollmar:
@zdc within a container there is no `/dev/log` unless it is manually volume mapped, independent of the log driver configured in Podman.

Podman collects logs written to stdout within the container, and depending on the log driver they are stored within a file per container (`k8s-file`) or forwarded to journald on the host (`journald`).

So to get logs from container into journald on the host there are these options:
* volume mount a directory and write logs from the container there, then have the host rsyslog reading those files
* volume mount /dev/log to the container and have logs written there (which can cause permission issues if the app runs as non-root in the container)
* use the journald log driver (which is the easiest, more modern way to do it)

The journald log driver is also the default for Podman 5.

dmbaturin (approved):
That logic is straightforward and easy to backport to LTS branches, no objections.

c-po:
> ```
> lrwxrwxrwx 1 root root 28 Apr 20 10:34 /dev/log -> /run/systemd/journal/dev-log
> ```

Yes 


```bash
cpo@LR1.wue3:~$ ls -al /dev/log
lrwxrwxrwx 1 root root 28 Apr 22 23:15 /dev/log -> /run/systemd/journal/dev-log
```

sever-sever (approved):
Fixes logging system

PR #4469, opened 2025-04-24
bgp: T7220: Add the option to disable enforce-first-as at peer level

canoziia:
## Change summary

Since FRR 10, `enforce-first-as` is enabled by default https://frrouting.org/release/10.0/ . When connecting to route servers at IXPs, we must disable the first AS enforcement (https://docs.frrouting.org/en/latest/bgp.html#clicmd-bgp-enforce-first-as).

To keep vyos's behavior consistent with previous, I think the default should still be kept as disabled.

## Types of changes

## Related Task(s)

https://vyos.dev/T7220

I have tested the command `no neighbor {{ neighbor }} enforce-first-as` in vtysh.

## Checklist:

### DIFF
--- a/data/templates/frr/bgpd.frr.j2
+++ b/data/templates/frr/bgpd.frr.j2
@@ -98,6 +98,8 @@
 {% endif %}
 {% if config.enforce_first_as is vyos_defined %}
  neighbor {{ neighbor }} enforce-first-as
+{% else %}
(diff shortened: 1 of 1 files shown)


c-po (approved):
Good explanation of the issue and proper solution. Thanks

dmbaturin (approved):
This just restores the old behavior, so I have no objections.

PR #4482, opened 2025-04-29
policy: T5069: large-community-list regex validator disallows whitespace

talmakion:
## Change summary

NOTE: Potentially hijacking an old ticket with a similar problem. 

T5816 improved 1.4/1.5's strict validation to allow full regex on a single large-community tuple. However, it is common to match multiple components of a large-community string via regex in earlier VyOS versions (not to mention Cisco IOS). T5816 is too strict about the form of the regex pattern, does not allow the common-separator wildcard '_' and whitespace, breaking old 1.3 configs and preventing simple pattern logic being used to match multiple communities. 

To resolve T5069's concerns and my own, this patch attempts to:
 * Perform basic, easy sanity checks that should always pass:
   * Contain at least one well-formed tuple
   * Permit only specific characters valid in community strings and POSIX regex
   * Rejecting obviously incorrect regular expressions, treating Python regex as a superset of POSIX
 * Run a final "sanity check" regex across the top, issue a warning if the overall pattern is not considered well formed

"Well formed" in this case would be one or more community segments roughly fitting the original check, separated by proper whitespace. "1:1:12:2:2" might be a mistake for "1:1:1 2:2:2", but there are still instances of useful patterns not following this form. 

Since there's no mechanism for providing warning feedback from a validator and validator output is thrown away unless an error, I've wrapped this in the console_warning() function and it just prints directly to the TTY. If this seems like overkill, if it could be more generally useful, or if there's a suitable mechanism I missed - I imagine validation success/warning/failure feedback could be useful at API level, not just CLI - let me know and I'll update it to match. 

If we retain the warning, I'll likely issue a followup patch for the documentation explaining the specific validation logic and what to check when the warning pops up. 

## Types of changes

## Related Task(s)

* https://vyos.dev/T5069
* https://vyos.dev/T5816 - only for reference

```
# Example tests via CLI:
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1"

  
  
  
  Malformed large-community-list
  Value validation failed
  Set failed

[edit]
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1:1"
[edit]
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1:12:2:2"
WARNING: Regex does not follow expected form and may not match as expected.
[edit]
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1:1 2:2:2"
[edit]
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1:1_.*_2:2:2"
[edit]
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1:1_.*_2:2:2:"
WARNING: Regex does not follow expected form and may not match as expected.
[edit]
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1:1_.*(_2:2:2)"
[edit]
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1:1_.*(_2:2:2))"

  
  
  
  Malformed large-community-list
  Value validation failed
  Set failed

[edit]
```

I've checked that these compound matches are accepted by FRR and work in a production BGP environment as the older versions of FRR in 1.3 did. 

Smoketests, test_policy.py:
```
Ran 28 tests in 375.194s

OK
```

test_protocols_bgp.py:
```
Ran 30 tests in 438.052s

OK
```

## Checklist:

### DIFF
--- a/src/validators/bgp-large-community-list
+++ b/src/validators/bgp-large-community-list
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 #
-# Copyright (C) 2021-2023 VyOS maintainers and contributors
+# Copyright (C) 2021-2025 VyOS maintainers and contributors
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 2 or later as
@@ -17,18 +17,27 @@
 import re
 import sys
 
-pattern = '(.*):(.*):(.*)'
-allowedChars = { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '.', '+', '*', '?', '^', '$', '(', ')', '[', ']', '{', '}', '|', '\\', ':', '-' }
+allowedChars = { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '.', '+', '*', '?', '^', '$', '(', ')', '[', ']', '{', '}', '|', '\\', ':', '-', '_', ' ' }
 
 if __name__ == '__main__':
     if len(sys.argv) != 2:
         sys.exit(1)
 
-    value = sys.argv[1].split(':')
-    if not len(value) == 3:
+    value = sys.argv[1]
+
+    # Require at least one well-formed large-community tuple in the pattern. 
+    tmp = value.split(':')
+    if len(tmp) < 3:
         sys.exit(1)
 
-    if not (re.match(pattern, sys.argv[1]) and set(sys.argv[1]).issubset(allowedChars)):
+    # Simple guard against invalid community & 1003.2 pattern chars
+    if not set(value).issubset(allowedChars):
+        sys.exit(1)
+
+    # Don't feed FRR badly formed regex
+    try:
+        re.compile(value)
+    except re.error:
         sys.exit(1)
 
     sys.exit(0)
--- a/src/conf_mode/policy.py
+++ b/src/conf_mode/policy.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import re
 from sys import exit
 
 from vyos.config import Config
@@ -24,8 +25,19 @@ from vyos.frrender import get_frrender_dict
 from vyos.utils.dict import dict_search
 from vyos.utils.process import is_systemd_service_running
 from vyos import ConfigError
+from vyos.base import Warning
 from vyos import airbag
 airbag.enable()
+
+# Sanity checks for large-community-list regex:
+# * Require complete 3-tuples, no blank members. Catch missed & doubled colons. 
+# * Permit appropriate community separators (whitespace, underscore)
+# * Permit common regex between tuples while requiring at least one separator
+#   (eg, "1:1:1_.*_4:4:4", matching "1:1:1 4:4:4" and "1:1:1 2:2:2 4:4:4",
+#        but not "1:1:13 24:4:4")
+# Best practice: stick with basic patterns, mind your wildcards and whitespace.
+# Regex that doesn't match this pattern will be allowed with a warning. 
+large_community_regex_pattern = r'([^: _]+):([^: _]+):([^: _]+)([ _]([^:]+):([^: _]+):([^: _]+))*'
 
 def community_action_compatibility(actions: dict) -> bool:
     """
@@ -146,6 +158,10 @@ def verify(config_dict):
                                    'large_community_list']:
                     if 'regex' not in rule_config:
                         raise ConfigError(f'A regex {mandatory_error}')
+
+                if policy_type == 'large_community_list':
+                    if not re.fullmatch(large_community_regex_pattern, rule_config['regex']):
+                        Warning(f'"policy large-community-list {instance} rule {rule} regex" does not follow expected form and may not match as expected.')
 
                 if policy_type in ['prefix_list', 'prefix_list6']:
                     if 'prefix' not in rule_config:


src/validators/bgp-large-community-list
```diff
@@ -17,18 +17,47 @@
 import re
 import sys
 
-pattern = '(.*):(.*):(.*)'
-allowedChars = { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '.', '+', '*', '?', '^', '$', '(', ')', '[', ']', '{', '}', '|', '\\', ':', '-' }
+def console_warning(txt: str):
+    with open('/dev/tty', 'w') as tty:
```
dmbaturin: We already have a thing specially for this purpose: [`vyos.base.Warning`](https://github.com/vyos/vyos-1x/blob/current/python/vyos/base.py#L47). Warnings also should be printed in non-interactive script calls as well, for testing, for example.

dmbaturin (changes requested):
The fix looks sensible but please use the standard warning mechanism instead of a custom alternative.

src/validators/bgp-large-community-list
talmakion: Cheers for pointing me in that direction - I wasn't sure if there was already a generic mechanism for warnings. 

The problem is that Warning() output is still eaten by the validation caller, same as any normal output on stdout/stderr, unless we exit non-zero. 

Change console_warning to a Warning instance, no more warning output:
```
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1:12:2:2"
[edit]
```

Running it raw:
```
vyos@TEST-VYOS-LEFT# python3 /usr/libexec/vyos/validators/bgp-large-community-list "1:1:12:2:2"

WARNING: Regex does not follow expected form and may not match as
expected.
```

Change the sys.exit(0) to sys.exit(1) (it's definitely hitting it):
```
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1:12:2:2"

  
  WARNING: Regex does not follow expected form and may not match as
  expected.
  
  
  Malformed large-community-list
  Value validation failed
  Set failed

[edit]
```

Not sure what the neatest way forward would be: 
* I can customise the Warning.print() handling for this specific use, also output to /dev/tty
* If it's useful elsewhere, add a flag to enable output to /dev/tty in the base implementation
* Get vyos1x-config to still display output without failing when flagged to do so, maybe a different exit code? Does anything else interact with the validators directly?

I don't particularly like the forced direct-to-TTY but it was the only way I could find to actually produce output for the user to see.

talmakion: I've gone with using exit codes from the validator to signal there's non-failing output and changed the warning to use the normal wrapper.

I've raised a [PR](https://github.com/vyos/vyos-utils/pull/32) to have validate_value handle this correctly. 

Combined, this works pretty close to what I was after:
```
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1:1 2:2:2"
[edit]
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1:12:2:2"
  
  WARNING: Regex does not follow expected form and may not match as
  expected.

[edit]
vyos@TEST-VYOS-LEFT# set policy large-community-list TEST rule 20 regex "1:1"
  
  
  Malformed large-community-list
  Value validation failed
  Set failed

[edit]
```

talmakion: Following on the comments on the [other PR](https://github.com/vyos/vyos-utils/pull/32), I've updated this to include a UserMessage with redirection capability and a validator-script specific warning wrapper to send everything to stderr.

I've made the assumption you probably don't want specific warning wrappers in vyos.base, and that a specific wrapper is better than a generic WarningStderr() or similar. 

I haven't had a chance yet to exercise other validators to find any that were chatty on stderr - I vaguely recall some were in older VyOS releases - but if this basic form suits, I can give them a look over.

src/validators/bgp-large-community-list
```diff
@@ -16,19 +16,45 @@
...
-    if not len(value) == 3:
+    value = sys.argv[1]
+
+    # Require at least one well-formed large-community tuple in the pattern. 
+    tmp = value.split(':')
+    if len(tmp) < 3:
+        sys.exit(1)
+
+    # Simple guard against invalid community & 1003.2 pattern chars
+    if not set(value).issubset(allowedChars):
```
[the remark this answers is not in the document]
talmakion: This is already present.

src/validators/bgp-large-community-list
```diff
@@ -16,19 +16,45 @@
...
+    try:
+        re.compile(value)
+    except re.error:
         sys.exit(1)
 
+    if not re.fullmatch(pattern, value):
+        # If you know what you're doing, you should be able to do it.
+        # It is, however, very easy to over- or under-restrict compound regex 
+        # logic against communities and ASNs, especially if they change. 
+        # Best practice: stick with basic patterns, mind your wildcards and whitespace.
```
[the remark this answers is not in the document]
talmakion: This is no longer part of the change.

talmakion:
Unsure why the smoketest failed - it is UserMessage/Warning-related, but I can't get it to fail in local. Comparing a vanilla VyOS with an updated one, manual smoketest runs and direct config sets work as expected. 

I did muff the previous force-push so I've just tweaked minor Ruff lint fails and some naming that didn't track changes in vyos.base to trigger another smoketest run. 

Edit: It was a vyos-configd interaction, need to use runtime values of sys.stdout.

talmakion:
Updated version: split off the warning-logic into conf verify and dropped output changes for validators. 

Re-ran smoketests for policy OK.

sever-sever (approved):
Fixes BGP large-community regex
Smoketest passed. Approved.

dmbaturin (approved):
The warning is now in the config scripts. No objections against the logic.

PR #4485, opened 2025-04-30
https: T7393: set listen-address bind fails silently without restart

jestabro:
## Change summary

The config-mode script `service_https.py` apply stage calls `systemctl reload-or-restart` on the https server, however, some settings require an explicit restart or will silently fail, since nginx drops privileges after start up.

Distinguish cases where a service restart may be needed and check in apply stage.

Add smoketest.

## Types of changes

## Checklist:

### DIFF
--- a/smoketest/scripts/cli/test_service_https.py
+++ b/smoketest/scripts/cli/test_service_https.py
@@ -16,6 +16,7 @@
 
 import unittest
 import json
+import psutil
 
 from requests import request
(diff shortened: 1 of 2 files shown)


sever-sever (approved):
Fixes restart NGINX if IP listen address was changed.

sarthurdev (approved):
Forces service restart if nginx listen address/port etc is changed

PR #4489, opened 2025-04-30
installer: T7420: pass image download credentials in environment variables

dmbaturin:
## Change summary

At the moment, we pass usernames and passwords in inline environment variables embedded in shell commands.

That approach has some security implications. Since credentials are a part of the command, they are visible in the process list, and can potentially leak into tech supports reports and the like when the stars are right those files are generated exactly at the moment when a download script is running. Not very likely, but not impossible.

Passing them in process environment variables is much safer in that regard.

As a side effect, this PR also removes the only remaining use of the `auth` argument of `vyos.utils.process` helpers and once it's merged, #4323 can also be merged without modifications.

## Types of changes

I used someone's [extension of `http.serve`](https://gist.github.com/lionelyoung/8cad668d4d30fa392842fa08d50d2bc6).

```
server# $ python ./serve.py -u dmbaturin -p qwerty -d .
Serving HTTP on 0.0.0.0 port 8000 (http://0.0.0.0:8000/) ...

vyos@vyos-lab-1:~$ add system image http://192.168.56.1:8000/vyos-2025.04.29-0019-rolling-generic-amd64.iso vrf MGMT username dmbaturin password qwerty
Validating signature
Signature is not valid. Do you want to continue with installation? [y/N] y
Validating image checksums
What would you like to name this image? (Default: 2025.04.29-0019-rolling) 
Would you like to set the new image as the default one for boot? [Y/n] 
An active configuration was found. Would you like to copy it to the new image? [Y/n] 
Copying SSH host keys
Copying system image files
Cleaning up
Unmounting target filesystems
Removing temporary files

```

## Checklist:

### DIFF
--- a/src/op_mode/image_installer.py
+++ b/src/op_mode/image_installer.py
@@ -565,21 +565,18 @@ def validate_signature(file_path: str, sign_type: str) -> None:
         print('Signature is valid')
 
 def download_file(local_file: str, remote_path: str, vrf: str,
-                  username: str, password: str,
                   progressbar: bool = False, check_space: bool = False):
-    environ['REMOTE_USERNAME'] = username
-    environ['REMOTE_PASSWORD'] = password
+    # Server credentials are implicitly passed in environment variables
+    # that are set by add_image
     if vrf is None:
         download(local_file, remote_path, progressbar=progressbar,
                  check_space=check_space, raise_error=True)
     else:
-        remote_auth = f'REMOTE_USERNAME={username} REMOTE_PASSWORD={password}'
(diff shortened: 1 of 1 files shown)


natali-rs1985:
Everything work as expected. And also since it was only one place where credentials were passed to cmd we can get rid of non used (already) `auth` argument in functions from `vyos.utils.process`

jestabro (approved):
A nice simplification with the benefit of allowing
https://github.com/vyos/vyos-1x/pull/4323

sever-sever (approved):

PR #4497, opened 2025-05-05
T7432: RPKI VRF Support

yzguy:
## Change summary

Add support for RPKI in a VRF

Op-mode commands

```
show rpki as-number <asn>  vrf <vrf>
show rpki cache-connection vrf <vrf>
show rpki cache-server vrf <vrf>
show rpki prefix <prefix> vrf <vrf>
show rpki prefix <prefix> as-number <asn> vrf <vrf>
show rpki prefix-table vrf <vrf>
reset rpki vrf <vrf>
```

Configuration Commands

```
set vrf name <name> protocols rpki cache <cache> port <port>
set vrf name <name> protocols rpki cache <cache> preference <pref>
set vrf name <name> protocols rpki cache <cache> source-address <src>
set vrf name <name> protocols rpki cache <cache> ssh username <username>
set vrf name <name> protocols rpki cache <cache> ssh key <key>
set vrf name <name> protocols rpki polling-period <rolling-period>
set vrf name <name> protocols rpki retry-interval <retry-interval>
```

## Types of changes

## Related Task(s)

https://vyos.dev/T7432

## Checklist:


yzguy:
Need to add to documentation
Need to add smoketests

c-po (changes requested):
You do not set a VRF table id in your smoketests which makes them fail:


```
ERROR: test_rpki_ssh (__main__.TestProtocolsRPKI.test_rpki_ssh)
  File "/usr/libexec/vyos/tests/smoke/cli/test_protocols_rpki.py", line 211, in test_rpki_ssh
    self.cli_commit()
  File "/usr/libexec/vyos/tests/smoke/cli/base_vyostest_shim.py", line 103, in cli_commit
    out = self._session.commit()
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/vyos/configsession.py", line 280, in commit
    out = self.__run_command([COMMIT])
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/vyos/configsession.py", line 210, in __run_command
    raise ConfigSessionError(output)
vyos.configsession.ConfigSessionError: [ vrf ]
VRF "blue" table id is mandatory!
[[vrf]] failed
[ pki ]
Embedded PKI openssh with name "rpki-smoketest" is still in use by CLI
path "protocols rpki cache 192.0.2.3 ssh key"
[[pki]] failed
[[vrf name blue protocols rpki]] failed
Commit failed
```

python/vyos/frrender.py
```diff
@@ -675,7 +690,7 @@ def inline_helper(config_dict) -> str:
                 output += render_to_string('frr/ripngd.frr.j2', config_dict['ripng'])
                 output += '\n'
             if 'rpki' in config_dict and 'deleted' not in config_dict['rpki']:
-                output += render_to_string('frr/rpki.frr.j2', config_dict['rpki'])
+                output += render_to_string('frr/rpki.frr.j2', {'rpki': config_dict['rpki']})
```
c-po, line 693: Why is this extra wrap required when other VRF aware implementations like BGP do not require it?

c-po (changes requested):
The proposed CLI op-mode commands differ from other VRF enabled implementations:

python/vyos/frrender.py
yzguy: It's intention is so the macro can be reused easily, it allows you to pass an object containing all the variables into the template rather than all the variables by themselves which means you'd have to pass every single one into the macro as a parameter as well

```jinja2
{% macro rpki_config(cache_servers, expire_interval, retry_interval, polling_period) %}
  {% for cache_server in cache_servers %}
  rpki cache tcp ....
  {% endfor %}
  {% if expire_interval is vyos_defined %}
  rpki expire_interval {{ expire_interval }}
  {% endif %}
{% endmacro %}

{{ rpki_config(cache_servers, expire_interval, retry_interval, polling period) }}
```

Similar use of it for the `zebra.route-map.frr.j2` and `zebra.vrf.route-map.frr.j2` would allow you to consolidate both of these to one template for global and non-global VRFs. You can see an attempt at it [here](https://github.com/yzguy/vyos-1x/commit/71156beb621a1f79d2d51fd5ac559000fb0ce718)

The current place the vrf aware RPKI config would go is into `zebra.vrf.route-map.frr.j2`, which loops through all VRFs in a single template, rather than looping through all VRFs in Python, and calling the template to be rendered for each VRF. I would repeat the same code for RPKI within `rpki.frr.j2` (for global VRF) and again in `zebra.vrf.route-map.frr.j2`. This way allows you to keep the RPKI specific code within one template, and allow it to be re-used for global and non-global VRFs

As to your question about why if BGP doesn't require it, BGP configuration in FRR is fundamentally different than RPKI, for example

```
router bgp 65000 vrf blue
```

vs

```
vrf blue
 rpki expire_interval 3600
exit-vrf
```

BGP configuration is not within a VRF block, the VRF is simply appended onto the `router bgp`. If it did, I would recommend doing what I did for RPKI, you define a `bgp_config` macro, pass in `bgp` whether it's the global BGP or vrf BGP, then everything else is the same.

c-po: Thanks for the detailed explanation!

c-po (commented):
The general implementation looks good and follows established patterns for FRR.

Smoketests still fail:

```
cpo@LR1.wue3:~$ /usr/libexec/vyos/tests/smoke/cli/test_protocols_rpki.py
test_rpki (__main__.TestProtocolsRPKI.test_rpki) ... ok
test_rpki_source_address (__main__.TestProtocolsRPKI.test_rpki_source_address) ... ok
test_rpki_ssh (__main__.TestProtocolsRPKI.test_rpki_ssh) ... ERROR
test_rpki_verify_preference (__main__.TestProtocolsRPKI.test_rpki_verify_preference) ... ok

ERROR: test_rpki_ssh (__main__.TestProtocolsRPKI.test_rpki_ssh)
  File "/usr/libexec/vyos/tests/smoke/cli/test_protocols_rpki.py", line 217, in test_rpki_ssh
    self.cli_commit()
  File "/usr/libexec/vyos/tests/smoke/cli/base_vyostest_shim.py", line 103, in cli_commit
    out = self._session.commit()
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/vyos/configsession.py", line 280, in commit
    out = self.__run_command([COMMIT])
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/vyos/configsession.py", line 210, in __run_command
    raise ConfigSessionError(output)
vyos.configsession.ConfigSessionError: [ pki ]
Embedded PKI openssh with name "rpki-smoketest" is still in use by CLI
path "protocols rpki cache 192.0.2.3 ssh key"
[[pki]] failed
[ vrf name blue protocols rpki ]
RPKI+SSH requires key to be defined!
[[vrf name blue protocols rpki]] failed
Commit failed


Ran 4 tests in 65.914s

FAILED (errors=1)
```

c-po (approved):
New feature addition, available with FRR10.2 - implementation follows established paths for FRR

dmbaturin (approved):
Now that smoke tests pass, I think it's good to go.

PR #4502, opened 2025-05-09
T7443: Un-restricting non-root logins after scheduled reboot/shutdown via pam_nologin

c-po:
## Change summary

When using reboot in, reboot at, or shutdown in, non-root users are prevented from logging in via SSH or console starting 5 minutes before the scheduled shutdown or reboot time.

This behavior is intended by pam_nologin.so, which is included in the SSH and login PAM stack (default on Debian). While expected, it may be inconvenient and could be reconsidered.

## Types of changes

## Related Task(s)

* https://vyos.dev/T7443

```
vyos@vyos:~$ reboot in 5
```

Now try to log in via SSH or locally.

## Checklist:

### DIFF
--- a/debian/vyos-1x.postinst
+++ b/debian/vyos-1x.postinst
@@ -50,6 +50,10 @@ if [[ -e /usr/share/pam-configs/tacplus ]]; then
     rm /usr/share/pam-configs/tacplus
 fi
 
+# Disable pam_nologin.so behavior for regular users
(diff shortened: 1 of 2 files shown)


dmbaturin (approved):
I agree that restricting logins is a bad idea. Probably was always bad, even in the time of VAX servers shared between end users through VT100 terminals — _warning users about an upcoming reboot_ is a good idea, but that should better be done through the login banner.

That AI review comment may have merits — if th PAM config ever has stray whitespace after the module, the regex will not match. I'm not against merging it as is, though.

jestabro (approved):
A reasonable fix.

PR #4525, opened 2025-05-22
vyos-1x-vmware: T3681: Fix Python bytecompile exclusion

yunzheng:
## Change summary

VMware resuming was broken again due to a change in the `Makefile` in the following commit: https://github.com/vyos/vyos-1x/commit/64c9fdef02323309e97b2bb682604ada52d651e8. That change added an exclusion to the `python -m compileall` regex, however the updated exclusion regex did not work and resulted in byte compiling and creating a `__pycache__` directory again which `open-vm-tools` tries to execute on resume.

Although a fix landed upstream in `open-vm-tools` to prevent executing `__pycache__` directories [here](https://github.com/vmware/open-vm-tools/pull/689.) There is no new version in Debian that has this fix yet.

This PR fixes the Python byte compile exclusion regex by specifying it as two separate regexes, I'm not sure why the second exclusion was added initially, as there is no `/ppp/` directory that python tries to compile but I have left it in there just in case.

## Types of changes

## Related Task(s)

https://vyos.dev/T3681

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ clean:
 
 .PHONY: test
 test: generate-configd-include-json
-	set -e; python3 -m compileall -q -x '/vmware-tools/scripts/, /ppp/' .
+	set -e; python3 -m compileall -q -x '/vmware-tools/scripts/' -x '/ppp/' .
 	PYTHONPATH=python/ python3 -m "nose" --with-xunit src --with-coverage --cover-erase --cover-xml --cover-package src/conf_mode,src/op_mode,src/completion,src/helpers,src/validators,src/tests --verbose
 
 .PHONY: check_migration_scripts_executable
--- a/debian/control
+++ b/debian/control
@@ -385,7 +385,7 @@ Description: VyOS configuration scripts and data
  VyOS configuration scripts, interface definitions, and everything
 
 Package: vyos-1x-vmware
-Architecture: amd64
+Architecture: all
 Depends:
  vyos-1x,
  open-vm-tools


c-po (approved):
LGTM, also the change in architecture mimucs Debians support for amd64, i386 and arm64 for the required open-vm-tools package

dmbaturin (approved):
The syntax was is incorrect here indeed. I honestly have no idea what's the deal with `ppp`. I think we can just remove that part later.

yunzheng:
> The syntax was is incorrect here indeed. I honestly have no idea what's the deal with `ppp`. I think we can just remove that part later.

Hi dmbaturin, it seems there can be only one `-x` defined, so `-x /ppp/` overrides the first `-x /vmware-tools/scripts/` :(

This PR didn't actually fix the issue so to say, can we amend this PR by removing `-x /ppp/`? Or should I create a new PR to just remove that?

With the `-x /ppp/` removed I can confirm the `__pycache__` is not there in the package:

```
$ dpkg -c vyos-1x-vmware_1.5dev0-3109-g08dc2e56b+dirty_all.deb
drwxr-xr-x root/root         0 2023-09-10 13:42 ./
drwxr-xr-x root/root         0 2023-09-10 13:42 ./etc/
drwxr-xr-x root/root         0 2023-09-10 13:42 ./etc/vmware-tools/
drwxr-xr-x root/root         0 2023-09-10 13:42 ./etc/vmware-tools/scripts/
drwxr-xr-x root/root         0 2023-09-10 13:42 ./etc/vmware-tools/scripts/resume-vm-default.d/
-rwxr-xr-x root/root      2079 2023-09-10 13:42 ./etc/vmware-tools/scripts/resume-vm-default.d/ether-resume.py
-rw-r--r-- root/root        33 2023-09-10 13:42 ./etc/vmware-tools/tools.conf
drwxr-xr-x root/root         0 2023-09-10 13:42 ./usr/
drwxr-xr-x root/root         0 2023-09-10 13:42 ./usr/share/
drwxr-xr-x root/root         0 2023-09-10 13:42 ./usr/share/doc/
drwxr-xr-x root/root         0 2023-09-10 13:42 ./usr/share/doc/vyos-1x-vmware/
-rw-r--r-- root/root       322 2023-09-10 13:42 ./usr/share/doc/vyos-1x-vmware/changelog.gz
-rw-r--r-- root/root      1261 2023-09-10 13:42 ./usr/share/doc/vyos-1x-vmware/copyright
```

PR #4551, opened 2025-06-09
op-mode: T7538: remove the obsolete "show login level" command

dmbaturin:
## Change summary

The old command `run show login level` is a legacy of the Vyatta days: it uses an implementation detail of the old scripts to determine if the current user is an admin or an operator, and also assumes that admin/operator distinctions are binary.

It's effectively useless now and will break when we introduce the new op mode permissions system, so we should remove it.

## Types of changes

- [x] Other (please describe): legacy functionality removal.

## Checklist:

### DIFF
--- a/op-mode-definitions/show-login.xml.in
+++ b/op-mode-definitions/show-login.xml.in
@@ -14,12 +14,6 @@
             </properties>
             <command>/usr/bin/id -Gn</command>
           </leafNode>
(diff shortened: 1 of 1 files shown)


sever-sever (approved):
We are not having levels at the moment

c-po (approved):
We no longer have login levels

PR #4555, opened 2025-06-11
op-mode: T7542: add support for "standalone" behavior of operational mode tag nodes

dmbaturin:
## Change summary

There are operational mode commands that can be called either with an argument or without. For example, `show interfaces ethernet` (show information about all Ethernet interfaces) and `show interfaces ethernet eth0` (only show information for `eth0`).

The old way to handle that was to create a `<node>` and a `<tagNode>` with the same name ("ethernet" for the example above) — the `<node>` would handle the call to `show interfaces ethernet` and the `<tagNode>` would handle the case with `ethX`.

That was a pretty nasty exploit of the way the old CLI backend works and causes multiple problems: from unnecessarily verbose XML files with a lot of duplication to big difficulties with implementing a new command lookup algorithm and a documentation generator (outlined in T7541).

There are two possible ways to think about such commands that can be called with or without arguments.

The first way is that `show interfaces ethernet` is like a normal node — it has a meaning of its own and can be called without an argument, but shares a property with tag nodes — it can also take an argument. In that approach, normal nodes need a way to include anonymous children to account for variable arguments. That was my initial idea that created significant  difficulties with the JSON representation that would be used for command lookup — how to mark a child node as that anonymous child? I was thinking of using some reserved name for that purpose like `"!tag"` (since `!` is not used in command names).

However, later I realized that there was another way to think about those commands. `show interfaces ethernet` is a tag node that shares a property with normal nodes — it has a special behavior when it's called without an argument. Most tag nodes command tell the user that an argument is required, this one show information for all Ethernet interfaces. That means that is just needs a way to define that behavior. And in practice that behavior is just a help string and a command.

This PR adds a new `<standalone>` element that can be included in `<tagNode>` to give it a help string and a command to call when it's executed without arguments:

```
<tagNode name="foo">
  <properties> ... </properties>
  <command>the command to run when the node is called with an argument</command>
  <standalone>
    <help>outer node help string</help>
    <command>command to run when the node is called without an argument</command>
  </standalone>
</tagNode>
```

That change is fully compatible with the old schema.

So far, commands are mandatory for standalone elements because I couldn't think of a case where it makes sense to add it just for the help strings. That restriction can be trivially relaxed later.

## Types of changes

## Checklist:

### DIFF
--- a/schema/op-mode-definition.rnc
+++ b/schema/op-mode-definition.rnc
@@ -1,6 +1,6 @@
 #    interface_definition.rnc: VyConf reference tree XML grammar
 #
-#    Copyright (C) 2014. 2017 VyOS maintainers and contributors <maintainers@vyos.net>
+#    Copyright (C) 2014-2025 VyOS maintainers and contributors <maintainers@vyos.net>
 #
 #    This library is free software; you can redistribute it and/or
 #    modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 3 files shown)


dmbaturin:
FIY: the CI failure is spurious and caused by yet another repo de-sync on a kernel upgrade. https://github.com/vyos/vyos-1x/actions/runs/15595434775/job/43924648393?pr=4555#step:7:1220

jestabro (approved):

c-po (approved):
Add new standalone XML node to ease op-mode CLI definitions.

PR #4559, opened 2025-06-13
vrf: T7506:  Do not use default table 254 for VRF

natali-rs1985:
## Types of changes

## Related Task(s)

* https://vyos.dev/T7506

```
set vrf name test table 254

vyos@vyos# commit
[ vrf ]
VRF "test" table id 254 connot be used!

[[vrf]] failed
Commit failed
[edit]
```
## Checklist:

### DIFF
--- a/src/conf_mode/vrf.py
+++ b/src/conf_mode/vrf.py
@@ -18,6 +18,8 @@ from sys import exit
 from jmespath import search
 from json import loads
 
+import vyos.defaults
+
 from vyos.config import Config
 from vyos.configdict import node_changed
 from vyos.configverify import verify_route_map
@@ -162,6 +164,11 @@ def verify(vrf):
             # table id is mandatory
             if 'table' not in vrf_config:
                 raise ConfigError(f'VRF "{name}" table id is mandatory!')
+
+            if int(vrf_config['table']) == vyos.defaults.rt_global_vrf:
+                raise ConfigError(
+                    f'VRF "{name}" table id {vrf_config["table"]} cannot be used!'
+                )
 
             # routing table id can't be changed - OS restriction
             if interface_exists(name):


sever-sever (approved):
Table 254 is a default table and must not be used for VRF

src/conf_mode/vrf.py
```diff
@@ -163,6 +164,11 @@ def verify(vrf):
             if 'table' not in vrf_config:
                 raise ConfigError(f'VRF "{name}" table id is mandatory!')
 
+            if vrf_config['table'] == default_table_id:
```
c-po: Please use the already present global definition `vyos.defaults.rt_symbolic_names`

https://github.com/vyos/vyos-1x/blob/current/python/vyos/defaults.py#L77

c-po (changes requested):

src/conf_mode/vrf.py
natali-rs1985: Done

c-po: Was this code tested? I do not think it will work as you compare a string with a dict.

natali-rs1985: Sorry! My mistake. Now it compares int with int https://github.com/vyos/vyos-1x/blob/current/python/vyos/defaults.py#L81

src/conf_mode/vrf.py
```diff
@@ -163,6 +164,11 @@ def verify(vrf):
             if 'table' not in vrf_config:
                 raise ConfigError(f'VRF "{name}" table id is mandatory!')
 
+            if int(vrf_config['table']) == rt_global_vrf:
+                raise ConfigError(
+                    f'VRF "{name}" table id {vrf_config["table"]} connot be used!'
```
dmbaturin:
```suggestion
                    f'VRF "{name}" table id {vrf_config["table"]} cannot be used!'
```

src/conf_mode/vrf.py
```diff
@@ -21,6 +21,7 @@
 from vyos.config import Config
 from vyos.configdict import node_changed
 from vyos.configverify import verify_route_map
+from vyos.defaults import rt_global_vrf
```
dmbaturin: I'd be inclined to either import it locally or import `vyos.defaults` and use a fully-qualified name. That's subjective, though.

dmbaturin (changes requested):
The idea is correct, I left some suggestions.

src/conf_mode/vrf.py
natali-rs1985: Made changes

c-po (approved):
Changes addressed and look good now

PR #4565, opened 2025-06-18
op-mode: T7560: add support for virtual tag nodes

dmbaturin:
## Change summary

We have a whole bunch of commands where support for both variable and fixed arguments like in `show ip route <x.x.x.x/y ... | bgp | vrf ...>` is implemented via specially constructed XML definitions that are written to produce a structure for the old runner: either a duplicate `<node>` and `<tagNode>` at the same level or a child node like `<node name="node.tag">`.

To make every op mode command path unique and implement a new wrapper, we need to eliminate both hacks.

This PR provides a native way to add variably-named children to nodes that also have fixed argument: a new `<virtualTagNode>` element that can be included in any `<node>` alongside other elements.

## Types of changes

- [x] Other (please describe): internal change.

## Related Task(s)

* https://vyos.dev/T7541 ­— the parent task.

## Checklist:

### DIFF
--- a/op-mode-definitions/clear-ip.xml.in
+++ b/op-mode-definitions/clear-ip.xml.in
@@ -7,12 +7,6 @@
           <help>Clear Internet Protocol (IP) statistics or status</help>
         </properties>
         <children>
-          <node name="prefix-list">
-            <properties>
-              <help>Clear prefix-list statistics or status</help>
-            </properties>
-            <command>vtysh -c "$*"</command>
-          </node>
           <tagNode name="prefix-list">
             <properties>
               <help>Clear prefix-list statistics or status for specified word</help>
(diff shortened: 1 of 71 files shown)


jestabro (commented):
Note that the task number is T7560: there is a typo in PR title and commit message.

c-po (approved):
Thanks for keeping the muscle memory intact and just replace the hack with a proper schema based solution.

dmbaturin:
Op mode definitions now pass the new `--check-path-ambiguity` test of the @jestabro's script, which means unambigious lookup is now possible:

```
$ ./python/vyos/xml_ref/generate_op_cache.py --check-path-ambiguity --xml-dir build/op-mode-definitions/ && echo 'No duplicates'
No duplicates
```

jestabro (approved):

PR #4569, opened 2025-06-20
installer: T6144: require at least 2GB of free space for image upgrade

dmbaturin:
## Change summary

Trying to add an image when the disk doesn't have enough space can break the system, so we need to make sure that the user can only initiate an upgrade when there's enough space to download the ISO image and to install its SquashFS image.

Our images are about 600M at the moment, so 2GB is a safe estimate.

This also raises the question if we should increase the requirement for installation, because at the moment it's 2GB so it's possible to create a system that cannot be upgraded. But it's a different question than this fix.

## Types of changes

## Checklist:

### DIFF
--- a/src/op_mode/image_installer.py
+++ b/src/op_mode/image_installer.py
@@ -19,7 +19,7 @@
 
 from argparse import ArgumentParser, Namespace
 from pathlib import Path
-from shutil import copy, chown, rmtree, copytree
+from shutil import copy, chown, rmtree, copytree, disk_usage
 from glob import glob
 from sys import exit
 from os import environ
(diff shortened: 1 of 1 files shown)


jestabro (approved):
Tested, and works as expected. I do question whether we want this for 1.4.3, as it may be an unpleasant surprise for those upgrading with the minimal free space, before deleting older images (as I often do). I defer to @sever-sever for his thoughts on whether it should be in 1.4.3 or delayed to 1.4.4. with advance warning to users.

sever-sever (approved):
I think it is a good idea to reserve minimal space for upgrading and good idea for 1.4.3

PR #4578, opened 2025-06-25
build: T7578: fail the package build if there are non-unique op mode nodes

dmbaturin:
## Change summary

PR #4565 eliminates the last non-unique op mode nodes. However, to ensure that the JSON cache is usable for command lookup, we need to ensure that no new duplicate nodes appear, so we need to make the check a part of the build process.

## Types of changes

- [x] Other (please describe): build script change.

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ op_mode_definitions: $(op_xml_obj)
 
 	find $(BUILD_DIR)/op-mode-definitions/ -type f -name "*.xml" | xargs -I {} $(CURDIR)/scripts/build-command-op-templates {} $(CURDIR)/schema/op-mode-definition.rng $(OP_TMPL_DIR) || exit 1
 
-	$(CURDIR)/python/vyos/xml_ref/generate_op_cache.py --xml-dir $(BUILD_DIR)/op-mode-definitions || exit 1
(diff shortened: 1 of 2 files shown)


jestabro (changes requested):
Note that `generate_op_cache.py `does not currently exit with error if there are duplicate paths, it simple prints the results and continues. One should add something like the following:
```
diff --git a/python/vyos/xml_ref/generate_op_cache.py b/python/vyos/xml_ref/generate_op_cache.py
index 117b080b4..19e27cd38 100755
--- a/python/vyos/xml_ref/generate_op_cache.py
+++ b/python/vyos/xml_ref/generate_op_cache.py
@@ -278,8 +278,8 @@ def main():
             with open(op_ref_json, 'w') as f:
                 json.dump(res, f, indent=2)
         else:
-            print('Found the following duplicate paths:\n')
-            print(out)
+            out = 'Found the following duplicate paths:\n' + out
+            sys.exit(out)
 
     with open(op_ref_cache, 'w') as f:
         f.write('from vyos.xml_ref.op_definition import NodeData\n')

```

c-po (approved):
Requested change from @jestabro included

jestabro (approved):

PR #4582, opened 2025-06-27
ipsec: T7545: Fix show vpn debug peer

hedrok:
## Types of changes

## Related Task(s)

https://vyos.dev/T7545

As I understand any setup with ipsec peers will allow testing.

I've used `/usr/libexec/vyos/tests/smoke/cli/test_vpn_ipsec.py`:

I had to fix interface number and remove `.interface` in several places so that it passed.

Added `from vyos.utils.process import cmd`

Added

````
        print("BEGIN show vpn ipsec connections")
        out = cmd(['/opt/vyatta/bin/vyatta-op-cmd-wrapper', 'show', 'vpn', 'ipsec', 'connections'])
        print(out)
        print("END")
        print("BEGIN show vpn debug peer main-branch")
        out = cmd(['/opt/vyatta/bin/vyatta-op-cmd-wrapper', 'show', 'vpn', 'debug', 'peer', 'main-branch'])
        print(out)
        print("END")
````

in `test_site_to_site` after `self.cli_commit()`

Before fix `python3 test_vpn_ipsec.py TestVPNIPsec.test_site_to_site` output was  `Peer not found, aborting` as in https://vyos.dev/T7545.

After fix:

````
### /usr/sbin/ipsec statusall | grep main-branch-tunnel-1 ###

### /usr/sbin/ipsec statusall | grep main-branch-tunnel-2 ###
````

## Open questions

1. I'm not sure the fix is robust. I've tested it under latest VyOS image, but maybe on other configurations `/etc/swanctl/swanctl.conf` contains other data (e.g. previous RE is appropriate for some configurations).
2. Maybe it would be better to use something like `_get_all_sitetosite_peers_name_list` from `src/op_mode/ipsec.py` instead of parsing Jinja-generated file `/etc/swanctl/swanctl.conf`?..
3. Maybe we should change output for vpn debug peer?.. Currently it just uses `/usr/sbin/ipsec statusall | grep {conn}` which outputs nothing in my case, and in origin logs in https://vyos.dev/T7545 the command also had no output.
4. Should I add test for `show vpn debug peer`?..

## Checklist:

### DIFF
--- a/src/op_mode/vpn_ipsec.py
+++ b/src/op_mode/vpn_ipsec.py
@@ -23,13 +23,13 @@ SWANCTL_CONF = '/etc/swanctl/swanctl.conf'
 
 
 def get_peer_connections(peer, tunnel, return_all = False):
-    search = rf'^[\s]*(peer_{peer}_(tunnel_[\d]+|vti)).*'
+    search = rf'^[\s]*({peer}-(tunnel-[\d]+|vti))[\s]*{{'
     matches = []
     with open(SWANCTL_CONF, 'r') as f:
         for line in f.readlines():
             result = re.match(search, line)
             if result:
-                suffix = f'tunnel_{tunnel}' if tunnel.isnumeric() else tunnel
+                suffix = f'tunnel-{tunnel}' if tunnel.isnumeric() else tunnel
                 if return_all or (result[2] == suffix):
                     matches.append(result[1])
     return matches
@@ -66,7 +66,8 @@ def debug_peer(peer, tunnel):
         return
 
     for conn in conns:
-        call(f'/usr/sbin/ipsec statusall | grep {conn}')
+        command = f'/usr/sbin/ipsec statusall | grep {conn}'
+        call(command)
 
 
 if __name__ == '__main__':

### CODE REVIEW

**Review by alexandr-san4ez — COMMENTED** [review]
I think the definition of the **tunnel suffix** adjustment is also important:
```python
# suffix = f'tunnel_{tunnel}' if tunnel.isnumeric() else tunnel
suffix = f'tunnel-{tunnel}' if tunnel.isnumeric() else tunnel      # line 32
```

Because this leads to a situation where debug data can only be obtained by fully specifying the tunnel name:
```
vyos@1af9e760225a:~$ show vpn debug peer PEER1 tunnel tunnel-1
### /usr/sbin/ipsec statusall | grep PEER1-tunnel-1 ###
```

As far as I understand, this method should work the same way, but at the moment it doesn't:
```
vyos@1af9e760225a:~$ show vpn debug peer PEER1 tunnel 1
Peer not found, aborting
```

hedrok:
@alexandr-san4ez Thank you, you are correct, I've pushed the fix.

src/op_mode/vpn_ipsec.py
```diff
@@ -66,7 +66,9 @@ def debug_peer(peer, tunnel):
         return
 
     for conn in conns:
-        call(f'/usr/sbin/ipsec statusall | grep {conn}')
+        command = f'/usr/sbin/ipsec statusall | grep {conn}'
+        print(f'\n### {command} ###')
```
dmbaturin: Please remove that stray debug print. Eventually we should remove the one at line 50 as well.

dmbaturin (changes requested):

alexandr-san4ez (approved):
Anyway in the future, it would be nice to cover this command by tests.

dmbaturin (approved):

PR #4628, opened 2025-07-29
conntrack: T7482: Fix custom timeouts

ryanzuwala:
Fix custom conntrack timeout rules and add smoketests

## Change summary

Custom timeout rules were previously not working at all.

The problem is that the firewall rules were originally configured to jump directly to the VYOS_CT_TIMEOUT chain from the PREROUTING and OUTPUT base chains. But those two base chains have a priority of raw (-300), and conntrack hooks at priority -200, therefore any rules that modify conntrack entries must have a priority above -200.

So I removed the VYOS_CT_TIMEOUT jumps from PREROUTING and OUTPUT, and created my own prerouting and output base chains with priority -199 to perform the jump to VYOS_CT_TIMEOUT. The two timeout jumping base chains are only created if custom timeout rules have been configured by an administrator.

To tidy up the nftables config, it also removes the empty VYOS_CT_TIMEOUT chain if there are no custom timeout rules created, since nothing else uses it.

I also added much more extensive smoke testing for this to ensure it doesn't accidentally break in the future. Custom timeout rules are very important to have for VoIP, and I'm sure there are plenty of other use cases for it as well.

## Types of changes

## Related Task(s)

https://vyos.dev/T7482

Experiment with creating custom conntrack timeout rules and verifying the records are being updated in conntrack.

Here is an IPv4 example using DNS traffic from a server with IP address 172.16.99.2:

```
configure

# UDP rule using DNS as an example
set system conntrack timeout custom ipv4 rule 1 destination port '53'
set system conntrack timeout custom ipv4 rule 1 protocol udp replied '700'
set system conntrack timeout custom ipv4 rule 1 protocol udp unreplied '700'
set system conntrack timeout custom ipv4 rule 1 source address '172.16.99.2'

commit ; save
```

Using this DNS example rule, you'll need to generate some DNS queries from your 172.16.99.2 server. You can just use `dig google.com` or ping a domain name.

After querying DNS records, run `conntrack -E` or `show conntrack table ipv4 | grep "172.16.99.2"`, and you should see your new timeouts of 700 seconds instead of the default of 30 for unreplied and 180 for stream.

## Checklist:

### DIFF
--- a/data/templates/conntrack/nftables-ct.j2
+++ b/data/templates/conntrack/nftables-ct.j2
@@ -16,17 +16,15 @@ table ip vyos_conntrack {
 {% endif %}
          return
     }
-    chain VYOS_CT_TIMEOUT {
 {% if timeout.custom.ipv4.rule is vyos_defined %}
+    chain VYOS_CT_TIMEOUT {
 {%     for rule, rule_config in timeout.custom.ipv4.rule.items() %}
         # rule-{{ rule }} {{ '- ' ~ rule_config.description if rule_config.description is vyos_defined }}
         {{ rule_config | conntrack_rule(rule, 'timeout', ipv6=False) }}
 {%     endfor %}
-{% endif %}
         return
     }
 
-{% if timeout.custom.ipv4.rule is vyos_defined %}
 {%     for rule, rule_config in timeout.custom.ipv4.rule.items() %}
     ct timeout ct-timeout-{{ rule }} {
         l3proto ip;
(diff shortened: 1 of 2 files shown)


ryanzuwala:

dmbaturin (approved):
@alexk37's testing shows that the basics work and I see no issues with the implementation.

jestabro (approved):
Merge for current; wait on backport.

dmbaturin:
Let's give it some time in current and see if anyone finds issues.

PR #4630, opened 2025-07-29
T7651: VPP use pid not ppid as default in vpp-failure-handler.service script

jestabro:
## Change summary

The `reset_section.py` script is used in a limited case for both (1) recovery from an apply error in a config mode script and (2) recovery by a service crash in a failure handler service. In the latter case, the default pid must be `os.getpid`, not `os.getppid`, so as not to be called by pid 1.

Without this fix, the current config can be damaged by the `vpp-failure-handler.service`, leading to symptoms such as missing system options settings, as described in the task.

## Types of changes

## Checklist:

### DIFF
--- a/src/helpers/reset_section.py
+++ b/src/helpers/reset_section.py
@@ -84,7 +84,7 @@ else:
 del test
 
 
-session_id = int(pid) if pid else os.getppid()
+session_id = int(pid) if pid else os.getpid()
 
 if in_session:
     # check hint left by vyshim when ConfigError is from apply stage


sever-sever (approved):
I checked this fix, it works for me.
I see VPP crashed and restarted, but I still connected to the console without issues.
Reboot the server and I see all expected `kerlnel options`

Thanks @jestabro

dmbaturin (approved):
Tested and works. no objections. A stark reminder of the subtle difference between PID and PPID...

PR #4638, opened 2025-07-31
serial: T7484: treat unavailable serial console devices as non-fatal

c-po:
## Change summary

Instead of failing or starting agetty on an invalid device, we now check whether the specified serial console device is a valid TTY. If it's not, we display a warning to the user and skip starting the agetty process.

## Types of changes

## Related Task(s)

* https://vyos.dev/T7484

```
vyos@vyos# set system console device ttyS0
[edit]
vyos@vyos# commit
[ system console ]

WARNING: "ttyS0" is not a TTY device and cannot be used as console!
```

## Checklist:

### DIFF
--- a/python/vyos/utils/serial.py
+++ b/python/vyos/utils/serial.py
@@ -116,3 +116,16 @@ def restart_login_consoles(prompt_user=False, quiet=True, devices: List[str]=[])
             cmd(f'systemctl stop {unit_name}')
 
     return True
+
+def is_tty(name: str) -> bool:
+    """ Check if a given device file (e.g. /dev/ttyS0) is a TTY (teletypewriter)
+    device in Linux
+    """
+    import os
+    path_tty = f'/dev/{name}'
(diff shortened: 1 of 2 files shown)


dmbaturin (approved):
The bug is annoying and the fix is straightforward.

jestabro (approved):
Logic is clear; tested by smoketest.

sever-sever (approved):
Do not fail on commit if console device is not TTY
Usefull for some platforms.

yunzheng:
Just curious, would this break if someone plugs in a serial cable at a later moment but the agetty daemon was not launched?

c-po:
> Just curious, would this break if someone plugs in a serial cable at a later moment but the agetty daemon was not launched?

One would need to restart that explicit serial line in order to come online and provide a TTY.

One could also think of a udev line triggering the restart automatically - but I think this is a very niche use-case and will cause more harm then good.

PR #4657, opened 2025-08-12
T7709: add explicit sync and atomic write on config file save

jestabro:
## Change summary

Add a context manager to explicitly sync file and use an atomic write in config save utility.

NOTE that this will need a (simple) manual backport to Sagitta, as the underlying component_version module differs.

## Types of changes

## Checklist:

### DIFF
--- a/python/vyos/component_version.py
+++ b/python/vyos/component_version.py
@@ -209,6 +209,16 @@ def version_info_prune_component(x: VersionInfo, y: VersionInfo) -> VersionInfo:
     x.component = {k: v for k, v in x.component.items() if k in y.component}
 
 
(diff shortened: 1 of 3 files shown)


jestabro:
Updated PR to use `os.replace` (== rename) for atomic write of config file: this is a better guarantee that on recovery from failure _either_ the new config _or_ the previous config will be available.

c-po (approved):
Add support for atomic write of a new configuration during save via a temporary file and atomic move operation.

dmbaturin (approved):
This logic seems good and should protect us from at least _some_ write failure scenarios. It may not protect us from everything, but it's still a useful change.

PR #4678, opened 2025-08-25
image: T5455: Add migration of SSH `known_hosts` files during upgrade

alexandr-san4ez:
## Change summary
During upgrade, the script now checks if any `known_hosts` files exist. If so, it prompts the user to save these SSH fingerprints, and upon confirmation, copies the files to the new image persistence directory.

```
Copying SSH host keys
Would you like to save the SSH known hosts (fingerprints) from your current configuration? [Y/n]
Copying SSH known hosts
Copying system image files
```

## Types of changes

## Related Task(s)
* https://vyos.dev/T5455

Ensure `known_hosts` files exist for root or users, then run the upgrade and confirm you’re prompted to migrate them. After choosing **Yes**, verify the files are copied; with **No** or no files, they should not be migrated or prompted. 

```
vyos@vyos:~$ touch /home/vyos/.ssh/known_hosts
vyos@vyos:~$ sudo touch /root/.ssh/known_hosts
vyos@vyos:~$ add system image vyos-1.5-rolling-202508251355-generic-amd64.iso
Validating image checksums
What would you like to name this image? (Default: 1.5-rolling-202508251355) test1
Would you like to set the new image as the default one for boot? [Y/n] Y
An active configuration was found. Would you like to copy it to the new image? [Y/n] Y
Would you like to copy SSH host keys? [Y/n] Y
Copying SSH host keys
Would you like to save the SSH known hosts (fingerprints) from your current configuration? [Y/n] Y
Copying SSH known hosts
Copying system image files
Cleaning up
Unmounting target filesystems
Removing temporary files
vyos@vyos:~$ sudo reboot now
...
vyos@vyos:~$ find /etc /home /root -name known_hosts 2>/dev/null
/home/vyos/.ssh/known_hosts
/root/.ssh/known_hosts
```

## Checklist:

### DIFF
--- a/python/vyos/utils/auth.py
+++ b/python/vyos/utils/auth.py
@@ -20,9 +20,35 @@ import string
 
 from enum import StrEnum
 from decimal import Decimal
+from pwd import getpwall
+from pwd import getpwnam
 from vyos.utils.process import cmd
 
-
+# Minimum UID used when adding system users
+MIN_USER_UID: int = 1000
+# Maximim UID used when adding system users
+MAX_USER_UID: int = 59999
+# List of local user accounts that must be preserved
+SYSTEM_USER_SKIP_LIST: frozenset = {
+    'radius_user',
+    'radius_priv_user',
+    'tacacs0',
+    'tacacs1',
+    'tacacs2',
+    'tacacs3',
+    'tacacs4',
+    'tacacs5',
+    'tacacs6',
+    'tacacs7',
+    'tacacs8',
+    'tacacs9',
+    'tacacs10',
+    'tacacs11',
+    'tacacs12',
+    'tacacs13',
+    'tacacs14',
+    'tacacs15',
+}
 DEFAULT_PASSWORD: str = 'vyos'
 LOW_ENTROPY_MSG: str = 'should be at least 8 characters long;'
 WEAK_PASSWORD_MSG: str = 'The password complexity is too low - @MSG@'
@@ -119,3 +145,24 @@ def get_current_user() -> str:
     elif 'USER' in os.environ:
         current_user = os.environ['USER']
     return current_user
+
+
+def get_local_users(min_uid=MIN_USER_UID, max_uid=MAX_USER_UID) -> list:
+    """Return list of dynamically allocated users (see Debian Policy Manual)"""
+    local_users = []
+
+    for s_user in getpwall():
+        if s_user.pw_uid < min_uid:
+            continue
+        if s_user.pw_uid > max_uid:
+            continue
+        if s_user.pw_name in SYSTEM_USER_SKIP_LIST:
+            continue
+        local_users.append(s_user.pw_name)
+
+    return local_users
+
+
+def get_user_home_dir(user: str) -> str:
+    """Return user's home directory"""
+    return getpwnam(user).pw_dir
--- a/src/conf_mode/system_login.py
+++ b/src/conf_mode/system_login.py
@@ -22,7 +22,6 @@ from copy import deepcopy
 from passlib.hosts import linux_context
 from psutil import users
 from pwd import getpwall
-from pwd import getpwnam
 from pwd import getpwuid
 from sys import exit
 from time import sleep
@@ -39,9 +38,13 @@ from vyos.template import is_ipv4
 from vyos.utils.auth import EPasswdStrength
 from vyos.utils.auth import evaluate_strength
 from vyos.utils.auth import get_current_user
+from vyos.utils.auth import get_local_users
+from vyos.utils.auth import get_user_home_dir
+from vyos.utils.auth import MIN_USER_UID
 from vyos.utils.configfs import delete_cli_node
 from vyos.utils.configfs import add_cli_node
 from vyos.utils.dict import dict_search
+from vyos.utils.file import move_recursive
 from vyos.utils.permission import chown
 from vyos.utils.process import cmd
 from vyos.utils.process import call
@@ -59,11 +62,7 @@ tacacs_nss_config_file = "/etc/tacplus_nss.conf"
 nss_config_file = "/etc/nsswitch.conf"
 login_motd_dsa_warning = r'/run/motd.d/92-vyos-user-dsa-deprecation-warning'
 
-# Minimum UID used when adding system users
-MIN_USER_UID: int = 1000
-# Maximim UID used when adding system users
-MAX_USER_UID: int = 59999
-# LOGIN_TIMEOUT from /etc/loign.defs minus 10 sec0
+# LOGIN_TIMEOUT from /etc/loign.defs minus 10 sec
 MAX_RADIUS_TIMEOUT: int = 50
 # MAX_RADIUS_TIMEOUT divided by 2 sec (minimum recomended timeout)
 MAX_RADIUS_COUNT: int = 8
@@ -71,29 +70,11 @@ MAX_RADIUS_COUNT: int = 8
 MAX_TACACS_COUNT: int = 8
 # Minimum USER id for TACACS users
 MIN_TACACS_UID = 900
-# List of local user accounts that must be preserved
-SYSTEM_USER_SKIP_LIST: list = ['radius_user', 'radius_priv_user', 'tacacs0', 'tacacs1',
-                              'tacacs2', 'tacacs3', 'tacacs4', 'tacacs5', 'tacacs6',
-                              'tacacs7', 'tacacs8', 'tacacs9', 'tacacs10',' tacacs11',
-                              'tacacs12', 'tacacs13', 'tacacs14', 'tacacs15']
 
 # As of OpenSSH 9.8p1 in Debian trixie, DSA keys are no longer supported
 SSH_DSA_DEPRECATION_WARNING: str = f'{SSH_DSA_DEPRECATION_WARNING} '\
 'The following users are using SSH-DSS keys for authentication.'
 
-def get_local_users(min_uid=MIN_USER_UID, max_uid=MAX_USER_UID):
-    """Return list of dynamically allocated users (see Debian Policy Manual)"""
-    local_users = []
-    for s_user in getpwall():
-        if getpwnam(s_user.pw_name).pw_uid < min_uid:
-            continue
-        if getpwnam(s_user.pw_name).pw_uid > max_uid:
-            continue
-        if s_user.pw_name in SYSTEM_USER_SKIP_LIST:
-            continue
-        local_users.append(s_user.pw_name)
-
-    return local_users
 
 def get_shadow_password(username):
     with open('/etc/shadow') as f:
@@ -389,9 +370,10 @@ def apply(login):
             tmp = dict_search('full_name', user_config)
             if tmp: command += f" --comment '{tmp}'"
 
-            tmp = dict_search('home_directory', user_config)
-            if tmp: command += f" --home '{tmp}'"
-            else: command += f" --home '/home/{user}'"
+            home_directory = dict_search('home_directory', user_config)
+            if not home_directory:
+                home_directory = f'/home/{user}'
+            command += f" --home '{home_directory}'"
 
             if 'operator' not in user_config:
                 command += f' --groups frr,frrvty,vyattacfg,sudo,adm,dip,disk,_kea'
@@ -404,7 +386,7 @@ def apply(login):
                 # crazy user will choose username root or any other system user which will fail.
                 #
                 # XXX: Should we deny using root at all?
-                home_dir = getpwnam(user).pw_dir
+                home_dir = get_user_home_dir(user)
                 # always re-render SSH keys with appropriate permissions
                 render(f'{home_dir}/.ssh/authorized_keys', 'login/authorized_keys.j2',
                        user_config, permission=0o600,
@@ -423,6 +405,17 @@ def apply(login):
 
             except Exception as e:
                 raise ConfigError(f'Adding user "{user}" raised exception: "{e}"')
+
+            # After invoking 'useradd' for each user, if /var/.users_backups/{user} exists, restore the
+            # backed up files to the newly created home directory. This reinstates the user's
+            # SSH environment and avoids loss of access or trust relationships due to the user
+            # creation process, which does not copy such custom files by default.
+            #
+            # More details: https://github.com/vyos/vyos-1x/pull/4678#pullrequestreview-3169648265
+            backup_directory = f"/var/.users_backups/{user}"
+            if command.startswith('useradd') and os.path.exists(backup_directory):
+                move_recursive(backup_directory, home_dir)
+                chown(home_dir, user=user, group='users', recursive=True)
 
             # T5875: ensure UID is properly set on home directory if user is re-added
             # the home directory will always exist, as it's created above by --create-home,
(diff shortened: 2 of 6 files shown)


src/op_mode/image_installer.py
```diff
@@ -1010,6 +1025,41 @@ def install_image() -> None:
...
+            files.append(root_known_hosts)
+
+    # for each non-system user
+    for user in getpwall():
+        home_dir = Path(user.pw_dir)
+        # skip system users
+        if home_dir.exists() and home_dir.as_posix().startswith('/home'):
+            known_hosts = home_dir / '.ssh' / 'known_hosts'
+            if known_hosts.exists():
+                files.append(known_hosts)
```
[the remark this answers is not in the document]
alexandr-san4ez: As I see in this repository we use `pwd.getpwall` [approach](https://github.com/vyos/vyos-1x/blob/968409218f914d180b2f9781edf3423ddb8ac738/src/op_mode/show_users.py#L82) instead of reading list of `/home` directories.

src/op_mode/image_installer.py
```diff
@@ -1010,6 +1025,41 @@ def install_image() -> None:
...
+    return files
+
+
+def migrate_known_hosts(target_dir: str):
+    """Copy `known_hosts` for root and all users to the new image directory"""
+
+    known_hosts_files = get_known_hosts_files()
+    for known_hosts_file in known_hosts_files:
+        # copy file to target directory
+        target_known_hosts = Path(f'{target_dir}{known_hosts_file}')
```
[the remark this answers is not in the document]
alexandr-san4ez: > If `known_hosts_file` is an absolute path like `/root/.ssh/known_hosts`, the result would be `{target_dir}/root/.ssh/known_hosts`

This is exactly the behavior that is necessary to implement `known_hosts` file saving:
`known_hosts_file = '/usr/lib/live/mount/persistence/boot/1.5-rolling-202508251355/rw/root/.ssh/known_hosts'`

src/op_mode/image_installer.py
```diff
@@ -1010,6 +1025,41 @@ def install_image() -> None:
...
+    for file_path in base_files:
+        root_known_hosts = Path(file_path)
+        if root_known_hosts.exists():
+            files.append(root_known_hosts)
+
+    # for each non-system user
+    for user in getpwall():
+        home_dir = Path(user.pw_dir)
+        # skip system users
+        if home_dir.exists() and home_dir.as_posix().startswith('/home'):
```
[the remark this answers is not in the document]
alexandr-san4ez: Ok, I will update this code

src/op_mode/image_installer.py
```diff
@@ -1010,6 +1025,46 @@ def install_image() -> None:
         exit(1)
 
 
+def get_known_hosts_files() -> list:
+    """Collect all existing `known_hosts` files for root and users under /home"""
+
+    # Minimum UID used when adding system users
+    MIN_USER_UID = 1000
```
c-po: Please do not re-define the values for MIN/MAX UIDs.

Please import them from `src/conf_mode/system_login.py` to avoid double definition.

src/op_mode/image_installer.py
```diff
@@ -1115,6 +1170,11 @@ def add_image(image_path: str, vrf: str = None, username: str = '',
             for host_key in host_keys:
                 copy(host_key, target_ssh_dir)
 
+        target_ssh_known_hosts_dir: str = f'{root_dir}/boot/{image_name}/rw'
+        if no_prompt or copy_ssh_known_hosts():
+            print('Copying SSH known hosts')
```
c-po:
```suggestion
            print('Copying SSH known_hosts files')
```

src/op_mode/image_installer.py
```diff
@@ -1010,6 +1025,46 @@ def install_image() -> None:
...
+
+    # for root
+    base_files = ('/root/.ssh/known_hosts', '/etc/ssh/ssh_known_hosts')
+    for file_path in base_files:
+        root_known_hosts = Path(file_path)
+        if root_known_hosts.exists():
+            files.append(root_known_hosts)
+
+    # for each non-system user
+    for user in getpwall():
```
c-po: Maybe you can reuse: `src/conf_mode/system_login.py::get_local_users()` or even make this a utility function in `vyos.utils.auth`?

c-po (changes requested):

src/op_mode/image_installer.py
alexandr-san4ez: Yes, it makes sense. I will change it

alexandr-san4ez: Implemented by using helper functions: `from vyos.utils.auth import get_local_users, get_user_home_dir`

src/conf_mode/system_login.py
```diff
@@ -34,6 +33,7 @@
 from vyos.utils.auth import EPasswdStrength
 from vyos.utils.auth import evaluate_strength
 from vyos.utils.auth import get_current_user
+from vyos.utils.auth import MIN_USER_UID, get_local_users, get_user_home_dir
```
c-po: Please use one import per line which makes grepping easier in the future.

python/vyos/utils/auth.py
```diff
@@ -20,9 +20,34 @@
 
 from enum import StrEnum
 from decimal import Decimal
+from pwd import getpwall, getpwnam
```
c-po: Please use one import per line which makes grepping easier in the future.

src/op_mode/image_installer.py
```diff
@@ -57,6 +57,7 @@
 from vyos.utils.file import read_file
 from vyos.utils.file import write_file
 from vyos.utils.process import cmd, run, rc_cmd
+from vyos.utils.auth import get_local_users, get_user_home_dir
```
c-po: Please use one import per line which makes grepping easier in the future.

c-po (changes requested):

c-po:
Other then the tree minor changes in placing imports on their own line - code looks good.

c-po (approved):
All requested changes added. Change looks good.

jestabro (changes requested):
I see one problem with this: by implicitly creating the home directories in the persistence directory
`/usr/lib/live/mount/persistence/boot/my-new-image-name/rw`
during add system image, the `useradd` in `system_login.py` on first boot into the new image will avoid copying skeleton files. The warning is not recorded, but an analogous situation is, say:
```
vyos@vyos:~$ sudo mkdir /home/bob
vyos@vyos:~$ sudo useradd --create-home --home /home/bob bob
useradd: warning: the home directory /home/bob already exists.
useradd: Not copying any file from skel directory into it.
```

The result in my test is that, on booting into the new image, one sees:
```
vyos@vyos:~$ show version

  Invalid command: [show]

vyos@vyos:~$ pwd
/home/vyos
vyos@vyos:~$ ls -la
total 16
drwxr-xr-x 3 vyos root  4096 Aug 29 17:56 .
drwxr-xr-x 1 root root  4096 Aug 29 17:41 ..
-rw------- 1 vyos users  224 Aug 29 17:56 .bash_history
drwxr-xr-x 2 vyos root  4096 Aug 29 17:47 .ssh
-rw-r--r-- 1 vyos users    0 Aug 29 17:47 .sudo_as_admin_successful
```
so, no .bashrc etc, hence no completions.

Updating the image _with_ this PR but _without_ copying the `known_hosts` files returns normal behavior in the new image.

So, I don't have a concrete suggestion yet, but do want to point out the issue.

alexandr-san4ez:
@jestabro thanks for finding this because I missed it. I will prepare fix for this PR.

alexandr-san4ez:
> @jestabro thanks for finding this because I missed it. I will prepare fix for this PR.

Fix: Preserve existing home directory when adding user by backing up and restoring its contents after `useradd` to prevent data loss.

jestabro:
Re-running failed jobs, as a check; all passed in a local test. Review pending.

jestabro:
The current version exposes an interesting issue (not directly related to the PR), as well as having a critical issue itself --- on a system with the PR, _before_ any upgrade, one has (with my added 'DEBUG print' lines):
```
vyos@vyos:~$ ls -lart
total 80
-rw-r--r-- 1 vyos users       675 Sep 10  2023 .profile
-rw-r--r-- 1 vyos users      4131 Sep 10  2023 .bashrc
-rw-r--r-- 1 vyos users       220 Apr 18 22:47 .bash_logout
-rw-r--r-- 1 vyos users         0 Sep  5 16:43 .sudo_as_admin_successful
-rw-r--r-- 1 vyos users       582 Sep  5 16:44 vyos-vimrc
-rw-r--r-- 1 vyos users       582 Sep  5 16:44 .vimrc
-rw------- 1 vyos users       923 Sep  5 16:45 .viminfo
-rw-r--r-- 1 vyos users      8070 Sep  5 18:20 file.py
drwxr-xr-x 2 vyos users      4096 Sep  5 18:40 .ssh
drwxr-xr-x 1 root root       4096 Sep  5 18:40 ..
-rw------- 1 vyos vyattacfg    20 Sep  5 18:44 .lesshst
drwxr-xr-x 3 vyos users      4096 Sep  5 18:44 .
-rw------- 1 vyos users     27764 Sep  5 18:44 .bash_history
vyos@vyos:~$ config
[edit]
vyos@vyos# set system login user jestabro authentication public-keys jestabro@homonoia type ssh-ed25519
[edit]
vyos@vyos# set system login user jestabro authentication public-keys jestabro@homonoia key AAAAC3NzaC1lZDI1NTE5AAAAIBLCU6nIgzNxgxPniBpwoOxS7uG9xCLCWf2sp9292Wti
[edit]
vyos@vyos# commit
[ system login ]
DEBUG print: command is useradd --create-home --no-user-group  --shell /bin/vbash --password '!' --home '/home/jestabro' --groups frr,frrvty,vyattacfg,sudo,adm,dip,disk,_kea jestabro
DEBUG print: command is usermod --shell /bin/vbash --home '/home/vyos' --groups frr,frrvty,vyattacfg,sudo,adm,dip,disk,_kea vyos

shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
[edit]
vyos@vyos# ls -lart
total 0
[edit]
```

(1) cd'ing away from the directory and back will find expected files
(2) note that `system_login.py` will apply to the full config under the node `['system', 'login']`, hence a change to one user will lead to move/restore of all home directories of current users, as seen in the example above (this is the exposed issue: independent of the current PR, we may want to consider a refinement here with `node_changed` or similar, later; note that does not avoid the issues here)
(3) one could consider the alternative of explicitly copying the skel files _only in the case of useradd_, however, I have no strong opinion either way, yet
(4) it is not surprising that this is non-trivial, however, it is a worthwhile problem to get right, so as to open the possibility of migrating (a limited amount of) other user home dir data

jestabro (changes requested):
Cf. comment above

alexandr-san4ez:
I've simplified the approach to saving user's files a bit. 
Now the files will be saved to a temporary directory (`/var/.users_backups/{user}` where `user` is name of user) and restored when the user is first created (`command == useradd`).

I'm not sure about naming of temporary directory (`/var/.users_backups`) and I would be grateful for your suggestions.

jestabro (approved):
Simple tests of upgrade now pass.

alexandr-san4ez:
@c-po could you look at this PR again because I have changed the original approach.

c-po:
@alexandr-san4ez maybe I am missing something. Building this PR form source and installing it results in no `known_hosts` file to be present on the target image.

```
vyos@vyos:~$ ls -al .ssh/known_hosts
-rw------- 1 vyos users 806 Oct  8 20:24 .ssh/known_hosts
vyos@vyos:~$ cat .ssh/known_hosts
|1|jmxK2OapO72...2EXY4trcLZ57oU= ssh-ed25519 AAA.XeV3...3F5jUZgrpulWOU
|1|EhS...5nWgjlsUUBbG/FJRtenY+8= ssh-rsa AAAAB3NzaC...LShmfZ
|1|LZl6...AzTVC1llrj331UfctS5/Hk= ecdsa-sha2-nistp256 AAAAE2V...qcQrXDLxiA=
```
```
vyos@vyos:~$ add sys im vyos-2025.10.07-0019-rolling-generic-amd64.iso
The file is 604.000 MiB.
[##################] 100%
Could not download minisig signature
Signature is not available. Do you want to continue with installation? [y/N] y
Validating image checksums
What would you like to name this image? (Default: 2025.10.07-0019-rolling) foo-bar-image
Would you like to set the new image as the default one for boot? [Y/n] y
An active configuration was found. Would you like to copy it to the new image? [Y/n] y
Would you like to copy SSH host keys? [Y/n] y
Copying SSH host keys
Would you like to save the SSH known hosts (fingerprints) from your current configuration? [Y/n] y
Copying SSH known_hosts files
Copying system image files
Cleaning up
Unmounting target filesystems
Removing temporary files
```
```
vyos@vyos:~$ reboot now
Warning: there are unsaved configuration changes!
Run 'save' command if you do not want to lose those changes after reboot/shutdown.

Broadcast message from root@vyos on pts/1 (Wed 2025-10-08 20:43:52 CEST):

The system will reboot now!
```

```
Welcome to VyOS!

   ┌── ┐
   . VyOS 2025.10.07-0019-rolling
   └ ──┘  current

 * Documentation:  https://docs.vyos.io/en/latest
 * Project news:   https://blog.vyos.io
 * Bug reports:    https://vyos.dev

You can change this banner using "set system login banner post-login" command.

VyOS is a free software distribution that includes multiple components,
you can check individual component licenses under /usr/share/doc/*/copyright

---
WARNING: This VyOS system is not a stable long-term support version and
         is not intended for production use.


vyos@vyos:~$ ls -al .ssh
total 12
drwxr-xr-x 2 vyos users 4096 Oct  8 20:44 .
drwxr-xr-x 3 vyos users 4096 Oct  8 20:44 ..
-rw------- 1 vyos users 1599 Oct  8 20:44 authorized_keys
```

jestabro:
The issue above is that if the target image does not also contain this PR, then the `known_hosts` file is never restored: recreating the above, booting into the new image, one sees:
```
vyos@vyos:~$ ls -la .ssh
total 12
drwxr-xr-x 2 vyos users 4096 Oct  9 00:38 .
drwxr-xr-x 3 vyos users 4096 Oct  9 00:38 ..
-rw------- 1 vyos users  150 Oct  9 00:38 authorized_keys
vyos@vyos:~$ ls -la /var/.users_backups/vyos/.ssh/known_hosts
-rw------- 1 root root 978 Oct  9 00:37 /var/.users_backups/vyos/.ssh/known_hosts
```

And since, in my setup, the PR was exactly a single commit ahead of current:
Source image:
```
vyos@vyos:~$ apt info vyos-1x
Package: vyos-1x
Version: 999.0-13250-g8571091b7
```
Target image:
```
vyos@vyos:~$ apt info vyos-1x
Package: vyos-1x
Version: 999.0-13249-g1c181499a
```

Note that although there are some hard restrictions on allowable upgrade images
https://github.com/vyos/vyos-1x/blob/current/src/op_mode/image_installer.py#L799-L800
https://github.com/vyos/vyos-1x/blob/current/src/op_mode/image_installer.py#L1068-L1070
system update is otherwise permissive.

In practice, once the changes were included, it would be a rare issue; nor should one consider stricter update version control based on this example (although we may want to consider stricter update version checks in general, that is a different matter).

c-po (approved):
Implementation looks good!

PR #4683, opened 2025-08-27
T7759: Revert "T7709: add atomic write for config file save"

jestabro:
## Change summary

The utility `write_file_atomic` does not correctly update the file in the bind mount directory `/config`.
Revert for investigation and a proper implementation.

## Types of changes

## Checklist:


sever-sever (approved):
Reverts atomic write config file

dmbaturin (approved):
Since it creates a different bug instead of the old one, I agree with reverting it for the time being.

However, I also wonder where the bind mount issue is. In Python's implementation of atomic writes, or deeper? Maybe it's something we can eventually contribute upstream so that the function stops being a footgun that needs non-intuitive workarounds.

PR #4693, opened 2025-09-03
T7485: ia-na ipv6 for PPPoE link

nvandamme:
## Change summary
If a pppoe interface define `ipv6 address autoconf`, this interface doesn't register the provided IA-NA address from RA's/DHPCv6 announces (dhcp6c).

By modifying the dhcp6c configuration template and handling pppoe interfaces edge cases, the right ia-na entry can be written and handled by dhcp6c to the pppoe interface.

See : https://vyos.dev/T7485#234216

### NOTE 
It doesn't handle requesting a specific IA-NA address (`send ia_na` is always defaulting to `0`).
_Is a new parameter in `interfaces pppoe [pppoeN] ipv6` xml definition should be handled has well?_

## Types of changes

## Related Task(s)

https://vyos.dev/T7485

## Checklist:

### DIFF
--- a/data/templates/dhcp-client/ipv6.j2
+++ b/data/templates/dhcp-client/ipv6.j2
@@ -6,8 +6,12 @@ interface {{ ifname }} {
     send client-id {{ dhcpv6_options.duid }};
 {% endif %}
 {% if address is vyos_defined and 'dhcpv6' in address %}
+{%     if dhcpv6_options.no_request_dns is not vyos_defined %}
     request domain-name-servers;
+{%     endif %}
+{%     if dhcpv6_options.no_request_domain_name is not vyos_defined %}
     request domain-name;
+{%     endif %}
 {%     if dhcpv6_options.parameters_only is vyos_defined %}
     information-only;
 {%     endif %}
--- a/interface-definitions/include/interface/dhcpv6-options.xml.i
+++ b/interface-definitions/include/interface/dhcpv6-options.xml.i
@@ -11,6 +11,18 @@
         <valueless/>
       </properties>
     </leafNode>
+    <leafNode name="no-request-domain-name">
+      <properties>
+        <help>Do not request domain name</help>
+        <valueless/>
+      </properties>
+    </leafNode>
+    <leafNode name="no-request-dns">
+      <properties>
+        <help>Do not request DNS servers</help>
+        <valueless/>
+      </properties>
+    </leafNode>
     <tagNode name="pd">
       <properties>
         <help>DHCPv6 prefix delegation interface statement</help>
--- a/interface-definitions/interfaces_pppoe.xml.in
+++ b/interface-definitions/interfaces_pppoe.xml.in
@@ -16,6 +16,21 @@
           </valueHelp>
         </properties>
         <children>
+          <leafNode name="address">
+            <properties>
+              <help>IP address</help>
+              <completionHelp>
+                <list>dhcpv6</list>
+              </completionHelp>
+              <valueHelp>
+                <format>dhcpv6</format>
+                <description>Dynamic Host Configuration Protocol for IPv6</description>
+              </valueHelp>
+              <constraint>
+                <regex>(dhcpv6)</regex>
+              </constraint>
+            </properties>
+          </leafNode>
           #include <include/pppoe-access-concentrator.xml.i>
           #include <include/interface/authentication.xml.i>
           #include <include/interface/dial-on-demand.xml.i>


data/templates/dhcp-client/ipv6.j2
```diff
@@ -17,6 +17,13 @@ interface {{ ifname }} {
 {%     if dhcpv6_options.rapid_commit is vyos_defined %}
     send rapid-commit; # wait for immediate reply instead of advertisements
 {%     endif %}
+{% elif 'pppoe' in ifname and ipv6.address is vyos_defined and 'autoconf' in ipv6.address %}
```
c-po: DHCPv6 client is only started under two conditions:
* "address" contains "dhcpv6" https://github.com/vyos/vyos-1x/blob/04d106a5e6cbe0c42adef3adddeed6ce2f1d0eee/python/vyos/ifconfig/interface.py#L1284-L1285
* PPPoE interface is used with DHCPv6-PD https://github.com/vyos/vyos-1x/blob/04d106a5e6cbe0c42adef3adddeed6ce2f1d0eee/python/vyos/ifconfig/pppoe.py#L119-L122

The best solution most likely would be to get rid of the if block `{% if address is vyos_defined and 'dhcpv6' in address %}` and use the same template code for both parts. Please give this a try as code simplification is better then adding another if-branch.

c-po (changes requested):

data/templates/dhcp-client/ipv6.j2
nvandamme: dhcpv6 doesn't seem to be a valid keyword under 'interfaces pppoe pppoe0 ipv6 address' ... I'll have to investigate.

c-po: My idea is to maybe entirely get rid of the `{% if address is vyos_defined and 'dhcpv6' in address %}` line. Could you give this a try if it solves your use-case? Then this change would be as easy as removing an if/endif block instead of adding an elif branch.

nvandamme: This line is toward ethernet interfaces only, pppoe is not covered by this case.
One possibility then would be to allow setting dhcpv6 as a valid keyword for pppoe ipv6 address, no?
--> It would allow full support of standard dhcp6c options towards pppoe interfaces in one go without having a special case

The other possibility would be to use an OR condition matching pppoe interfaces instead of using a second case:
```diff
- {% if address is vyos_defined and 'dhcpv6' in address %}
+ {% if (address is vyos_defined and 'dhcpv6' in address) or ('pppoe' in ifname and ipv6.address is vyos_defined and 'autoconf' in ipv6.address) %}
```

What's the best approach?

nvandamme: I've added 'dhcpv6' as a valid keyword in `set interface pppoe pppoeN ipv6 address`.

As dhcpv6 options were also missing dns request control, i've also added to new keys in `interface/dhcpv6-options.xml.i`:
- `no-request-domain-name`: do not request client domain name from dhcpv6 server
- `no-request-dns`: do not request dns servers from dhcpv6 server 

Changed the templates to detect these new keys.

data/templates/dhcp-client/ipv6.j2
```diff
@@ -5,9 +5,13 @@ interface {{ ifname }} {
 {% if dhcpv6_options.duid is vyos_defined %}
     send client-id {{ dhcpv6_options.duid }};
 {% endif %}
-{% if address is vyos_defined and 'dhcpv6' in address %}
+{% if (address is vyos_defined and 'dhcpv6' in address) or (ipv6 is vyos_defined and ipv6.address is vyos_defined and 'dhcpv6' in ipv6.address) %}
```
c-po: Please try if it is possible to drop the entire line 8

nvandamme: - `(address is vyos_defined and 'dhcpv6' in address)` might be removed because the user explicitly expects to have an ipv6 address provided by dhcpv6 for common interfaces.

- the removal of `(ipv6 is vyos_defined and ipv6.address is vyos_defined and 'dhcpv6' in ipv6.address)` will always provide an IA_NA address to the user when dhcpv6-pd is setup on a pppoe interface: **leaving this part ensure the user have the choice to request an ipv6 address for the pppoe interface with or without dhcpv6-pd and vice-versa**.

I very much prefer to leave this line (and line 29) as it is proposed on commit https://github.com/vyos/vyos-1x/pull/4693/commits/e0cefe0f71c62d42a0989e4440d5f256cb868ef9, because it is specifically handling these scenarios: 
- requesting *explicitly* an ipv6 via dhcpv6 on a common interface (not pppoe) via `set interface type name address dhcpv6`
- requesting *explicitly* an ipv6 via dhcpv6 on a pppoe interface ***independently*** of any dhcpv6-pd configuration via `set interface pppoe pppoeN ipv6 address dhcpv6`
- requesting *explicitly* dhcpv6-pd on a pppoe interface ***without*** requesting a dedicated ipv6 for this pppoe interface.

Note: `set interface type name dhcpv6-options temporary` is not explicit and might be confusing for pppoe's case, in my opinion.

@c-po what do you think about it?

c-po: Hi @nvandamme I have not forgotten about this and setup a lab on my side to check this out. I think part of the issue that you now try to solve/revert was introduced in commit ec1cf7dd1508e4a84d99818c7e34d093242b3331 by noone else then myself.

c-po: In addition `set interface pppoe pppoeN ipv6 address dhcpv6` should be `set interface pppoe pppoeN address dhcpv6` to keep all the interface related muscle memory in sync.

c-po:
@nvandamme I've rebased your changes and kept the additions for no domain name/DNS server which is a nice idea.

I also incorporated a different design. PPPoE interfaces now behave like DHCPv6 ethernet interfaces - you can define `set interfaces pppoe pppoe0 address dhcpv6` to also send IA_ND to the DHCPv6 server.

Is this something which will fix your issue, too? I try to keep a common CLI design wherever possible.

nvandamme:
> @nvandamme I've rebased your changes and kept the additions for no domain name/DNS server which is a nice idea.
> 
> I also incorporated a different design. PPPoE interfaces now behave like DHCPv6 ethernet interfaces - you can define `set interfaces pppoe pppoe0 address dhcpv6` to also send IA_ND to the DHCPv6 server.
> 
> Is this something which will fix your issue, too? I try to keep a common CLI design wherever possible.

Nice! Seems to cover all previous explained cases, go for it!

Also, not tried it yet, but we might want to send other values for IA_NA: the user do not want a basic EUI64 generated address, but a specific one with a corresponding IAID (note that wide-dhcpv6-client only support integer and not standard hex based notation though...)

sever-sever (approved):
Adds `dhcpv6` option to PPPoE address
Add the ability to opt out of domain name options.

c-po (approved):
Implementation solved initial bug report from author. Tested by both me and the bug report author.

nvandamme:
@c-po `set system pppoe pppoeN address dhcpv6` seems to error when `interface.update()` is called:

```
[ interfaces pppoe pppoe0 ]
  File "/usr/libexec/vyos/services/vyos-configd", line 156, in run_
script
    script.apply(c)
  File "/usr/libexec/vyos/conf_mode/interfaces_pppoe.py", line 132,
 in apply
    p.update(pppoe)
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/pppoe.py", lin
e 124, in update
    super().update(config)
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/interface.py",
 line 1831, in update
    self.add_addr(addr, vrf_changed=vrf_changed)
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/interface.py",
 line 1286, in add_addr
    elif not is_intf_addr_assigned(self.ifname, addr, netns=netns):
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/vyos/utils/network.py", line
 454, in is_intf_addr_assigned
    if ip_interface(addr) == interface or address == addr:
       ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/ipaddress.py", line 117, in ip_interfac
e
    raise ValueError(f'{address!r} does not appear to be an IPv4 or
 IPv6 interface')
ValueError: 'd' does not appear to be an IPv4 or IPv6 interface

[[interfaces pppoe pppoe0]] failed
Commit failed
[edit]
```

see bug report in https://vyos.dev/T7485

I haven't encountered this bug with the earlier proposed setting path `set system pppoe pppoeN ipv6 address dhcpv6` (and my patched vyos proxmox image is working with it for weeks now).

Is there anything we missed in `interface.py` resulting in this bug? Or maybe a simple malformed string user side?

Also, on `pppoe.py` i was also forcing `self.set_dhcpv6(True)` when `system pppoe pppoeN ipv6 address dhcpv6` was set, but a quick glance to `interface.py` shows that is indeed set later on: https://github.com/vyos/vyos-1x/blob/1b16870043774cf66c5d37ffb281476a42ce8b8b/python/vyos/ifconfig/interface.py#L1284

PR #4717, opened 2025-09-13
dhcpv6: T7646: restore missing default route after upgrade

c-po:
## Change summary

Commit de44c6aef249 ("interface: T7379: do not request SLAAC default route when only DHCPv6 is set") introduced a behavioral change while addressing an earlier issue fixed in e9fb2078d5ea ("interface: T7375: SLAAC assigned address is not cleared when removing SLAAC").

This change caused unintended connectivity loss after upgrading to VyOS 1.4.3. The underlying reason is that VyOS now requires explicit configuration to determine how IPv6 addressing and routing should be handled, rather than making assumptions: Requesting a DHCPv6 address alone will correctly assign an address.

However, since the IPv6 default route is typically advertised via SLAAC (and not DHCPv6), SLAAC must also be explicitly enabled to receive the default route.

Historically, this distinction was implicit and did not require additional configuration.

To preserve backward compatibility, a configuration migrator has been added. It inspects existing configurations that only request a DHCPv6 address and automatically appends the required CLI node to also enable SLAAC, ensuring that the default route is restored after upgrade.

## Types of changes

## Related Task(s)

* https://vyos.dev/T7646

## Checklist:

### DIFF
--- a/src/migration-scripts/interfaces/33-to-34
+++ b/src/migration-scripts/interfaces/33-to-34
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+#
+# Copyright VyOS maintainers and contributors <maintainers@vyos.io>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or later as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# T6318: WiFi country-code should be set system-wide instead of per-device
+
+from vyos.configtree import ConfigTree
+
+base = ['interfaces', 'wireless']
+
+def migrate(config: ConfigTree) -> None:
+    if not config.exists(base):
+        # Nothing to do
+        return
+
+    installed = False
+    for interface in config.list_nodes(base):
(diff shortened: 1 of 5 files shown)


sever-sever (approved):
Restore IPv6 default route via autoconf

src/migration-scripts/interfaces/33-to-34
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
```
dmbaturin, line 1: I have no objections against the main part of the PR related to the missing `autoconf` option. However, I don't understand why it includes a migration script for a task about wireless interface country codes that seems completely unrelated to the issue of T7646. Is it an accidental `git add` or I'm missing something here?

dmbaturin (changes requested):

src/migration-scripts/interfaces/33-to-34
c-po: Since resolving this issue requires a backport to VyOS 1.4, we need to adjust the migration script versioning by incrementing it by one. The `current` and `circinus` branches are already one version ahead of `sagitta`, so aligning the version numbers requires this shift.

We've handled similar versioning shifts in the past, typically during active development cycles when the target branch (in this case, `current` and `circinus`) has not yet reached LTS status. That's why you’re seeing the wireless-related migrations already in place - those changes had their migrator version bumped to leave room for a clean backport to 1.4.

This approach allows us to avoid conflicts while keeping versioning consistent across branches.

sarthurdev (approved):
Restores previous expected behaviour for IPv6 default route

src/migration-scripts/interfaces/33-to-34
dmbaturin: I understand it now, thanks!

dmbaturin (approved):

PR #4728, opened 2025-09-17
nose: T7825: Use nose2 for unit tests

sarthurdev:
## Change summary

Update to use nose2 for unit tests

## Types of changes

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1035

## Checklist:

### DIFF
--- a/python/vyos/airbag.py
+++ b/python/vyos/airbag.py
@@ -24,6 +24,8 @@ from vyos.defaults import airbag_noteworthy_size
 
 
 def enable(log=True):
+    if 'nose2' in sys.modules:
+        return
     if log:
         _intercepting_logger()
     _intercepting_exceptions()
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ clean:
 .PHONY: test
 test: generate-configd-include-json
 	set -e; python3 -m compileall -q -x '/vmware-tools/scripts/' .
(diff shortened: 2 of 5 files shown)


python/vyos/airbag.py
```diff
@@ -157,7 +159,7 @@ def _intercepting_exceptions(_singleton=[False]):
 {instructions}
 
 When reporting problems, please include as much information as possible:
-- do not obfuscate any data (feel free to contact us privately if your
```
dmbaturin: Is it just a formatting change or the message is messed up by that extra space somehow? If it's just formatting, I don't think we should mix it in the same commit with logic changes.

dmbaturin (approved):
python3-nose is no longer shipped in Debian Trixie. python3-nose2 is backward-compatible and does exist in Debian Bookworm already, so I think we should indeed update to it to make our life simpler in the future.

The companion PR in vyos-build is now merged. I only have a simple question about whitespace change.

c-po (approved):

c-po (approved):
Move to a new backwards compatible version python3-nose2. Added `circinus` backport label to make our code more maintainable in the future when backporting other stuff

python/vyos/airbag.py
sarthurdev: Think IDE automatically did this, will revert line.

PR #4737, opened 2025-09-20
T7850: make op_mode_config_dict edit level aware

jestabro:
## Change summary

A simple extension to allow op_mode_config_dict to be edit-level aware when called from within a config session with, say,
`vyos@vyos# run some op path`

Recall that `op_mode_config_dict` is an isolated function to provide a (slightly) faster config dict at a path, when one does not need the full capability, and overhead, of Config instantiation. This PR is provided as more of a feature request than a bug fix, since `op_mode_config_dict` is not an extensible solution, and has limited usefulness until we can drop the overhead of _any_ call to `cli-shell-api.`

## Types of changes

- [X] Other (please describe): feature request

## Checklist:

### DIFF
--- a/python/vyos/utils/list.py
+++ b/python/vyos/utils/list.py
@@ -18,3 +18,26 @@ def is_list_equal(first: list, second: list) -> bool:
     if len(first) != len(second) or len(first) == 0:
         return False
     return sorted(first) == sorted(second)
+
+
+def list_strip(lst: list, sub: list, right: bool = False) -> list:
+    """Remove list 'sub' from beginning (right=False), resp., end of list 'lst'"""
+
+    if not right:
+        while sub:
+            if lst[:1] == sub[:1]:
+                lst = lst[1:]
+                sub = sub[1:]
+            else:
+                lst = []
+                sub = []
+    else:
+        while sub:
+            if lst[-1:] == sub[-1:]:
+                lst = lst[:-1]
+                sub = sub[:-1]
+            else:
+                lst = []
+                sub = []
+
+    return lst
--- a/python/vyos/configquery.py
+++ b/python/vyos/configquery.py
@@ -39,6 +39,7 @@ from vyos.xml_ref import is_tag
 from vyos.base import Warning
 from vyos.utils.backend import vyconf_backend
 from vyos.configsource import ConfigSourceVyconfSession
+from vyos.utils.list import list_strip
 
 config_file = os.path.join(directories['config'], 'config.boot')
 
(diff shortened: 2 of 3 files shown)


python/vyos/utils/list.py
```diff
@@ -18,3 +18,26 @@ def is_list_equal(first: list, second: list) -> bool:
     if len(first) != len(second) or len(first) == 0:
         return False
     return sorted(first) == sorted(second)
+
+
+def list_strip(l: list, k: list, right: bool = False) -> list:  # noqa: E741
+    """Remove list k from beginning (right=False), resp., end of list l"""
```
alexandr-san4ez:
```suggestion
def list_strip(lst: list, sub: list, right: bool = False) -> list:
    """Remove list `sub` from beginning (`right=False`), resp., end of list `lst`"""
```

In my opinion `l` and `k` are not descriptive name of parameters

python/vyos/utils/list.py
```diff
@@ -18,3 +18,26 @@ def is_list_equal(first: list, second: list) -> bool:
...
+
+    if not right:
+        while k:
+            if l[:1] == k[:1]:
+                l = l[1:]
+                k = k[1:]
+            else:
+                l = []
+                k = []
```
alexandr-san4ez: I think would be great to write unit tests for this function. Because logic is unclear if `k` is longer than `l`.

alexandr-san4ez (changes requested):

jestabro:
Thanks for the suggestions @alexandr-san4ez : var name changes made and nosetest added (and rebased over current).

jestabro:
Separate commits.

alexandr-san4ez (approved):
A practical improvement for this use cases - looks good to merge.

dmbaturin (approved):
The implementation seems good and Oleksandr's suggestions are now implemented.

sever-sever (approved):
Works as expected
Manually checked
```
vyos@r14# edit interfaces ethernet eth0
[edit interfaces ethernet eth0]
vyos@r14# sudo python3
Python 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0] on linux
>>> 
>>> 
>>> import os
>>> import subprocess
>>> from vyos.utils.list import list_strip
>>> 
>>> edit_level = os.environ.get('VYATTA_EDIT_LEVEL', '')
>>> path = []
>>> 
>>> if edit_level:
...     tmp = edit_level.split('/')
...     edit_path = [el for el in tmp if el]
...     relative_path = list_strip(path, edit_path)
... else:
...     relative_path = path
... 
>>> def op_mode_run(cmd):
...     """low-level to avoid overhead"""
...     p = subprocess.Popen(cmd, stdout=subprocess.PIPE)
...     out = p.stdout.read()
...     p.wait()
...     return p.returncode, out.decode()
... 
>>> command = ['/bin/cli-shell-api', '--show-active-only', 'showConfig']
>>> 
>>> op_mode_run(command + relative_path)
(0, 'address 192.168.122.14/24\ndescription WAN\nhw-id 52:54:00:77:fa:36\nipv6 {\n    address {\n    }\n}\n')
>>> 

```

PR #4739, opened 2025-09-20
T7852: Switch to yescrypt password encryption

roedie:
## Change summary

Swith the default password encryptions to yescrypt since this is the defalt since Bookworm.

## Types of changes

## Related Task(s)

https://vyos.dev/T7852

No test was done

## Checklist:

### DIFF
--- a/python/vyos/utils/auth.py
+++ b/python/vyos/utils/auth.py
@@ -90,7 +90,7 @@ def evaluate_strength(passwd: str) -> dict[str, str]:
 def make_password_hash(password):
     """ Makes a password hash for /etc/shadow using mkpasswd """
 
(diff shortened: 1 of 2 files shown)


dmbaturin (approved):
I'm inclined to agree with the change. Yescrypt was proven to be more resistant to cracking than SHA-512, and offline cracking attacks are arguably more relevant for VyOS than for many other distros because password hashes are stored in the config, and configs are often backed up or shared, and not always in an encrypted form.

Debian, Ubuntu, Fedora, and Arch have already switched to yescrypt, so I think we can make it the default as well.

sarthurdev (approved):
Sensible default, follows debian standard.

PR #4757, opened 2025-09-25
firewall: T7781: Firewall chains are created when unneeded

l0crian1:
## Change summary

This fixes incorrect behavior where the `input/output/forward/prerouting/etc...` chains were created regardless if a user configured them or not. The checks to prevent this were already present in the `nftables.j2` Jinja2 template, but it was being skipped due to the `<defaultValue>` field for `default-action` making the checks always pass.

This behavior would cause filtering to occur on traffic when there was no intention to filter, effectively slowing down processing.

## Types of changes

## Related Task(s)

https://vyos.dev/T7781

Configure a single chain:
```
vyos@vyos# set firewall ipv4 output filter rule 1 action accept 
vyos@vyos# commit
```
#### Before fix:
```
vyos@vyos# sudo nft list table vyos_filter
table ip vyos_filter {
        chain VYOS_FORWARD_filter {
                type filter hook forward priority filter; policy accept;
                counter packets 0 bytes 0 accept comment "FWD-filter default-action accept"
        }

        chain VYOS_INPUT_filter {
                counter packets 0 bytes 0 accept comment "INP-filter default-action accept"
        }

        chain VYOS_OUTPUT_filter {
                counter packets 0 bytes 0 accept comment "ipv4-OUT-filter-1"
                counter packets 0 bytes 0 accept comment "OUT-filter default-action accept"
        }

        chain VYOS_OUTPUT_raw {
                type filter hook output priority raw; policy accept;
                counter packets 0 bytes 0 accept comment "OUT-raw default-action accept"
        }

        chain VYOS_PREROUTING_raw {
                type filter hook prerouting priority raw; policy accept;
                counter packets 0 bytes 0 accept comment "PRE-raw default-action accept"
        }

        chain VYOS_FRAG_MARK {
                type filter hook prerouting priority -450; policy accept;
                ip frag-off & 0x3fff != 0x0 meta mark set 0x000ffff1 return
        }
}
```
#### After fix:
```
vyos@vyos# sudo nft list table vyos_filter
table ip vyos_filter {
        chain VYOS_OUTPUT_filter {
                counter packets 370 bytes 38689 accept comment "ipv4-OUT-filter-1"
                counter packets 0 bytes 0 accept comment "OUT-filter default-action accept"
        }

        chain VYOS_FRAG_MARK {
                type filter hook prerouting priority -450; policy accept;
                ip frag-off & 0x3fff != 0x0 meta mark set 0x000ffff1 return
        }
}

```
#### Smoketest results:
```
test_bridge_firewall (__main__.TestFirewall.test_bridge_firewall) ... ok
test_cyclic_jump_validation (__main__.TestFirewall.test_cyclic_jump_validation) ... ok
test_disable_conntrack_per_chain (__main__.TestFirewall.test_disable_conntrack_per_chain) ... ok
test_gre_match (__main__.TestFirewall.test_gre_match) ... ok
test_ipsec_metadata_match (__main__.TestFirewall.test_ipsec_metadata_match) ... ok
test_ipv4_remote_group (__main__.TestFirewall.test_ipv4_remote_group) ... ok
test_ipv6_remote_group (__main__.TestFirewall.test_ipv6_remote_group) ... ok
test_remote_group (__main__.TestFirewall.test_remote_group) ... ok
test_zone_with_vrf (__main__.TestFirewall.test_zone_with_vrf) ... ok

Ran 29 tests in 562.998s

OK
```
## Checklist:

### DIFF
--- a/interface-definitions/include/firewall/default-action-base-chains.xml.i
+++ b/interface-definitions/include/firewall/default-action-base-chains.xml.i
@@ -17,6 +17,5 @@
       <regex>(drop|accept)</regex>
     </constraint>
   </properties>
-  <defaultValue>accept</defaultValue>
 </leafNode>
 <!-- include end -->
--- a/data/templates/firewall/nftables.j2
+++ b/data/templates/firewall/nftables.j2
@@ -34,13 +34,13 @@ table ip6 raw {
 delete table ip vyos_filter
 {% endif %}
 table ip vyos_filter {
-{% if ipv4 is vyos_defined %}
-{%     if flowtable is vyos_defined %}
-{%         for name, flowtable_conf in flowtable.items() %}
+{% if flowtable is vyos_defined %}
+{%     for name, flowtable_conf in flowtable.items() %}
 {{ offload_tmpl.flowtable(name, flowtable_conf) }}
-{%         endfor %}
-{%     endif %}
+{%     endfor %}
+{% endif %}
 
+{% if ipv4 is vyos_defined %}
 {%     set ns = namespace(sets=[]) %}
 {%     if ipv4.forward is vyos_defined %}
 {%         for prior, conf in ipv4.forward.items() %}
@@ -222,13 +222,13 @@ table ip vyos_filter {
 delete table ip6 vyos_filter
 {% endif %}
 table ip6 vyos_filter {
-{% if ipv6 is vyos_defined %}
-{%     if flowtable is vyos_defined %}
-{%         for name, flowtable_conf in flowtable.items() %}
+{% if flowtable is vyos_defined %}
+{%     for name, flowtable_conf in flowtable.items() %}
 {{ offload_tmpl.flowtable(name, flowtable_conf) }}
-{%         endfor %}
-{%     endif %}
+{%     endfor %}
+{% endif %}
 
+{% if ipv6 is vyos_defined %}
 {%     set ns = namespace(sets=[]) %}
 {%     if ipv6.forward is vyos_defined %}
 {%         for prior, conf in ipv6.forward.items() %}
--- a/python/vyos/template.py
+++ b/python/vyos/template.py
@@ -600,7 +600,7 @@ def nft_rule(rule_conf, fw_hook, fw_name, rule_id, ip_name='ip'):
 @register_filter('nft_default_rule')
 def nft_default_rule(fw_conf, fw_name, family):
     output = ['counter']
-    default_action = fw_conf['default_action']
+    default_action = fw_conf.get('default_action', 'accept')
     #family = 'ipv6' if ipv6 else 'ipv4'
 
     if 'default_log' in fw_conf:
--- a/smoketest/scripts/cli/test_firewall.py
+++ b/smoketest/scripts/cli/test_firewall.py
@@ -650,6 +650,14 @@ class TestFirewall(VyOSUnitTestSHIM.TestCase):
         self.cli_set(['firewall', 'global-options', 'state-policy', 'related', 'action', 'accept'])
         self.cli_set(['firewall', 'global-options', 'state-policy', 'invalid', 'action', 'drop'])
 
+        self.cli_set(['firewall', 'ipv4', 'input', 'filter', 'rule', '1', 'action', 'accept'])
+        self.cli_set(['firewall', 'ipv4', 'input', 'filter', 'rule', '1', 'protocol', 'tcp'])
+        self.cli_set(['firewall', 'ipv4', 'input', 'filter', 'rule', '1', 'destination', 'port', '22'])
+
+        self.cli_set(['firewall', 'ipv4', 'forward', 'filter', 'rule', '1', 'action', 'accept'])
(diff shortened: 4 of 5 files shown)


interface-definitions/include/firewall/default-action-base-chains.xml.i
```diff
@@ -17,6 +17,5 @@
       <regex>(drop|accept)</regex>
     </constraint>
   </properties>
-  <defaultValue>accept</defaultValue>
```
c-po, line 20: The `defaultValue` not only serves internally as the default for the code, but also for our documentation. `defaultValues` will be automatically attached to the `<help>` message during CLI completion.

The is only one "better/ugly/you name it" solution to the problem you identified:

Mangling the dict and removing defaults if required.

https://github.com/vyos/vyos-1x/blob/5845c4b1c50bc0335284cfb3306e0a91de3efd40/python/vyos/frrender.py#L78-L83

l0crian1: Generating unnecessary config, just to turn around and delete it so it doesn't cause unintended behavior seems a little backwards to me. It's treating a symptom rather than solving the problem.

Is the only additional concern the help string? I can just manually update it to:
```
Default-action for rule-set (default: accept)
```
The `defaultValue` has a lot of value within the code, (like for the firewall global settings), but to enable a default value of `accept` in python, all that is required was this change (along with deleting the `defaultValue`):
``` python
    default_action = fw_conf.get('default_action', 'accept')
```
All of the other changes (smoketests, verify(), Jinja2, etc...) would be required regardless of the solution. Does keeping the default value in the XML really buy anything at that point?

c-po: > Is the only additional concern the help string? I can just manually update it to:

Currently this is true - but it will (unfortunately until we find a generic solution to this issue) decouple the CLI help string from the internal implementation. I am a fan of a generic solution but yet have no proper idea how to handle this in the firewall context. Thus I tend to agree we go with your implementation.

Please add a `<!-- XML COMMENT -->` to the XML help string indicating that the default you define here in the help string is implemented in a different file, so we have a future reference by comment in case anyone starts wondering.

jestabro:
Following conversations with @l0crian1 , an alternative to dropping the `defaultValue` element is a simple pruning step (diff below). His additional suggestion to codify this approach by adding hints within the XML, for _not_ including paths _solely_ of default values bears consideration; that is an ongoing discussion ... In the interim, we may consider the suggested pruning step as a solution to inform any later revision, as the key idea is to remove (or not include) certain paths not explicitly set --- N.B. neither this nor the original PR should be adopted without careful review: it is unclear if there are cases where we would want to retain the default paths. Note that with the changes to the smoketest and nftables.j2 in this PR, firewall smoketests pass with the alternative patch below. Tested version here: https://github.com/vyos/vyos-1x/compare/current...jestabro:prune-default-chains
```
From cefef39c2846813ff5aa43c0ef2f28fd4f5d473c Mon Sep 17 00:00:00 2001
From: John Estabrook <jestabro@vyos.io>
Date: Mon, 20 Oct 2025 15:38:01 -0500
Subject: [PATCH 1/3] firewall: T7781: prune config_dict chains added solely by
 defaultValue

---
 src/conf_mode/firewall.py | 16 ++++++++++++++++
 1 file changed, 16 insertions(+)

diff --git a/src/conf_mode/firewall.py b/src/conf_mode/firewall.py
index 6630b811d..ad899ef20 100755
--- a/src/conf_mode/firewall.py
+++ b/src/conf_mode/firewall.py
@@ -22,6 +22,7 @@
 from sys import exit
 from vyos.base import Warning
+from vyos.config import ConfigDict
 from vyos.configdict import is_node_changed
 from vyos.configdiff import get_config_diff, Diff
 from vyos.configdep import set_dependents, call_dependents
@@ -117,6 +118,20 @@ def geoip_updated(conf, firewall):

     return False

+
+def prune_chains_added_by_default(d: ConfigDict):
+    for p in ('ipv4', 'ipv6', 'bridge'):
+        for q in ('forward', 'input', 'output', 'prerouting'):
+            for r in ('filter', 'raw'):
+                path = [p] + [q] + [r]
+                if d.from_defaults(path):
+                    del d[p][q][r]
+                    if not d[p][q]:
+                        del d[p][q]
+        if p in d and not d[p]:
+            del d[p]
+
+
 def get_config(config=None):
     if config:
         conf = config
@@ -129,6 +144,7 @@ def get_config(config=None):
                                     get_first_key=True,
                                     with_recursive_defaults=True)

+    prune_chains_added_by_default(firewall)

     firewall['group_resync'] = bool('group' in firewall or is_node_changed(conf, base + ['group']))
     if firewall['group_resync']:
-- 
2.39.5

```

jestabro (changes requested):
See summary of discussion above: this is not a request for changes, but more of a request for further discussion.

c-po:
As discussed @l0crian1 let's proceed with closing this PR and rather opt for a general solution to this issue.

PR #4758, opened 2025-09-26
op-mode: T7871: add support for op mode command argument constraints

dmbaturin:
## Change summary

To allow non-admin users to execute most op mode commands, we need to ensure they can't enter arguments that can lead to shell escapes.

A blanket ban on all special characters is too restrictive, so we need a way to relax it where needed.

One way to do that is to ensure that arguments follow specific formats. That can be done using constraint checks similar to those we already have in configuration mode.

We can use the same syntax as we use for configuration mode definitions, with `<regex>` and `<validator>` tags:

```
<interfaceDefinition>
  <tagNode name="ping">
    <properties>
      <help>Send Internet Control Message Protocol (ICMP) echo request</help>
      <completionHelp>
        <list>&lt;hostname&gt; &lt;x.x.x.x&gt; &lt;h:h:h:h:h:h:h:h&gt;</list>
      </completionHelp>
      <constraint>
        <validator name="ip-address"/>
        <validator name="fqdn"/>
      </constraint>
    </properties>
    ...
```

We can probably omit support for constraint groups until we find a real case for that.

In the cache, it's probably a good idea to keep regex and validator constraints separate so that the runner can evaluate regexes first — it can do that internally. If none of the regexes match, then it can jump to much more expensive external checks.

```
      "constraints": {
        "regexes": [],
        "validators": [
          {
            "name": "ip-address",
            "argument": null
          },
          {
            "name": "fqdn",
            "argument": null
          }
        ]
      },
```

## Types of changes

## Checklist:

### DIFF
--- a/python/vyos/xml_ref/generate_op_cache.py
+++ b/python/vyos/xml_ref/generate_op_cache.py
@@ -139,6 +139,39 @@ def sort_op_data(obj):
     key_func = cmp_to_key(compare_keys)
     return sort_func(obj, key_func)
 
+def get_constraints(props):
+    if props is None:
+        return None
+
+    # Put regexes and validators into separate dict fields.
+    # Since multiple constraints work like logical OR,
+    # it's better for the runner can evaluate regexes internally first,
+    # which is much faster than calling external validators.
+    constraints = {
+      'regexes': [],
+      'validators': []
+    }
+
+    constraint_elem = props.find('constraint')
+
+    if constraint_elem is not None:
+        constraint_elems = list(constraint_elem)
+        if constraint_elems:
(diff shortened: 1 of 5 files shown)


jestabro (approved):
This looks quite reasonable to me. Until constraints are integrated into the op-mode-runner itself, there is a limit on what can be tested here. However, one can check that the cache generation is sound: a simple test of the ping example given in the PR, and an analogous check of a regex for `show interfaces ethernet`, give the correct results.

c-po (approved):
The implemenation looks good and it adds a feature I was long missing. Thanks

PR #4771, opened 2025-10-03
T7903: set airbag enable to default log=False

jestabro:
## Change summary

The airbag module logging has been implicated in recent bugs, notably [T7117](https://vyos.dev/T7117). The recommendation is to set the default log=False, meaning that sys.stderr is not by default redirected by a singleton class. Pending further analysis of the root cause of errors, this should be uncontentious as a temporary measure, since

- the current default does not add much information, as stderr is not commonly used for script/module output
- an author can always choose to selectively enable for specific scripts as needed
- revision of airbag logging should be considered after analysis and in conjunction with updates to the logger module

## Types of changes

## Checklist:

### DIFF
--- a/python/vyos/airbag.py
+++ b/python/vyos/airbag.py
@@ -23,7 +23,7 @@ from vyos.version import get_full_version_data
 from vyos.defaults import airbag_noteworthy_size
 
 
-def enable(log=True):
+def enable(log=False):
     if 'nose2' in sys.modules:
         return
     if log:


sever-sever (approved):
Disabling log fixes the `recursionError` issue.
I think it's a good change until we have another fix.

c-po (approved):
Disabel stderr logging by default.

PR #4774, opened 2025-10-03
vyos.utils.network: T7898: check if system UUID is available before trying to use it for host identity generation

dmbaturin:
## Change summary

The original implementation unconditionally assumed that `/sys/class/dmi/id/product_uuid` always exists. As [T7898](https://vyos.dev/T7898) shows, it's not always the case. I could reproduce that in a KVM guest spawned manually with `qemu-kvm`, and I think I have seen hardware boxes without UUIDs even if long time ago — but as long as such machines exist at all, and they do, we need to consider the situation when product UUID is unavailable.

## Types of changes

## Checklist:

### DIFF
--- a/python/vyos/utils/network.py
+++ b/python/vyos/utils/network.py
@@ -55,7 +55,8 @@ def get_host_identity() -> str:
     Build a stable host identity string for deterministic MAC generation.
 
     Combines:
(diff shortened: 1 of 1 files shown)


sever-sever (approved):
Use `hostname` if the `product_uuid` does not exist

jestabro (approved):
Handle case of uuid_file not present.

PR #4799, opened 2025-10-17
T7942: consistent naming of "memory" in op-mode

c-po:
## Change summary

We have the following memory related commands:
* `show hardware mem`
* `show system memory <cache | detail | routing-daemons>`

Most of the time we call it memory so this should be also made consistent in the op-mode CLI. This is NOT a breaking change as mem is an abbreviation to memory, so it's backwards compatible.

## Types of changes

- [x] Other (please describe): Cosmetic change

## Related Task(s)

* https://vyos.dev/T7942

## Checklist:

### DIFF
--- a/op-mode-definitions/show-hardware.xml.in
+++ b/op-mode-definitions/show-hardware.xml.in
@@ -33,7 +33,7 @@
             </properties>
             <command>dmidecode</command>
           </node>
-          <node name="mem">
+          <node name="memory">
             <properties>
               <help>Show system RAM details</help>
             </properties>


sever-sever (approved):
Use `memory` keyword instead of `mem`

dmbaturin (approved):
It is indeed a very old inconsistency and since it's not exposed in machine-readable op mode yet, we can just change it. For human users, `run show hardware mem` will still work because `mem` is an unambiguous shortening of `memory`.

PR #4807, opened 2025-10-23
T7910: Standardize vyconf session resource management

jestabro:
## Change summary

Standardize VyconfSession instance lifetime management across its use in the various config modules.
In summary:
- use weakref.finalize for reliable session teardown
- op-mode instances are ephemeral: a new session on init; teardown with gc
- config mode instances are persistent, with lifetime managed by CLI session, resp. ConfigSession
- add request `show_sessions` to return list of dicts of internal session record data structure, for data tracking and proof of correctness

## Types of changes

## Related PR(s)

https://github.com/vyos/vyconf/pull/32

A screenshot showing the persistence and lifetime of a config-mode session:

Screenshot_20251104_154153

## Checklist:


jestabro:
Rebase over current to pull in hash update for `show_sessions`. Ready for merge upon approval.

jestabro:
Note that darker-ruff-lint will never be happy with the auto-generated files, `python/vyos/proto/vyconf_pb2.py`, `python/vyos/proto/vyconf_proto.py`. We will exclude them from linting in the future.

python/vyos/proto/vyconf_proto.py
```diff
@@ -405,3 +411,9 @@ def set_request_reload_reftree(token: str = None, on_behalf_of: int = None):
     req = Request(reload_reftree=reqi)
     req_env = RequestEnvelope(token, req)
     return req_env
+
+def set_request_show_sessions(token: str = None, exclude_self: bool = False, exclude_other: bool = False):
+    reqi = ShowSessions (exclude_self, exclude_other)
```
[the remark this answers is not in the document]
jestabro: This is an artifact of the auto-generation script; as it causes no harm, it will be fixed in a separate task.

python/vyos/vyconf_session.py
```diff
@@ -36,45 +38,75 @@ class VyconfSessionError(Exception):
...
+                    raise ValueError(f'No existing session for token: {token}')
+                self.__token = token
+
+        if not self.in_config_session and not extant:
+            self._finalizer = weakref.finalize(self, self._teardown, self.__token)
 
         self.on_error = on_error
 
+    def _teardown(self, token):
+        vyconf_client.send_request('teardown', token)
```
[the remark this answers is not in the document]
jestabro: No, it purposely ignores any response (hence error handling) as the session is destroyed, hence can not respond.

python/vyos/vyconf_session.py
```diff
@@ -95,6 +127,21 @@ def get_config(self):
...
+        self, exclude_self: bool = False, exclude_other: bool = False
+    ) -> list | dict:
+        out = vyconf_client.send_request(
+            'show_sessions',
+            token=self.__token,
+            exclude_self=exclude_self,
+            exclude_other=exclude_other,
+        )
+
+        lst = json.loads(out.output)
```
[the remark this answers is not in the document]
jestabro: No, the JSON of out.output is generated from automatic serialization; this is not the point to check for errors (as they would have been caught early).

python/vyos/configsession.py
```diff
@@ -223,12 +229,13 @@ def __del__(self):
...
-                self._vyconf_session.discard()
-            self._vyconf_session.exit_config_mode()
-            self._vyconf_session.teardown()
+
+    def finalize_vyconf(self, session: VyconfSession):
+        if session.session_changed():
+            Warn('Exiting with uncommitted changes')
+            session.discard()
+        session.exit_config_mode()
+        session.teardown()
```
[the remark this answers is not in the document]
jestabro: Similar issue with teardown mentioned in an earlier copilot comment: we are purposely ignoring errors in this function.

src/helpers/teardown-config-session.py
```diff
@@ -23,5 +23,10 @@
 
 pid = sys.argv[1]
 
-vc = VyconfSession(pid=pid)
-vc.teardown()
+try:
+    vc = VyconfSession(pid=int(pid), extant=True)
```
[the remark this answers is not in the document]
jestabro: No the provenance of the pid in this limited use function is straight from the bash shell -> python script arg -> back to int.

jestabro:
All Copilot suggestions resolved above (none appropriate, but for one issue of formatting in the dataclass generation script, to be handled elsewhere).

python/vyos/configsession.py
```diff
@@ -199,10 +200,15 @@ def __init__(self, session_id, app=APP, shared=False):
 
         self.shared = shared
 
+        if not self.shared and self._vyconf_session:
+            self._finalizer = weakref.finalize(
```
hedrok, line 204: From weakref.finalize documentation (https://docs.python.org/3/library/weakref.html#weakref.finalize):

> Note It is important to ensure that func, args and kwargs do not own any references to obj, either directly or indirectly, since otherwise obj will never be garbage collected. In particular, func should not be a bound method of obj.

In such case object self is stored till end of script run, so it is same as using atexit. Is this intentional?.. If so, I suggest writting a comment about that.

I've checked this on simple example:

```python
#!/usr/bin/env python3

import weakref

class C:
    def __init__(self, n):
        self.n = n
    def finalizer(self, msg: str):
        print(f"In finalizer! Storing number: {self.n}")
        print(f"Final meesage: {msg}")

a = C(32)
b = C(17)

weakref.finalize(a, a.finalizer, 'a finalizer')
weakref.finalize(b, print, 'b finalizer')

print("registered")

del a
del b

print("del all")
```

Output on my machine:

```shell
vyos ~/reviews/VD-1893$ ./test.py
registered
b finalizer
del all
In finalizer! Storing number: 32
Final meesage: a finalizer
```
`b` is finalized after `del`, but `a` only at exit.

python/vyos/vyconf_session.py
```diff
@@ -36,45 +38,75 @@ class VyconfSessionError(Exception):
...
+    )
+    return out.output
+
+
 class VyconfSession:
     def __init__(
-        self, token: str = None, pid: int = None, on_error: Type[Exception] = None
+        self,
+        pid: int = None,
+        token: str = None,
```
hedrok, line 55: Order of `pid` and `token` is changed, is this intentional?..

python/vyos/vyconf_session.py
```diff
@@ -95,6 +127,21 @@ def get_config(self):
...
+        out = vyconf_client.send_request(
+            'show_sessions',
+            token=self.__token,
+            exclude_self=exclude_self,
+            exclude_other=exclude_other,
+        )
+
+        lst = json.loads(out.output)
+        if len(lst) == 1:
+            return lst[0]
```
hedrok, line 143: Maybe it would be better to return list always?.. I think it can be quite unexpected, and it will be hard to use in scripts if there will be such need.

hedrok (changes requested):

python/vyos/configsession.py
jestabro: Yes, thanks: the finalizer functions need to be class methods; I will make those changes.

python/vyos/vyconf_session.py
jestabro: Yes, that is the content of:
```
commit 3022334f79b5e34c26bac7ce1c6e716663061293
Author: John Estabrook <jestabro@vyos.io>
Date:   Wed Oct 22 13:45:10 2025 -0500

    T7910: switch keyword order pid/token for consistency and intuition
    
    pid is the more commonly passed argument, and this maintains consistency
    with configsession arg use.
```

python/vyos/vyconf_session.py
jestabro: Yes, it's odd, but for certain functions designed for internal use, it can make some sense; cf.
https://github.com/vyos/vyos-1x/blob/current/python/vyos/utils/config.py#L27-L45

python/vyos/configsession.py
jestabro: Thanks again for that review, @hedrok : an important point. The PR has been updated; your test script with analogous modifications and a sanity check would look like:
```python
#!/usr/bin/env python3

import weakref
from time import sleep

class C:
    def __init__(self, n):
        self.n = n
        weakref.finalize(self, self.finalizer, self.n)
    @classmethod
    def finalizer(cls, val: int):
        print(f"In finalizer!")
        print(f"Final message: {str(val)}")

a = C(32)
b = C(17)

# Create a weak reference to the object
weak_ref_a = weakref.ref(a)
weak_ref_b = weakref.ref(b)

# Check if the object is still referenced
if weak_ref_a():
    print("Object 'a' is still alive.")
else:
    print("Object 'a' has been garbage collected.")

if weak_ref_b():
    print("Object 'b' is still alive.")
else:
    print("Object 'b' has been garbage collected.")

print("registered")

del a
del b

print('deleted both objects')
print('calling garbage collection:')

import gc
gc.collect()

# Check if the object is still referenced
if weak_ref_a():
    print("Object 'a' is still alive.")
else:
    print("Object 'a' has been garbage collected.")

if weak_ref_b():
    print("Object 'b' is still alive.")
else:
    print("Object 'b' has been garbage collected.")

print("leaving in a moment ...")
sleep(1)
```

hedrok: I approved the PR as I wouldn't write anything if it was classmethod originally.
But as we already have discussion I would like to add one consideration: I suggest changing `self.finalizer` to `ConfigSession.finalizer` or to `type(self).finalizer` if you think we need to support `finalizer` overriding.

This way anyone who reads code won't have to recheck that it is classmethod, and if someone doesn't think about this and removes `@classmethod` or overrides with object method they will get an error.

Yet again updated demo:

```python
#!/usr/bin/env python3

import weakref
from time import sleep

class C:
    def __init__(self, n):
        self.n = n
        weakref.finalize(self, self.finalizer, self.n)
    @classmethod
    def finalizer(cls, val: int):
        print(f"In finalizer!")
        print(f"Final meesage: {str(val)}")

class D(C):
    @classmethod
    def finalizer(cls, val: int):
        print(f"D.finalizer with {val=}")

class OhNo(C):
    def finalizer(cls, val: int):
        print(f"OhNo.finalizer with {val=}")

a = C(32)
b = D(17)
c = OhNo(131)

print("registered")

del a
del b
del c

print("del all")
sleep(1)
```

`c` is destructed only at exit.

If we add `type()`, we get an error in stdout (this is the way weakref.finalize work):

```
Exception ignored in: <finalize object at 0x7ffa4f148ae0; dead>
  File "/nix/store/bbyp6vkdszn6a14gqnfx8l5j3mhfcnfs-python3-3.12.11/lib/python3.12/weakref.py", line 590, in __call__
    return info.func(*info.args, **(info.kwargs or {}))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: OhNo.finalizer() missing 1 required positional argument: 'val'
```

hedrok (approved):

python/vyos/configsession.py
jestabro: That's a good point. I'll adopt (some version of) those checks going forward. Thanks, @hedrok

jestabro:
Smoketest errors above unrelated to PR.

PR #4811, opened 2025-10-25
T7957: filter stderr when deleting container images

c-po:
## Change summary

The call to `rc_cmd('podman inspect ...')` determines whether a container image scheduled for deletion has any ancestor containers still using it. Previously, if the podman command wrote output to stderr, `rc_cmd()` would return that error message alongside or instead of the ancestor container ID.

This caused subsequent podman calls to fail, as the error string was incorrectly treated as a valid command argument. This change ensures only valid ancestor IDs are returned.

This is a fix for commit a99ca6d11b5 ("op-mode: T7403: add option for forcefully remove a container image")

## Types of changes

## Related Task(s)

* https://vyos.dev/T7957

## Related PR(s)

* https://github.com/vyos/vyos-1x/pull/4590

```
vyos@vyos:~$ delete container image cd91c0f1d8b3
Cannot delete image "cd91c0f1d8b3" because it is currently being used by container "3b42e5405dec"!
```

## Checklist:

### DIFF
--- a/src/op_mode/container.py
+++ b/src/op_mode/container.py
@@ -120,11 +120,11 @@ def delete_image(name: str, force: typing.Optional[bool] = False):
 
     for image in name:
         # convert the truncated image ID to a full image ID
-        rc, ancestor = rc_cmd(f'podman inspect {image} --format "{{{{.Id}}}}"')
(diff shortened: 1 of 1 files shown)


sarthurdev (approved):
Fixes issue with container op-mode due to stderr output.

sever-sever (approved):
Only the actual command output is captured in ancestor.

PR #4818, opened 2025-10-29
wlb: T7966: Restore default route when interface disconnects/reconnects

l0crian1:
## Change summary

When an interface disconnects, like if a user needs to unplug interfaces, the default routes within the tables used for `ip rule` policy routing are removed. When the interface returns to an operational state, the default routes are not repopulated since the nexthop hasn't changed.

This change adds a check to ensure when the default route is dropped, it can be restored.
## Types of changes

## Related Task(s)

https://vyos.dev/T7966
#### Configure WAN Load Balancing:
```
set interfaces ethernet eth0 vif 20 address 'dhcp'
set interfaces ethernet eth0 vif 101 address 'dhcp'
set interfaces ethernet eth1 address '10.5.5.1/24'

set load-balancing wan enable-local-traffic
set load-balancing wan interface-health eth0.20 nexthop 'dhcp'
set load-balancing wan interface-health eth0.101 nexthop 'dhcp'
set load-balancing wan rule 10 inbound-interface 'eth1'
set load-balancing wan rule 10 interface eth0.20
set load-balancing wan rule 10 interface eth0.101
```
#### Check the `ip rule` and routing tables:
```
vyos@vyos# sudo ip rule show
0:      from all lookup local
32764:  from all fwmark 0xca lookup 202
32765:  from all fwmark 0xc9 lookup 201
32766:  from all lookup main
32767:  from all lookup default

vyos@vyos# sudo ip route show default table 201
default via 192.168.2.1 dev eth0.20 

vyos@vyos# sudo ip route show default table 202
default via 10.0.101.1 dev eth0.101 
```
#### Disconnect the interface, and then reconnect the interface. After, check the default routes in the tables again, they will not be present. This will lead to traffic failures as traffic leaves incorrect interfaces, and is not source NAT-ed:
```
vyos@vyos# sudo ip route show default table 201

vyos@vyos# sudo ip route show default table 202

```
#### Disconnect the interface, and then reconnect the interface. After, check the default routes in the tables again, they will now be repopulated:
```
vyos@vyos# sudo ip route show default table 201
default via 192.168.2.1 dev eth0.20 

vyos@vyos# sudo ip route show default table 202
default via 10.0.101.1 dev eth0.101 
```
## Checklist:

### DIFF
--- a/src/helpers/vyos-load-balancer.py
+++ b/src/helpers/vyos-load-balancer.py
@@ -103,15 +103,36 @@ def get_ipv4_address(ifname):
                     return addr_json['addr_info'][0]['local']
     return None
 
+def get_dynamic_nexthop(ifname: str) -> str | None | bool:
+    '''
+    Resolve the dynamic next-hop for a WAN interface.
+
+    Determines the current default gateway learned dynamically on the interface:
+    - PPPoE interfaces (`pppoe*`): uses `parse_ppp_nexthop`.
+    - Other interfaces (e.g. DHCP): uses `parse_dhcp_nexthop`.
+
+    Return values:
+    - str: IPv4 next-hop address
+    - None: when DHCP lease has no router value
+    - False: when PPPoE nexthop state file is missing
+
+    Args:
+        ifname: Interface name (e.g. 'pppoe0', 'eth0').
+
+    Returns:
+        See above for possible values.
+    '''
+    if ifname.startswith('pppoe'):
+        return parse_ppp_nexthop(ifname)
+    else:
+        return parse_dhcp_nexthop(ifname)
+
 def dynamic_nexthop_update(lb, ifname):
     # Update on DHCP/PPP address/nexthop changes
     # Return True if nftables needs to be updated - IP change
 
     if 'dhcp_nexthop' in lb['health_state'][ifname]:
-        if ifname[:5] == 'pppoe':
-            dhcp_nexthop_addr = parse_ppp_nexthop(ifname)
-        else:
-            dhcp_nexthop_addr = parse_dhcp_nexthop(ifname)
+        dhcp_nexthop_addr = get_dynamic_nexthop(ifname)
 
         table_num = lb['health_state'][ifname]['table_number']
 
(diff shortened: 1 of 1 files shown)


src/helpers/vyos-load-balancer.py
```diff
@@ -125,6 +126,43 @@ def dynamic_nexthop_update(lb, ifname):
...
+        return
+
+    rc, out = rc_cmd(f'ip -j route show default table {table_num}')
+    if rc == 0:
+        rt_table = json.loads(out)
+        if len(rt_table) > 0:
+            return
+        else:
+            if 'dhcp_nexthop' in lb['health_state'][ifname]:
+                if ifname[:5] == 'pppoe':
```
c-po:
```suggestion
                if ifname.startswith('pppoe'):
```

src/helpers/vyos-load-balancer.py
```diff
@@ -295,6 +333,9 @@ def handle_sigterm(signum, frame):
                     if dynamic_nexthop_update(lb, ifname):
                         ip_change = True
 
+                    restore_default_route(lb, ifname)
+
```
c-po, line 355: One newline should be sufficient.

c-po (changes requested):

src/helpers/vyos-load-balancer.py
l0crian1: @c-po i was reusing a block of code found elsewhere in that file. There would be a difference in those blocks of code if I changed that, though I agree `startswith` is proper. 

I can also move that block into its own function and update the `dynamic_nexthop_update()` function. I try not to refactor much when doing a fix, but maybe that’s more proper in this case. 

Just let me know if you want me to:
1. Keep the slicing like in the other function
2. Update the new function with `startswith`
3. Refactor that block into a `get_dynamic_nexthop()` function and update the existing `dynamic_nexthop_update()` function to use that

I’m fine with anything, but am partial to option 3.

c-po: Please proceed with option 3

l0crian1: Done!

src/helpers/vyos-load-balancer.py
l0crian1: Done!

c-po (approved):
Restore WLB route if interface (e.g. pppoe) reconnected and next-hop needs update.

sever-sever (approved):
Re-add the default route or a table X after reconnection for WLB

PR #4835, opened 2025-11-05
T7731: Static ARP entries are missing after an interface status change

natali-rs1985:
## Types of changes

## Related Task(s)

* https://vyos.dev/T7731

```
set interfaces ethernet eth1 address 10.0.0.1/24
set protocols static arp interface eth1 address 10.0.0.2 mac 00:11:22:33:44:55
commit

vyos@vyos# sudo ip neighbor
10.0.0.2 dev eth1 lladdr 00:11:22:33:44:55 PERMANENT
10.20.40.1 dev eth2 lladdr 52:54:00:96:c3:5d STALE
```
Change interface status (VPP configuration requires interface reinitialization)
```
commit
```
Before the fix ARP entry was missing
```
vyos@vyos# sudo ip neighbor
10.20.40.1 dev eth2 lladdr 52:54:00:96:c3:5d STALE
```
After the fix it's in place
```
vyos@vyos# sudo ip neighbor
10.0.0.2 dev eth1 lladdr 00:11:22:33:44:55 PERMANENT
10.20.40.1 dev eth2 lladdr 52:54:00:96:c3:5d STALE
```

```
vyos@vyos# /usr/libexec/vyos/tests/smoke/cli/test_vpp.py -k test_21_static_arp
test_21_static_arp (__main__.TestVPP.test_21_static_arp) ... ok

Ran 1 test in 34.444s

OK
```

## Checklist:

### DIFF
--- a/src/conf_mode/interfaces_bonding.py
+++ b/src/conf_mode/interfaces_bonding.py
@@ -175,6 +175,10 @@ def get_config(config=None):
             tmp = has_vrf_configured(conf, interface)
             if tmp: bond['member']['interface'][interface].update({'has_vrf' : ''})
 
+    # Protocols static arp dependency
+    if 'static_arp' in bond:
+        set_dependents('static_arp', conf)
+
     return bond
 
 
@@ -279,7 +283,7 @@ def apply(bond):
     else:
         b.update(bond)
 
-    if dict_search('member.interface_remove', bond):
+    if dict_search('member.interface_remove', bond) or 'static_arp' in bond:
         try:
             call_dependents()
         except ConfigError:
--- a/src/conf_mode/interfaces_bridge.py
+++ b/src/conf_mode/interfaces_bridge.py
@@ -124,6 +124,10 @@ def get_config(config=None):
         if len(bridge['member']) == 0:
             del bridge['member']
 
+    # Protocols static arp dependency
+    if 'static_arp' in bridge:
+        set_dependents('static_arp', conf)
+
     return bridge
 
 def verify(bridge):
@@ -213,12 +217,20 @@ def apply(bridge):
         if 'interface' in bridge['member']:
             tmp.extend(bridge['member']['interface'])
 
-    for interface in tmp:
-        if interface.startswith(tuple(['vxlan', 'wlan'])) and interface_exists(interface):
-            try:
-                call_dependents()
-            except ConfigError:
-                raise ConfigError(f'Error updating member interface {interface} configuration after changing bridge!')
+    # collect member interfaces that require dependent updates
+    interfaces_need_update = [
+        iface
+        for iface in tmp
+        if iface.startswith(('vxlan', 'wlan')) and interface_exists(iface)
+    ]
+
+    if interfaces_need_update or 'static_arp' in bridge:
+        try:
+            call_dependents()
+        except ConfigError:
+            raise ConfigError(
+                'Error updating member interface configuration after changing bridge!'
+            )
 
     return None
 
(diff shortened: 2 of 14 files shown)


src/conf_mode/interfaces_bonding.py
```diff
@@ -175,6 +175,16 @@ def get_config(config=None):
             tmp = has_vrf_configured(conf, interface)
             if tmp: bond['member']['interface'][interface].update({'has_vrf' : ''})
 
+    # Protocols static arp dependency
+    static_arp = conf.get_config_dict(
```
c-po: This is highly duplicated code. Why not simply place one instance of it into: `get_interface_dict()`

c-po (changes requested):

src/conf_mode/interfaces_bonding.py
natali-rs1985: Made changes.

src/conf_mode/interfaces_ethernet.py
```diff
@@ -20,6 +20,7 @@
 
 from vyos.base import Warning
 from vyos.config import Config
+from vyos.configdep import set_dependents, call_dependents
```
c-po: Please use one import per line - this makes it easier `grep` - able for refactorings

```suggestion
from vyos.configdep import set_dependents
from vyos.configdep import call_dependents
```

src/conf_mode/interfaces_wireless.py
```diff
@@ -84,7 +85,7 @@ def get_config(config=None):
         conf = Config()
     base = ['interfaces', 'wireless']
 
-    _, wifi = get_interface_dict(conf, base)
+    ifname, wifi = get_interface_dict(conf, base)
```
c-po: Where is `ifname` needed?

`_` was used to indicate this variable is unused.

c-po (changes requested):

c-po:
Two more minor changes, rest looks good to me. Happy to approve after the changes were made.

src/conf_mode/interfaces_wireless.py
natali-rs1985: I just forgot to change it back.

c-po (approved):
Implementation looks sane. All requested changes got incorporated into this PR. Thanks!

src/conf_mode/interfaces_bridge.py
```diff
@@ -220,6 +224,9 @@ def apply(bridge):
             except ConfigError:
                 raise ConfigError(f'Error updating member interface {interface} configuration after changing bridge!')
 
+    if 'static_arp' in bridge:
+        call_dependents()
+
```
zdc, line 234: I may be wrong, but I think this will not work.

If dependencies are already called in line 223, we should not do this again. So, we can update the condition for calling them, but `call_dependents()` should be only one.

src/conf_mode/interfaces_vxlan.py
```diff
@@ -249,6 +255,8 @@ def apply(vxlan):
         v = VXLANIf(**vxlan)
         v.update(vxlan)
 
+    call_dependents()
```
zdc: Is it safe to use it unconditionally?

Same in a few other files.

zdc (changes requested):
Two main concerns:

1. In one instance, `call_dependents()` is called conditionally, while in others, it is unconditional. One of these approaches is likely incorrect. If both are valid, I would prefer consistency across all instances.

2. Calling `call_dependents()` multiple times within the same file seems risky.

src/conf_mode/interfaces_vxlan.py
natali-rs1985: In this case yes. We call only one dependency (static_arp)

zdc (approved):
The code looks good for me now.

PR #4836, opened 2025-11-06
smoketest: T7996: fix hardcoded "vrf red" search path in FRR config

c-po:
## Change summary

Commit 186900f7165b2 (`smoketest: T7927: test DHCP route preservation`) which was added to validate a bugfix for DHCP default routes added it's own little regression. Tests defined a VRF named red15, but when reading in the FRR configuration we have had a hardcoded search for `vrf red` instead. 

This has been fixed to re-use the defined VRF variable name we perform our test on.

The original VRF DHCP test case, introduced in commit 186900f7165b2 (`smoketest: T7927: test DHCP route preservation`), used a fixed delay of 8 seconds to wait for the DHCP server and client to initialize. This approach was unreliable and could cause unnecessary test delays.

Since commit 354517677f (`wlb: T7977: Updated smoketest to validate nft vmap weight buckets`), the `vyos.utils.misc.wait_for()` helper provides a more efficient busy-wait loop with an early-exit condition. This test case now uses `wait_for()` to detect when the DHCP-assigned default route becomes available, eliminating the fixed sleep and reducing test runtime.

## Types of changes

## Related Task(s)

* https://vyos.dev/T7996

Local smoketests runs of `make testc`, `make test-no-interfaces-no-vpp`  and `make test-interfaces` completed successfully

## Checklist:

### DIFF
--- a/smoketest/scripts/cli/base_vyostest_shim.py
+++ b/smoketest/scripts/cli/base_vyostest_shim.py
@@ -140,12 +140,14 @@ class VyOSUnitTestSHIM:
                 pprint.pprint(out)
             return out
 
-        def getFRRconfig(self, start_section:str=None, stop_section='^!',
+        def getFRRconfig(self, start_section:str=None, end_marker='$', stop_section='^!',
                          start_subsection:str=None, stop_subsection='^ exit') -> str:
             """
             Retrieve current "running configuration" from FRR
 
             start_section:    search for a specific start string in the configuration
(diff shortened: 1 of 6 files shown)


sever-sever (approved):
Fix VRF smoketest
Dehardcode VRF
Use `wait_for`

jestabro (approved):
Several subtle fixes for logic of smoketests/warnings: all vrf smoketests now pass in local test.

PR #4866, opened 2025-11-24
T8030: VPP: Check support for changed driver too

natali-rs1985:
## Types of changes

## Related Task(s)

* https://vyos.dev/T8030

NIC e1000
```
commit
```
Before the fix, when we changed driver type there was an error and VPP reloaded to previous configuration because original NIC does not support XDP:
```
vyos@vyos# set vpp settings interface eth1 driver xdp
[edit]
vyos@vyos# commit
[ vpp ]
WARNING: offload option in eth1 settings is not supported by VPP interfaces. It will be ignored.

An error occurred: VPP API call failed: -12. VPP service will be
restarted with the previous configuration
[[vpp]] failed
Commit failed
[ vpp ]
WARNING: offload option in eth1 settings is not supported by VPP interfaces. It will be ignored.

[edit]
```
After the fix:
```
vyos@vyos# set vpp settings interface eth1 driver xdp
[edit]
vyos@vyos# commit
[ vpp ]
WARNING: offload option in eth1 settings is not supported by VPP interfaces. It will be ignored.
Driver xdp is not compatible with interface eth1!
[[vpp]] failed
Commit failed
[edit]
```
## Checklist:

### DIFF
--- a/python/vyos/vpp/config_verify.py
+++ b/python/vyos/vpp/config_verify.py
@@ -22,7 +22,6 @@ from vyos import ConfigError
 from vyos.base import Warning
 from vyos.utils.cpu import get_core_count as total_core_count, get_cpus
 
-from vyos.vpp.control_host import get_eth_driver
 from vyos.vpp.config_resource_checks import cpu as cpu_checks, memory as mem_checks
 from vyos.vpp.config_resource_checks.resource_defaults import default_resource_map
 from vyos.vpp.utils import human_memory_to_bytes, bytes_to_human_memory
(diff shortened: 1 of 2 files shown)


alexk37:
Fix works for me, nic driver gets verified now

sever-sever (approved):
Add verification driver check on the step of the changing driver 
DPDK -> XDP, XDP -> DPDK

zdc (commented):
Can it be a situation when `original_driver` in `persist_config` is unknown at all?

I am recalling that a few months ago, we had a change (or discussion only) to store driver details unconditionally, not only for NICs that need to override a driver, specifically because we need it in multiple places.

If so, then we can drop from the verification `get_eth_driver()`

natali-rs1985:
> Can it be a situation when `original_driver` in `persist_config` is unknown at all?

I haven't seen this situation yet.

> I am recalling that a few months ago, we had a change (or discussion only) to store driver details unconditionally, not only for NICs that need to override a driver, specifically because we need it in multiple places.

We store information about `original_driver` for each interface unconditionally

zdc (approved):
I have not tested it, but the logic and code looks totally OK for me now.

PR #4883, opened 2025-12-01
interfaces: T8054: use --is-valid-intf-address for validating interface addresses

dmbaturin:
## Change summary

Currently interface address command templates use the `ip-host` validator. The problem with that is that not every valid _host address_ is a valid _network interface address_.

For example, 224.0.0.1/24 is a "host address" in the sense that it's not a network address, but it's the multicast group address of all routers and should never be assigned to an interface manually — it must be joined correctly via IGMP to actually work as expected.

The curious part is that we in fact have had `interface-address` validator since the introduction of ipaddrcheck-based validators in  2018, but for some reason never used it. Now it refers to `--is-any-host` but ipaddrcheck already had an option specially for this purpose: `--is-valid-intf-address`.

The function behind it (https://github.com/vyos/ipaddrcheck/blob/current/src/ipaddrcheck_functions.c#L416-L442) excludes all special-purpose addresses.

It disallows IPv4 loopback addresses by default — ideally we should only allow them on `lo`, but we don't have a mechanism to pass parent node names to validators yet, so here it's allowed unconditionally.

If this PR is merged, we can then deal with https://vyos.dev/T7973 (the question about all-zero host part IPv6 addresses) in that function, if we decide to allow them.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8054

## Checklist:

### DIFF
--- a/interface-definitions/include/interface/address-ipv4-ipv6-dhcp.xml.i
+++ b/interface-definitions/include/interface/address-ipv4-ipv6-dhcp.xml.i
@@ -22,7 +22,7 @@
       <description>Dynamic Host Configuration Protocol for IPv6</description>
     </valueHelp>
     <constraint>
-      <validator name="ip-host"/>
+      <validator name="interface-address"/>
       <regex>(dhcp|dhcpv6)</regex>
     </constraint>
     <multi/>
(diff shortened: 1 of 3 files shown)


c-po (approved):
Add new IP address validator for interface based IP addresses. Smoketests confirmed it does not blow up.

sever-sever (approved):
Add the correct validator `interface-address` and use it in the `interface x address` instead of `ip-host`

PR #4892, opened 2025-12-10
login: T8086: replace getpwall() based user enumeration to avoid NSS/TACACS timeouts

c-po:
## Change summary

The previous implementation of `system login` relied on Python's `pwd.getpwall()` to enumerate user accounts. This forces a full walk through the NSS stack, which is acceptable in general but problematic for our use-case. VyOS only needs information about locally created accounts and not remote accounts provided via AAA backends such as TACACS or RADIUS.

When TACACS servers are unreachable, NSS lookups become extremely slow due to repeated timeouts. As a result, any operation triggering `pwd.getpwall()` (including configuration commits) can stall for several minutes.

This change introduces a dedicated helper, `get_local_passwd_entries()`, which reads `/etc/passwd` directly and avoids NSS entirely. Since only local UIDs are relevant, this provides all required data with no external dependencies.

Performance improvement on VyOS 1.4.3 with two unreachable TACACS servers:

```
set system login tacacs server 192.168.1.50 key test123
set system login tacacs server 192.168.1.51 key test123
time commit
```

### Before

```
real    3m29.825s
user    0m0.329s
sys     0m0.246s
```

### After

```
real    0m1.464s
user    0m0.337s
sys     0m0.195s
```

This significantly improves commit performance and removes sensitivity to AAA server outages.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8086

All smoketests passed:

* `make test-interfaces`
* `make test-no-interfaces-no-vpp`
* `make testc`

## Checklist:

### DIFF
--- a/python/vyos/utils/auth.py
+++ b/python/vyos/utils/auth.py
@@ -18,10 +18,13 @@ import math
 import re
 import string
 
-from enum import StrEnum
+from dataclasses import dataclass
 from decimal import Decimal
-from pwd import getpwall
 from pwd import getpwnam
+from enum import StrEnum
+from typing import List
+from typing import Optional
+
 from vyos.utils.process import cmd
 
 # Minimum UID used when adding system users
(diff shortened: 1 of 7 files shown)


sarthurdev (approved):
Migrates away from `pwd` module, uses a vyos.utils helper for reading /etc/passwd.

Smoketests pass, not tested locally.

jestabro (changes requested):
As a replacement for use of getpwall() this works in a basic test; not tested at scale for performance comparison. Note that the op_mode script does have a bug, as pointed out by the copilot review on non-iterability of the dataclass object:
```
vyos@vyos:~$ /usr/libexec/vyos/op_mode/show_users.py
  File "/usr/libexec/vyos/op_mode/show_users.py", line 114, in <module>
    main()
  File "/usr/libexec/vyos/op_mode/show_users.py", line 107, in main
    for user in list_users():
                ^^^^^^^^^^^^
  File "/usr/libexec/vyos/op_mode/show_users.py", line 82, in list_users
    for (name, _, uid, _, _, _, _) in get_local_passwd_entries():
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: cannot unpack non-iterable PasswdEntry object
```
This is not called by the op-mode command 'show users', but is by 'show system login users'.

c-po:
This fixes the mentioned bug from Copilot and also mentioned by @jestabro 
```
cpo@LR1.wue3:~$ show system login users
Username    Type    Locked    Tty    From           Last login
----------  ------  --------  -----  -------------  ------------------------
cpo         vyos    False     pts/0  172.16.33.123  Wed Dec 17 21:18:28 2025
```

jestabro (approved):
The issue mentioned above is resolved. All looks good, and was previously tested.

PR #4899, opened 2025-12-14
T8026: Fixed session commit for the generate router

xTITUSMAXIMUSX:
## Change summary

Fix session commit in `/generate` route and ensuring generated router data is applied as expected.

## Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Code style update (formatting, renaming)
* [ ] Refactoring (no functional changes)
* [ ] Migration from an old Vyatta component to vyos-1x, please link to related PR inside obsoleted component
* [ ] Other (please describe):

## Related Task(s)

https://vyos.dev/T8026

## Related PR(s)

N/A

### Test environment

* VyOS current
* Local development environment

### Testing performed

1. Triggered the `/generate` route before the change and observed that generated router data was not persisted.
2. Added `session.commit()` in `routers.py` under the `/generate` route.
3. Triggered the `/generate` route again and verified that the generated router data is now correctly committed and persists as expected.

No regressions observed.

## Checklist:

* [x] I have read the [**[CONTRIBUTING](https://github.com/vyos/vyos-1x/blob/current/CONTRIBUTING.md)**](https://github.com/vyos/vyos-1x/blob/current/CONTRIBUTING.md) document
* [x] I have linked this PR to one or more Phabricator Task(s)
* [x] I have run the components [**[SMOKETESTS](https://github.com/vyos/vyos-1x/tree/current/smoketest/scripts/cli)**](https://github.com/vyos/vyos-1x/tree/current/smoketest/scripts/cli) if applicable
* [x] My commit headlines contain a valid Task id
* [ ] My change requires a change to the documentation
* [ ] I have updated the documentation accordingly

PR #4901, opened 2025-12-15
T8103: add root to those allowed to call op-run commands directly

jestabro:
## Change summary

When op-mode commands are called via http-api they do not register as being called by admin in vyatta-op-run; consequently, certain checks fail, breaking, for example, the `install_into_config` functionality of the `/generate` endpoint.

Fix the check for admin to include those having euid == 0, as is the case for the vyos-http-api service.

The related, lower priority, question of appropriate env vars to test `in_session` for non-privileged calls to vyos-op-run will be handled separately. The current fix resolves the regression of T8103.

Note that this is related to, but independent of, https://github.com/vyos/vyos-1x/pull/4899

## Types of changes

Before:

```
└──> curl -k -X POST -d '{"key": "baz", "op": "generate", "path": ["pki", "wireguard", "key-pair", "install", "interface", "wg0"]}' https://192.168.122.31/generate 
{"success": true, "data": "You are not in configure mode, commands to install manually from configure mode:\nset interfaces wireguard wg0 private-key 'UP3TXQCoexDTDnbmuNdjzR/y1HjQUFAnJs9tyVpzh1k='\nCorresponding public-key to use on peer system is: '8ER4u63NkO489Rcm9zY2BOs6bkADFnVjL2G8bdN3LDM='\n", "error": null}
```

After:

```
└──> curl -k -X POST -d '{"key": "baz", "op": "generate", "path": ["pki", "wireguard", "key-pair", "install", "interface", "wg0"]}' https://192.168.122.146/generate
{"success": true, "data": "1 value(s) installed. Use \"compare\" to see the pending changes, and \"commit\" to apply.\nCorresponding public-key to use on peer system is: 'phce83BicbcwVFMWn0xyrNA0X6FADQwj18EkZmMEJV8='\n", "error": null}
```

## Checklist:

### DIFF
--- a/src/opt/vyatta/share/vyatta-op/functions/interpreter/vyatta-op-run
+++ b/src/opt/vyatta/share/vyatta-op/functions/interpreter/vyatta-op-run
@@ -218,7 +218,7 @@ _vyatta_op_run ()
     run_cmd=$(_vyatta_op_conv_run_cmd "$run_cmd") # convert the positional parameters
     local ret=0
 
-    if groups "$(whoami)" | grep -q -E ' vyattacfg(\s|$)'; then
+    if [[ $EUID -eq 0 ]] || groups "$(whoami)" | grep -q -E ' vyattacfg(\s|$)'; then
       # If the user is an admin,
       # use sudo directly for now
       op_runner="sudo"


dmbaturin (approved):
Certainly a missed case from my side, no objections here.

c-po (approved):
As root is - with the exception to capabilities/SELinux - allowed to do everything I see no harm in adding `root` to the allowed user list.

PR #4906, opened 2025-12-18
T8105: Modify workflows failed due to pullrequest_target change

asklymenko:
## Change summary

The pull_request_target workflows are now evaluated against the base repository’s default branch for branch protections and environment rules, not the pull request’s target branch.

It is required to change the structure to have only 1 common file for all branches. This workflow should call a branch-specific workflow to run tests. In these changes I use a local workflow of a branch (the branch is not specified in the workflow path). According to the GitHub documentation it should call a workflow of the same branch, if a local workflow specified. Let's try this approach to fix current problem.

## Types of changes

## Related Task(s)

* https://vyos.dev/TT8105

## Related PR(s)

* https://github.com/vyos/.github/pull/107

## Checklist:


asklymenko:
@kumvijaya Could you please review this approach?

kumvijaya (approved):
Can approve as this needs quick-fix.
But reusable workflow to be kept in .github repo (not on same repo)

asklymenko:
@kumvijaya 
Thank you. If it is better to keep reusable workflows in .github repo I will move it to that repo later.

c-po (commented):
The package smoketest workflow is ONLY required in vyos-1x repo - is it really necessary to make it an overcomplicated re-usable workflow?

kumvijaya:
> The package smoketest workflow is ONLY required in vyos-1x repo - is it really necessary to make it an overcomplicated re-usable workflow?

Agree. If it is used only in this repo, can avoid creating reusable workflow.

asklymenko:
@c-po @kumvijaya I think we should have the same logic for other branches, in this case it is ok to move that file to other repo. To have the same location for all repos.

c-po:
> @c-po @kumvijaya I think we should have the same logic for other branches, in this case it is ok to move that file to other repo. To have the same location for all repos.

I strongly disagree as the logic becomes overly complex for no reason.

asklymenko:
I am closing this PR, this issue will be resolved by [another PR](https://github.com/vyos/vyos-1x/pull/4914).

PR #4910, opened 2025-12-19
T8010: Added the vyos_vpp plugin to the accel-pppd's template

AndriiFullroot:
## Change summary

Added the vyos_vpp plugin for accel-ppp-ng.

## Types of changes

## Related Task(s)

https://vyos.dev/T8010

## Checklist:

### DIFF
--- a/data/templates/accel-ppp/pppoe.config.j2
+++ b/data/templates/accel-ppp/pppoe.config.j2
@@ -20,6 +20,7 @@ connlimit
 sigchld
 pppd_compat
 {% endif %}
+vyos_vpp
 
 [core]
 thread-count={{ thread_count }}


AndriiFullroot:

sever-sever (approved):
Add `vyos_vpp` plugin for the PPPoE server
It depends on https://github.com/accel-ppp/accel-ppp-ng/pull/19

data/templates/accel-ppp/pppoe.config.j2
```diff
@@ -20,6 +20,7 @@ connlimit
 sigchld
 pppd_compat
 {% endif %}
+vyos_vpp
```
c-po, line 23: I do wonder if it is a good idea to add this plugin by default. I would make this dynamic based on the fact if VPP service is enabled, or not on the CLI.

AndriiFullroot: It was my initial idea, the issue is that the option "vpp-cp" is "per-interface" and there is no option to check if vpp-cp is set without iterating through.
Vyacheslav and I discussed that, and we agreed to enable the plugin by default for now.

dmbaturin: Loading the plugin has no implications or negative effects, so I'd go with this now.

dmbaturin (approved):
Loading the plugin has no implications or negative effects, so I'd go with this now.

PR #4916, opened 2025-12-25
T7832: Add support for assigning ACLs on subinterfaces

inetman28:
## Change summary
Support assigning ACLs to subinterfaces (e.g., eth0.100, eth5.200)

## Types of changes

## Related Task(s)

- https://vyos.dev/T7832

```
set interfaces ethernet eth2 vif 1000 address '172.16.0.251/24'
set vpp settings interface eth2 driver 'dpdk'
set vpp settings interface eth2 rx-mode 'polling'
set vpp acl ip tag-name drop-all rule 1000 action 'deny'
set vpp acl ip interface eth2.1000 input acl-tag 100 tag-name 'drop-all'

vyos# commit

vyos@vyos# run show vpp acl ip interface 
Interface    Input ACLs    Output ACLs
-----------  ------------  -------------
eth2.1000     drop-all 
[edit]
vyos@vyos#
```

## Checklist:

### DIFF
--- a/src/conf_mode/vpp_acl.py
+++ b/src/conf_mode/vpp_acl.py
@@ -214,9 +214,17 @@ def verify(config):
 
             for iface, iface_config in acl.get('interface', {}).items():
                 if iface not in config.get('vpp_ifaces'):
-                    raise ConfigError(
-                        f'{iface} must be a VPP interface for ACL interface'
-                    )
+                    # Check if it is a sub-interface
+                    if '.' in iface:
+                        parent = iface.split('.')[0]
+                        if parent not in config.get('vpp_ifaces'):
+                            raise ConfigError(
+                                f'{iface} must be a VPP interface for ACL interface'
+                            )
+                    else:
+                        raise ConfigError(
+                            f'{iface} must be a VPP interface for ACL interface'
+                        )
 
     if 'ip' in config:
         acl = config.get('ip')


inetman28:

zdc (changes requested):
Thank you for the contribution! This feature expansion would indeed be very useful, but it requires a bit more code to function properly.

In this specific case, simply relaxing the protection is not sufficient.

There is a critical distinction between Ethernet and all other interfaces in this context. Ethernet has two key characteristics:
- It cannot be removed from the system while it is running (with rare exceptions, which we can disregard here).
- VPP configures it at a very early stage, ensuring that there is never a situation where an Ethernet interface is configured in VPP but not properly connected to it.

This does not apply to other interfaces. For example:
1. Configure your example configuration.
2. Remove a VLAN from the Ethernet configuration.
3. Re-add the same VLAN.

The configuration may appear valid, but the ACL will not be applied to the interface anymore.

To ensure this works correctly, you must also establish the proper connection between VLANs and VPP ACLs.

natali-rs1985 (changes requested):
We have a function that rertieves interfaces along with their sub-interfaces:
https://github.com/vyos/vyos-1x/blob/current/python/vyos/vpp/utils.py#L104
You can use the `cli_ethernet_with_vifs_ifaces()` function instead of `cli_ifaces_list()` when you specifically need all available sub-interfaces (such as for assigning ACLs)

sever-sever:
There has been no activity for several months, and there have been many conflicts.
I closed the PR

PR #4920, opened 2025-12-26
T7098: Fix VPP MTU misconfiguration

natali-rs1985:
## Types of changes

## Related Task(s)

* https://vyos.dev/T7098

```
commit
```
Before the fix
```
vyos@vyos# sudo vppctl show interface
              Name               Idx    State  MTU (L3/IP4/IP6/MPLS)     Counter          Count
eth1                              1      up          1500/0/0/0     tx-error                       2
local0                            0     down          0/0/0/0       drops                          1
tap4096                           2      up          9000/0/0/0     rx packets                     1
                                                                    rx bytes                      90
                                                                    drops                          1
                                                                    ip6                            1
```
After the fix
```
vyos@vyos# sudo vppctl show interface
              Name               Idx    State  MTU (L3/IP4/IP6/MPLS)     Counter          Count
eth1                              1      up          1500/0/0/0     tx-error                       2
local0                            0     down          0/0/0/0       drops                          1
tap4096                           2      up          1500/0/0/0     rx packets                     1
                                                                    rx bytes                      90
                                                                    drops                          1
                                                                    ip6                            1
```

```
vyos@vyos# /usr/libexec/vyos/tests/smoke/cli/test_vpp.py -k test_01_vpp_basic
test_01_vpp_basic (__main__.TestVPP.test_01_vpp_basic) ... ok

Ran 1 test in 41.221s

OK
```
## Checklist:

### DIFF
--- a/src/conf_mode/interfaces_ethernet.py
+++ b/src/conf_mode/interfaces_ethernet.py
@@ -46,6 +46,7 @@ from vyos.utils.dict import dict_to_paths_values
 from vyos.utils.dict import dict_set
 from vyos.utils.dict import dict_delete
 from vyos.utils.process import is_systemd_service_running
+from vyos.vpp.control_vpp import VPPControl
 from vyos import ConfigError
 from vyos import airbag
 airbag.enable()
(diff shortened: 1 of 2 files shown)


src/conf_mode/interfaces_ethernet.py
```diff
@@ -428,13 +429,34 @@ def generate(ethernet):
...
+    # If the interface is managed by the VPP DPDK driver, synchronize runtime
+    # parameters between Linux and the corresponding VPP LCP interface
+    if (
+        ethernet.get('vpp', {})
+        .get('settings', {})
+        .get('interface', {})
+        .get(ifname, {})
+        .get('driver')
+        == 'dpdk'
+    ):
```
c-po: This reads very hard - please change it to:


```suggestion
    if dict_search('vpp.settings.interface.{ifname}.driver', ethernet) == dpdk:
```

src/conf_mode/interfaces_ethernet.py
```diff
@@ -428,13 +429,34 @@ def generate(ethernet):
...
+        vpp_api = VPPControl()
```
c-po, line 452: We have always kept VPP outside of `interfaces ethernet` - if we now include VPP stuff here, too - why not combine the entire VPP stuff and the Kernel part under one unified CLI tree?

Can this code path be made conditional to be NOT executed if VPP is not configured?

c-po (changes requested):

src/conf_mode/interfaces_ethernet.py
natali-rs1985: Thank you for the suggestion. I've made the change.

src/conf_mode/interfaces_ethernet.py
natali-rs1985: The code path is already conditional: it will only be executed if VPP is configured for the interface and the driver is set to DPDK. If VPP is not configured, this part will not execute.

c-po (approved):
After discussing the upcoming plans in the maintainers meeting I am happy to see this change coming.

dmbaturin (approved):
I see no issues with the logic and it fixes a real issue.

PR #4921, opened 2025-12-30
T7664: support FRR 10.5

hedrok:
## Support FRR 10.5

Support for the following changes:

* Removal of 'connected', 'kernel' and 'table' protocols in 'ip protocol'. Added migration script that just removes commands with these protocols (added config test). Updated smoke tests.
* In segment-routing FRR now doesn't output `block-len`, `node-len` and `func-bits` if they have default value. Updated smoke test accordingly.
* FRR 10.5 changed default value for `bgp reject-as-sets` to on by default. I haven't changed default, just added `no bgp reject-as-sets` in FRR configuration template.

## Types of changes

- [X] Other (please describe): Support FRR 10.5

## Related Task(s)

* https://vyos.dev/T7664

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1090

With vyos-build PR and this PR all smoke tests (VPP not tested) pass.

## Checklist:

- [X] My change requires a change to the documentation: https://docs.vyos.io/en/latest/configuration/system/ip.html need to update protocol list
- [X] I have updated the documentation accordingly: https://github.com/vyos/vyos-documentation/pull/1730


c-po (changes requested):
Can you please create or extend `configtest` from VyOS 1.4 which makes use of the now no longer supported `redistribute` statements for both the default and any other VRF? This way we have a build in test that the CLI entries in question will be migrated away.

See for details https://github.com/vyos/vyos-1x/blob/current/smoketest/README.md

hedrok:
@c-po Thanks, I've added config test, and it showed one mistake (I've fixed that too).

smoketest/configs/basic-system-ip-protocol
```diff
@@ -0,0 +1,436 @@
...
+            protocol table {
+                route-map stub
+            }
+        }
+        table 1234
+    }
+}
+/* Warning: Do not remove the following line. */
+/* === vyatta-config-version: "broadcast-relay@1:cluster@1:config-management@1:conntrack-sync@1:conntrack@1:dhcp-relay@2:dhcp-server@5:dns-forwarding@1:firewall@5:ipsec@5:l2tp@1:mdns@1:nat@4:ntp@1:pptp@1:qos@1:quagga@6:snmp@1:ssh@1:system@27:vrf@3:vrrp@2:wanloadbalance@3:webgui@1:webproxy@1:webproxy@2:zone-policy@1" === */
+/* Release version: 1.2.6 */
```
c-po: This version line is wrong as VRF support for any protocol was only intruduced recently in VyOS 1.4

c-po (changes requested):

smoketest/configs/basic-system-ip-protocol
hedrok: Thank you, I've updated it.
I've edited previous file (with '1.2.6' version) manually to get result I needed - I've just added sections with protocols and added versions of system and vrf.
The updated file is file saved by VyOS 1.4.3 without modifications.
To create it I've:
1. Loaded `basic-vyos` configuration on VyOS 1.4.3
2. Ran commands:
```
set vrf name red table 1234
set policy route-map stub
set system ip protocol any       route-map stub
set system ip protocol babel     route-map stub
set system ip protocol bgp       route-map stub
set system ip protocol connected route-map stub
...
set system ip protocol table     route-map stub
set system ipv6 ...
set vrf name red ip protocol ...
set vrf name red ipv6 protocol ...
```
3. Saved configuration file.

smoketest/configs/basic-system-ip-protocol
```diff
@@ -0,0 +1,388 @@
...
+                port "8882"
+            }
+        }
+        service my-tcp-api {
+            backend "bk-01"
+            mode "tcp"
+            port "8888"
+        }
+    }
+}
```
c-po: I see no benefit in a load-balancing testcase here - we should declutter it
```suggestion
```

smoketest/configs/basic-system-ip-protocol
```diff
@@ -0,0 +1,388 @@
...
+        ciphers "chacha20-poly1305@openssh.com"
+        ciphers "rijndael-cbc@lysator.liu.se"
+        key-exchange "curve25519-sha256@libssh.org"
+        key-exchange "diffie-hellman-group1-sha1"
+        key-exchange "diffie-hellman-group-exchange-sha1"
+        key-exchange "diffie-hellman-group-exchange-sha256"
+        listen-address "192.168.0.1"
+        port "22"
+    }
+}
```
c-po: I see no benefit in testing with those services enabled. Sometimes it is also a good idea to test with less config elements and remove all services.

```suggestion
```

c-po (changes requested):

hedrok:
Thanks for decluttering, as I've described, I just took `basic-vyos`, added needed configuration and changed nothing else as I'm not sure which parts of `basic-vyos` are needed.

c-po (approved):
Manually tested multiple times. All smoketests now pass as expected.
Requires FRR patches which are currently pending upstream.

sever-sever (approved):
Remove, deleted by upstream  zebra`ip protocols connected|kernel|table` options
Add migration scripts

PR #4934, opened 2026-01-03
T8145: LUKS Encryption Passphrase Observable

giga1699:
## Change summary

Do not display the LUKS recovery key while typing, as it is a secret value.

## Types of changes

## Related Task(s)

https://vyos.dev/T8145

## Related PR(s)

https://github.com/vyos/vyos-1x/pull/4935

Try to encrypt without a TPM, and make sure the passphrase doesn't show up on the terminal while typing.

## Checklist:

### DIFF
--- a/src/helpers/vyos-config-encrypt.py
+++ b/src/helpers/vyos-config-encrypt.py
@@ -299,7 +299,7 @@ if __name__ == '__main__':
 
     if need_recovery or (args.enable and not ask_yes_no(f'Automatically generate a {question_key_str}?', default=True)):
         while True:
-            recovery_key = ask_input(f'Enter {question_key_str}:', default=None).encode()
+            recovery_key = ask_input(f'Enter {question_key_str}:', default=None, no_echo=True).encode()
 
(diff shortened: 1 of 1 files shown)


sarthurdev (approved):
Small change to hide key output for TPM unlock.

I think we could further improve this with one extra step:
If we extend `ask_input` to also make available getpass `echo_char` argument. Setting to '*' and you would then get feedback that the input is registered and shows length of entered string.

giga1699:
Just looked at that, getpass echo_char isn't included in python 3.11 which is on VyOS 1.4 and 1.5. It wasn't added until python 3.14

https://docs.python.org/3/library/getpass.html

> Changed in version 3.14: Added the echo_char parameter for keyboard feedback.

c-po:
> If we extend ask_input to also make available getpass echo_char argument. Setting to '*' and you would then get feedback that the input is registered and shows length of entered string.

Nice idea!

c-po (approved):
Hide passphrase input

PR #4935, opened 2026-01-03
T8146: Confirm the key when config encryption is configured without TPM

giga1699:
## Change summary

Force the user to confirm that they have the correct encryption key when not using a TPM. Help mitigate user mistyping key a single time, and getting locked out of their config.

## Types of changes

## Related Task(s)

https://vyos.dev/T8146

## Related PR(s)

https://github.com/vyos/vyos-1x/pull/4934

Try to use encrypt enable without a TPM present. Should ask you to confirm key.

## Checklist:

### DIFF
--- a/src/helpers/vyos-config-encrypt.py
+++ b/src/helpers/vyos-config-encrypt.py
@@ -326,6 +326,9 @@ if __name__ == '__main__':
             print('Backup the recovery key in a safe place!')
             print('Recovery key: ' + recovery_key.decode())
         elif args.enable:
+            if recovery_key != ask_input('Confirm key:', default=None, no_echo=True).encode():
+                raise ValueError("Keys did not match!")
+
             encrypt_config(recovery_key, is_tpm=False)
 
             print('Encrypted config volume has been enabled without TPM')


sarthurdev:
Feels a bit redundant given we echo the recovery key in the same code block?

https://github.com/vyos/vyos-1x/blob/current/src/helpers/vyos-config-encrypt.py#L333

giga1699:
I guess I was just looking at it from the perspective of someone typing something in wrong a single time since this only applies if the user is creating the only usable decryption key. I agree it's less relevant when you're mainly relying on the TPM to handle the keys for you. Yes, you *may* see it in the echo after encryption is enabled, but you may look past the small error or misread the output thinking that's what you intended to type. Actually typing it a second time might pickup a misread or mistype. Surely if someone is just pasting from a vault or something, it's a minor problem.

Confirmation input is pretty standard for initial creation of a password/secret if you have to type it manually. Just figured it was a simple thing to help mitigate a potential loss of configs if someone isn't really paying enough attention during the creation.

sarthurdev (approved):
Minor change. Adds extra confirm step to enabling encrypted config without TPM.

dmbaturin (approved):
I think the check is useful.

PR #4936, opened 2026-01-06
T8153: PPPoE IPv6 autoconf: no Router Solicitation sent and default IPv6 route not installed after successful dial-up

jd82k:
## Change summary

When PPPoE comes up with IPv6 SLAAC (ipv6 address autoconf), VyOS may not send Router Solicitation and therefore fails to learn/install the IPv6 default route.

Trigger rdisc6 after PPPoE is up when 'ipv6.address autoconf' is configured. This makes the IPv6 default route appear immediately without manual intervention.

Workaround of linking ip-up.d scripts into ipv6-up.d is also used to ensure the IPv6 PPP hook scripts are executed consistently.
## Types of changes

## Related Task(s)

https://vyos.dev/T8153
## Checklist:

### DIFF
--- a/python/vyos/ifconfig/pppoe.py
+++ b/python/vyos/ifconfig/pppoe.py
@@ -15,6 +15,7 @@
 
 from vyos.ifconfig.interface import Interface
 from vyos.utils.assertion import assert_range
+from vyos.utils.dict import dict_search
 from vyos.utils.network import get_interface_config
 
 @Interface.register
@@ -139,3 +140,7 @@ class PPPoEIf(Interface):
             self._cmd(f'vtysh -c "conf t" {vrf} -c "ip route 0.0.0.0/0 {self.ifname} tag 210 {distance}"')
             if 'ipv6' in config:
                 self._cmd(f'vtysh -c "conf t" {vrf} -c "ipv6 route ::/0 {self.ifname} tag 210 {distance}"')
+
+        # kick RS when IPv6 is up.
+        if 'autoconf' in dict_search('ipv6.address', config):
+            self._cmd(f'rdisc6 --single --retry 3 {self.ifname}')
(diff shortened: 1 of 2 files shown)


jd82k:

jd82k:
I can't use this `'autoconf' in dict_search('ipv6.address', config):` because there is no dict_search in the context.

c-po:
> I can't use this `'autoconf' in dict_search('ipv6.address', config):` because there is no dict_search in the context.

`from vyos.utils.dict import dict_search`

jd82k:
> > I can't use this `'autoconf' in dict_search('ipv6.address', config):` because there is no dict_search in the context.
> 
> `from vyos.utils.dict import dict_search`

Ok

jd82k:
> > I can't use this `'autoconf' in dict_search('ipv6.address', config):` because there is no dict_search in the context.
> 
> `from vyos.utils.dict import dict_search`

I changed the code and now it should work.

c-po (approved):
I changed the `rdisc6` command-line to use full length arguments which makes it easier for maintainers to understand what could happen.

I also dropped `--quiet` as there is no need for it as the output is not redirected to log/stdout. I also dropped `--wait 1000` which is the deafult.

This is an improvement in terms of IPv6 speedup when retrieving an address. Adding backports.

jd82k:
> I changed the `rdisc6` command-line to use foll length arguments which makes it easier for maintainers to understand what could happen.
> 
> I also dropped `--quiet` as there is no need for it as the output is not redirected to log/stdout. I also dropped `--wait 1000` which is the deafult.
> 
> This is an improvement in terms of IPv6 speedup when retrieving an address. Adding backports.

Thanks for your job!

sever-sever (approved):
Call rdisc6 after PPPoE is up when 'ipv6.address autoconf' is configured.

dmbaturin (approved):
I think we may want to make the number of retries configurable at some point, but I have no objections against merging the PR as it is.

PR #4950, opened 2026-01-21
T8196: Add adjust-mss forwarded from interface

Firefishy:
## Change summary
Also adjust-mss forwarded from an interface.

## Types of changes

## Related Task(s)
https://vyos.dev/T8196

## Related PR(s)
N/A

Updated existing smoketests.

## Checklist:

### DIFF
--- a/data/vyos-firewall-init.conf
+++ b/data/vyos-firewall-init.conf
@@ -20,6 +20,10 @@ table raw {
         type filter hook postrouting priority -300; policy accept;
     }
 
+    chain VYOS_TCP_MSS_FWD {
+        type filter hook forward priority -300; policy accept;
+    }
+
     chain vyos_global_rpfilter {
         return
     }
@@ -38,6 +42,10 @@ table raw {
 table ip6 raw {
     chain VYOS_TCP_MSS {
         type filter hook postrouting priority -300; policy accept;
+    }
+
+    chain VYOS_TCP_MSS_FWD {
+        type filter hook forward priority -300; policy accept;
     }
 
     chain vyos_global_rpfilter {
--- a/python/vyos/ifconfig/interface.py
+++ b/python/vyos/ifconfig/interface.py
@@ -692,13 +692,24 @@ class Interface(Control):
         return self.set_interface('ipv6_cache_tmo', tmo)
 
     def _cleanup_mss_rules(self, table, ifname):
-        commands = []
         results = self._cmd(f'nft -a list chain {table} VYOS_TCP_MSS').split("\n")
         for line in results:
             if f'oifname "{ifname}"' in line:
                 handle_search = re.search('handle (\d+)', line)
                 if handle_search:
-                    self._cmd(f'nft delete rule {table} VYOS_TCP_MSS handle {handle_search[1]}')
+                    self._cmd(
+                        f'nft delete rule {table} VYOS_TCP_MSS handle {handle_search[1]}'
+                    )
+
+    def _cleanup_mss_fwd_rules(self, table, ifname):
+        results = self._cmd(f'nft -a list chain {table} VYOS_TCP_MSS_FWD').split("\n")
+        for line in results:
+            if f'iifname "{ifname}"' in line:
+                handle_search = re.search('handle (\d+)', line)
+                if handle_search:
+                    self._cmd(
+                        f'nft delete rule {table} VYOS_TCP_MSS_FWD handle {handle_search[1]}'
+                    )
 
     def set_tcp_ipv4_mss(self, mss):
         """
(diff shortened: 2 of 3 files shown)


l0crian1:
This should not be implemented in this way. MSS can be different in each direction, and this eliminates that possibility and is not in line with the implementation within industry.

Typically, each side will clamp the MSS if needed from their direction and would not be done on a single device. If it is desired to control it on a single device, an additional command to account for that should be added, not having the single command do both directions. Or the MSS can be clamped using a `policy` rule.

Firefishy:
> This should not be implemented in this way. MSS can be different in each direction, and this eliminates that possibility and is not in line with the implementation within industry.
> 

I think you have misunderstood the issue. 
This is widely used. I confirmed the implementation with what OpenWRT does. 

It is also recommended per the linked nftables document. 

I really need to include a proper diagram to explain the issue, currently vyos implementation will not clamp mss for subnet behind a tunnel where output subnet has "normal" MTU, but the input tunnel has restricted MTU.

> Typically, each side will clamp the MSS if needed from their direction and would not be done on a single device. If it is desired to control it on a single device, an additional command to account for that should be added, not having the single command do both directions. Or the MSS can be clamped using a `policy` rule.

Typically ICMP should just work and MSS fiddling shouldn't but necessary, but the internet is imperfect and hacks like mudging MSS are needed.

l0crian1:
> I think you have misunderstood the issue.
> This is widely used. I confirmed the implementation with what OpenWRT does.

I'm pretty sure I understand the issue, this doesn't seem to be an obscure edge case. One thing to note is that `prerouting` and `postrouting` already apply to forwarded traffic. You don't need a forward hook for the MSS clamping. You can see in this diagram that packets that are forwarded will already traverse those hooks.

image

I have no issue with creating a mechanism to control to announcements in both directions, but it shouldn't set it for both directions when simply doing `adjust-mss`, this is not widely used and would be unnecessarily limiting (users who need asymmetric MSS values won't be able to). Control for both directions is more typically controlled via separate touch points, this is done on the majority of NOS' (even if OpenWRT differs; I haven't checked the rules their FW creates). The nftables note doesn't state to match the clamped value to the SAME value and even calls out that the values may differ in each direction. It is simply pointed out the common misconception that MSS is negotiated when it is in fact a unidirectional announcement, and as such, the clamping would need to be applied in both directions if you desire each direction to be clamped to that value.

Anybody who has ever used a GRE based DDoS service will tell you the issues with how you're trying to do this. Or a lot of other use cases where traffic is intentionally asymmetric across paths of different technologies.

IMO, if control for each direction is desired, then something like this would make more sense:
```
set interfaces ethernet <interface> ip adjust-mss-ingress <mss>
set interfaces ethernet <interface> ip adjust-mss-egress <mss>
set interfaces ethernet <interface> ip adjust-mss-both <mss>
```
But realistically, this can already be done. For instance, if I want traffic that is forwarded from eth1 to eth2 to be clamped to 1300, I can already do this:
```
set interfaces ethernet eth1 ip adjust-mss 1300
set interfaces ethernet eth2 ip adjust-mss 1300
```
That should set the MSS announcement for both directions of the socket.

It's possible that you're trying to solve some other problem and I just don't get it, but after looking at the code it doesn't seem to be.

Perhaps it'd help if you do some configs with their nftables output as examples in the `How to test` section.

Firefishy:
Thanks. I will create a proper test case and validate... I would be happy to be at fault here or learn if I have made some mistake.

I can understand the desire for MSS to differ in direction, but I have not seen such a connection in practice... but equally I don't have much beyond basic tunnel experience.

sarthurdev:
I've had a quick look now and I haven't seen any issues in the past being insufficient applied on the one direction. 

I think that @l0crian1 might have it right here though. If this is an edge case, using PBR to adjust MSS with more granular control might be the better solution.

zdc (commented):
I agree that we should not allow changing the MSS in both directions with the `set interfaces ethernet <interface> ip adjust-mss` command, as this would introduce a breaking change for existing users.

Additionally, implementing `clamp-mss-to-pmtu` is not feasible with the suggested approach, because the outgoing interface is not resolved when the rule is applied.

A solution to this problem already exists through policy-based routing (PBR) using the command `set policy route <name> rule <num> set tcp-mss`.

In my view, while the original reason for creating the PR remains valid, the proposed solution is overly complex. In practice, VyOS users handle this situation using PBR rules.

jd82k:
This PR is a better choice. https://github.com/vyos/vyos-1x/pull/4973

Firefishy:
> This PR is a better choice. #4973

Awesome! Thank you. I am going to close this PR in favour of your PR. I have unfortunately not had sufficient time recently to work on the issue.

jd82k:
> > This PR is a better choice. #4973
> 
> Awesome! Thank you. I am going to close this PR in favour of your PR. I have unfortunately not had sufficient time recently to work on the issue.

I’m sorry my PR was closed. Vyos team think there is no real use case for this PR.

PR #4952, opened 2026-01-23
vpp: T8182: Disallow configuration of IPoE server on VPP interface

natali-rs1985:
## Types of changes

## Related Task(s)

* https://vyos.dev/T8182

```
vyos@vyos# show vpp
 settings {
     interface eth1 {
         driver dpdk
     }
 }
[edit]
vyos@vyos#


set service ipoe-server client-ip-pool POOL range '192.168.2.2-192.168.2.100'
set service ipoe-server default-pool 'POOL'
set service ipoe-server gateway-address '192.168.1.1/24'
set service ipoe-server name-server '1.1.1.1'
set service ipoe-server name-server '198.168.2.1'
set service ipoe-server interface eth1

vyos@vyos# commit
[ service ipoe-server ]
eth1 is a VPP interface and cannot be used for IPoE!
[[service ipoe-server]] failed
Commit failed
[edit]
```

## Checklist:

### DIFF
--- a/src/conf_mode/vpp.py
+++ b/src/conf_mode/vpp.py
@@ -429,6 +429,13 @@ def get_config(config=None):
                     iface_filter_eth(conf, iface)
                 set_dependents(dependency, conf, iface)
 
+    config['ipoe_conf'] = conf.get_config_dict(
+        ['service', 'ipoe-server'],
+        key_mangling=('-', '_'),
+        get_first_key=True,
+        no_tag_node_value_mangle=True,
+    )
+
     # kernel-interfaces dependency
     if effective_config.get('kernel_interfaces'):
         for iface in config.get('kernel_interfaces', {}):
@@ -480,6 +487,20 @@ def verify(config):
 
     if 'interface' not in config['settings']:
         raise ConfigError('"settings interface" is required but not set!')
+
+    ipoe_ifaces = list(
+        {
+            iface
+            for iface in config['settings']['interface']
+            for ipoe_iface in config.get('ipoe_conf', {}).get('interface', {})
+            if iface == ipoe_iface or ipoe_iface.startswith(f'{iface}.')
+        }
+    )
+    if ipoe_ifaces:
+        raise ConfigError(
+            f'Interface(s) {", ".join(ipoe_ifaces)} cannot be added to VPP because '
+            'IPoE is already configured. An interface cannot be used by both VPP and IPoE!'
+        )
 
     # check if the system meets minimal requirements
     verify_vpp_minimum_memory()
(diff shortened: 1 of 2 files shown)


sever-sever (approved):
Disallow to configure VPP with IPoE interfaces as they are not compatible 
I have not tested, but the logic seems straightforward

alexk37:
It was difficult to read two different messages displayed as one: the warning and the commit failed error.

```
vyos@vyos# set vpp set int eth1 dr dpdk
[edit]
vyos@vyos# commit
[ vpp ]
WARNING: offload option in eth1 settings is not supported by VPP interfaces. It will be ignored.
Interface(s) eth1 cannot be added to VPP because IPoE is already
configured. An interface cannot be used by both VPP and IPoE!
[[vpp]] failed
Commit failed
[edit]
vyos@vyos#
```

In other cases vpp commit output has a new line between messages:

```
vyos@vyos# commit
[ vpp ]
WARNING: offload option in eth1 settings is not supported by VPP interfaces. It will be ignored.

[edit]
vyos@vyos#
```

Restriction works for both cases, trying to enable vpp with ipoe configured and vice versa.

```
vyos@vyos# set service ipoe-server name-server '198.168.2.1'
[edit]
vyos@vyos# commit
[ service ipoe-server ]
eth1 is a VPP interface and cannot be used for IPoE!
[[service ipoe-server]] failed
Commit failed
[edit]
vyos@vyos#
```

So we assume that no config migration is needed as no one would use such combination?

dmbaturin (approved):
I agree that configurations with mutually exclusive features must raise commit errors.

Regarding comments from @alexk37, I agree that the warning and the error shown together without a line break are hard to read. Fixing that is probably out of scope of this PR, though.

Regarding config migration, since this PR doesn't introduce config _syntax_ changes, formally it doesn't need a migration script. It also doesn't affect any configuration from older LTS releases so I think we can simply mention it in the release post but don't need to make sure old configurations with that already non-working option combination load without commit errors.

src/conf_mode/vpp.py
```diff
@@ -492,6 +499,20 @@ def verify(config):
     if 'interface' not in config['settings']:
         raise ConfigError('"settings interface" is required but not set!')
 
+    ipoe_ifaces = list(
+        {
+            iface
+            for iface in config['settings']['interface']
+            for ipoe_iface in config.get('ipoe_conf', {}).get('interface', {})
+            if iface == ipoe_iface or ipoe_iface.startswith(f'{iface}.')
```
zdc, line 496: Would it not match `eth1` in VPP and `eth11` in IPoE?

zdc (changes requested):

src/conf_mode/vpp.py
natali-rs1985: No. This could match only `eth1` in VPP and `eth1` or `eth1.XX` in IPoE

zdc: Yes, right. I have not spotted the dot. :-)

zdc (approved):
Everything looks valid to me.

PR #4959, opened 2026-01-29
firewall: T7569: Add custom nftables rules from a file

l0crian1:
## Change summary

This change adds the ability to add raw nftables rules from `.nft` files. This provides a few benefits:

- Features that are not currently in the VyOS CLI can be applied.
  - This provides not only a way to handle edge cases for firewall configs, but acts as a measure for what features could be added to the CLI in the future.
- Allows commit/boot times to improve since the config is applied along with the atomic load of `nftables.conf` without needing the config to be directly in VyOS.
### Use Case example:
Imagine a customer with a very strict policy for allowed traffic. If they have 10k rules, that would not only make the processing of traffic very slow, but the config load and commit times would be horrible in VyOS. They could instead load the config from a file, and apply that to a map, which is a feature not currently in VyOS.

They could simply manage a CSV with values for: `srcip, dstip, dstport, proto, verdict` and run that against a python script that will output their nft file. All of those would then get added to a map which has O(1) lookup. So this would solve both the long commit/load times of the config, as well as the slow processing of rules.

The output file would look something like:
```
delete table ip my_custom_table
table ip my_custom_table {
        map flow_vmap {
                type ipv4_addr . ipv4_addr . inet_service . inet_proto : verdict
                flags interval
                counter
                elements = { 10.5.0.100 . 10.100.0.55 . 22 . tcp : accept,
                             10.5.0.101 . 10.100.0.55 . 23 . tcp : accept,
                             10.5.0.102 . 10.100.0.55 . 443 . tcp : accept }
        }

        chain my_custom_chain {
                type filter hook forward priority filter+15; policy drop;
                ip saddr . ip daddr . tcp dport . meta l4proto vmap @flow_vmap comment "TCP based Verdict Map"
        }
}
```
These were the times to generate a file with 10k elements in a map, as well as the time to apply that to nftables:
```
root@vyos:/home/vyos# time python3 test.py -n 10000
Wrote 10000 elements to flow_vmap.nft (table inet vyos_test, map flow_vmap)

real    0m0.047s
user    0m0.045s
sys     0m0.002s

root@vyos:/home/vyos# time nft -f flow_vmap.nft 

real    0m0.095s
user    0m0.034s
sys     0m0.055s
```
### Config tree:
```
├── firewall
│   ├── custom-rules
│   │   └── from-file
│   │   │   └── <file path>
│   │           └── revision
```
## Types of changes

## Related Task(s)

https://vyos.dev/T7569

#### Create a file containing valid nftables config and place it somewhere within the `/config` directory:
```
table ip test {
        map flow_vmap {
                type ipv4_addr . ipv4_addr . inet_service . inet_proto : verdict
                flags interval
                counter
                elements = { 192.0.2.10 . 198.51.100.20 . 22 . tcp  : accept,
                             0.0.0.0/0 . 203.0.113.5 . 23 . tcp  : accept }
        }

        chain my_custom_chain {
                ip saddr . ip daddr . tcp dport . meta l4proto vmap @flow_vmap comment "My Custom Rule"
        }
}
```
#### Apply the configuration:
```
set firewall custom-rules from-file /config/test.nft revision 1
```
#### Verify config was applied to nftables:
```
vyos@vyos# sudo nft list table test
table ip test {
        map flow_vmap {
                type ipv4_addr . ipv4_addr . inet_service . inet_proto : verdict
                flags interval
                counter
                elements = { 192.0.2.10 . 198.51.100.20 . 22 . tcp counter packets 0 bytes 0 : accept,
                             0.0.0.0/0 . 203.0.113.5 . 23 . tcp counter packets 0 bytes 0 : accept }
        }

        chain my_custom_chain {
                ip saddr . ip daddr . tcp dport . meta l4proto vmap @flow_vmap comment "My Custom Rule"
        }
}
```
#### Smoketest results:
```
test_bridge_firewall (__main__.TestFirewall.test_bridge_firewall) ... ok
test_custom_nftables_rules (__main__.TestFirewall.test_custom_nftables_rules) ... ok
test_cyclic_jump_validation (__main__.TestFirewall.test_cyclic_jump_validation) ... ok
test_disable_conntrack_per_chain (__main__.TestFirewall.test_disable_conntrack_per_chain) ... ok
test_gre_match (__main__.TestFirewall.test_gre_match) ... ok
test_ipsec_metadata_match (__main__.TestFirewall.test_ipsec_metadata_match) ... ok
test_ipv4_remote_group (__main__.TestFirewall.test_ipv4_remote_group) ... ok
test_ipv6_remote_group (__main__.TestFirewall.test_ipv6_remote_group) ... ok
test_remote_group (__main__.TestFirewall.test_remote_group) ... ok
test_wildcard_interfaces (__main__.TestFirewall.test_wildcard_interfaces) ... ok
test_zone_with_default_firewall (__main__.TestFirewall.test_zone_with_default_firewall) ... ok
test_zone_with_vrf (__main__.TestFirewall.test_zone_with_vrf) ... ok
test_zone_without_member (__main__.TestFirewall.test_zone_without_member) ... ok

Ran 33 tests in 81.218s

OK
```

#### Handling errors in user created `.nft` files:
There is already a `nft -c` operation applied to `nftables.conf`, which will catch any errors in the user's syntax

## Checklist:

### DIFF
--- a/interface-definitions/firewall.xml.in
+++ b/interface-definitions/firewall.xml.in
@@ -7,6 +7,41 @@
     </properties>
     <children>
       #include <include/firewall/global-options.xml.i>
+      <node name="custom-rules">
+        <properties>
+          <help>Add custom nftables rules from a file (advanced users only)</help>
+        </properties>
+        <children>
+          <tagNode name="from-file">
+            <properties>
+              <help>Filename that contains custom nftables rules</help>
+              <valueHelp>
+                <format>txt</format>
+                <description>/path/to/nft-config.nft</description>
+              </valueHelp>
+              <constraint>
+                <validator name="path-check"/>
+              </constraint>
+              <constraintErrorMessage>Filename that contains custom nftables rules not found</constraintErrorMessage>
+            </properties>
+            <children>
+              <leafNode name="revision">
+                <properties>
+                  <help>Unique revision number for the custom rule changes</help>
+                  <valueHelp>
+                    <format>u32:0-999999</format>
+                    <description>Unique revision number for the custom rule changes</description>
+                  </valueHelp>
+                  <constraint>
+                    <validator name="numeric" argument="--range 0-999999"/>
+                  </constraint>
+                  <constraintErrorMessage>Revision number must be between 0 and 999999</constraintErrorMessage>
+                </properties>
+              </leafNode>
+            </children>
+          </tagNode>
+        </children>
+      </node>
       <tagNode name="flowtable">
         <properties>
           <help>Flowtable</help>
--- a/smoketest/scripts/cli/test_firewall.py
+++ b/smoketest/scripts/cli/test_firewall.py
@@ -19,6 +19,8 @@ import unittest
 from glob import glob
 from time import sleep
 
+from pathlib import Path
+
 from base_vyostest_shim import VyOSUnitTestSHIM
 
 from vyos.configsession import ConfigSessionError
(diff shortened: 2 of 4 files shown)


zdc:
I have several arguments in favor of implementing this feature, but there are two critical reasons why it might not be the best approach:

1. If functionality is missing, it is better to implement it through a feature request in the long run, even if it is slower. This ensures proper integration and maintenance within the system.

2. We already have configuration hooks that can achieve the same result as this PR. Although they are not as straightforward to configure, they offer more control, dynamic capabilities, and a clear division of responsibility between what is maintained by VyOS and what is managed by users.

Despite the very strong temptation to approve this PR, I must refrain and acknowledge that it is not the right approach in the long run.

l0crian1:
> 1. If functionality is missing, it is better to implement it through a feature request in the long run, even if it is slower. This ensures proper integration and maintenance within the system.

I agree, it is always better to have the feature implemented into VyOS, but the reality is that "slower" could mean years, and just as realistically can (and often does) mean never. This wouldn't be meant to replace the feature request pipeline, but there is a wide divide between the features of nftables and what can be done in VyOS. Allowing for that divide to be bridged (until a feature is officially added) using a mechanism tied to the commit process is important, rather than relying on a post-hook script that will be disconnected from that process.

> 2. We already have configuration hooks that can achieve the same result as this PR. Although they are not as straightforward to configure, they offer more control, dynamic capabilities, and a clear division of responsibility between what is maintained by VyOS and what is managed by users.

I'm assuming you're meaning the post-commit hooks here. These would behave very differently than this feature. This feature does a single atomic load of both the VyOS generated config, as well as any custom config files. Using the post-commits hooks means that nftables would exist in a partially configured state for a period. With a very large VyOS config, the post-commit script could take 1m+ after the apply stage in a conf-mode script to actually execute due to the long commit times with large configs.

With this, the entire config is loaded at once, which both avoids that partially committed state, but also ensures the commit doesn't complete if the `nft -c` check fails, which prevents the config applied by VyOS and the config applied by a post-commit script from getting out of sync.

sarthurdev:
I agree with @zdc here. We've been long since moving away from using CLI nodes to reference user-defined raw variables, configs etc. While it will impose a performance cost, features should instead be directly built into the CLI with proper validation and error handling.

PR #4962, opened 2026-01-30
isis: T6978: add IS-IS SRv6 node-msd configuration

jvoss:
## Change summary
Add SRv6 node-msd configuration options to IS-IS to support uSID implementations.

## Types of changes

## Related Task(s)

* https://vyos.dev/T6978

```
$ /usr/libexec/vyos/tests/smoke/cli/test_protocols_isis.py
[...]
test_isis_14_segment_routing_srv6_advanced (__main__.TestProtocolsISIS.test_isis_14_segment_routing_srv6_advanced) ... ok
```

## Checklist:

### DIFF
--- a/interface-definitions/include/isis/protocol-common-config.xml.i
+++ b/interface-definitions/include/isis/protocol-common-config.xml.i
@@ -298,25 +298,6 @@
         #include <include/segment-routing-label-value.xml.i>
       </children>
     </node>
-    <node name="srv6">
-      <properties>
-        <help>Segment-Routing SRv6 configuration</help>
-      </properties>
-      <children>
-        <leafNode name="locator">
-          <properties>
-            <help>Specify SRv6 locator</help>
-            <valueHelp>
-              <format>txt</format>
-              <description>SRv6 locator name</description>
-            </valueHelp>
-            <constraint>
-              #include <include/constraint/alpha-numeric-hyphen-underscore.xml.i>
-            </constraint>
-          </properties>
-        </leafNode>
-      </children>
-    </node>
     <leafNode name="maximum-label-depth">
       <properties>
         <help>Maximum MPLS labels allowed for this router</help>
@@ -410,6 +391,94 @@
         </node>
       </children>
     </tagNode>
+    <node name="srv6">
+      <properties>
+        <help>Segment-Routing over IPv6 (SRv6) configuration</help>
+      </properties>
+      <children>
+        <leafNode name="locator">
+          <properties>
+            <help>Specify SRv6 locator</help>
+            <valueHelp>
+              <format>txt</format>
+              <description>SRv6 locator name</description>
+            </valueHelp>
+            <completionHelp>
+              <script>${vyos_completion_dir}/list_srv6_locators.sh</script>
+            </completionHelp>
+            <constraint>
+              #include <include/constraint/alpha-numeric-hyphen-underscore.xml.i>
+            </constraint>
+          </properties>
+        </leafNode>
+        <node name="node-msd">
+          <properties>
+            <help>SRv6 Maximum Segment ID (SID) Depth (MSD)</help>
+          </properties>
+          <children>
+            <leafNode name="max-end-d">
+              <properties>
+                <help>Maximum Segment Identifier for End-of-Data</help>
+                <valueHelp>
+                  <format>u32:0-255</format>
+                  <description>Maximum End D MSD</description>
+                </valueHelp>
+                <constraint>
+                  <validator name="numeric" argument="--range 0-255"/>
+                </constraint>
+              </properties>
+            </leafNode>
+            <leafNode name="max-end-pop">
+              <properties>
+                <help>Maximum number of SIDs in Segment-Routing Header (SRH) for segment pop</help>
+                <valueHelp>
+                  <format>u32:0-255</format>
+                  <description>Maximum End Pop MSD</description>
+                </valueHelp>
+                <constraint>
+                  <validator name="numeric" argument="--range 0-255"/>
+                </constraint>
+              </properties>
+            </leafNode>
+            <leafNode name="max-h-encaps">
+              <properties>
+                <help>Maximum Headend Encapsulation in MSD</help>
+                <valueHelp>
+                  <format>u32:0-255</format>
+                  <description>Maximum Headend encaps in MSD</description>
+                </valueHelp>
+                <constraint>
+                  <validator name="numeric" argument="--range 0-255"/>
+                </constraint>
+              </properties>
+            </leafNode>
+            <leafNode name="max-segs-left">
+              <properties>
+                <help>Maximum Segments left in MSD</help>
+                <valueHelp>
+                  <format>u32:0-255</format>
+                  <description>Maximum Segments Left</description>
+                </valueHelp>
+                  <constraint>
+                  <validator name="numeric" argument="--range 0-255"/>
+                </constraint>
+              </properties>
+            </leafNode>
+          </children>
+        </node>
+        <leafNode name="interface">
+          <properties>
+            <help>Interface for Segment-Routing over IPv6 (SRv6)</help>
+            <completionHelp>
+              <script>${vyos_completion_dir}/list_interfaces</script>
+            </completionHelp>
+            <constraint>
+              #include <include/constraint/interface-name.xml.i>
+            </constraint>
+          </properties>
+        </leafNode>
+      </children>
+    </node>
   </children>
 </node>
 <node name="redistribute">
--- a/src/conf_mode/protocols_isis.py
+++ b/src/conf_mode/protocols_isis.py
@@ -261,6 +261,13 @@ def verify(config_dict):
             f'max-lsp-lifetime must be greater or equal to lsp-refresh-interval + 300'
         )
 
+    # Check IS-IS SRv6
+    if dict_search('segment_routing.srv6', isis):
+        # The interface used to install SRv6 SIDs in the Linux data plane.
+        # https://docs.frrouting.org/en/stable-10.2/isisd.html#clicmd-interface-NAME
+        if not dict_search('segment_routing.srv6.interface', isis):
+            raise ConfigError('Missing interface used for installing SRv6 SIDs')
+
     return None
 
 def generate(config_dict):
--- a/data/templates/frr/isisd.frr.j2
+++ b/data/templates/frr/isisd.frr.j2
@@ -152,9 +152,30 @@ advertise-passive-only
 {%     if segment_routing.maximum_label_depth is vyos_defined %}
  segment-routing node-msd {{ segment_routing.maximum_label_depth }}
 {%     endif %}
-{%     if segment_routing.srv6.locator is vyos_defined %}
+{%     if segment_routing.srv6 is vyos_defined %}
  segment-routing srv6
+{%         if segment_routing.srv6.locator is vyos_defined %}
   locator {{ segment_routing.srv6.locator }}
+{%         endif %}
+{%         if segment_routing.srv6.node_msd is vyos_defined %}
+  node-msd
+{%             if segment_routing.srv6.node_msd.max_end_d is vyos_defined %}
+   max-end-d {{ segment_routing.srv6.node_msd.max_end_d }}
+{%             endif %}
+{%             if segment_routing.srv6.node_msd.max_end_pop is vyos_defined %}
+   max-end-pop {{ segment_routing.srv6.node_msd.max_end_pop }}
+{%             endif %}
+{%             if segment_routing.srv6.node_msd.max_h_encaps is vyos_defined %}
+   max-h-encaps {{ segment_routing.srv6.node_msd.max_h_encaps }}
+{%             endif %}
+{%             if segment_routing.srv6.node_msd.max_segs_left is vyos_defined %}
+   max-segs-left {{ segment_routing.srv6.node_msd.max_segs_left }}
+{%             endif %}
+  exit
+{%         endif %}
+{%         if segment_routing.srv6.interface is vyos_defined %}
+  interface {{ segment_routing.srv6.interface }}
+{%         endif %}
  exit
 {%     endif %}
 {%     if segment_routing.global_block is vyos_defined %}
--- a/op-mode-definitions/include/isis-common.xml.i
+++ b/op-mode-definitions/include/isis-common.xml.i
@@ -143,7 +143,7 @@
 </node>
 <node name="segment-routing">
   <properties>
-    <help>Show IS-IS Segment-Routing (SPRING) information</help>
+    <help>Show IS-IS Segment Routing (SPRING) information</help>
   </properties>
   <children>
     <leafNode name="node">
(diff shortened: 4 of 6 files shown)


src/conf_mode/protocols_isis.py
```diff
@@ -220,6 +222,17 @@ def verify(config_dict):
...
+    if dict_search('segment_routing.srv6', isis):
+        # The interface used to install SRv6 SIDs in the Linux data plane.
+        # https://docs.frrouting.org/en/stable-10.2/isisd.html#clicmd-interface-NAME
+        srv6_iface = dict_search('segment_routing.srv6.interface', isis)
+        if not srv6_iface and not interface_exists('sr0'):
+            call('ip link add sr0 type dummy && ip link set sr0 up')
+        elif srv6_iface and interface_exists('sr0'):
+            call('ip link del sr0')
+    elif interface_exists('sr0'):
+        call('ip link del sr0')
```
sever-sever: Adding and deleting interfaces shouldn't be in the `verify` section.
It should be in the `apply`

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,91 @@
...
+                <valueHelp>
+                  <format>u32:0-255</format>
+                  <description>Specify Maximum End Pop MSD</description>
+                </valueHelp>
+                <constraint>
+                  <validator name="numeric" argument="--range 0-255"/>
+                </constraint>
+              </properties>
+            </leafNode>
+            <leafNode name="max-h-encaps">
```
fett0: same here with the explanation: max-h-encaps : Maximum H.Encaps MSD Type: specifies the maximum number of SIDs that can be included as part of the H.Encaps operation.

src/conf_mode/protocols_isis.py
jvoss: Moved to `apply`

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,91 @@
...
+              </properties>
+            </leafNode>
+          </children>
+        </node>
+        <leafNode name="interface">
+          <properties>
+            <help>Interface for Segment Routing over IPv6 (SRv6) (default: sr0)</help>
+            <completionHelp>
+              <script>${vyos_completion_dir}/list_interfaces</script>
+            </completionHelp>
```
c-po, line 474: Is there a reason to have no CLI interface name validation?

```suggestion
            </completionHelp>
            <constraint>
              #include <include/constraint/interface-name.xml.i>
            </constraint>            
```

src/conf_mode/protocols_isis.py
```diff
@@ -269,6 +271,28 @@ def generate(config_dict):
...
+        vrf
+        and dict_search(f'vrf.name.{vrf}.protocols.isis', config_dict)
+        or config_dict.get('isis', {})
+    )
+
+    if dict_search('segment_routing.srv6', isis):
+        # The interface used to install SRv6 SIDs in the Linux data plane.
+        # https://docs.frrouting.org/en/stable-10.2/isisd.html#clicmd-interface-NAME
+        srv6_iface = dict_search('segment_routing.srv6.interface', isis)
+        if not srv6_iface and not interface_exists('sr0'):
```
c-po: I have no good feeling using this bare CLI interface bringup method. Is there any compelling reason to not have the user specify the interface used properly on the VyOS CLI and if it's not present simply `raise ConfigError('Missing interface used for installing SRv6 SIDs')`?

c-po (changes requested):

src/conf_mode/protocols_isis.py
jvoss: The default according to the documentation is to have a dummy interface named `sr0` for this purpose. Are you saying you would rather require the user to deviate from this and require the user to create their own `dumX` interface?

I believe that adds to confusion of what the purpose of a random `dumX` interface is for. This is why I added specifying an interface as an option or accept the default.

interface-definitions/include/isis/protocol-common-config.xml.i
jvoss: ~~Yes; in this case the interface `sr0` would not exist prior to the commit if they accepted the expected default `sr0` interface.~~

It's been a bit since I've looked at this. I will change this to include the constraint.

src/conf_mode/protocols_isis.py
jvoss: After some experimenting, I've refactored this around the intent that `sr0` will never be a thing on VyOS. I believe this should address your concerns about that interface creation/deletion and agree it likely makes more sense here with the appropriate checks.

Keep in mind that SRv6 for IS-IS always requires an interface for the end SID. It likely does not work today with the current configuration before this PR. I have added a check to ensure an interface is configured but `sr0` is no longer automatically managed when one is not specified.

Here is now the relevant minimal SRv6 configuration for IS-IS:
```
set protocols router isis segment-routing srv6 locator LOC1
set protocols router isis segment-routing srv6 interface dum0
```

interface-definitions/include/isis/protocol-common-config.xml.i
jvoss: This constraint has now been added as requested.

While the intent is to use a dummy interface as described in the [FRR documentation](https://docs.frrouting.org/en/stable-10.2/isisd.html#clicmd-interface-NAME), it does appear that any interface will technically work. Therefore I have left the constraint to any valid interface.

c-po:
Rebased to latest `current` `HEAD` to re-run CI

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+              #include <include/constraint/alpha-numeric-hyphen-underscore.xml.i>
+            </constraint>
+          </properties>
+        </leafNode>
+        <node name="node-msd">
+          <properties>
+            <help>Segment Routing over IPv6 (SRv6) Maximum SRv6 SID Depths</help>
+          </properties>
+          <children>
+            <leafNode name="max-end-d">
```
sever-sever, line 419: This name of the option not looks good for VyOS CLI
Consider to rename it to something more helpful

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+                <valueHelp>
+                  <format>u32:0-255</format>
+                  <description>Specify Maximum End Pop MSD</description>
+                </valueHelp>
+                <constraint>
+                  <validator name="numeric" argument="--range 0-255"/>
+                </constraint>
+              </properties>
+            </leafNode>
+            <leafNode name="max-h-encaps">
```
sever-sever, line 443: This name of the option not looks good for VyOS CLI
Consider to rename it to something more helpful

interface-definitions/include/isis/protocol-common-config.xml.i
jvoss: Commands is duplicated syntax directly from FRR. Can you provide an example of what you would like to see in VyOS?

interface-definitions/include/isis/protocol-common-config.xml.i
jvoss: Commands is duplicated syntax directly from FRR. Can you provide an example of what you would like to see in VyOS?

interface-definitions/include/isis/protocol-common-config.xml.i
l0crian1: The name should probably stay the same in my opinion, since it's the name in the RFC (it's a really badly written RFC in parts). But I think the help string can be improved from what FRR had. Maybe something like:
```
Maximum number of SRv6 End.D decapsulations supported on this node
```
Or:
```
Limits the maximum number of SRv6 End.D decapsulations this node may perform
```
...just my thoughts!

interface-definitions/include/isis/protocol-common-config.xml.i
c-po: IMHO we shoudls tick with the common FRR provided names for this "new" and not yet that widely used technology.

interface-definitions/include/isis/protocol-common-config.xml.i
c-po: Lets start of with the FRR help string. We should definately keep the CLI node names for this "new" and yet not that much used technology to not get people confused. If anyone want's to put in a nice PR for CLI help string changes - be my guest - happily appreciated.

c-po (approved):
Implementation looks good to me - smoketests pass locally so FRR configuration is validated. Not tested function wise due to lack of SRv6 knowlege in IS-IS here.

@jvoss it would be very much appreciated if you can also extend the documentation here: https://github.com/vyos/vyos-documentation/blob/current/docs/configuration/protocols/isis.rst

jvoss:
> Implementation looks good to me - smoketests pass locally so FRR configuration is validated. Not tested function wise due to lack of SRv6 knowlege in IS-IS here.
> 
> @jvoss it would be very much appreciated if you can also extend the documentation here: https://github.com/vyos/vyos-documentation/blob/current/docs/configuration/protocols/isis.rst

Sure! See https://github.com/vyos/vyos-documentation/pull/1760

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+              <script>${vyos_completion_dir}/list_srv6_locators.sh</script>
+            </completionHelp>
+            <constraint>
+              #include <include/constraint/alpha-numeric-hyphen-underscore.xml.i>
+            </constraint>
+          </properties>
+        </leafNode>
+        <node name="node-msd">
+          <properties>
+            <help>Segment Routing over IPv6 (SRv6) Maximum SRv6 SID Depths</help>
```
c-po:
```suggestion
            <help>Segment Routing over IPv6 (SRv6) Maximum SID Depth (MSD)</help>
```

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+          </properties>
+          <children>
+            <leafNode name="max-end-d">
+              <properties>
+                <help>Specify Maximum End D MSD</help>
```
c-po:
```suggestion
                <help>Maximum Segment Identifier for End-of-Data</help>
```

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+                <valueHelp>
+                  <format>u32:0-255</format>
+                  <description>Specify Maximum End D MSD</description>
```
c-po:
```suggestion
                  <description>Maximum End D MSD</description>
```

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+                </valueHelp>
+                <constraint>
+                  <validator name="numeric" argument="--range 0-255"/>
+                </constraint>
+              </properties>
+            </leafNode>
+            <leafNode name="max-end-pop">
+              <properties>
+                <help>Specify Maximum End Pop MSD</help>
```
c-po:
```suggestion
                <help>Maximum number of SIDs in Segment Routing Header (SRH) for segment pop</help>
```

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+                <valueHelp>
+                  <format>u32:0-255</format>
+                  <description>Specify Maximum End Pop MSD</description>
```
c-po:
```suggestion
                  <description>Maximum End Pop MSD</description>
```

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+                </valueHelp>
+                <constraint>
+                  <validator name="numeric" argument="--range 0-255"/>
+                </constraint>
+              </properties>
+            </leafNode>
+            <leafNode name="max-h-encaps">
+              <properties>
+                <help>Specify Maximum H.Encaps MSD</help>
```
c-po:
```suggestion
                <help>Maximum Headend Encapsulation in MSD</help>
```

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+                <valueHelp>
+                  <format>u32:0-255</format>
+                  <description>Specify Maximum H.Encaps MSD</description>
```
c-po:
```suggestion
                  <description>Maximum Headend encaps in MSD</description>
```

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+                </valueHelp>
+                <constraint>
+                  <validator name="numeric" argument="--range 0-255"/>
+                </constraint>
+              </properties>
+            </leafNode>
+            <leafNode name="max-segs-left">
+              <properties>
+                <help>Specify Maximum Segments Left MSD</help>
```
c-po:
```suggestion
                <help>Maximum Segments left in MSD</help>
```

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+                <valueHelp>
+                  <format>u32:0-255</format>
+                  <description>Specify Maximum Segments Left MSD</description>
```
c-po:
```suggestion
                  <description>Maximum Segments Left</description>
```

c-po (changes requested):
Rename CLI nodes. Discussed with Fernando

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
         </node>
       </children>
     </tagNode>
+    <node name="srv6">
+      <properties>
+        <help>Segment-Routing SRv6 configuration</help>
```
c-po:
```suggestion
        <help>Segment Routing over IPv6 (SRv6) configuration</help>
```

interface-definitions/include/isis/protocol-common-config.xml.i
```diff
@@ -410,6 +391,94 @@
...
+              <script>${vyos_completion_dir}/list_srv6_locators.sh</script>
+            </completionHelp>
+            <constraint>
+              #include <include/constraint/alpha-numeric-hyphen-underscore.xml.i>
+            </constraint>
+          </properties>
+        </leafNode>
+        <node name="node-msd">
+          <properties>
+            <help>Segment Routing over IPv6 (SRv6) Maximum SID Depth (MSD)</help>
```
c-po:
```suggestion
            <help>SRv6 Maximum Segment ID (SID) Depth (MSD)</help>
```

c-po (changes requested):
Help string reformat

c-po (approved):
Incorporated CLI help string changes and gathered feedback from Fernando about SRv6 CLI node names (we should stick to the FRR names which are also referenced in the RFC, even if they sound weird).

Brancehd rebased on latest `current`.

sever-sever (approved):
Add IS-IS SRv6 `node-msd` configuration option
I have not tested this feature; I rely on the author's knowledge and smoketests.

PR #4973, opened 2026-02-08
T8242: interfaces add bidirectional support for interface adjust-mss

jd82k:
## Change summary

Add a new optional CLI leaf `adjust-mss-direction` under both
`interfaces ... ip` and `interfaces ... ipv6`.

Supported values:
- outbound (default, existing behavior)
- inbound
- both

Behavior:
- outbound: always installs MSS rewrite rules in postrouting using `oifname`
- inbound:
  - `clamp-mss-to-pmtu`: installs in postrouting using `iifname`
  - numeric MSS: installs in prerouting using `iifname` (so input traffic is covered)
- both: installs outbound + inbound rules according to the logic above

Implementation details:
- Extend MSS rule generation in `Interface` to be direction- and mode-aware
  (match key + nft chain selection).
- Add prerouting MSS chains (`VYOS_TCP_MSS_PREROUTING`) for IPv4/IPv6 raw tables.
- Keep backward compatibility by treating missing direction as `outbound`.
- Improve MSS cleanup to remove stale interface rules from both postrouting and
  prerouting chains, for both `oifname` and `iifname` matches.
- Wire direction config into the IPv4 and IPv6 interface update paths.
- Extend interface smoketests to validate:
  - direction behavior (`outbound` / `inbound` / `both`)
  - correct chain placement (postrouting vs prerouting)
  - both IPv4 and IPv6 MSS rule generation.

## Types of changes

## Related Task(s)

https://vyos.dev/T8242

## Checklist:

### DIFF
--- a/data/vyos-firewall-init.conf
+++ b/data/vyos-firewall-init.conf
@@ -20,6 +20,10 @@ table raw {
         type filter hook postrouting priority -300; policy accept;
     }
 
+    chain VYOS_TCP_MSS_PREROUTING {
+        type filter hook prerouting priority -300; policy accept;
+    }
+
     chain vyos_global_rpfilter {
         return
     }
@@ -38,6 +42,10 @@ table raw {
 table ip6 raw {
     chain VYOS_TCP_MSS {
         type filter hook postrouting priority -300; policy accept;
+    }
+
+    chain VYOS_TCP_MSS_PREROUTING {
+        type filter hook prerouting priority -300; policy accept;
     }
 
     chain vyos_global_rpfilter {
--- a/interface-definitions/include/interface/adjust-mss.xml.i
+++ b/interface-definitions/include/interface/adjust-mss.xml.i
@@ -20,4 +20,28 @@
     </constraint>
   </properties>
 </leafNode>
+<leafNode name="adjust-mss-direction">
+  <properties>
+    <help>Adjust TCP MSS direction</help>
+    <completionHelp>
+      <list>outbound inbound both</list>
+    </completionHelp>
+    <valueHelp>
+      <format>outbound</format>
+      <description>Adjust MSS for packets leaving this interface</description>
+    </valueHelp>
+    <valueHelp>
+      <format>inbound</format>
+      <description>Adjust MSS for packets entering this interface</description>
+    </valueHelp>
+    <valueHelp>
+      <format>both</format>
+      <description>Adjust MSS for both inbound and outbound packets</description>
+    </valueHelp>
+    <constraint>
+      <regex>(outbound|inbound|both)</regex>
+    </constraint>
+  </properties>
+  <defaultValue>outbound</defaultValue>
+</leafNode>
 <!-- include end -->
--- a/python/vyos/ifconfig/interface.py
+++ b/python/vyos/ifconfig/interface.py
@@ -691,19 +691,84 @@ class Interface(Control):
             return None
         return self.set_interface('ipv6_cache_tmo', tmo)
 
-    def _cleanup_mss_rules(self, table, ifname):
-        commands = []
-        results = self._cmd(f'nft -a list chain {table} VYOS_TCP_MSS').split("\n")
-        for line in results:
-            if f'oifname "{ifname}"' in line:
-                handle_search = re.search('handle (\d+)', line)
-                if handle_search:
-                    self._cmd(f'nft delete rule {table} VYOS_TCP_MSS handle {handle_search[1]}')
+    _TCP_MSS_POSTROUTING_CHAIN = 'VYOS_TCP_MSS'
+    _TCP_MSS_PREROUTING_CHAIN = 'VYOS_TCP_MSS_PREROUTING'
+    _TCP_MSS_CHAIN_HOOKS = {
+        'raw': {
+            _TCP_MSS_POSTROUTING_CHAIN: 'type filter hook postrouting priority -300; policy accept;',
+            _TCP_MSS_PREROUTING_CHAIN: 'type filter hook prerouting priority -300; policy accept;',
+        },
+        'ip6 raw': {
+            _TCP_MSS_POSTROUTING_CHAIN: 'type filter hook postrouting priority -300; policy accept;',
+            _TCP_MSS_PREROUTING_CHAIN: 'type filter hook prerouting priority -300; policy accept;',
+        },
+    }
 
-    def set_tcp_ipv4_mss(self, mss):
+    def _ensure_mss_chain(self, table, chain):
+        try:
+            self._cmd(f'nft list chain {table} {chain}')
+        except OSError:
+            chain_definition = self._TCP_MSS_CHAIN_HOOKS[table][chain]
+            self._cmd(f"nft add chain {table} {chain} '{{ {chain_definition} }}'")
+
+    def _ensure_mss_chains(self, table):
+        for chain in self._TCP_MSS_CHAIN_HOOKS[table]:
+            self._ensure_mss_chain(table, chain)
+
+    def _cleanup_mss_rules(self, table, ifname):
+        self._ensure_mss_chains(table)
+        for chain in self._TCP_MSS_CHAIN_HOOKS[table]:
+            results = self._cmd(f'nft -a list chain {table} {chain}').split('\n')
+            for line in results:
+                if f'oifname "{ifname}"' in line or f'iifname "{ifname}"' in line:
+                    handle_search = re.search(r'handle (\d+)', line)
+                    if handle_search:
+                        self._cmd(
+                            f'nft delete rule {table} {chain} handle {handle_search[1]}'
+                        )
+
+    def _get_mss_match_options(self, direction, mss):
+        inbound_chain = (
+            self._TCP_MSS_POSTROUTING_CHAIN
+            if mss == 'clamp-mss-to-pmtu'
+            else self._TCP_MSS_PREROUTING_CHAIN
+        )
+        # Keep pre-direction configs working during upgrades.
+        if not direction:
+            direction = 'outbound'
+
+        direction_map = {
+            'outbound': [(self._TCP_MSS_POSTROUTING_CHAIN, 'oifname')],
+            'inbound': [(inbound_chain, 'iifname')],
+            'both': [
+                (self._TCP_MSS_POSTROUTING_CHAIN, 'oifname'),
+                (inbound_chain, 'iifname'),
+            ],
+        }
+
+        if direction not in direction_map:
+            raise ValueError(f'Invalid adjust-mss-direction "{direction}"')
+
+        return direction_map[direction]
+
+    def _add_fixed_mss_rule(self, table, chain, match_option, mss):
+        nft_prefix = f'nft add rule {table} {chain}'
+        base_cmd = f'{match_option} "{self.ifname}" tcp flags & (syn|rst) == syn'
+        low_mss = str(int(mss) + 1)
+        self._cmd(
+            f"{nft_prefix} '{base_cmd} tcp option maxseg size "
+            f"{low_mss}-65535 tcp option maxseg size set {mss}'"
+        )
+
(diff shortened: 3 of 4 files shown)


jd82k:
Please help me change to PR title. I don't know why it doesn't meet the requirements.

l0crian1:
Ingress would need to be applied to a new chain with a `prerouting` hook.

Default values are canonically not done in code, but rather using the `<defaultValue>` element in the XML, but a maintainer can comment more if they see that as an issue in this PR.

jd82k:
> Ingress would need to be applied to a new chain with a `prerouting` hook.
> 
> Default values are canonically not done in code, but rather using the `<defaultValue>` element in the XML, but a maintainer can comment more if they see that as an issue in this PR.

Thanks for the review. I have addressed both points in the latest update:

1. Ingress MSS is now applied in a dedicated chain with a `prerouting` hook (`VYOS_TCP_MSS_PREROUTING`), instead of using the existing postrouting-only path.  
   - Outbound still uses route-based clamp (`tcp option maxseg size set rt mtu`) in postrouting.  
   - Inbound now supports `clamp-mss-to-pmtu` by dynamically computing an interface-based MSS at commit time (IPv4: `MTU - 40`, IPv6: `MTU - 60`) and programming a fixed MSS rule in prerouting.

2. I removed the hardcoded default from Python.  
   - The default direction is now defined canonically in XML via `<defaultValue>outbound</defaultValue>` under `adjust-mss-direction`.

l0crian1:
> Thanks for the review. I have addressed both points in the latest update:
> 
> 1. Ingress MSS is now applied in a dedicated chain with a `prerouting` hook (`VYOS_TCP_MSS_PREROUTING`), instead of using the existing postrouting-only path.
>    
>    * Outbound still uses route-based clamp (`tcp option maxseg size set rt mtu`) in postrouting.
>    * Inbound now supports `clamp-mss-to-pmtu` by dynamically computing an interface-based MSS at commit time (IPv4: `MTU - 40`, IPv6: `MTU - 60`) and programming a fixed MSS rule in prerouting.
> 2. I removed the hardcoded default from Python.
>    
>    * The default direction is now defined canonically in XML via `<defaultValue>outbound</defaultValue>` under `adjust-mss-direction`.

I want to note that I'm not a maintainer, so these are just my opinions and may not necessarily mate up with the actual VyOS maintainers.

- I don't think "dynamically computing an interface-based MSS" really helps that much. MSS often needs to be clamped due to additional encapsulation (VxLAN, GRE, IPIP, IPsec, etc...), so taking a blind guess at what the MSS should be clamped to doesn't really help. I think it's better to raise a config error when `PMTUD` is configured and the direction is either ingress or both.
- I think the implementation seems overly busy for what is actually being done. For instance, you can just extend the existing functions to use something like this and it would achieve the same thing:
```
direction_map = {
    'inbound': [["VYOS_TCP_MSS_PREROUTING", "iifname"]],
    'outbound': [["VYOS_TCP_MSS", "oifname"]],
    'both': [["VYOS_TCP_MSS_PREROUTING", "iifname"], ["VYOS_TCP_MSS", "oifname"]]
}

for dir_list in direction_map[direction]:
    chain, ifdirection = dir_list
```

jd82k:
> > Thanks for the review. I have addressed both points in the latest update:
> > 
> > 1. Ingress MSS is now applied in a dedicated chain with a `prerouting` hook (`VYOS_TCP_MSS_PREROUTING`), instead of using the existing postrouting-only path.
> >    
> >    * Outbound still uses route-based clamp (`tcp option maxseg size set rt mtu`) in postrouting.
> >    * Inbound now supports `clamp-mss-to-pmtu` by dynamically computing an interface-based MSS at commit time (IPv4: `MTU - 40`, IPv6: `MTU - 60`) and programming a fixed MSS rule in prerouting.
> > 2. I removed the hardcoded default from Python.
> >    
> >    * The default direction is now defined canonically in XML via `<defaultValue>outbound</defaultValue>` under `adjust-mss-direction`.
> 
> I want to note that I'm not a maintainer, so these are just my opinions and may not necessarily mate up with the actual VyOS maintainers.
> 
> * I don't think "dynamically computing an interface-based MSS" really helps that much. MSS often needs to be clamped due to additional encapsulation (VxLAN, GRE, IPIP, IPsec, etc...), so taking a blind guess at what the MSS should be clamped to doesn't really help. I think it's better to raise a config error when `PMTUD` is configured and the direction is either ingress or both.
> * I think the implementation seems overly busy for what is actually being done. For instance, you can just extend the existing functions to use something like this and it would achieve the same thing:
> 
> ```
> direction_map = {
>     'inbound': [["VYOS_TCP_MSS_PREROUTING", "iifname"]],
>     'outbound': [["VYOS_TCP_MSS", "oifname"]],
>     'both': [["VYOS_TCP_MSS_PREROUTING", "iifname"], ["VYOS_TCP_MSS", "oifname"]]
> }
> 
> for dir_list in direction_map[direction]:
>     chain, ifdirection = dir_list
> ```

Thanks for the feedback. I agree that MSS tuning can be affected by multiple encapsulation layers (VXLAN/GRE/IPIP/IPsec), and a static guess is not universally correct.

That said, for interface-scoped MSS adjustment, there is still practical value in deriving ingress MSS from the interface MTU when clamp-mss-to-pmtu is requested:

A common real-world case is tunnel interfaces (for example WireGuard), where operators explicitly set interface MTU (e.g. 1420).
In that case, deriving MSS from that MTU provides predictable behavior and avoids forcing users to manually duplicate the same arithmetic in config.
Conceptually, each interface with MTU < 1500 should own its MSS behavior, especially when the feature is configured under that interface node.
So the intent is not to “blindly guess” globally, but to honor the operator’s interface MTU choice as the closest available source of truth for ingress-side clamping semantics.

If maintainers still prefer strict behavior, I can switch to validation that rejects clamp-mss-to-pmtu for inbound/both and requires an explicit numeric MSS instead.

l0crian1:
> That said, for interface-scoped MSS adjustment, there is still practical value in deriving ingress MSS from the interface MTU when clamp-mss-to-pmtu is requested:

PMTUD has a well-defined, RFC based behavior, so I think trying to create a non-standards based behavior when configuring PMTUD is ultimately wrong. That's not to say you couldn't do something like `clamp-mss-to-imtu` for "Clamp MSS to interface MTU", but as implemented in this PR, deviating from the established behavior could get confusing.

I think `clamp-mss-to-pmtu` should only be contextually valid when doing PTMUD outbound. That's just my 2¢!

jd82k:
I reverted the ingress implementation back to the existing postrouting MSS chain.

Reasoning:
- `clamp-mss-to-pmtu` is route-PMTU semantics (`rt mtu`), which is fundamentally route-context based.
- Implementing ingress via a dedicated prerouting path required non-standard behavior (deriving a fixed MSS from interface MTU), which diverges from PMTUD semantics and can be confusing.
- Keeping everything in postrouting preserves the established VyOS behavior and keeps `clamp-mss-to-pmtu` standards-aligned.

What remains:
- `adjust-mss-direction` (`outbound|inbound|both`) is still supported.
- Default direction is still defined in XML (`<defaultValue>outbound</defaultValue>`), not hardcoded in Python.

So this change prioritizes semantic correctness, predictability, and lower maintenance risk.

l0crian1:
> I reverted the ingress implementation back to the existing postrouting MSS chain.

`iifname` is not valid within the `postrouting` hook for local traffic. This is why I mentioned the `prerouting` hook. Currently, clamping MSS works for both local as well as forwarded traffic, so this would have different behavior between the existing feature and the new feature, depending on ingress (forward only) and egress (forward and local).

jd82k:

jd82k:
> > I reverted the ingress implementation back to the existing postrouting MSS chain.
> 
> `iifname` is not valid within the `postrouting` hook for local traffic. This is why I mentioned the `prerouting` hook. Currently, clamping MSS works for both local as well as forwarded traffic, so this would have different behavior between the existing feature and the new feature, depending on ingress (forward only) and egress (forward and local).

Thanks for the clarification. You are correct that `iifname` in `postrouting` does not apply to locally originated traffic.

My intent here is:

- `outbound`: keep existing behavior (applies to both forwarded and local egress traffic).
- `inbound`: target transit/forwarded traffic only.

For traffic terminated on the router (`input`), MSS clamping is usually not needed. In TCP, each endpoint advertises MSS during SYN/SYN-ACK based on its local egress path. The sender then uses:

`effective_send_mss = min(local_send_mss, peer_advertised_mss)`

So for local services, the router TCP stack already advertises an MSS consistent with its own MTU/path, and the peer will limit segment size accordingly.

So the behavior difference is intentional: ingress (`iifname`) is forward-only, while egress keeps the existing forward+local behavior.

l0crian1:
> So for local services, the router TCP stack already advertises an MSS consistent with its own MTU/path, and the peer will limit segment size accordingly.

That isn't quite accurate. It will advertise an MSS consistent with it's MTU, but the path relies on PMTUD working end-to-end, where it's not uncommon for unreachables being blocked somewhere along the path (especially with many security requirements/guides recommending them to be disabled).

One major issue with ignoring local traffic is if someone wanted a single touch point to clamp BGP sessions. They would be left troubleshooting why it's not working.

I don't personally see any sane reason to not keep parity with the existing behavior (outbound), when adding the new behavior (inbound). Do you have a specific reason you don't want to keep parity in the behaviors?

jd82k:
> > So for local services, the router TCP stack already advertises an MSS consistent with its own MTU/path, and the peer will limit segment size accordingly.
> 
> That isn't quite accurate. It will advertise an MSS consistent with it's MTU, but the path relies on PMTUD working end-to-end, where it's not uncommon for unreachables being blocked somewhere along the path (especially with many security requirements/guides recommending them to be disabled).
> 
> One major issue with ignoring local traffic is if someone wanted a single touch point to clamp BGP sessions. They would be left troubleshooting why it's not working.
> 
> I don't personally see any sane reason to not keep parity with the existing behavior (outbound), when adding the new behavior (inbound). Do you have a specific reason you don't want to keep parity in the behaviors?

Thanks, this is a fair point, especially for local services (for example BGP sessions) where operators expect a single control point.

I’ve updated the implementation accordingly:

- For **numeric MSS values**, inbound rules are now installed in a **prerouting** MSS chain (`iifname`), so inbound handling also covers local input traffic.
- For **`clamp-mss-to-pmtu`**, inbound remains in the existing **postrouting** MSS chain, aligned with the current clamp behavior and kernel-supported `rt mtu` usage.

So the behavior is now:
- `outbound`: postrouting (`oifname`)
- `inbound`:
  - numeric MSS -> prerouting (`iifname`)
  - clamp-mss-to-pmtu -> postrouting (`iifname`)
- `both`: combines outbound plus inbound behavior above.

This keeps backward compatibility for clamp while making fixed-MSS inbound behavior useful for local services.

jd82k:
Please review this PR. @c-po @dmbaturin @sarthurdev @sever-sever

jd82k:
Please assign this PR to me. Thanks!

sarthurdev (approved):
Much improved approach, code looks good, adds smoketest coverage and passes.

jd82k:
Please review this PR. @c-po @dmbaturin @sever-sever

dmbaturin:
I discussed it with @c-po and other maintainers and our conclusion is that there's no real use case for this, so I'm closing this PR.

jd82k:
> I discussed it with @c-po and other maintainers and our conclusion is that there's no real use case for this, so I'm closing this PR.

I do not think it is fair to say there is no use case here. There are real cases where this is useful, especially with tunnels, WireGuard, or other lower-MTU paths, and also in setups where people want one clean place to handle MSS adjustment instead of adding extra rules everywhere.

I understand the concern that the current implementation may not be the best way to expose it in VyOS, and that is a separate discussion. But that is different from saying the feature has no real use case. To me, the current workaround is basically to solve the same problem with more manual and redundant configuration, which is not very clean either.

PR #4984, opened 2026-02-12
T8257: image install search previous needs to consider legacy bind mount

jestabro:
## Change summary

This is the converse problem to https://vyos.dev/T7994: normalization of the config bind mount in https://vyos.dev/T7836 means that when searching for previous installations one searches in `{mount_path}/opt/vyatta/etc/config`. However, when searching for previous installations _that were themselves created_ with the legacy bind mount (<= 1.4.4), the `config.boot` file resident there is not the saved version, but rather the baseline created by the original installation. One must therefore search both the legacy path `{mount_path}/config` as well as the normalized path. This is well defined, as the legacy path only exists in the `{mount_path}` in the case of legacy bind mount.

Note that this is not an issue for upgrades using add system image.

## Types of changes

Test by mounting ISO with this PR to the cdrom drive of an existing 1.4.4 vm and install from booted cd. The correct config.boot is copied. Similarly, installing over existing 1.5 images behave correctly after https://vyos.dev/T7994.

## Checklist:

### DIFF
--- a/src/op_mode/image_installer.py
+++ b/src/op_mode/image_installer.py
@@ -272,12 +272,18 @@ def search_previous_installation(disks: list[str]) -> None:
     print('Searching for data from previous installations')
     image_data = []
     encrypted_configs = []
+    legacy_bind_mount = False
     for disk_name in disks:
         for partition in disk.partition_list(disk_name):
             if disk.partition_mount(partition, mnt_tmp):
                 if Path(mnt_tmp + '/boot').exists():
                     for path in Path(mnt_tmp + '/boot').iterdir():
-                        if path.joinpath('rw/opt/vyatta/etc/config/.vyatta_config').exists():
+                        if path.joinpath('rw/config/.vyatta_config').exists():
+                            legacy_bind_mount = True
+                            image_data.append((path.name, partition))
+                        elif path.joinpath(
+                            'rw/opt/vyatta/etc/config/.vyatta_config'
+                        ).exists():
                             image_data.append((path.name, partition))
                 if Path(mnt_tmp + '/luks').exists():
                     for path in Path(mnt_tmp + '/luks').iterdir():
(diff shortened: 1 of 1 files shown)


alexk37:
@jestabro it didn't detect the previous installation, updating from 1.4.4 to ISO with this PR on a GNS3 instance

```
vyos@vyos:~$ in im
Welcome to VyOS installation!
This command will install VyOS to your permanent storage.
Would you like to continue? [y/N] y
What would you like to name this image? (Default: 2026.02.12-0404-integration) 
Please enter a password for the "vyos" user: 

WARNING: Default password used. Consider changing it on next login.

Please confirm password for the "vyos" user: 
What console should be used by default? (K: KVM, S: Serial)? (Default: S) 
Probing disks
1 disk(s) found
The following disks were found:
Drive: /dev/sda (4.0 GB)
Which one should be used for installation? (Default: /dev/sda) 
Installation will delete all data on the drive. Continue? [y/N] y
Searching for data from previous installations
No previous installation found
Would you like to use all the free space on the drive? [Y/n]
```

The same but on a vmware vm works, previous installation is detected and config is copied over to the the system.

Screenshot 2026-02-12 at 12 06 46<br><br>


The difference could be generic vs kvm flavor of 1.4.4 installation? GNS3 uses kvm flavor and vmware - generic.
Or maybe qcow2 vs iso installation of base 1.4.4 system?

jestabro:
@alexk37 thanks for testing --- for a sanity check, I rebuilt from this PR, and tested on kvm vm 1.4.4: all works as expected --- previous installation was detected, and config was correctly copied. I'll have to take a closer look tomorrow, as I don't see offhand why a flavor difference should matter in this case; a qcow2 difference would be interesting, but note that the change simply looks in the old and new locations, so any odd corner case for the legacy construction must have previously been present (one would reason).

alexk37:
> @alexk37 thanks for testing --- for a sanity check, I rebuilt from this PR, and tested on kvm vm 1.4.4: all works as expected --- previous installation was detected, and config was correctly copied. I'll have to take a closer look tomorrow, as I don't see offhand why a flavor difference should matter in this case; a qcow2 difference would be interesting, but note that the change simply looks in the old and new locations, so any odd corner case for the legacy construction must have previously been present (one would reason).

must be qcow2 yes, there is no .vyatta_config file is present in such installation

jestabro:
Interesting @alexk37, thanks --- I'll investigate that as a separate matter.

sever-sever (approved):
Add legacy bind mount checks

jestabro:
...
> 
> must be qcow2 yes, there is no .vyatta_config file is present in such installation

Issue for installation from qcow2 here:
https://github.com/vyos/vyos-build/pull/1117

c-po (approved):
Take care about legacy bind mount paths during image upgrade

PR #4999, opened 2026-02-18
T8279: recover config data provided by legacy image upgrade tools

jestabro:
## Change summary

When upgrading from systems using the legacy image update tools (1.3.x), config data is copied to a now non-standard path. Add check on first boot into target image to recover source image config data.

This is a basic workaround, however, it will provide the expected continuity of config data when upgrading 1.3.x -> 1.4.5/1.5 (those versions with the corrected /config bind mount).

## Types of changes

## Checklist:

### DIFF
--- a/src/init/vyos-router
+++ b/src/init/vyos-router
@@ -312,6 +312,23 @@ bind_mount_boot ()
     fi
 }
 
+# this is called before migration and bind_mount_slash_config to check for
+# upgrade data installed by legacy image update tools
+copy_legacy_config_path ()
+{
+    if [ -d /config ] ; then
+        if [ -f /config/.upgraded ] ; then
+            rm -f /config/.upgraded
+            cp -a /config /opt/vyatta/etc
+            chgrp -R vyattacfg /opt/vyatta/etc/config
+            chmod -R 775 /opt/vyatta/etc/config
+            chmod 660 /opt/vyatta/etc/config/config.boot
+            touch /opt/vyatta/etc/config/.vyatta_config
+            rm -rf /config
+        fi
+    fi
+}
+
 bind_mount_slash_config ()
 {
     if [ -d /opt/vyatta/etc/config ]
@@ -577,6 +594,8 @@ start ()
     # If for any reason FRR was not started by system_frr.py - start it anyways.
     # This is a safety net!
     systemctl start frr.service
+
+    disabled copy_legacy_config_path || copy_legacy_config_path
 
     disabled bootfile || init_bootfile
 


alexk37:
Verified upgrading from 1.3.8 to iso from this pr, both via `add system image` and from live-mode of mounted iso. Configuration and config folder contents are migrated

src/init/vyos-router
```diff
@@ -312,6 +312,27 @@ bind_mount_boot ()
...
+            do
+                cp -a /config/$i /opt/vyatta/etc/config
+                chgrp -R vyattacfg /opt/vyatta/etc/config/$i
+                chmod -R 775 /opt/vyatta/etc/config/$i
+            done
+            cp /config/config.boot /opt/vyatta/etc/config/config.boot
+            chgrp vyattacfg /opt/vyatta/etc/config/config.boot
+            chmod 660 /opt/vyatta/etc/config/config.boot
+            touch /opt/vyatta/etc/config/.vyatta_config
+            rm -rf /config
```
sarthurdev, line 327: Would this not potentially remove user files in non-standard locations under /config?

dmbaturin: It's an interesting question... we technically never promised to preserve data outside of the well-known subdirectories mentioned on line 321. However, in practice we always copied the entire `/config` so people could have come to rely on that behavior indeed.

jestabro: Yes, I was of two minds on this: I initially went for the conservative approach, however, I now agree with @sarthurdev and will instead reproduce what the (modern) upgrade tools do: copy the whole directory, after pruning the .upgraded hint. I'll push a revised version ...

c-po: Another input from my side. I store all my persistent container configuration files in `/config/container` so after an image upgrade all the configuration files my containers need - which run on VyOS - will start again with the exact same configuration prior to the image upgrade.

jestabro:
Move to draft to test above mentioned change and investigate a separate issue.

jestabro:
Updated to copy full config directory, in case of legacy upgrade; fix misplacement of call.

sarthurdev (approved):
Move config data from legacy path to expected path, resolving migration from 1.3.x

alexk37:
For some reason after upgrade system load stops at this step, I'm checking if something is wrong with me env

```
[  OK  ] Started Deferred execution scheduler.
[  OK  ] Finished Update GRUB loader configuration structure.
[  OK  ] Started VyOS Router.
         Starting Permit User Sessions...
[  OK  ] Finished Permit User Sessions.
[  OK  ] Started Getty on tty1.
[  OK  ] Started Serial Getty on ttyS0.
[  OK  ] Reached target Login Prompts.
[  OK  ] Reached target Multi-User System.
[  OK  ] Reached target VyOS target.
         Starting Record Runlevel Change in UTMP...
[  OK  ] Finished Record Runlevel Change in UTMP.
[  114.923893] vyos-router[1162]: Starting VyOS router.
[  121.809101] vyos-router[1162]: Waiting for NICs to settle down: settled in 0sec..
[  153.112783] vyos-router[1162]: Mounting VyOS Config...done.
[  185.849683] vyos-router[1162]:  migrate system configure.
[  186.180330] vyos-config[1168]: Configuration success
```

alexk37:
Tested in another environment - config and /config/custom folder with its content migrated to 1.5

dmbaturin (approved):
That logic looks fine to me and I trust @alexk37's testing.

PR #5003, opened 2026-02-21
T8291: Add explicit CLI support for marking DNS forwarding upstreams as DoT and remove ambiguous implicit DoT behavior on port 853.

jd82k:
### Change summary
Add explicit CLI support for marking DNS forwarding upstreams as DoT and remove ambiguous implicit DoT behavior on port 853.

- Add new `dot` node under upstream DNS server config (shared include, reused in both paths):
  - `service dns forwarding name-server <ip> dot`
  - `service dns forwarding domain <domain> name-server <ip> dot`
- Keep using the common `port-number` include (no duplicated port XML node).
- Refactor upstream canonicalization in `service_dns_forwarding.py`:
  - default port `53`
  - if `dot` is set and no port is provided, default to `853`
  - explicit `port` is respected (including non-853)
- Remove the previous validation that rejected `dot` when port was not `853`.
- Disable implicit recursor behavior `port 853 => DoT` by setting:
  - `dot-to-port-853=no` in generated `recursor.conf`
- Extend smoketests for:
  - global upstream `dot`
  - domain upstream `dot`
  - custom DoT ports
  - `dot-to-port-853=no` rendering check

### Types of changes

### Related Task(s)
https://vyos.dev/T8291
### Related PR(s)
N/A

### How to test / Smoketest result
1. Configure plain upstream on port 853 (without `dot`) and commit:
   - `set service dns forwarding name-server 192.0.2.1 port 853`
   - Verify generated `recursor.conf` contains `dot-to-port-853=no`
   - Verify `recursor.forward-zones.conf` contains `192.0.2.1:853`
2. Configure explicit `dot` upstream with default port:
   - `set service dns forwarding name-server 192.0.2.2 dot`
   - Verify `recursor.forward-zones.conf` contains `192.0.2.2:853`
3. Configure explicit `dot` upstream with custom port:
   - `set service dns forwarding name-server 192.0.2.3 dot`
   - `set service dns forwarding name-server 192.0.2.3 port 1053`
   - Commit should succeed
   - Verify `recursor.forward-zones.conf` contains `192.0.2.3:1053`
4. Repeat `dot` checks for `service dns forwarding domain <domain> name-server ...`

### Checklist:
- [ ] I have run the components SMOKETESTS if applicable

### DIFF
--- a/data/templates/dns-forwarding/recursor.conf.j2
+++ b/data/templates/dns-forwarding/recursor.conf.j2
@@ -35,6 +35,9 @@ local-port={{ port }}
 # dnssec
 dnssec={{ dnssec }}
 
+# disable implicit DoT for forwarders on port 853
+dot-to-port-853=no
+
 {% if dns64_prefix is vyos_defined %}
 # dns64-prefix
 dns64-prefix={{ dns64_prefix }}
@@ -70,4 +73,3 @@ ecs-ipv4-bits={{ options.ecs_ipv4_bits }}
 {% if options.edns_subnet_allow_list is vyos_defined %}
 edns-subnet-allow-list={{ options.edns_subnet_allow_list | join(',') }}
 {% endif %}
-
--- a/interface-definitions/include/name-server-ipv4-ipv6-port.xml.i
+++ b/interface-definitions/include/name-server-ipv4-ipv6-port.xml.i
@@ -16,8 +16,11 @@
   </properties>
   <children>
     #include <include/port-number.xml.i>
-    <leafNode name="port">
-      <defaultValue>53</defaultValue>
+    <leafNode name="dot">
+      <properties>
+        <help>Use DNS over TLS (DoT) for this upstream DNS server</help>
+        <valueless/>
+      </properties>
     </leafNode>
   </children>
 </tagNode>
--- a/interface-definitions/include/version/dns-forwarding-version.xml.i
+++ b/interface-definitions/include/version/dns-forwarding-version.xml.i
@@ -1,3 +1,3 @@
 <!-- include start from include/version/dns-forwarding-version.xml.i -->
-<syntaxVersion component='dns-forwarding' version='4'></syntaxVersion>
+<syntaxVersion component='dns-forwarding' version='5'></syntaxVersion>
 <!-- include end -->
--- a/smoketest/scripts/cli/test_service_dns_forwarding.py
+++ b/smoketest/scripts/cli/test_service_dns_forwarding.py
@@ -41,6 +41,14 @@ def get_config_value(key, file=CONFIG_FILE):
     tmp = re.findall(r'\n{}=+(.*)'.format(key), tmp)
     return tmp[0]
 
+
+def get_upstream_port(upstream_data):
+    if 'port' in upstream_data:
+        return upstream_data['port']
+    if 'dot' in upstream_data:
+        return '853'
+    return '53'
+
 class TestServicePowerDNS(VyOSUnitTestSHIM.TestCase):
     @classmethod
     def setUpClass(cls):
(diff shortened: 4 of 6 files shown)


jd82k:

sarthurdev:
What is the use case for this change? DoT on non-standard ports?

The changes look fine, just trying to understand why it's needed.

jd82k:
> What is the use case for this change? DoT on non-standard ports?
> 
> The changes look fine, just trying to understand why it's needed.

Thanks for checking.

Yes, the main use case is DoT on non-standard ports in enterprise/private deployments. In practice, internal DNS infrastructures are often shared by multiple routers/tenants behind load balancers or proxies, and port-based separation/policy is used. In those environments, forcing DoT to 853 is too restrictive.

There are also networks where 853 is blocked, redirected, or heavily filtered, so encrypted DNS has to run on an alternative port for operational reasons.

So the intent of this change is:
- `dot` explicitly selects the protocol (DoT)
- `port` explicitly selects the transport port
- no implicit “853 => DoT” behavior

This makes behavior predictable and supports real deployment constraints without changing default plain-DNS behavior unless `dot` is configured.

jd82k:
Please review this PR. @sever-sever @c-po @sarthurdev @dmbaturin

data/templates/dns-forwarding/recursor.conf.j2
```diff
@@ -35,6 +35,9 @@ local-port={{ port }}
 # dnssec
 dnssec={{ dnssec }}
 
+# disable implicit DoT for forwarders on port 853
+dot-to-port-853=no
```
indrajitr, line 39: Won't we need to make this conditional for the scenario where DoT is actually enabled? I am not sure if PowerDNS Recursor supports non-standard port for DoT.

indrajitr (commented):
We probably would require a migration for this one.

data/templates/dns-forwarding/recursor.conf.j2
jd82k: Good catch. `dot-to-port-853` is a global implicit switch in Recursor, not a per-upstream setting.  
If we make it conditional (e.g., enabled when any DoT upstream exists), plain upstreams on port 853 could still be silently upgraded to DoT, which breaks explicit CLI intent.

So the direction is: keep `dot-to-port-853=no` and make DoT explicit per upstream via the `dot` node.  
I agree we should ensure the backend serialization is explicit as well (including non-standard ports), and I’ll add/follow up with that in this PR.

jd82k:
> We probably would require a migration for this one.

I've added the migration script.

data/templates/dns-forwarding/recursor.conf.j2
indrajitr: So if we have:
```
dot-to-port-853=no
+.=9.9.9.9:853
```
Will we still have all queries to 9.9.9.9 over DoT?

NB: I apologize if I am understanding this incorrectly; it has been quite a while since I looked into PowerDNS Recursor :)

jd82k: It depends on if you set this CLI or not `set service dns forwarding name-server 9.9.9.9 dot`

jd82k: PowerDNS Recursor provides two ways to use DoT for upstream queries.

The first is implicit behavior: `dot-to-port-853=yes` (default). In that mode, any upstream configured on port 853 is automatically treated as DoT.

I changed this to an explicit model because, in many real deployments, DoT does not run on the standard port 853. To keep semantics consistent, I set `dot-to-port-853=no` and use an explicit `dot` option to indicate DoT.

With this model:
- `9.9.9.9:853` alone is treated as plain DNS.
- If `dot` is configured, Recursor uses DoT (including non-standard ports).

I believe this is a better implementation and makes operations/configuration clearer for administrators.

indrajitr: > It depends on if you set this CLI or not `set service dns forwarding name-server 9.9.9.9 dot`

How will the generated config file look/differ in the following two cases:

- `set service dns forwarding name-server 1.2.3.4 dot` (DoT over 853)
- `set service dns forwarding name-server 4.5.6.7 port 853` (plain DNS over 853)

jd82k: After checking this further, I noticed PowerDNS Recursor does not support DoT forwarding on non-standard ports. Because of that limitation, this PR has no practical value right now, so I’m closing it.

jd82k:
After checking this further, I noticed PowerDNS Recursor does not support DoT forwarding on non-standard ports. Because of that limitation, this PR has no practical value right now, so I’m closing it.

PR #5046, opened 2026-03-11
T8269: add option to silence individual validator output

jestabro:
## Change summary

DRAFT until merge of:
https://github.com/vyos/vyos-utils/pull/51
https://github.com/vyos/vyos1x-config/pull/67
AND update of vyos1x-config commit hash: until then, integration will fail.

NOTE that this will need a manual backport after manual backport of above PR for vyos1x-config.

Add option to silence individual validator output, in cases where it is misleading; for example (from task):

```
vyos@vyos# set protocols bgp neighbor 172.18.20

  Incorrect path /sys/class/net/172.18.20: no such file or directory

  Error: 172.18.20 is not a valid IP address

  Invalid value
  Value validation failed
  Set failed

```

After these PRs, one may make the following adjustment to the interface definition:

```
diff --git a/interface-definitions/include/bgp/protocol-common-config.xml.i b/interface-definitions/include/bgp/protocol-common-config.xml.i
index e859fe804..f4bcf1509 100644
--- a/interface-definitions/include/bgp/protocol-common-config.xml.i
+++ b/interface-definitions/include/bgp/protocol-common-config.xml.i
@@ -1036,6 +1036,8 @@
       <validator name="ip-address"/>
       #include <include/constraint/interface-name.xml.i>
     </constraint>
+    <constraintSilenceOutput/>
+    <constraintErrorMessage>BGP neighbor must be one of: IP address, IPv6 address, or interface name</constraintErrorMessage>
   </properties>
   <children>
     <node name="address-family">
```

resulting in:

```
vyos@vyos# set protocols bgp neighbor 172.18.20

  BGP neighbor must be one of: IP address, IPv6 address, or interface name
  Value validation failed
  Set failed

```

## Types of changes

- [X] Other (please describe): add element to optionally silence individual validator output

## Related PR(s)

https://github.com/vyos/vyos-utils/pull/51
https://github.com/vyos/vyos1x-config/pull/67

Described in summary, above.

## Checklist:

### DIFF
--- a/interface-definitions/include/bgp/protocol-common-config.xml.i
+++ b/interface-definitions/include/bgp/protocol-common-config.xml.i
@@ -1036,6 +1036,8 @@
       <validator name="ip-address"/>
       #include <include/constraint/interface-name.xml.i>
     </constraint>
+    <constraintSilenceOutput/>
+    <constraintErrorMessage>BGP neighbor must be one of: IP address, IPv6 address, or interface name</constraintErrorMessage>
   </properties>
   <children>
     <node name="address-family">
--- a/schema/interface_definition.rnc
+++ b/schema/interface_definition.rnc
@@ -96,6 +96,7 @@ properties = element properties
     constraintGroup* &
     valueHelp* &
     (element constraintErrorMessage { text })? &
+    (element constraintSilenceOutput { empty })? &
     completionHelp* &
     
     # "docs" is used to store documentation for a node in a structured format
--- a/schema/interface_definition.rng
+++ b/schema/interface_definition.rng
@@ -174,6 +174,11 @@
             <text/>
           </element>
         </optional>
+        <optional>
+          <element name="constraintSilenceOutput">
+            <empty/>
+          </element>
+        </optional>
         <zeroOrMore>
           <ref name="completionHelp"/>
         </zeroOrMore>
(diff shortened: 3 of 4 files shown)


sever-sever (approved):
Add a "silence" option for validators

dmbaturin (commented):
I'm not strictly against this implementation but I now wonder: are there cases when we want both a constraint error message from XML _and_ the output of the validator to be displayed?

Could we silence all validators where the user supplied a constraint error message explicitly?

jestabro:
> I'm not strictly against this implementation but I now wonder: are there cases when we want both a constraint error message from XML _and_ the output of the validator to be displayed?
> 
> Could we silence all validators where the user supplied a constraint error message explicitly?

Good point. We can choose that approach (easily implemented), if desired: I'll ask @c-po for his preference here, following on our earlier discussions. Do we want all validators silent _whenever_ there is a constraint error message ? We would lose any value-specific error information not covered by a general message --- is there any case where that matters ?

dmbaturin (approved):
After some thinking and discussions with @jestabro and other maintainers, I think I can approve the current design.

Silencing all validator outputs when a user-defined constraint error message exists can silence useful errors, and we need to freedom to choose if we want to silence everything or not.

A long-term solution will require making communication with validators more flexible, so that they can explicitly separate validation error messages for users from debug outputs, internal errors, etc.

dmbaturin:
The prerequisite PRs are now merged so we can merge this one as well.

PR #5071, opened 2026-03-20
T8405: fix noipv6 emitted when dhcpv6-options configured without ipv6 node

AlectoTheFirst:
## Change summary

When a PPPoE interface has dhcpv6-options (e.g. prefix delegation) configured but no explicit ipv6 node, the Jinja template writes noipv6 into the PPP peer configuration. This prevents IPv6CP negotiation, causing the ISP to silently ignore all DHCPv6 traffic. Extending the peer template to also check for dhcpv6_options before emitting noipv6, and add dhcpv6-options and ipv6 to the list of config nodes that trigger a PPPoE session restart so the change takes effect without manual disconnect/reconnect as it is not immediately clear that this is necessary

## Types of changes

## Related Task(s)
https://vyos.dev/T8405

  1. Configure a PPPoE interface with DHCPv6-PD but **no** `ipv6` node:
```
  set interfaces pppoe pppoe0 authentication username user
  set interfaces pppoe pppoe0 authentication password pass
  set interfaces pppoe pppoe0 source-interface eth0
  set interfaces pppoe pppoe0 dhcpv6-options pd 0 length 56
  set interfaces pppoe pppoe0 dhcpv6-options pd 0 interface eth1 sla-id 1
  commit
```

  2. Before fix: the generated peer config at 
 ```
  /etc/ppp/peers/pppoe0
```
contains
```
     noipv6 
```
preventing IPv6CP negotiation. 

DHCPv6 Solicit messages are sent but never receive a response from the BNG.

  5. After fix: peer config contains `+ipv6`, IPv6CP negotiates successfully, and DHCPv6-PD works as expected.

  ### Smoketest
  $ /usr/libexec/vyos/tests/smoke/cli/test_interfaces_pppoe.py

  Existing `test_pppoe_dhcpv6pd` always sets `ipv6 address autoconf`
  alongside DHCPv6-PD (line 267), so it does not hit this bug but
  confirms the change does not regress the existing path.

## Checklist:

### DIFF
--- a/data/templates/pppoe/peer.j2
+++ b/data/templates/pppoe/peer.j2
@@ -59,7 +59,7 @@ mru {{ mru }}
 
 {{ "usepeerdns" if no_peer_dns is not vyos_defined }}
 
-{% if ipv6 is vyos_defined %}
++{% if ipv6 is vyos_defined or dhcpv6_options.pd is vyos_defined or (address is vyos_defined and 'dhcpv6' in address) %}
 +ipv6 {{ 'ipv6cp-use-ipaddr' if ipv6.address.autoconf is vyos_defined }}
 {% else %}
 noipv6
--- a/src/conf_mode/interfaces_pppoe.py
+++ b/src/conf_mode/interfaces_pppoe.py
@@ -49,9 +49,18 @@ def get_config(config=None):
     # We should only terminate the PPPoE session if critical parameters change.
     # All parameters that can be changed on-the-fly (like interface description)
     # should not lead to a reconnect!
-    for options in ['access-concentrator', 'connect-on-demand', 'service-name',
-                    'source-interface', 'vrf', 'no-default-route',
-                    'authentication', 'host_uniq']:
+    for options in [
+        'access-concentrator',
+        'connect-on-demand',
+        'service-name',
+        'source-interface',
+        'vrf',
+        'no-default-route',
+        'authentication',
+        'host-uniq',
+        'dhcpv6-options',
+        'ipv6',
+    ]:
         if is_node_changed(conf, base + [ifname, options]):
             pppoe.update({'shutdown_required': {}})
             # bail out early - no need to further process other nodes

### CODE REVIEW

File: src/conf_mode/interfaces_pppoe.py [hunk 1]
```diff
@@ -49,9 +49,18 @@ def get_config(config=None):
...
-                    'authentication', 'host_uniq']:
+    for options in [
+        'access-concentrator',
+        'connect-on-demand',
+        'service-name',
+        'source-interface',
+        'vrf',
+        'no-default-route',
+        'authentication',
+        'host_uniq',
```
c-po: Calls to `is_node_changed()` require the original CLI path not a mangled one.
You were lucky spotting another minor bug.

```suggestion
        'host-uniq',
```

data/templates/pppoe/peer.j2
```diff
@@ -59,7 +59,7 @@ mru {{ mru }}
 
 {{ "usepeerdns" if no_peer_dns is not vyos_defined }}
 
-{% if ipv6 is vyos_defined %}
+{% if ipv6 is vyos_defined or dhcpv6_options is vyos_defined %}
```
c-po: I think we should change it to:

```suggestion
+{% if ipv6 is vyos_defined or dhcpv6_options.pd is vyos_defined or (address is vyos_defined and 'dhcpv6' in address) %}
```

So only enable it if dhcpv6 PD is requested, and DHCPv6 address is requested or the ipv6 node is set.

c-po (changes requested):

data/templates/pppoe/peer.j2
AlectoTheFirst: that would cause the same bug for "parameters-only" and "temporary" as the IPv6CP would be disabled again. Though i am wondering if there is a scenario where somebody would use this without a PD config...

src/conf_mode/interfaces_pppoe.py
AlectoTheFirst: yes, just changed it

data/templates/pppoe/peer.j2
c-po: Yeah, the downside on PPPoE interfaces is that they are different then ethernet or other interfaces in the linux Kernel.

Assume you set dhcpv6-options on an ethernet interface, nothing will happen unless `address dhcpv6` is set, too. We should try to be as consistent as possible with minimal exceptions.

If a user only wants `dhcpv6-options parameters-only` or `dhcpv6-options temporary`, the user should clearly also set `address dhcpv6` to be consistent.

AlectoTheFirst: yes, you are right and even if there are some exotic edge cases it could still be extended later. I'd think it would make sense to add a note to the docs about this specific pppoe / ppp behavior and also extend the [pppoe ipv6 example](https://docs.vyos.io/en/latest/configexamples/pppoe-ipv6-basic.html) accordingly for a PD only scenario

c-po (approved):
All requested changes addressed. Tested locally.

sarthurdev (commented):
Perhaps worth extending this test to add coverage for this?

https://github.com/vyos/vyos-1x/blob/current/smoketest/scripts/cli/test_interfaces_pppoe.py#L253

sever-sever (approved):
Fix IPv6 PD PPPoE client template

PR #5079, opened 2026-03-24
T8418: adjust permissions setting of config.boot for case of cloud-init

jestabro:
## Change summary

Group owner and write permission were set within a conditional block that cloud-init does not see. Move settings outside of unneeded conditional protection.

This became apparent after T8185 restored the correct use of custom `config.boot.default` as defined in flavor files.

## Types of changes

Tested locally with cloud-init deployed in qemu. @alexk37 will be providing further tests.

## Checklist:

### DIFF
--- a/src/init/vyos-router
+++ b/src/init/vyos-router
@@ -164,6 +164,8 @@ init_bootfile () {
         else
             $vyos_libexec_dir/add-system-version.py > $BOOTFILE
         fi
+    fi
+    if [ -f $BOOTFILE ] ; then
         chgrp ${GROUP} $BOOTFILE
         chmod 660 $BOOTFILE
     fi


sever-sever (changes requested):
The task [T8148](https://vyos.dev/T8148) about `SLAAC assigned address` changes, not about `config.boot/vyos-router`

Seems to refer to the wrong task.

jestabro:
Fixed: T8418 Thanks @sever-sever

sever-sever (approved):
Fix the permissions on the `config.boot` file

c-po (approved):
Adjust bootfile permission only if bootfile exists. If there is no bootfile, it will be created from the default boot.

cloud-init installs its own file

dmbaturin (approved):
I see no reason why the fix would have any negative effects and the bug is real.

PR #5086, opened 2026-03-25
vpp: T8419: Disallow cross-membership between VPP and kernel bonding/bridge interfaces

natali-rs1985:
## Types of changes

## Related Task(s)

* https://vyos.dev/T8419

```
vyos@vyos# run show configuration commands | grep bridge
set interfaces bridge br1 member interface eth1
[edit]
[edit]
vyos@vyos# commit
[ vpp ]
Interface eth1 cannot be added to VPP - it or its VLAN(s) is used as a
member of bridge(s): br1
[[vpp]] failed
Commit failed
[edit]
vyos@vyos# del interfaces bridge
[edit]
vyos@vyos# commit
[ vpp ]

[edit]
vyos@vyos#
```
## Checklist:

### DIFF
--- a/src/conf_mode/interfaces_bonding.py
+++ b/src/conf_mode/interfaces_bonding.py
@@ -44,6 +44,7 @@ from vyos.configdict import has_address_configured
 from vyos.configdict import has_vrf_configured
 from vyos.configdep import set_dependents
 from vyos.configdep import call_dependents
+from vyos.vpp.utils import cli_ifaces_list
 from vyos import ConfigError
 from vyos import airbag
 airbag.enable()
@@ -170,6 +171,8 @@ def get_config(config=None):
     if 'static_arp' in bond:
         set_dependents('static_arp', conf)
 
+    bond['vpp_ifaces'] = cli_ifaces_list(conf)
+
     return bond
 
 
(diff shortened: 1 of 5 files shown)


c-po (dismissed):
Please add a testcase which triggers the error condition and checks for the ConfigError assertion.

src/conf_mode/interfaces_bonding.py
```diff
@@ -239,6 +242,12 @@ def verify(bond):
                         continue
                     raise ConfigError(error_msg + f'it has a "{option_path.replace(".", " ")}" assigned!')
 
+            iface_base = interface.split('.')[0]  # get the parent interface name
+            if iface_base in bond['vpp_ifaces']:
+                raise ConfigError(
+                    error_msg + f'it is already configured as VPP interface'
```
sever-sever: Do we need fstring here?

c-po: No, we do not.

natali-rs1985: Fixed

natali-rs1985:
> Please add a testcase which triggers the error condition and checks for the ConfigError assertion.

Added a smoketest

dmbaturin (approved):
I don't see any issues in the latest version of the PR. I also got a confirmation from @c-po that @natali-rs1985 addressed his concerns.

sever-sever (approved):
Though VPP interfaces could be bridged with non-VPP interfaces, we agree not to allow it (at least on this step, and probably rethinking about this in the future.)

PR #5089, opened 2026-03-26
bgp: T7338: Add support for "parameters as-notation <asdot|asdot+>"

robinchrist:
## Change summary

This PR adds support in BGP for `parameters as-notation <asdot|asdot+|plain>` 

With this option the user can control how FRR prints AS Numbers, which can be particularly useful in datacenter scenarios with lots very large of 4-byte AS numbers, as the reserved range starts at 4200000000.

## Types of changes

## Related Task(s)

* https://vyos.dev/T7338

```
DEBUG - test_bgp_28_peer_group_member_all_internal_or_external (__main__.TestProtocolsBGP.test_bgp_28_peer_group_member_all_internal_or_external) ... ok
DEBUG - test_bgp_29_peer_group_remote_as_equal_local_as (__main__.TestProtocolsBGP.test_bgp_29_peer_group_remote_as_equal_local_as) ... ok
DEBUG - test_bgp_30_import_vrf_routemap (__main__.TestProtocolsBGP.test_bgp_30_import_vrf_routemap) ... ok
DEBUG - test_bgp_31_as_notation (__main__.TestProtocolsBGP.test_bgp_31_as_notation) ... ok
DEBUG - test_bgp_99_bmp (__main__.TestProtocolsBGP.test_bgp_99_bmp) ... ok


DEBUG - 
DEBUG - Ran 31 tests in 120.663s
DEBUG - 
DEBUG - OK
```
+ some manual tests, looked fine.

## Checklist:

### DIFF
--- a/interface-definitions/include/bgp/protocol-common-config.xml.i
+++ b/interface-definitions/include/bgp/protocol-common-config.xml.i
@@ -1153,6 +1153,25 @@
     <help>BGP parameters</help>
   </properties>
   <children>
+    <leafNode name="as-notation">
+      <properties>
+        <help>BGP AS-notation output format</help>
+        <completionHelp>
+          <list>asdot asdot+</list>
+        </completionHelp>
+        <valueHelp>
+          <format>asdot</format>
+          <description>Use asdot notation only for 4 byte AS numbers</description>
+        </valueHelp>
+        <valueHelp>
+          <format>asdot+</format>
+          <description>Use asdot notation for all AS numbers</description>
+        </valueHelp>
+        <constraint>
+          <regex>(asdot\+|asdot)</regex>
+        </constraint>
+      </properties>
+    </leafNode>
     <leafNode name="allow-martian-nexthop">
       <properties>
         <help>Allow Martian nexthops to be received in the NLRI from a peer</help>
--- a/smoketest/scripts/cli/test_protocols_bgp.py
+++ b/smoketest/scripts/cli/test_protocols_bgp.py
@@ -1638,6 +1638,57 @@ class TestProtocolsBGP(VyOSUnitTestSHIM.TestCase):
         self.assertIn(f' bgp router-id {router_id}', frr_vrf_config)
 
 
+    def test_bgp_31_as_notation(self):
+        # Test BGP AS-notation output format for both global and VRF BGP instances.
+        # Verify all three notation types render correctly on the router bgp line.
+        router_id = '127.0.0.4'
+        vrf = 'red'
+
+        for vrf in ['', 'red']:
+            vrf_base = ['vrf', 'name', vrf] if vrf else []
+            vrf_frr_bit = f' vrf {vrf}' if vrf else ''
+            bgp_path = vrf_base + base_path
+
+            if vrf_base:
+                self.cli_set(vrf_base + ['table', '2000'])
+
+            self.cli_set(bgp_path + ['system-as', ASN])
+            self.cli_set(bgp_path + ['parameters', 'router-id', router_id])
+
+            for notation in ['asdot', 'asdot+']:
+                self.cli_set(bgp_path + ['parameters', 'as-notation', notation])
+                self.cli_commit()
+
+                # our config options are called 'asdot' and 'asdot+' as in
+                # RFC 5396
+                # but FRR calls them 'dot' and 'dot+' in the router bgp line
+                # e.g. router bgp 12345 as-notation dot
+                # or
+                # router bgp 12345 vrf red as-notation dot
+                router_str = f'router bgp {ASN}{vrf_frr_bit} as-notation {notation.replace("as", "")}'
+                # getFRRConfig interprets the arguments as regex, so escape '+' if present
+                frrconfig = self.getFRRconfig(
+                    router_str.replace('+', r'\+'), stop_section='^exit'
+                )
+                self.assertIn(router_str, frrconfig)
+
+            # Verify removing as-notation works
+            self.cli_delete(bgp_path + ['parameters', 'as-notation'])
+            self.cli_commit()
+
+            router_str = f'router bgp {ASN}{vrf_frr_bit}'
+
+            frrconfig = self.getFRRconfig(router_str, stop_section='^exit')
+            self.assertIn(router_str, frrconfig)
+            self.assertNotIn('as-notation', frrconfig.splitlines()[0])
+
+            # Cleanup
+            self.cli_delete(bgp_path)
+            if vrf_base:
+                self.cli_delete(vrf_base)
+
+            self.cli_commit()
+
     def test_bgp_99_bmp(self):
         target_name = 'instance-bmp'
         target_address = '127.0.0.1'
--- a/data/templates/frr/bgpd.frr.j2
+++ b/data/templates/frr/bgpd.frr.j2
@@ -259,7 +259,7 @@
 {# j2lint: disable=jinja-statements-delimeter #}
 {%- endmacro -%}
 !
-router bgp {{ system_as }} {{ 'vrf ' ~ vrf if vrf is vyos_defined }}
+router bgp {{ system_as }} {{ 'vrf ' ~ vrf if vrf is vyos_defined }} {{ 'as-notation ' ~ parameters.as_notation | replace('as', '') if parameters.as_notation is vyos_defined }}
 {% if parameters.ebgp_requires_policy is vyos_defined %}
  bgp ebgp-requires-policy
 {% else %}


robinchrist:

interface-definitions/include/bgp/protocol-common-config.xml.i
```diff
@@ -1153,6 +1153,29 @@
...
   <children>
+    <leafNode name="as-notation">
+      <properties>
+        <help>BGP AS-notation output format</help>
+        <completionHelp>
+          <list>dot dot+ plain</list>
+        </completionHelp>
+        <valueHelp>
+          <format>dot</format>
+          <description>Use AS4B dot notation only for AS numbers larger than 65535</description>
```
c-po: Use names as per https://datatracker.ietf.org/doc/html/rfc5396
```suggestion
          <description>Use asdot notation only for 4 byte AS numbers</description>
```

interface-definitions/include/bgp/protocol-common-config.xml.i
```diff
@@ -1153,6 +1153,29 @@
...
   </properties>
   <children>
+    <leafNode name="as-notation">
+      <properties>
+        <help>BGP AS-notation output format</help>
+        <completionHelp>
+          <list>dot dot+ plain</list>
+        </completionHelp>
+        <valueHelp>
+          <format>dot</format>
```
c-po:
```suggestion
          <format>asdot</format>
```

interface-definitions/include/bgp/protocol-common-config.xml.i
```diff
@@ -1153,6 +1153,29 @@
     <help>BGP parameters</help>
   </properties>
   <children>
+    <leafNode name="as-notation">
+      <properties>
+        <help>BGP AS-notation output format</help>
+        <completionHelp>
+          <list>dot dot+ plain</list>
```
c-po: Use names as per https://datatracker.ietf.org/doc/html/rfc5396

I know that this requires some fiddling with the bgpd.j2 template - just `parameters.as_notation.replace('as', '')` to strip the leading `as` string.

```suggestion
          <list>asdot asdot+ plain</list>
```

interface-definitions/include/bgp/protocol-common-config.xml.i
```diff
@@ -1153,6 +1153,29 @@
...
+        </completionHelp>
+        <valueHelp>
+          <format>dot</format>
+          <description>Use AS4B dot notation only for AS numbers larger than 65535</description>
+        </valueHelp>
+        <valueHelp>
+          <format>dot+</format>
+          <description>Use AS4B dot notation for all AS numbers</description>
```
c-po:
```suggestion
          <description>Use asdot notation for all AS numbers</description>
```

interface-definitions/include/bgp/protocol-common-config.xml.i
```diff
@@ -1153,6 +1153,29 @@
...
+        </valueHelp>
+        <valueHelp>
+          <format>plain</format>
+          <description>Use plain notation for all AS numbers</description>
+        </valueHelp>
+        <constraint>
+          <regex>(dot\+|dot|plain)</regex>
```
c-po:
```suggestion
          <regex>(asdot\+|asdot|plain)</regex>
```

interface-definitions/include/bgp/protocol-common-config.xml.i
```diff
@@ -1153,6 +1153,29 @@
...
+        </constraint>
+      </properties>
```
c-po, line 1173: Explicitly define `plain` on the CLI which is FRRs implicit internal default, too - make it obvious during completion help.
```suggestion
      </properties>
      <defaultValue>plain</defaultValue>
```

smoketest/scripts/cli/test_protocols_bgp.py
```diff
@@ -1638,6 +1638,54 @@ def test_bgp_30_import_vrf_routemap(self):
...
+    def test_bgp_31_as_notation(self):
+        # Test BGP AS-notation output format for both global and VRF BGP instances.
+        # Verify all three notation types render correctly on the router bgp line.
+        router_id = '127.0.0.4'
+        vrf = 'red'
+
+        for vrf in ['', 'red']:
+            vrf_base = ['vrf', 'name', vrf] if vrf else []
+            vrf_frr_bit = f' vrf {vrf}' if vrf else ''
+            bgp_path = vrf_base + base_path
```
c-po, line 1650: Nice trick ;)

smoketest/scripts/cli/test_protocols_bgp.py
```diff
@@ -1638,6 +1638,54 @@ def test_bgp_30_import_vrf_routemap(self):
...
+
+            if vrf_base:
+                self.cli_set(vrf_base + ['table', '2000'])
+
+            self.cli_set(bgp_path + ['system-as', ASN])
+            self.cli_set(bgp_path + ['parameters', 'router-id', router_id])
+
+            for notation in ['dot', 'dot+', 'plain']:
```
c-po:
```suggestion
            for notation in ['asdot', 'asdot+', 'plain']:
```

c-po (changes requested):

smoketest/scripts/cli/test_protocols_bgp.py
robinchrist: I'm just lazy :D

interface-definitions/include/bgp/protocol-common-config.xml.i
robinchrist: Interesting, I didn't know a default property exists! Thanks

robinchrist: Actually - no, we can't do this. This will break pretty much all existing test cases, as now the generated section is
```
router bgp 12345 as-notation plain
```
and this does not work with `getFRRconfig` (as it does an `^<content>$` match)

I'd suggest to remove the `defaultValue` again.

Another option would be to filter the Jinja template and not emit if the setting is `plain`? Which would kinda defeat the purpose.

sever-sever (approved):
Add `asdot|asdot+` ASN format for BGP

interface-definitions/include/bgp/protocol-common-config.xml.i
c-po: That sounds reasonable. Before changing all testcases, why not simply have the `asdot` and `asdot+` CLI option without `plain`? If nothing is configured, this will then default back to the old `plain` behavior. Best of both worlds.

robinchrist: > If nothing is configured, this will then default back to the old plain behavior. Best of both worlds.

Isn‘t this the case anyway if the user does not configure it explicitly, regardless of whether we offer the option or not?

Offering all 3 options allows the user to be more explicit about intent, which I think is a plus.

(side note: I don’t understand why FRR implemented this parameter the way they did: as part of the router statements.
Should‘ve been a normal parameter in the router section.

I might actually file a PR suggesting to move this as a parameter to the router section and deprecate specifying it as part of the router section)

c-po: I am not a fan of implicit default values which have an option to explicitly overwriting it, but not mentioned on the CLI.

`plain` is the default when not specified on the CLI, and as to my understanding has the same impact when defining it explicitly.

We could either remove `plain` from the CLI options in this PR, or adjust all smoketests to add the new `<defaultValue>plain</defaultValue>` and check the newly rendered `router bgp ASN`string. This could be done in a separate PR as it requires much more fiddeling which I can work on - but we should agree on an implementation first.

Having an explicit default value will also automatically improve the CLI strings to: `BGP AS-notation output format (default: plain)`

robinchrist: I would be fine with removing `plain` from this PR for the sake of getting it merged, although I‘m not a fan of that.

I would like to avoid emitting `plain` into the FRR config by default, as it appears in the router statement and makes the config a bit ugly. 

This would be a different thing if it didn’t appear in the router statement.

c-po: Please go ahead then and remove plain. I am then more then happy to include it!

robinchrist: Done. Rebased and pushed.

c-po (approved):
Requested changes addressed. Add new display option to also allow asdot notation when working with BGP paths

PR #5103, opened 2026-04-02
firewall: T8281: Normalize network prefixes in rule configuration

alexandr-san4ez:
## Change summary
Convert non-canonical IPv4 and IPv6 prefixes in 'source/destination address' fields to proper network addresses (e.g. `10.10.10.1/30 -> 10.10.10.0/30`).

This prevents invalid migrations of configuration and ensures consistent prefix handling.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8281

### Manual test
1. Configure VyOS 1.3 and get the configuration:
```
set firewall name TEST rule 10 action 'accept'
set firewall name TEST rule 10 source address '10.10.10.1/30'
```

2. Verify migration of the configuration:
```
vyos@vyos# load vyos-13.conf
vyos@vyos# comp commands | match "source address"
set firewall ipv4 name TEST rule 10 source address '10.10.10.0/30'
```

### Smoketest
```
root@e5b402c81d55:/vyos/build# scripts/check-qemu-install --debug --configd --match="firewall-rules-address" --cpu 2 --memory 7 --configtest --iso ./build/live-image-amd64.hybrid.iso
DEBUG - Loaded migration result test for config "firewall-rules-address"
DEBUG - ... completed: 0.011284
DEBUG - test_firewall_rules_address (__main__.TestConfigFirewallRulesAddress.test_firewall_rules_address) ...  time: 17.160
DEBUG - ok
DEBUG - Ran 1 test in 17.187s
DEBUG - OK
DEBUG - vyos@vyos:~$echo EXITCODE:$?
DEBUG -  echo EXITCODE:$?
 INFO - Configtest finished successfully!
```

## Checklist:


interface-definitions/include/version/firewall-version.xml.i
```diff
@@ -1,3 +1,3 @@
 <!-- include start from include/version/firewall-version.xml.i -->
-<syntaxVersion component='firewall' version='20'></syntaxVersion>
+<syntaxVersion component='firewall' version='21'></syntaxVersion>
```
c-po: So `sagitta` currently uses firewall version 15, which means if we do want to get it in there we need to change the sagitta version to 16, and then bump all other migration scripts by one.

As VyOS 1.5 is already out this could become a breaking change and needs to be discussed.

c-po (approved):
The logic of the migrator is straight forward by extracting the network portion of a prefix and setting this back into the CLI.

interface-definitions/include/version/firewall-version.xml.i
alexandr-san4ez: I also see another way: update `15-to-16` migration script for `circinus` and `sagitta`. But I'm not sure that a good solution.

alexandr-san4ez: Or even `10-to-11` migration script because it has more related changes.

dmbaturin: I think I'm in favor of adding this to an older migration script. Effectively, this is a missed case of 1.3/iptables to 1.4/nftables migration that should have been in 1.4 all along, so integrating this into an older script restores that justice, in a sense.

alexandr-san4ez: I have updated the PR and moved the code into `10-to-11` migration script.

dmbaturin (approved):
The logic seems good to me and the tests seem thorough enough. We should certainly fix that to keep old configs functional.

PR #5106, opened 2026-04-03
nat64: T8456: add constraint for translation port range

c-po:
## Change summary

NAT64 requires dedicated transport address space for translation, similar to exclusive port ownership during socket binding. If a local process and Jool translation both use the same transport tuple (for example, 192.0.2.1:5000), traffic conflicts can occur.

Jool does not prevent pool4 from overlapping with other port allocations, so avoiding conflicts is an operator responsibility.

In addition to service ports already in use, account for Linux ephemeral range (`net.ipv4.ip_local_port_range`), which defaults to 32768-60999. This default is why, when pool4 is empty, Jool uses 61001-65535 on the node's primary global addresses.

One can adjust the ephemeral range via sysctl, and Jools translation range via pool4 add/remove.

```
vyos@vyos:~$ sysctl net.ipv4.ip_local_port_range
net.ipv4.ip_local_port_range = 32768    60999
```

VyOS now uses `verify()` in NAT64 to check that the supplied tranlation port range does not overlap with the Kernels ephemeral port range.

Change is required in order to update the Jool Kernel Module

## Types of changes

## Related Task(s)

* https://vyos.dev/T8456

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1153

```
cpo@LR1.wue3:~$ /usr/libexec/vyos/tests/smoke/cli/test_nat64.py
test_snat64 (__main__.TestNAT64.test_snat64) ... ok

Ran 1 test in 4.131s

OK
```

## Checklist:

### DIFF
--- a/smoketest/scripts/cli/test_nat64.py
+++ b/smoketest/scripts/cli/test_nat64.py
@@ -20,6 +20,9 @@ import unittest
 
 from base_vyostest_shim import VyOSUnitTestSHIM
 
+from vyos.configsession import ConfigSessionError
+from vyos.utils.system import sysctl_read
+
 base_path = ['nat64']
 src_path = base_path + ['source']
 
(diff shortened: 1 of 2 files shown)


dmbaturin (approved):
I'm a bit uneasy about constraints that depend on the system state (the ephemeral port range in sysctl, in this case), but I agree that the issue is real and I don't see a better fix.

sever-sever (approved):
Extend the port verification for the new Jool version

PR #5119, opened 2026-04-10
wan: T8480: add suppress_prefixlength ip rules for internal routing

doctorpangloss:
## Change summary

WLB per-interface routing tables only contain a default route. When LAN traffic is fwmarked by WLB nftables rules, ip rule policy routes it to these tables where internal destinations (BGP, connected, DNAT'd) have no matching route and incorrectly exit via WAN.

Add `ip rule add fwmark <mark> table main suppress_prefixlength 0` before each per-interface table rule. This checks the main routing table first for specific routes (BGP, connected, static) but suppresses the default route (prefix length 0), so only internet-bound traffic falls through to WLB per-interface tables.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8480
* https://vyos.dev/T2574 (same root cause, closed as Not Applicable on older version)

## How to test

Configure WLB with BGP or DNAT and verify internal traffic is routed correctly:

```
set protocols bgp system-as 65000
set protocols bgp peer-group calico remote-as 65414
set protocols bgp listen range 10.2.0.0/24 peer-group calico

set load-balancing wan interface-health eth0 nexthop dhcp
set load-balancing wan interface-health eth1 nexthop 192.168.1.1
set load-balancing wan rule 1 failover
set load-balancing wan rule 1 interface eth0 weight 10
set load-balancing wan rule 1 interface eth1 weight 1
set load-balancing wan rule 1 inbound-interface eth2
set load-balancing wan rule 1 protocol all

set nat destination rule 110 destination address <wan-ip>
set nat destination rule 110 translation address 10.152.184.99
```

Before fix: LAN traffic to BGP prefixes and DNAT destinations exits via WAN.
After fix: `ip rule show` shows suppress_prefixlength rules before per-interface table rules. Internal traffic is routed via main table, WAN traffic via per-interface tables.

```
$ ip rule show
201: from all fwmark 0xc9 lookup main suppress_prefixlength 0
301: from all fwmark 0xc9 lookup 201
```

## Checklist:

- [x] I have read the **CONTRIBUTING** document
- [ ] I have run the components **SMOKETESTS** if applicable

### DIFF
--- a/interface-definitions/load-balancing_wan.xml.in
+++ b/interface-definitions/load-balancing_wan.xml.in
@@ -29,6 +29,12 @@
               <valueless/>
             </properties>
           </leafNode>
+          <leafNode name="only-default-route">
+            <properties>
+              <help>Prefer specific routes in the main routing table over WAN load balancing</help>
+              <valueless/>
+            </properties>
+          </leafNode>
           <leafNode name="hook">
             <properties>
               <help>Script to be executed on interface status change</help>
--- a/smoketest/scripts/cli/test_load-balancing_wan.py
+++ b/smoketest/scripts/cli/test_load-balancing_wan.py
@@ -138,6 +138,23 @@ class TestLoadBalancingWan(VyOSUnitTestSHIM.TestCase):
         tmp = cmd('sudo ip route show table 202')
         self.assertEqual(tmp, original)
 
+        tmp = cmd('sudo ip rule show')
+        self.assertIn('from all fwmark 0xc9 lookup 201', tmp)
+        self.assertIn('from all fwmark 0xca lookup 202', tmp)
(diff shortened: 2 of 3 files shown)


sarthurdev (changes requested):
I think this might be a good addition, especially where exclude rules would not easily work. But because it changes behaviour and may affect existing setups it will require a CLI node to enable.

Perhaps something like `set load-balancing wan only-default-route`?

Input from other maintainers is welcome @dmbaturin @c-po @sever-sever

sarthurdev:
@doctorpangloss If we are to move forward with the PR could please sign CLA and address requested changes? Thanks

sarthurdev (approved):
Support load balance only via default routes. Adds to WLB smoketest and passes.

c-po (approved):
Add ability to use default only route in WLB rules - otherwise fallback to main routing table for more specifics.

PR #5120, opened 2026-04-10
wan: T8481: scope flush-connections to WLB interfaces only

doctorpangloss:
## Change summary

`flush-connections` runs `conntrack --delete` which destroys all conntrack entries system-wide, including unrelated NAT translations (hairpin SNAT, DNAT port forwards) and established connections that have nothing to do with WLB.

WLB tags every connection it manages with a conntrack mark via the `wlb_mangle_isp_<ifname>` chain (`ct mark set 0xc9`, `0xca`, etc.). Scope the flush to these marks using `conntrack -D -m <mark>` so only WLB-managed connections are affected.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8481
* https://vyos.dev/T1311 (related: conntrack flush hanging with conntrackd, resolved)

## How to test

1. Enable WLB with `set load-balancing wan flush-connections`
2. Establish a long-lived TCP connection on a non-WLB path (e.g. SSH from LAN to LAN host)
3. Trigger a WLB state change (fail a health check)
4. Before fix: all connections drop. After fix: only connections marked by WLB are flushed.

## Checklist:

- [x] I have read the **CONTRIBUTING** document
- [ ] I have run the components **SMOKETESTS** if applicable

### DIFF
--- a/src/helpers/vyos-load-balancer.py
+++ b/src/helpers/vyos-load-balancer.py
@@ -278,8 +278,8 @@ if __name__ == '__main__':
         run('ip route flush cache')
 
         if 'flush_connections' in lb:
-            run('conntrack --delete')
-            run('conntrack -F expect')
+            for _state in lb['health_state'].values():
+                run(f'conntrack --delete --mark {_state["table_number"]}')
 
         with open(wlb_status_file, 'w') as f:
             f.write(json.dumps(lb['health_state']))
--- a/src/helpers/vyos-load-balancer.py
+++ b/src/helpers/vyos-load-balancer.py
@@ -204,8 +204,11 @@ def cleanup(lb):
         index = 1
         for ifname, health_conf in lb['interface_health'].items():
             table_num = lb['mark_offset'] + index
+            suppress_prio = lb['mark_offset'] + index
+            table_prio = suppress_prio + 100
             run(f'ip route del table {table_num} default')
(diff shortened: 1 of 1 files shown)


src/helpers/vyos-load-balancer.py
```diff
@@ -272,8 +272,9 @@ def get_config():
         run('ip route flush cache')
 
         if 'flush_connections' in lb:
-            run('conntrack --delete')
-            run('conntrack -F expect')
+            for _ifname in lb['interface_health']:
+                run(f'conntrack -D -o {_ifname}')
+                run(f'conntrack -D -r {_ifname}')
```
c-po: Please use - if possible - full cmdline arguments which makes reading the code easier. There seems to be none for `-D`.

```suggestion
                run(f'conntrack -D --output {_ifname}')
                run(f'conntrack -D --reply-src {_ifname}')
```

c-po (changes requested):

sarthurdev (changes requested):
Have you tested this on a system to verify?

Manpage hints that conntrack matches only on source/destination IP addresses, not interfaces.

Also `-o` / `--output` refers to returned output from the command.

Ref: https://manpages.debian.org/trixie/conntrack/conntrack.8.en.html

doctorpangloss:
> Have you tested this on a system to verify?

i have now :)

PR #5122, opened 2026-04-12
op-mode: T8483: fix /show_sensors.py: No such file or directory

c-po:
## Change summary

After rewriting the op-mode handlind code and introducing `<virtualTagNodes>` for op-mode, also all `<command>` statements will be executed by a runner. Executing `bash -c ''` code within that runner lacks the proper environment.

This can be verified by adding `env` into the `bash -c ''` executions string.

Solve this issue by moving the hypervisor detection code to `show_sensors.py`.

No backport for `circinus` needed as this code only lives in `current`.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8483

```
vyos@vyos:~$ show environment sensors
No sensors found
```

## Checklist:

### DIFF
--- a/op-mode-definitions/show-environment.xml.in
+++ b/op-mode-definitions/show-environment.xml.in
@@ -11,8 +11,7 @@
             <properties>
               <help>Show hardware monitoring results</help>
             </properties>
-            <!-- Linux always adds "hypervisor" to CPU flags -->
-	    <command>bash -c 'if ! grep -q hypervisor /proc/cpuinfo; then ${vyos_op_scripts_dir}/show_sensors.py; else echo "VyOS running under hypervisor, no sensors available"; fi'</command>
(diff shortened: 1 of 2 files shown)


dmbaturin (approved):
This PR fixes an unintentionally introduced error so I'm not against merging it as is.

However, I wonder if we need the check at all.

First, running the sensors script in a VM doesn't produce any unsightly errors. I tried in a local KVM machine and on a cloud instance and the result it:

```
$ /usr/libexec/vyos/op_mode/show_sensors.py 
No sensors found
```

Second, it's unlikely but not inconceivable for a VM to have simulated sensors for testing or access to host sensors. I'm not aware of any such things existing at the moment but should we block them from working if they appear, if there's no cost for us?

sever-sever (approved):
Fix for `show environment sensors` due to new backend behavior we cannot use bash in this case.

PR #5129, opened 2026-04-14
VPP: T8495: fix issue with assigning  acl on sub interfaces

inetman28:
## Change summary

Fix VPP ACL handling for Ethernet subinterfaces.

`vpp acl` validation used `cli_ifaces_list()`, which only included base VPP interfaces and rejected Ethernet subinterfaces such as `eth1.200`. As a result, ACLs could be applied to physical interfaces like `eth0`/`eth1`, but not to VLAN subinterfaces, even though applying the ACL directly in VPP worked.

This change updates ACL validation to include Ethernet VIF/subinterfaces and also adds the required dependency handling so `vpp_acl` is re-applied when `vif` / `vif-s` configuration changes on a VPP-managed Ethernet interface. This prevents the case where a subinterface is removed and re-created and the ACL remains in config but is no longer attached in VPP runtime.

A smoketest was also added to verify ACL assignment to an Ethernet subinterface.

## Types of changes

## Related Task(s)

- https://vyos.dev/T8495

## Related PR(s)

- None

Test environment:
- Built custom VyOS ISO with this patch applied
- Verified behavior on a running image with VPP enabled

Manual verification:
1. Configure VPP on an Ethernet interface
2. Create an Ethernet VLAN subinterface
3. Create an IP ACL
4. Apply the ACL to the subinterface
5. Commit configuration
6. Verify ACL is attached in VPP

Example:

```
set system option kernel memory hugepage-size 2M hugepage-count '1800'
set vpp settings allow-unsupported-nics   
set vpp settings interface eth0
set vpp settings interface eth1
set vpp settings interface eth2
set vpp settings resource-allocation memory main-heap-size '1G' 
set interface ethernet eth1 vif 200 address '100.64.200.1/24'
set interface ethernet eth2 vif 100 address '100.64.100.1/24'
set vpp acl ip tag-name STATEFUL rule 10 action 'permit-reflect'
set vpp acl ip tag-name STATEFUL rule 10 protocol 'all'
set vpp acl ip interface eth1.200 input acl-tag 10 tag-name 'STATEFUL'
commit

vppctl
show acl-plugin interface 
sw_if_index 0:
sw_if_index 1:
sw_if_index 2:
sw_if_index 3:
sw_if_index 4:
sw_if_index 5:
sw_if_index 6:
sw_if_index 7:
sw_if_index 8:
sw_if_index 9:
  input acl(s): 2
vpp# 
```

I've built ISO image by vyos-build and tested that in GNS3 with above config. 

Expected result:

- Commit succeeds
- ACL is attached to eth1.200

Additional regression coverage:

- Added smoketest test_20_3_vpp_acl_subinterface in smoketest/scripts/cli/test_vpp.py


data/config-mode-dependencies/vyos-vpp.json
```diff
@@ -15,6 +15,9 @@
         "vpp_sflow": ["vpp_sflow"],
         "pppoe_server": ["service_pppoe-server"]
     },
+    "interfaces_ethernet": {
+        "vpp_acl": ["vpp_acl"]
+    },
```
natali-rs1985:
```suggestion
```
We can remove this dependency since subinterface deletion is already validated/guarded by the existing VPP feature config removal checks, so we don’t need a hard dependency here. Also, changing interface config does not change the VPP interface index (sw_if_index), so ACL application remains consistent across reconfiguration.

src/conf_mode/interfaces_ethernet.py
```diff
@@ -175,6 +175,8 @@ def get_config(config=None):
     if tmp: ethernet.update({'frr_dict' : get_frrender_dict(conf)})
 
     ethernet['flowtable_interfaces'] = get_flowtable_interfaces(conf)
+    vif_changed = is_node_changed(conf, base + [ifname, 'vif'])
+    vif_s_changed = is_node_changed(conf, base + [ifname, 'vif-s'])
```
natali-rs1985:
```suggestion
```
These could be removed as well.

src/conf_mode/interfaces_ethernet.py
```diff
@@ -190,6 +192,13 @@ def get_config(config=None):
             get_first_key=True,
             no_tag_node_value_mangle=True,
         )
+        if (
+            'acl' in vpp_config
+            and (vif_changed or vif_s_changed)
+            and dict_search(f'settings.interface.{ifname}', vpp_config) is not None
+        ):
+            ethernet['vpp_acl_dependent'] = True
+            set_dependents('vpp_acl', conf)
```
natali-rs1985:
```suggestion
```
And this

src/conf_mode/interfaces_ethernet.py
```diff
@@ -441,7 +450,7 @@ def apply(ethernet):
         e.remove()
     else:
         e.update(ethernet)
-    if 'static_arp' in ethernet:
+    if 'static_arp' in ethernet or 'vpp_acl_dependent' in ethernet:
```
natali-rs1985:
```suggestion
    if 'static_arp' in ethernet:
```
And this

smoketest/scripts/cli/test_vpp.py
```diff
@@ -1392,6 +1392,46 @@ def test_20_2_vpp_ipfix_bond(self):
         _, out = rc_cmd('sudo vppctl show flowprobe feature')
         self.assertIn(required_str, out)
 
+    def test_20_3_vpp_acl_subinterface(self):
```
natali-rs1985:
```suggestion
    def test_23_vpp_acl_subinterface(self):
```

inetman28:
it looks like there are issues with check-qemu-install. 

Also I see the same issue on another PR tests (for example here : https://github.com/vyos/vyos-1x/pull/5130/checks )

The failure happens in the external `check-qemu-install` when:
```
show configuration commands | match "system console"
```
actual result:
```set system console device ttyS0 speed '115200'```

but it expects:
```
set system console device ttyS0 kernel
set system console device ttyS0 speed '115200'
```

any suggestions ?

**Review by natali-rs1985 — CHANGES_REQUESTED** [review]
Thanks for the implementation — the code looks good overall. I just have two small change requests:

1. **Dependency cleanup:** We should drop the `interfaces_ethernet` -> `vpp_acl` dependency and remove all related code in `interfaces_ethernet.py`.
2. **Test numbering:** Please use a different test number for ACL tests (e.g., 23 — `test_23_vpp_acl_subinterface`). `20` is reserved for IPFIX tests. Also please update the test ordering accordingly.

**Comment by inetman28** [comment]
Hi Natalia,

Thanks for your comments. I applied your suggestions. Thanks.

File: data/config-mode-dependencies/vyos-vpp.json [hunk 6]
```diff
@@ -51,4 +51,3 @@
         "vpp_nat_cgnat": ["vpp_nat_cgnat"]
     }
 }
-
```
natali-rs1985, line 54: This change at the end of the file looks unintended; please remove it from the PR.

smoketest/scripts/cli/test_vpp.py
```diff
@@ -1392,6 +1392,46 @@ def test_20_2_vpp_ipfix_bond(self):
...
+        acl_interfaces = [
+            entry
+            for entry in vpp.api.acl_interface_list_dump()
+            if entry.sw_if_index == subif_index and entry.count != 0
+        ]
+        self.assertEqual(len(acl_interfaces), 1)
+        self.assertEqual(acl_interfaces[0].n_input, 1)
+        self.assertEqual(
+            list(acl_interfaces[0].acls)[: acl_interfaces[0].count], [acl_index]
+        )
```
natali-rs1985, line 1433: Could you please place this test after the existing test_22...  so the numeric ordering remains sequential?

natali-rs1985 (changes requested):
2 more changes before this is ready.
Also, could you please rebase this branch on the current base and force-push? CI is failing, and rebasing will ensure the tests run against the latest baseline and help avoid failures caused by an outdated base.

natali-rs1985 (approved):
Thanks for the implementation. All review comments addressed

inetman28:
Hi Natalia,

I rebased this branch, resolved comments and performed config push.

Thank you.

sever-sever (approved):
Allow the assignment of VPP ACLs to subinterfaces.

PR #5131, opened 2026-04-15
T8499: add ndp-proxy route-sync for dynamic /128 host routes

jd82k:
## Change summary and Use case

Add an optional route-sync feature under service ndp-proxy ... prefix ... to keep /128 host routes in sync with active LAN IPv6 neighbors when NDP proxy is used in mode interface.

This is meant to solve the common IPv6 passthrough case where ndppd can answer NDP correctly, but return traffic is still routed to WAN because the delegated/on-link prefix exists on the upstream interface.

Use case:
When a VyOS router is connected downstream of a modem that only provides IPv6 via SLAAC, and VyOS is used in router mode for IPv6 passthrough, NDP proxy is required.
Configuration:
```bash
set service ndp-proxy interface eth0 prefix '::/0' mode interface
set service ndp-proxy interface eth0 prefix '::/0' interface 'br0'
```

Router Advertisement is also required:

```bash
set service router-advert interface 'br0' prefix '::/64' base-interface 'eth0'
```

The problem is that although br0 can advertise and provide upstream eth0 (modem) SLAAC-derived IPv6 addresses to downstream clients, return routes for those client IPv6 addresses are not installed. As a result, hosts behind br0 cannot use IPv6 connectivity correctly.

This PR addresses that issue by introducing a new route-sync option, which dynamically maintains a downstream IPv6 host route table and installs routes for those addresses via br0.

```bash
set service ndp-proxy interface eth0 prefix '::/0' route-sync
```

## Types of changes

## Related Task(s)

https://vyos.dev/T8499

```shell
vyos@vyos:~$ sudo /usr/libexec/vyos/tests/smoke/cli/test_service_ndp-proxy.py -v
test_basic (__main__.TestServiceNDPProxy.test_basic) ... ok
test_disabled_interface_skips_validation (__main__.TestServiceNDPProxy.test_disabled_interface_skips_validation) ... ok
test_disabled_prefix_skips_validation (__main__.TestServiceNDPProxy.test_disabled_prefix_skips_validation) ... ok
test_prefix_mode_auto_rejects_interface (__main__.TestServiceNDPProxy.test_prefix_mode_auto_rejects_interface) ... ok
test_prefix_mode_interface_requires_interface (__main__.TestServiceNDPProxy.test_prefix_mode_interface_requires_interface) ... ok
test_prefix_mode_interface_with_interface (__main__.TestServiceNDPProxy.test_prefix_mode_interface_with_interface) ... ok
test_route_sync_removed_cleans_runtime_config (__main__.TestServiceNDPProxy.test_route_sync_removed_cleans_runtime_config) ... ok
test_route_sync_renders_config (__main__.TestServiceNDPProxy.test_route_sync_renders_config) ... ok
test_route_sync_requires_interface_mode (__main__.TestServiceNDPProxy.test_route_sync_requires_interface_mode) ... ok

Ran 9 tests in 91.954s

OK
```
## Checklist:

### DIFF
--- a/interface-definitions/service_ndp-proxy.xml.in
+++ b/interface-definitions/service_ndp-proxy.xml.in
@@ -111,6 +111,12 @@
                     </properties>
                     <defaultValue>static</defaultValue>
                   </leafNode>
+                  <leafNode name="route-sync">
+                    <properties>
+                      <help>Synchronize active neighbor /128 routes for this prefix</help>
+                      <valueless/>
+                    </properties>
+                  </leafNode>
                   #include <include/generic-interface.xml.i>
                 </children>
               </tagNode>
--- a/smoketest/scripts/cli/test_service_ndp-proxy.py
+++ b/smoketest/scripts/cli/test_service_ndp-proxy.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
 import unittest
 
 from base_vyostest_shim import VyOSUnitTestSHIM
@@ -25,6 +26,8 @@ from vyos.utils.process import process_named_running
 
 PROCESS_NAME = 'ndppd'
 NDPPD_CONF = '/run/ndppd/ndppd.conf'
+ROUTE_SYNC_CONF = '/run/ndppd/route-sync.conf'
+ROUTE_SYNC_PROTO = '/etc/iproute2/rt_protos.d/ndp_proxy_sync.conf'
 base_path = ['service', 'ndp-proxy']
 
 def getConfigSection(string=None, end=' {', endsection='^}'):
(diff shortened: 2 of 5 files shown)


dmbaturin:
We discussed the PR among maintainers and the conclusion is that it looks like a solution in search of a problem, to put it bluntly.

jd82k:
I do not think “solution in search of a problem” is accurate here.

The problem is that NDP proxying and kernel forwarding are two separate things. ndppd can make the upstream believe that the downstream client address is reachable via VyOS, but it does not necessarily make the VyOS kernel route packets for that address back to the downstream interface when the same /64 is on-link on the upstream interface.

A more-specific /128 route is the missing piece. Without it, route lookup can still select the upstream interface. With it, forwarding works. This can be demonstrated with `ip -6 route get <client_gua>` and a simple before/after test.

I agree that DHCPv6-PD is the correct solution when available, but this feature is specifically for networks where only SLAAC/on-link /64 is available and PD is not provided.

PR #5137, opened 2026-04-19
T8534: refactor sysctl_(read|write) to accept key parts

c-po:
## Change summary

Interface names can contain dots (e.g. VLAN subinterfaces like `eth0.10`), which conflicts with sysctl's dot-separated key syntax.
    
Change `sysctl_read()` and `sysctl_write()` to take key components as a list and normalize each component by replacing `.` with `/` before invoking sysctl. This fixes sysctl lookups/updates for VLAN subinterfaces.
    
Extend the SR-TE smoketest to cover a VLAN subinterface.
    
Previous error raising this issue:
```
vyos-configd: sysctl: cannot stat /proc/sys/net/ipv6/conf/eth0/10/seg6_enabled: No such file or directory
vyos-configd: sysctl: cannot stat /proc/sys/net/ipv6/conf/eth0/201/seg6_enabled: No such file or directory
```

## Types of changes

## Related Task(s)

* https://vyos.dev/T8534

## Checklist:

### DIFF
--- a/python/vyos/utils/system.py
+++ b/python/vyos/utils/system.py
@@ -16,28 +16,34 @@
 import os
 from subprocess import run
 
-def sysctl_read(name: str) -> str:
+def _sysctl_key(parts: list[str]) -> str:
+    normalized = (p.replace('.', '/') for p in parts)
+    return '.'.join(normalized)
+
+def sysctl_read(name: list[str]) -> str:
     """Read and return current value of sysctl() option
 
     Args:
-        name (str): sysctl key name
+        name (list[str]): sysctl key name components
 
     Returns:
         str: sysctl key value
     """
-    tmp = run(['sysctl', '-nb', name], capture_output=True)
-    return tmp.stdout.decode()
+    key = _sysctl_key(name)
+    tmp = run(['sysctl', '-nb', key], capture_output=True)
+    return tmp.stdout.decode().strip()
 
-def sysctl_write(name: str, value: str | int) -> bool:
+def sysctl_write(name: list[str], value: str | int) -> bool:
     """Change value via sysctl()
 
     Args:
-        name (str): sysctl key name
+        name (list[str]): sysctl key name components
         value (str | int): sysctl key value
 
     Returns:
         bool: True if changed, False otherwise
     """
+    key = _sysctl_key(name)
     # convert other types to string before comparison
     if not isinstance(value, str):
         value = str(value)
@@ -45,7 +51,7 @@ def sysctl_write(name: str, value: str | int) -> bool:
     if sysctl_read(name) == value:
         return True
     # return False if sysctl call failed
-    if run(['sysctl', '-wq', f'{name}={value}']).returncode != 0:
+    if run(['sysctl', '-wq', f'{key}={value}']).returncode != 0:
         return False
     # compare old and new values
     # sysctl may apply value, but its actual value will be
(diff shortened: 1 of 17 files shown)


hedrok (changes requested):
It disturbs me that it is very easy to not see this when adding new code.
In `src/conf_mode/protocols_mpls.py` we've already had
```
                    system_interface = system_interface.replace('.', '/')
```
but here not.
I suggest changing `sysctl_read` and `sysctl_write` to accept iterables only (not strings), and for them to convert list `['net', 'ipv6', 'conf', interface, 'seg6_enabled']` to appropriate key with replacing '.' to '/' in each element.
Yes, it is less convenient to use, but shall help with preventing this problem in future + we have only 59 `sysctl_read` and 37 `sysctl_write` usages currently, so I think it shouldn't be a problem.

If you don't think it is a good option, I suggest to at least make it optional to pass iterable, check type in `sysctl_read`/`sysctl_write` and use lists instead of manually changing '.' to '/' before usage of sysctl_read/sysctl_write when variable is used.

natali-rs1985 (approved):
The fix and the inline comment are clear.

c-po:
> It disturbs me that it is very easy to not see this when adding new code. In src/conf_mode/protocols_mpls.py we've already had

Yes I also do not like it

> I suggest changing sysctl_read and sysctl_write to accept iterables only (not strings), and for them to convert list ['net', 'ipv6', 'conf', interface, 'seg6_enabled']

That might actually be a good idea!

c-po:
> ❌ Typos detected in PR (1 found)
> 
> File	Typo	Suggestion
> [src/conf_mode/system_ipv6.py:75](https://github.com/vyos/vyos-1x/blob/2e6ed83398d34879fde3cbd37ab2f8e8b832bfbd/src/conf_mode/system_ipv6.py#L75)	`ND`	`AND`

ND stands for neighbor discovery

src/conf_mode/protocols_segment-routing.py
```diff
@@ -69,25 +69,33 @@ def apply(config_dict):
     sr_interfaces = list(sr.get('interface', {}).keys())
 
     for interface in list_diff(current_interfaces, sr_interfaces):
+        # Interface names can contain a dot that conflicts with sysctl's
+        # dot-as-separator convention. Pass VLAN interface with / as separator
+        # which is properly converted by sysctl
+        interface = interface.replace('.', '/')
```
hedrok, line 75: Please remove.

src/conf_mode/protocols_segment-routing.py
```diff
@@ -69,25 +69,33 @@ def apply(config_dict):
...
         # Disable processing of IPv6-SR packets
-        sysctl_write(f'net.ipv6.conf.{interface}.seg6_enabled', '0')
+        sysctl_write(['net', 'ipv6', 'conf', interface, 'seg6_enabled'], '0')
 
     for interface, interface_config in sr.get('interface', {}).items():
+        # Interface names can contain a dot that conflicts with sysctl's
+        # dot-as-separator convention. Pass VLAN interface with / as separator
+        # which is properly converted by sysctl
+        interface = interface.replace('.', '/')
```
hedrok, line 83: Please remove.

python/vyos/utils/system.py
```diff
@@ -16,36 +16,42 @@
 import os
 from subprocess import run
 
-def sysctl_read(name: str) -> str:
+def _sysctl_key(parts: list[str]) -> str:
+    normalized = [p.replace('.', '/') for p in parts]
+    return '.'.join(normalized)
```
hedrok, line 21: Really minor, ignore if you think it shouldn't be fixed.
This way first we create full list, then pass it to `join`.
Just replacing with generator comprehension:
```python
    normalized = (p.replace('.', '/') for p in parts)
    return '.'.join(normalized)
```
Will not create full extra copy of list without drawbacks - elements will be given to `join` one by one.

hedrok (changes requested):
Thanks for update! I've added several minor comments.

python/vyos/utils/system.py
c-po: Thanks

hedrok (approved):
Approve. Thanks, all my previous concerns are addressed.

sever-sever (approved):
Sysctl read/write refactoring
The TPM tests fail due to known reasons.

PR #5142, opened 2026-04-22
bgp: T8535: Fix FRR rendering for BGP-LS with next-hop-self

natali-rs1985:
When a BGP neighbor has 'next-hop-self' configured under 'address-family link-state', the rendered FRR config is rejected by frr-reload.py with:

  % Unknown command:  exit-address-family

FRR's link-state SAFI parser accepts 'next-hop-self' inside the AF block but does not recognize 'exit-address-family' after it. Since 'exit-address-family' is optional when the block is followed by '!', we omit it in this specific case.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8535
## Checklist:

### DIFF
--- a/data/templates/frr/bgpd.frr.j2
+++ b/data/templates/frr/bgpd.frr.j2
@@ -160,6 +160,7 @@
 {%         elif afi == 'link_state' %}
  address-family link-state
 {%         endif %}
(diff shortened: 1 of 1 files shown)


natali-rs1985:
This bug should be fixed on the FRR's side

PR #5154, opened 2026-04-28
general: T8595: add AGENTS.md and specific implementations

andamasov:
## Summary

Adds a single source of truth for repo-level agent instructions, plus three symlinks so different agentic tools all read the same content from one canonical file.

**Files added:**

| Path | Type | Consumed by |
|---|---|---|
| `CLAUDE.md` | real file | Claude Code, Anthropic agents |
| `AGENTS.md` | symlink → `CLAUDE.md` | tools that follow the AGENTS.md convention (e.g. OpenAI Codex) |
| `.cursorrules` | symlink → `CLAUDE.md` | Cursor |
| `.github/copilot-instructions.md` | symlink → `../CLAUDE.md` | GitHub Copilot ([docs](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions-in-your-ide/add-repository-instructions-in-your-ide)) |

**Rationale**

- One source of truth — guidance lives in `CLAUDE.md`, the symlinks are zero-content. No drift between four files.
- Each agentic tool reads its canonical filename and gets the same content.
- Mac/Linux dev environments handle symlinks natively. Windows checkouts may render them as text files containing the target path; the canonical `CLAUDE.md` is unaffected and still rendered/served correctly by GitHub.

**Out of scope**

Path-specific Copilot instructions (`.github/instructions/*.instructions.md`) need YAML frontmatter (`applyTo:`) and target specific globs, so they cannot be symlinks. Not added here; they can be authored as separate small files per-repo if/when narrow path-scoped guidance is needed.

Tracked under [T8595](https://vyos.dev/T8595).

## Test plan

- [x] `CLAUDE.md` renders on GitHub.
- [x] `AGENTS.md`, `.cursorrules`, `.github/copilot-instructions.md` resolve to CLAUDE.md content (`readlink` returns the right target).
- [ ] CI passes.

### DIFF
--- a/.cursorrules
+++ b/.cursorrules
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,120 @@
+# CLAUDE.md
+
+## Project purpose
+
+The user-visible VyOS package: command definitions (XML), conf-mode and op-mode
+scripts, Jinja2 templates, validators, migration scripts, and the Python
+`vyos.*` library imported by all of the above.
+
+This is the largest single VyOS package and the primary surface for new feature
+work in 1.4+.
+
+## Tech stack
+
+- Python 3 (>=3.11) with the `vyos.*` library under `python/vyos/`.
+- XML interface definitions for configure mode (conf-mode) are located in
+  `interface-definitions/`. XML CLI building blocks can be split into include
+  files to reduce duplication. Include files are in
+  `interface-definitions/include/`.
+- XML definitions for operational mode (op-mode) CLI commands are stored in
+  `op-mode-definitions/`. They use the same include pattern as conf-mode.
+  Include files are in `op-mode-definitions/include/`.
+- Both conf-mode and op-mode make heavy use of Jinja2 as a templating engine.
+  Jinja2 templates are either inline Python strings or stored as discrete files
+  under `data/templates/`. Storing templates as files is preferred.
+- C wrapper `libvyosconfig` (vendored at `libvyosconfig/`) builds
+  `libvyosconfig.so.0` (shared library) using OCaml ctypes bindings against
+  [`vyos/vyos1x-config`](https://github.com/vyos/vyos1x-config).
+- Build: Debian packaging via `debhelper` + `dh-python`. Build dependencies are
+  in `debian/control` (for example: `protobuf-compiler`, `libpcre2-dev`,
+  `libffi-dev`, `python3-vici`, `python3-fastapi`, ...).
+- Tests: `nose2` (`nose2.cfg`), Python `pylint`, and ruff (`ruff.toml`).
+- Runtime smoketests are located under `smoketest/`. These tests are used by
+  `vyos-build` when assembling and testing ISO images.
+
+## Build instructions
+
+```bash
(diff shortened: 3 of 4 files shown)


andamasov:
@copilot review

andamasov:
@copilot review

andamasov:
@copilot review

zdc:
@andamasov, to my best knowledge, `AGENTS.md` would be preferred, because it is a well-established universal entry point, not exclusive to Claude.

The right way is to create `AGENTS.md` and link `CLAUDE.md` to it, if you want Claude-specific instructions: https://code.claude.com/docs/en/memory#agents-md

andamasov:
> @andamasov, to my best knowledge, `AGENTS.md` would be preferred, because it is a well-established universal entry point, not exclusive to Claude.
> 
> The right way is to create `AGENTS.md` and link `CLAUDE.md` to it, if you want Claude-specific instructions: https://code.claude.com/docs/en/memory#agents-md
i will do better, i will symlink agents.md and .cursorrules to claude.md

zdc (changes requested):
It should be opposite: the instructions should be stored in `AGENTS.md` file, and the `CLAUDE.md` should be one of:

- a real symlink to the `AGENTS.md`
- contain a [reference text](https://code.claude.com/docs/en/memory#agents-md) to use instructions from `AGENTS.md`

Also, Cursor does not need `.cursorrules` in this case - it [works with `AGENTS.md` directly](https://cursor.com/docs/rules#agentsmd).

For the content itself, one important addition to have there is an instruction to agents to keep the `AGENTS.md` file updated when changes in the repository are made that affect information stored there.

I would also prefer to have a nested `AGENTS.md` with more detailed information about at least core components, for example:

- `data/templates/AGENTS.md`
- `interface-definitions/AGENTS.md`
- `op-mode-definitions/AGENTS.md`
- `python/AGENTS.md`
- `src/conf_mode/AGENTS.md`
- `src/migration-scripts/AGENTS.md`
- `src/op_mode/AGENTS.md`

One file is not suitable to store all the information for proper AI operations. We also need to have cross-links between files, or at least between the root one and the nested, so agents can understand what instructions are available and for what purposes/components.

Pay attention: creating `AGENTS.md` is a crucial task for future contributions. It should be done only with a full repository analysis, with indexing and semantic search available for AI during this process, and using the most capable AI model available with maximum reasoning. And also include mandatory developers' input/review after.

c-po (approved):
Initial start providing a common AGENTS.md file which we can later extend for specific content by ammending e.g. CLAUDE.md

andamasov:
> I would also prefer to have a nested `AGENTS.md` with more detailed information about at least core components, for example:
> 
> * `data/templates/AGENTS.md`
> * `interface-definitions/AGENTS.md`
> * `op-mode-definitions/AGENTS.md`
> * `python/AGENTS.md`
> * `src/conf_mode/AGENTS.md`
> * `src/migration-scripts/AGENTS.md`
> * `src/op_mode/AGENTS.md`

This will not work for copilot review, as it uses other approach

andamasov:
Closing in favor of [vyos/vyos-1x#5177](https://github.com/vyos/vyos-1x/pull/5177) under [T8595](https://vyos.dev/T8595).

Schema changed: `AGENTS.md` is now the tool-neutral primary instructions file. `.github/copilot-instructions.md` is a symlink to `../AGENTS.md` (verified working with Copilot code review). No `CLAUDE.md`, no `.cursorrules` — both tools fall back to `AGENTS.md` natively.

@zdc @c-po — please re-review the new PR; the schema change addresses the previous review.

PR #5156, opened 2026-04-30
pseudo-ethernet: T8540: Add anycast-gateway support for EVPN

alexandr-san4ez:
## Change summary
Introduce `anycast-gateway` option for pseudo-ethernet interfaces. When set, a local FDB entry is installed on the parent bridge to prevent the shared anycast MAC from leaking over the VXLAN overlay.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8540

## Related PR(s)
* https://github.com/vyos/vyos-documentation/pull/1865
* https://github.com/vyos/vyos-build/pull/1199

### Manual test
```
set interfaces bridge br0 enable-vlan
set interfaces bridge br0 member interface eth2 native-vlan '100'
set interfaces bridge br0 member interface eth3 native-vlan '200'
set interfaces bridge br0 member interface vxlan0
set interfaces bridge br0 vif 100
set interfaces pseudo-ethernet peth100 address '192.168.1.1/24'
set interfaces pseudo-ethernet peth100 ip enable-arp-accept
set interfaces pseudo-ethernet peth100 mac '00:aa:aa:aa:aa:aa'
set interfaces pseudo-ethernet peth100 source-interface 'br0.100'
set interfaces pseudo-ethernet peth100 anycast-gateway
commit
```

```
vyos@vyos1# sudo bridge fdb show dev br0 | grep 00:aa:aa:aa:aa:aa
00:aa:aa:aa:aa:aa self permanent
00:aa:aa:aa:aa:aa vlan 100 master br0 permanent
00:aa:aa:aa:aa:aa vlan 1 master br0 permanent
00:aa:aa:aa:aa:aa master br0 permanent
[edit]
```

```
del interfaces pseudo-ethernet peth100 anycast-gateway
commit
```

```
vyos@vyos1# sudo bridge fdb show dev br0 | grep 00:aa:aa:aa:aa:aa
00:aa:aa:aa:aa:aa self permanent
[edit]
```

```
del interfaces pseudo-ethernet peth100
commit
```

```
vyos@vyos1# sudo bridge fdb show dev br0 | grep 00:aa:aa:aa:aa:aa
[edit]
```

### Smoketest
```
vyos@vyos1:~$ /usr/libexec/vyos/tests/smoke/cli/test_interfaces_pseudo-ethernet.py -k anycast_gateway
test_anycast_gateway (__main__.PEthInterfaceTest.test_anycast_gateway) ... ok
```

## Checklist:

### DIFF
--- a/python/vyos/ifconfig/bridge.py
+++ b/python/vyos/ifconfig/bridge.py
@@ -13,6 +13,8 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 
+import json
+
 from vyos.ifconfig.interface import Interface
 from vyos.utils.assertion import assert_boolean
 from vyos.utils.assertion import assert_list
@@ -97,12 +99,25 @@ class BridgeIf(Interface):
         },
     }}
 
+    _command_get = {**Interface._command_get, **{
+        'fdb_entries': {
+            'shellcmd': 'bridge -json -detail fdb show dev {ifname}',
+            'format': json.loads,
+        },
+    }}
+
     _command_set = {**Interface._command_set, **{
         'add_port': {
             'shellcmd': 'ip link set dev {value} master {ifname}',
         },
         'del_port': {
             'shellcmd': 'ip link set dev {value} nomaster',
+        },
+        'add_local_fdb_entry': {
+            'shellcmd': 'bridge fdb add {value} dev {ifname} self local',
+        },
+        'del_local_fdb_entry': {
+            'shellcmd': 'bridge fdb del {value} dev {ifname} self local',
         },
     }}
 
@@ -272,6 +287,32 @@ class BridgeIf(Interface):
 
         return self.set_interface('vlan_protocol', map[protocol])
 
+    def get_fdb_entries(self):
+        """
+        Get the bridge forwarding database (FDB) entries
+        """
+        return self.get_interface('fdb_entries')
+
+    def add_local_fdb_entry(self, mac_address: str):
+        """
+        Add a local FDB entry for the given MAC address on the bridge interface.
+
+        Example:
+        >>> from vyos.ifconfig import BridgeIf
+        >>> BridgeIf('br0').add_local_fdb_entry('cc:38:2e:bf:7b:0d')
+        """
+        self.set_interface('add_local_fdb_entry', mac_address.lower())
+
+    def del_local_fdb_entry(self, mac_address: str):
+        """
+        Remove a local FDB entry for the given MAC address on the bridge interface.
+
+        Example:
+        >>> from vyos.ifconfig import BridgeIf
+        >>> BridgeIf('br0').del_local_fdb_entry('cc:38:2e:bf:7b:0d')
+        """
+        self.set_interface('del_local_fdb_entry', mac_address.lower())
+
     def update(self, config):
         """ General helper function which works on a dictionary retrieved by
         get_config_dict(). It's main intention is to consolidate the scattered
@@ -331,6 +372,16 @@ class BridgeIf(Interface):
                 # Remove old VLANs from the bridge
                 cmd = f'bridge vlan del dev {self.ifname} vid {vlan} self'
                 self._cmd(cmd)
+
+                # After deleting vif, the FDB entry for that VLAN can linger.
+                # We should remove it manually:
+                fdb_entries = self.get_fdb_entries()
+                for entry in fdb_entries:
+                    mac = entry.get('mac')
+                    entry_vlan = entry.get('vlan')
+                    if entry_vlan and mac and str(entry_vlan) == vlan:
+                        cmd = f'bridge fdb del {mac} dev {self.ifname} vlan {vlan}'
+                        self._cmd(cmd)
 
             for vlan in config.get('vif', {}):
                 cmd = f'bridge vlan add dev {self.ifname} vid {vlan} self'
--- a/smoketest/scripts/cli/test_interfaces_pseudo-ethernet.py
+++ b/smoketest/scripts/cli/test_interfaces_pseudo-ethernet.py
@@ -19,6 +19,8 @@ import unittest
 
 from base_interfaces_test import BasicInterfaceTest
 from base_vyostest_shim import VyOSUnitTestSHIM
+from vyos.configsession import ConfigSessionError
+from vyos.utils.process import cmd
 
 from vyos.ifconfig import Section
 
@@ -43,6 +45,56 @@ class PEthInterfaceTest(BasicInterfaceTest.TestCase):
         cls._interfaces = list(cls._options)
         # call base-classes classmethod
         super(PEthInterfaceTest, cls).setUpClass()
+
+    def test_anycast_gateway(self):
+        # Create the underlying bridge and sub-interface in the test
+
+        for i, peth in enumerate(self._interfaces):
+            eth = peth[1:]  # Convert peth0 -> eth0
+            br = f'br{i}'
+            vlan = str(i + 100)
+
+            # Format the MAC using index as a two-digit hexadecimal
+            mac_address = f'00:aa:aa:aa:aa:{i:02x}'
+
+            with self.subTest(peth=peth, eth=eth, mac_address=mac_address, i=i):
+                base_bridge_path = ['interfaces', 'bridge', br]
+                base_br_member_path = base_bridge_path + ['member', 'interface']
+
+                self.cli_set(base_bridge_path + ['enable-vlan'])
+                self.cli_set(base_br_member_path + [eth, 'native-vlan', vlan])
+                self.cli_set(base_br_member_path + [f'vxlan{i}'])
+                self.cli_set(base_bridge_path + ['vif', vlan])
+
+                self.cli_set(
+                    self._base_path + [peth, 'source-interface', f'{br}.{vlan}']
+                )
+                self.cli_set(self._base_path + [peth, 'anycast-gateway'])
+
+                # Anycast gateway requires MAC
+                with self.assertRaises(ConfigSessionError):
+                    self.cli_commit()
+
+                self.cli_set(self._base_path + [peth, 'mac', mac_address])
+                self.cli_commit()
+
+                # Verify FDB entry exists with flag
+                fdb = cmd(f'bridge fdb show dev {br}')
+                self.assertIn(f'{mac_address} master {br} permanent', fdb)
+                self.assertIn(f'{mac_address} self permanent', fdb)
+
+                # Then remove just the anycast-gateway flag
+                self.cli_delete(self._base_path + [peth, 'anycast-gateway'])
+                self.cli_commit()
+
+                fdb = cmd(f'bridge fdb show dev {br}')
+                self.assertNotIn(f'{mac_address} master {br} permanent', fdb)
+
+                # Clean up temp bridge and peth
+                self.cli_delete(self._base_path + [peth])
+                self.cli_delete(base_bridge_path)
+                self.cli_commit()
+
 
 if __name__ == '__main__':
     unittest.main(verbosity=2, failfast=VyOSUnitTestSHIM.TestCase.debug_on())
--- a/src/conf_mode/interfaces_pseudo-ethernet.py
+++ b/src/conf_mode/interfaces_pseudo-ethernet.py
@@ -68,6 +68,35 @@ def get_config(config=None):
 
     return peth
 
+def _verify_anycast_gateway(peth: dict):
+    """Validate anycast-gateway requirements."""
+
+    if 'anycast_gateway' not in peth:
+        return
+
+    ifname = peth['ifname']
+
+    # Requirement 1: MAC address must be explicitly configured
+    if 'mac' not in peth:
+        raise ConfigError(
+            f'Anycast-gateway requires an explicit MAC address to be set on interface {ifname}. '
+            f'Use: set interfaces pseudo-ethernet {ifname} mac <mac>'
+        )
+
+    # Requirement 2: source-interface must be a bridge or bridge sub-interface
+    source_iface = peth.get('source_interface')
+    if not source_iface:
+        raise ConfigError(
+            f'Anycast-gateway requires source-interface to be set on interface {ifname}'
+        )
+
+    if not source_iface.startswith('br'):
+        raise ConfigError(
+            'Anycast-gateway requires source-interface to be a bridge '
+            'or a bridge vlan interface (e.g. br0 or br0.100), but '
+            f'"{source_iface}" is neither of these two.'
+        )
+
 def verify(peth):
     if 'deleted' in peth:
         verify_bridge_delete(peth)
@@ -81,6 +110,8 @@ def verify(peth):
     verify_mirror_redirect(peth)
     # use common function to verify VLAN configuration
     verify_vlan_config(peth)
+
+    _verify_anycast_gateway(peth)
 
     return None
 
--- a/interface-definitions/interfaces_pseudo-ethernet.xml.in
+++ b/interface-definitions/interfaces_pseudo-ethernet.xml.in
@@ -62,6 +62,12 @@
           #include <include/interface/redirect.xml.i>
           #include <include/interface/vif-s.xml.i>
           #include <include/interface/vif.xml.i>
+          <leafNode name="anycast-gateway">
+            <properties>
+              <help>Use the interface as EVPN Anycast Gateway</help>
+              <valueless/>
+            </properties>
+          </leafNode>
         </children>
       </tagNode>
     </children>
--- a/python/vyos/ifconfig/macvlan.py
+++ b/python/vyos/ifconfig/macvlan.py
@@ -14,6 +14,10 @@
 # License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 
 from vyos.ifconfig.interface import Interface
+from vyos.ifconfig import BridgeIf
+from vyos.utils.network import get_interface_config
+from vyos.utils.network import interface_exists
+from vyos.utils.network import split_interface_vlans
 
 @Interface.register
 class MACVLANIf(Interface):
@@ -40,6 +44,72 @@ class MACVLANIf(Interface):
         # interface is always A/D down. It needs to be enabled explicitly
         self.set_admin_state('down')
 
+    def _get_bridge_by_source(self, source_interface: str) -> BridgeIf | None:
+        """
+        Resolve a source interface name to its root BridgeIf object.
+
+        Handles plain bridge names (e.g. 'br0') and bridge sub-interfaces
+        with one or two VLAN suffixes (e.g. 'br0.100', 'br0.100.200').
+        """
+
+        bridge = None
+        if source_interface.startswith('br'):
+            # We only need the root bridge name, so unpack with *_ to
+            # discard the VLAN parts.
+            bridge_ifname, *_ = split_interface_vlans(source_interface)
+            if interface_exists(bridge_ifname):
+                bridge = BridgeIf(bridge_ifname)
+
+        return bridge
+
+    def _create_anycast_gateway(self, source_interface, mac):
+        """Install a local FDB entry on the parent bridge for the anycast MAC"""
+
+        if source_interface and mac:
+            bridge = self._get_bridge_by_source(source_interface)
+            if bridge:
+                bridge.add_local_fdb_entry(mac)
+
+    def _delete_anycast_gateway(self, source_interface, mac):
+        """Remove the local FDB entry from the parent bridge for the anycast MAC"""
+
+        if source_interface and mac:
+            bridge = self._get_bridge_by_source(source_interface)
+            if bridge:
+                try:
+                    bridge.del_local_fdb_entry(mac)
+                except OSError:
+                    pass  # Bridge may already be gone, that is fine
+
+    def update(self, config):
+        # Always attempt to remove any existing anycast FDB entry before
+        # applying the new config. It reads the currently live
+        # state of the interface (before update).
+        self._delete_anycast_gateway(self.get_source_interface(), self.get_mac())
+
+        # Apply all standard interface configuration
+        super().update(config)
+
+        # After the interface is fully updated, install the new FDB entry if
+        # anycast-gateway is set in the incoming config.
+        if 'anycast_gateway' in self.config:
+            self._create_anycast_gateway(
+                self.config.get('source_interface'), self.get_mac()
+            )
+
+    def remove(self, skip_delete=False):
+        # Before tearing down the MACVLAN interface, clean up the anycast
+        # gateway FDB entry from the parent bridge.
+        self._delete_anycast_gateway(
+            self.get_source_interface(), self.get_mac()
+        )
+
+        return super().remove(skip_delete=skip_delete)
+
     def set_mode(self, mode):
         cmd = f'ip link set dev {self.ifname} type macvlan mode {mode}'
         return self._cmd(cmd)
+
+    def get_source_interface(self):
+        interface_config = get_interface_config(self.ifname)
+        return interface_config['link'] if interface_config is not None else None
--- a/python/vyos/utils/network.py
+++ b/python/vyos/utils/network.py
@@ -158,6 +158,41 @@ def get_vrf_tableid(interface: str):
         table = tmp['linkinfo']['info_slave_data']['table']
     return table
 
+
+def split_interface_vlans(interface: str) -> tuple:
+    """
+    Parse a interface name (with optional VLAN suffixes) into its
+    component parts.
+
+    Handles three input forms:
+      - 'br0'         -> root bridge interface only
+      - 'br0.100'     -> root bridge interface + one VLAN sub-interface level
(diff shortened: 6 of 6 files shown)


python/vyos/ifconfig/bridge.py
```diff
@@ -272,6 +272,26 @@ def set_vlan_protocol(self, protocol):
...
 
+    def add_local_fdb_entry(self, mac_address: str):
+        """
+        Add a local FDB entry for the given MAC address on the bridge interface.
+
+        Example:
+        >>> from vyos.ifconfig import BridgeIf
+        >>> BridgeIf('br0').add_local_fdb_entry('cc:38:2e:bf:7b:0d')
+        """
+        self._cmd(f'bridge fdb add {mac_address.lower()} dev {self.ifname} self local')
```
c-po: Please do not call the commands directly and wrap them into the common code template so it blends in.

```python
    _command_set = {**Interface._command_set, **{
        'add_port': {
            'shellcmd': 'ip link set dev {value} master {ifname}',
        },
        'del_port': {
            'shellcmd': 'ip link set dev {value} nomaster',
        },
    }}
```

https://github.com/alexandr-san4ez/vyos-1x/blob/87425763dfd9c1129bb1967de1f89cf3238dd204/python/vyos/ifconfig/bridge.py#L100-L107

python/vyos/ifconfig/bridge.py
```diff
@@ -272,6 +272,26 @@ def set_vlan_protocol(self, protocol):
...
+
+    def del_local_fdb_entry(self, mac_address: str):
+        """
+        Remove a local FDB entry for the given MAC address on the bridge interface.
+
+        Example:
+        >>> from vyos.ifconfig import BridgeIf
+        >>> BridgeIf('br0').del_local_fdb_entry('cc:38:2e:bf:7b:0d')
+        """
+        self._cmd(f'bridge fdb del {mac_address.lower()} dev {self.ifname} self local')
```
c-po: Please do not call the commands directly and wrap them into the common code template so it blends in.

```python
    _command_set = {**Interface._command_set, **{
        'add_port': {
            'shellcmd': 'ip link set dev {value} master {ifname}',
        },
        'del_port': {
            'shellcmd': 'ip link set dev {value} nomaster',
        },
    }}
```

https://github.com/alexandr-san4ez/vyos-1x/blob/87425763dfd9c1129bb1967de1f89cf3238dd204/python/vyos/ifconfig/bridge.py#L100-L107

src/conf_mode/interfaces_pseudo-ethernet.py
```diff
@@ -68,6 +82,35 @@ def get_config(config=None):
...
+    if not source_iface:
+        raise ConfigError(
+            f'Anycast-gateway requires source-interface to be set on interface {ifname}'
+        )
+
+    if not source_iface.startswith('br'):
+        raise ConfigError(
+            'Anycast-gateway requires source-interface to be a bridge '
+            'interface or a bridge sub-interface (e.g. br0 or br0.100). '
+            f'"{source_iface}" does not resolve to a bridge master.'
```
c-po:
```suggestion
            'Anycast-gateway requires source-interface to be a bridge '
            'or a bridge vlan interface (e.g. br0 or br0.100), but '
            f'"{source_iface}" is neither of these two.'
```

c-po (changes requested):

l0crian1:
One issue I foresee is if someone disables the bridge interface, then the FDB entry will be purged. When they reenable the bridge interface, there isn't a mechanism to add it back. It'd only be added back if/when the `peth` interface is modified.

python/vyos/ifconfig/bridge.py
alexandr-san4ez: Thanks for your feedback. I didn't notice this possibility. I will fix that.

alexandr-san4ez:
> One issue I foresee is if someone disables the bridge interface, then the FDB entry will be purged. When they reenable the bridge interface, there isn't a mechanism to add it back. It'd only be added back if/when the `peth` interface is modified.

It is a good point. I will try to fix that by some approach.

l0crian1:
> It is a good point. I will try to fix that by some approach.

It’d probably make more sense to make this a normal leafNode within a bridge interface. That would not only solve the issue I stated, but it would eliminate some of the checks that you created, like the hardcoded MAC and source-interface checks. 

So instead of the `anycast-gateway` config being under the peth, it’d be within the bridge interface:

```
set interfaces bridge br0 vif 100 anycast-gateway '00:aa:aa:aa:aa:aa'
```

alexandr-san4ez:
> > It is a good point. I will try to fix that by some approach.
> 
> It’d probably make more sense to make this a normal leafNode within a bridge interface. That would not only solve the issue I stated, but it would eliminate some of the checks that you created, like the hardcoded MAC and source-interface checks.
> 
> So instead of the `anycast-gateway` config being under the peth, it’d be within the bridge interface:
> 
> ```
> set interfaces bridge br0 vif 100 anycast-gateway '00:aa:aa:aa:aa:aa'
> ```

It is interesting idea. @c-po, @aapostoliuk, @sever-sever what do you think about it?

sever-sever:
> > > It is a good point. I will try to fix that by some approach.
> > 
> > 
> > It’d probably make more sense to make this a normal leafNode within a bridge interface. That would not only solve the issue I stated, but it would eliminate some of the checks that you created, like the hardcoded MAC and source-interface checks.
> > So instead of the `anycast-gateway` config being under the peth, it’d be within the bridge interface:
> > ```
> > set interfaces bridge br0 vif 100 anycast-gateway '00:aa:aa:aa:aa:aa'
> > ```
> 
> It is interesting idea. @c-po, @aapostoliuk, @sever-sever what do you think about it?

Not sure why we need the word `anycast` there?
For me, it looks like
```
set fdb address <aa:aa:bb:bb:cc:cc:dd:dd> interface br0
```
Or
```
set fdb interface br0 address <aa:aa:bb:bb:cc:cc:dd:dd> 
```
Without additional magic words.

src/conf_mode/interfaces_pseudo-ethernet.py
```diff
@@ -82,11 +125,71 @@ def verify(peth):
...
+    #   2. the MAC address changed (old entry is now stale)
+    #   3. the source interface changed (old entry is now stale)
+    #   4. the interface is being deleted
+    if (
+        (was_anycast and not is_anycast)
+        or mac_changed
+        or source_changed
+        or 'deleted' in peth
+    ):
+        old_source = effective_config.get('source_interface')
```
natali-rs1985: `old_source` is already defined in R162. can be removed.

alexandr-san4ez: Yes, your are right. I have fixed in the recent update. Thanks you.

src/conf_mode/interfaces_pseudo-ethernet.py
```diff
@@ -82,11 +125,70 @@ def verify(peth):
...
+                with contextlib.suppress(OSError):
+                    bridge.del_local_fdb_entry(old_mac)
+
+    if 'deleted' not in peth:
+        # Add new FDB entry if anycast-gateway is active
+        if is_anycast and new_source and new_mac:
+            bridge = _get_bridge_by_source(new_source)
+            if bridge:
+                bridge.add_local_fdb_entry(new_mac)
+
```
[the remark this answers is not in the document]
alexandr-san4ez: > If the bridge is disabled/re-enabled, kernel FDB state can be purged

I couldn't confirm this:
```
vyos@vyos4# sudo bridge fdb show dev br0 | grep 00:aa:aa:aa:aa:aa
00:aa:aa:aa:aa:aa self permanent
00:aa:aa:aa:aa:aa vlan 100 master br0 permanent
00:aa:aa:aa:aa:aa vlan 1 master br0 permanent
00:aa:aa:aa:aa:aa master br0 permanent
[edit]
vyos@vyos4# set interfaces bridge br0 disable
[edit]
vyos@vyos4# commit
[edit]
vyos@vyos4# del interfaces bridge br0 disable
[edit]
vyos@vyos4# commit
[edit]
vyos@vyos4# sudo bridge fdb show dev br0 | grep 00:aa:aa:aa:aa:aa
00:aa:aa:aa:aa:aa self permanent
00:aa:aa:aa:aa:aa vlan 100 master br0 permanent
00:aa:aa:aa:aa:aa vlan 1 master br0 permanent
00:aa:aa:aa:aa:aa master br0 permanent
[edit]
```

smoketest/scripts/cli/test_interfaces_pseudo-ethernet.py
```diff
@@ -28,21 +30,64 @@ def setUpClass(cls):
...
+
+        self.cli_set(base_bridge_path + ['enable-vlan'])
+        self.cli_set(
+            base_bridge_path + ['member', 'interface', test_eth, 'native-vlan', '100']
+        )
+        self.cli_set(base_bridge_path + ['member', 'interface', 'vxlan0'])
+        self.cli_set(base_bridge_path + ['vif', '100'])
+
+        mac_address = '00:aa:aa:aa:aa:aa'
+        self.cli_set(self._base_path + ['peth0', 'source-interface', 'br0.100'])
```
c-po: The string `peth0` is repeated 5 times and should be moved to a variable.

In addition to this duplication, you should instead of defining a single `peth0` interface iterate through `cls._interfaces` which is done for all other smoketests (see base class).

smoketest/scripts/cli/test_interfaces_pseudo-ethernet.py
```diff
@@ -28,21 +30,64 @@ def setUpClass(cls):
...
             for tmp in os.environ['TEST_ETH'].split():
                 cls._options.update({f'p{tmp}' : [f'source-interface {tmp}']})
+                cls._test_interfaces.append(tmp)
 
         else:
             for tmp in Section.interfaces('ethernet'):
                 if '.' in tmp:
                     continue
                 cls._options.update({f'p{tmp}' : [f'source-interface {tmp}']})
+                cls._test_interfaces.append(tmp)
```
c-po: This line is superfluous, as all the information is present in: `cls._interfaces` - see below.

smoketest/scripts/cli/test_interfaces_pseudo-ethernet.py
```diff
@@ -28,21 +30,64 @@ def setUpClass(cls):
         cls._base_path = ['interfaces', 'pseudo-ethernet']
 
         cls._options = {}
+        cls._test_interfaces = []
         # we need to filter out VLAN interfaces identified by a dot (.)
         # in their name - just in case!
         if 'TEST_ETH' in os.environ:
             for tmp in os.environ['TEST_ETH'].split():
                 cls._options.update({f'p{tmp}' : [f'source-interface {tmp}']})
+                cls._test_interfaces.append(tmp)
```
c-po: This line is superfluous, as all the information is present in: `cls._interfaces` - see below. Remove this line.

smoketest/scripts/cli/test_interfaces_pseudo-ethernet.py
```diff
@@ -28,21 +30,64 @@ def setUpClass(cls):
         cls._base_path = ['interfaces', 'pseudo-ethernet']
 
         cls._options = {}
+        cls._test_interfaces = []
```
c-po: Unneeded, please remove

src/conf_mode/interfaces_pseudo-ethernet.py
```diff
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import contextlib
```
c-po: Why do we need this? It would be the first use of it in the interfaces subsystem - and there is only a single occurance which could be written in a more effective way to spare this import.

src/conf_mode/interfaces_pseudo-ethernet.py
```diff
@@ -49,6 +52,17 @@ def get_config(config=None):
     base = ['interfaces', 'pseudo-ethernet']
     ifname, peth = get_interface_dict(conf, base)
 
+    # Get effective config as we need full dictionary per interface delete
```
c-po: Why do we need to read in the effective configuration? There is no reason to do so - all other interface implementations rather consult the source of truth - Linux Kernel - about state info. Please drop this and work with the Kernels FDP table

```python
>>> pprint.pprint(vyos.utils.network.get_bridge_fdb('br0'))
[{'flags': ['self'], 'mac': '33:33:00:00:00:01', 'state': 'permanent'},
 {'flags': ['self'], 'mac': '33:33:00:00:00:02', 'state': 'permanent'},
 {'flags': ['self'], 'mac': '33:33:ff:29:7e:92', 'state': 'permanent'},
 {'flags': ['self'], 'mac': '01:00:5e:00:00:01', 'state': 'permanent'},
 {'flags': ['self'], 'mac': '33:33:ff:00:00:00', 'state': 'permanent'},
 {'flags': [],
  'mac': '00:aa:aa:aa:aa:aa',
  'master': 'br0',
  'state': 'permanent',
  'vlan': 1},
 {'flags': [],
  'mac': '00:aa:aa:aa:aa:aa',
  'master': 'br0',
  'state': 'permanent'},
 {'flags': [],
  'mac': 'c6:11:aa:29:7e:92',
  'master': 'br0',
  'state': 'permanent',
  'vlan': 1},
 {'flags': [],
  'mac': 'c6:11:aa:29:7e:92',
  'master': 'br0',
  'state': 'permanent'}]
```

c-po (changes requested):

c-po:
> One issue I foresee is if someone disables the bridge interface, then the FDB entry will be purged. When they reenable the bridge interface, there isn't a mechanism to add it back. It'd only be added back if/when the `peth` interface is modified.

Could be done using proper dependencies, like if peth is a bridge member, add this peth instance as bridge dependency.

l0crian1:
> > One issue I foresee is if someone disables the bridge interface, then the FDB entry will be purged. When they reenable the bridge interface, there isn't a mechanism to add it back. It'd only be added back if/when the `peth` interface is modified.
> 
> Could be done using proper dependencies, like if peth is a bridge member, add this peth instance as bridge dependency.

That can definitely be done, but I think the suggestion @sever-sever made with making this a generic FDB entry mechanism makes the most sense. That's all that is really being done here. The ability to statically add FDB entries is missing from VyOS anyways, so you'd be adding utility beyond just `anycast-gateway`. If someone needed to statically define an entry for any other reason, they'd be able to do it. Syntax could be something like:

```
set interfaces bridge br0 fdb-entry 'aa:bb:cc:dd:ee:f0'
```

By attaching the entry directly to the bridge, `interfaces_bridge.py` can easily ensure that the system state and the config state stay in lockstep with each other.  FDB entries will generally be attached to the bridge, but the syntax could also let it define the interface within that bridge as well. Something like:

```
set interfaces bridge br0 fdb-entry 'aa:bb:cc:dd:ee:f0' parent 'eth0'
```
Basically, `anycast-gateway` would simply remain a concept rather than a config item. If you need `anycast-gateway`, the documentation would simply state:

1. Create a `peth` interface with the bridge as a parent.
2. Create a static `fdb-entry` for the `peth` interface's MAC address on that bridge.

src/conf_mode/interfaces_pseudo-ethernet.py
alexandr-san4ez: We should read the effective configuration to detect changing of `interfaces pseudo-ethernet pethN source-interface` or `interfaces pseudo-ethernet pethN mac`. I don't see any ability to get previous value using `vyos.utils.network.get_bridge_fdb`.

src/conf_mode/interfaces_pseudo-ethernet.py
alexandr-san4ez: I have fixed this in the recent changes.

src/conf_mode/interfaces_pseudo-ethernet.py
c-po: You get the current in-use mac and source-interface by:

```python
>>> from vyos.ifconfig import MACVLANIf
>>> iface = MACVLANIf('peth0')
>>> iface.get_mac()
'00:aa:00:00:00:01'
```

and

```python
>>> from vyos.utils.network import get_interface_config
>>> get_interface_config('peth0')['link']
'eth0'
```

For simplicity you should simply extend `vyos.ifconfig.macvlan.MACVLANIf` with a `get_source_interface()` method. Thus you do not need to read anything from the effective configuration and can solely rely on the operating system itself. This also makes debugging easier if you change something hardcore on the OS CLI, and the scripts will still account for this during your debug session.

In addition ditching the effective config code path would relax the amount of added lines to this module which in th elong run will ease maintenance.

alexandr-san4ez: @c-po thanks for your rather detailed feedback, I have updated the PR to take this approach into account.

src/conf_mode/interfaces_pseudo-ethernet.py
```diff
@@ -82,12 +113,75 @@ def verify(peth):
...
+            bridge = _get_bridge_by_source(new_source)
+            if bridge:
+                bridge.add_local_fdb_entry(new_mac)
+
+
 def apply(peth):
+    # Obtain current MAC address and source interface if pseudo-ethernet exists.
+    # It is neccessary to apply anycast gateway.
+    current_mac = current_source = None
+    if 'deleted' not in peth:
```
c-po: It’s a strong improvement over the previous implementation, especially removing the effective-config path.

I still see one design concern before merge:

At lines 177 and 193, we check `if 'deleted' not in peth:` twice. The block around line 177 appears to only gather interface state, while line 193 is where behavior changes happen. Could these be consolidated?

Related to that: if `p.remove()` is called at line 191, I’d expect `remove()` in `MACVLANIf` to handle anycast-gateway FDB cleanup as part of interface teardown. Similarly, anycast-gateway activation/deactivation logic seems better owned inside `MACVLANIf().update()`, since that class already has the required internal state and lifecycle hooks.

If we move this logic into `MACVLANIf,` we can likely remove the external delta/workaround checks around lines 155 and 166 and make the flow easier to reason about.

Overall: very good direction - once this design point is addressed, I think it’ll be in great shape.

c-po (changes requested):

src/conf_mode/interfaces_pseudo-ethernet.py
alexandr-san4ez: I have updated the PR and the latest version looks much better. Thanks for your comment.

c-po (approved):
All requested changes adressed

natali-rs1985:
Maybe we need to add a dependency here `interfaces_bridge -> interfaces_pseudo-ethernet`?
```
set interfaces bridge br0 enable-vlan
set interfaces bridge br0 member interface eth2 native-vlan '100'
set interfaces bridge br0 member interface eth3 native-vlan '200'
set interfaces bridge br0 member interface vxlan0
set interfaces bridge br0 vif 100
set interfaces pseudo-ethernet peth100 address '192.168.1.1/24'
set interfaces pseudo-ethernet peth100 ip enable-arp-accept
set interfaces pseudo-ethernet peth100 mac '00:aa:aa:aa:aa:aa'
set interfaces pseudo-ethernet peth100 source-interface 'br0.100'
set interfaces pseudo-ethernet peth100 anycast-gateway
commit

vyos@vyos# set interfaces bridge br0 vif 200
[edit]
vyos@vyos# commit
[edit]
vyos@vyos# set interfaces pseudo-ethernet peth100 source-interface 'br0.200'
[edit]
vyos@vyos# commit
[edit]
vyos@vyos# sudo bridge fdb show dev br0 | grep 00:aa:aa:aa:aa:aa
00:aa:aa:aa:aa:aa self permanent
00:aa:aa:aa:aa:aa vlan 200 master br0 permanent
00:aa:aa:aa:aa:aa vlan 100 master br0 permanent
00:aa:aa:aa:aa:aa vlan 1 master br0 permanent
00:aa:aa:aa:aa:aa master br0 permanent
[edit]
```
For now looks good.
Then delete vlan 100 from bridge:
```
vyos@vyos# del interfaces bridge br0 vif 100
[edit]
vyos@vyos# commit
[edit]
vyos@vyos# sudo bridge fdb show dev br0 | grep 00:aa:aa:aa:aa:aa
00:aa:aa:aa:aa:aa vlan 200 master br0 permanent
00:aa:aa:aa:aa:aa vlan 100 master br0 permanent
00:aa:aa:aa:aa:aa vlan 1 master br0 permanent
00:aa:aa:aa:aa:aa master br0 permanent
[edit]
```
vlan 100 is still in the FDB

alexandr-san4ez:
@natali-rs1985 I fixed this by updating the BridgeIf class because I think using dependencies for this approach is overkill:

```python
# add VLAN interfaces to local 'parent' bridge to allow forwarding
if 'enable_vlan' in config:
    fdb_entries = self.get_fdb_entries()

    for vlan in config.get('vif_remove', {}):
        # Remove old VLANs from the bridge
        cmd = f'bridge vlan del dev {self.ifname} vid {vlan} self'
        self._cmd(cmd)

        # After deleting vif, the FDB entry for that VLAN can linger.
        # We should remove it manually:
        for entry in fdb_entries:
            mac = entry.get('mac')
            entry_vlan = entry.get('vlan')
            if entry_vlan and mac and str(entry_vlan) == vlan:
                cmd = f'bridge fdb del {mac} dev {self.ifname} vlan {vlan}'
                self._cmd(cmd)
```

natali-rs1985:
@alexandr-san4ez it fails now with OSError:
```
vyos@vyos# sudo systemctl restart vyos-configd
[edit]
vyos@vyos# set interfaces bridge br0 enable-vlan
[edit]
vyos@vyos# set interfaces bridge br0 member interface eth2 native-vlan '100'
[edit]
vyos@vyos# set interfaces bridge br0 member interface eth3 native-vlan '200'
[edit]
vyos@vyos# set interfaces bridge br0 member interface vxlan0
[edit]
vyos@vyos# set interfaces bridge br0 vif 100
[edit]
vyos@vyos# set interfaces pseudo-ethernet peth100 address '192.168.1.1/24'
[edit]
vyos@vyos# set interfaces pseudo-ethernet peth100 ip enable-arp-accept
[edit]
vyos@vyos# set interfaces pseudo-ethernet peth100 mac '00:aa:aa:aa:aa:aa'
[edit]
vyos@vyos# set interfaces pseudo-ethernet peth100 source-interface 'br0.100'
[edit]
vyos@vyos# set interfaces pseudo-ethernet peth100 anycast-gateway
[edit]
vyos@vyos# commit
[edit]
vyos@vyos# sudo bridge fdb show dev br0 | grep 00:aa:aa:aa:aa:aa
00:aa:aa:aa:aa:aa self permanent
00:aa:aa:aa:aa:aa vlan 100 master br0 permanent
00:aa:aa:aa:aa:aa vlan 1 master br0 permanent
00:aa:aa:aa:aa:aa master br0 permanent
[edit]
vyos@vyos# set interfaces bridge br0 vif 200
[edit]
vyos@vyos# commit
[edit]
vyos@vyos# sudo bridge fdb show dev br0 | grep 00:aa:aa:aa:aa:aa
00:aa:aa:aa:aa:aa self permanent
00:aa:aa:aa:aa:aa vlan 100 master br0 permanent
00:aa:aa:aa:aa:aa vlan 1 master br0 permanent
00:aa:aa:aa:aa:aa master br0 permanent
[edit]
vyos@vyos# set interfaces pseudo-ethernet peth100 source-interface 'br0.200'
[edit]
vyos@vyos# commit
[edit]
vyos@vyos# sudo bridge fdb show dev br0 | grep 00:aa:aa:aa:aa:aa
00:aa:aa:aa:aa:aa self permanent
00:aa:aa:aa:aa:aa vlan 200 master br0 permanent
00:aa:aa:aa:aa:aa vlan 100 master br0 permanent
00:aa:aa:aa:aa:aa vlan 1 master br0 permanent
00:aa:aa:aa:aa:aa master br0 permanent
[edit]
vyos@vyos# del interfaces bridge br0 vif 100
[edit]
vyos@vyos# commit
[ interfaces bridge br0 ]
  File "/usr/libexec/vyos/services/vyos-configd", line 157, in run_script
    script.apply(c)
  File "/usr/libexec/vyos/conf_mode/interfaces_bridge.py", line 220, in apply
    br.update(bridge)
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/bridge.py", line 385, in update
    self._cmd(cmd)
  File "/usr/lib/python3/dist-packages/vyos/ifconfig/control.py", line 66, in _cmd
    return cmd(command, self.debug, env=env)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/vyos/utils/process.py", line 208, in cmd
    raise OSError(code, feedback)
OSError: [Errno 255] failed to run command: None bridge fdb del f6:4c:6d:8d:5e:5f dev br0 vlan 100
returned: 
exit code: 255

[[interfaces bridge br0]] failed
Commit failed
[edit]
vyos@vyos# 
```
Maybe the problem is with the ordering (VLAN deleted before FDB entry)

alexandr-san4ez:
@natali-rs1985 check again please. The reason was in deleting not existing FDB record (twice).

natali-rs1985 (approved):
@alexandr-san4ez Now it works correct. My concerns were addressed, thank you! Have no objections!

PR #5157, opened 2026-05-01
geoip: T8590: fix initialization failure and set clobbering on boot and commit

xTITUSMAXIMUSX:
## Change summary

Three related bugs prevented GeoIP nftables sets from being populated correctly at boot and when incrementally modifying firewall or policy route rules.

1. geoip_updated() always returned False

   node_changed() returns a flat list of changed key name strings (e.g. ['ipv4', 'rule', 'geoip', 'country_code']).  The prior code passed that list to dict_search_recursive(), which searches for a key inside a dict and therefore never matches anything in a plain list. geoip_updated() consequently always returned False, meaning geoip_update() was never triggered by an incremental add or change of a GeoIP rule — only by explicit "update geoip" or by the fallback path when geoip_refresh() fails.

   Fix: replace the dict_search_recursive loop with a direct membership
   test: 'geoip' in changes.

2. GeoIP block executed unconditionally, poisoning the boot sequence

   geoip_sets() always returns {'name': [], 'ipv6_name': []}.  A non-empty dict is truthy in Python regardless of whether its values are empty lists, so the guards "if geoip_sets:" and "if 'name' in geoip_sets:" were always True.

   On boot, when policy_route.py is invoked as a dependent of firewall.py (triggered by group_resync), it entered the GeoIP block even with no policy route GeoIP rules configured.  Because /run/nftables-geoip.conf did not yet exist, geoip_refresh() returned False, and geoip_update(policy=policy) was called with an empty policy.  This created /run/nftables-geoip.conf containing only empty table stubs.  When firewall.py subsequently called geoip_refresh(), the file existed and nft loaded it successfully — so the geoip_update(firewall) call was never reached and the firewall GeoIP sets were left empty for the entire uptime of the router.

   Fix: check the actual list contents instead of the container dict: if geoip_sets['name'] or geoip_sets['ipv6_name'].

3. geoip_update() clobbered the other caller's sets

   geoip_update() renders /run/nftables-geoip.conf from both firewall_sets and policy_sets in a single pass.  When called with only one argument (as firewall.py and policy_route.py each do), the other argument defaulted to None and that half of the file was rendered empty, erasing whatever the other script had written.  The geoip-update helper (run by "update geoip" and the weekly cron) was unaffected because it always passes both arguments, which masked this bug in normal operation.

   Fix: when either argument is absent, read the missing config from the live Config session before building the set tables, so that every invocation writes the complete combined firewall + policy file.

## Types of changes

## Related Task(s)

https://vyos.dev/T8590

## Checklist:

### DIFF
--- a/python/vyos/geoip.py
+++ b/python/vyos/geoip.py
@@ -217,6 +217,9 @@ def geoip_update(firewall=None, policy=None):
 
         if firewall:
             for codes, path in dict_search_recursive(firewall, 'country_code'):
+                if path[0] == 'policy':
+                    continue
+
                 version = 6 if path[0] == 'ipv6' else 4
                 vprefix = '6' if version == 6 else ''
                 set_name = f'GEOIP_CC{vprefix}_{path[1]}_{path[2]}_{path[4]}'
@@ -224,6 +227,9 @@ def geoip_update(firewall=None, policy=None):
 
         if policy:
             for codes, path in dict_search_recursive(policy, 'country_code'):
+                if path[0] == 'firewall':
+                    continue
+
                 version = 6 if path[0] == 'route6' else 4
                 vprefix = '6' if version == 6 else ''
                 set_name = f'GEOIP_CC{vprefix}_{path[0]}_{path[1]}_{path[3]}'
--- a/src/conf_mode/firewall.py
+++ b/src/conf_mode/firewall.py
@@ -22,8 +22,8 @@ from glob import glob
 from sys import exit
 from vyos.base import Warning
 from vyos.config import Config
-from vyos.configdict import is_node_changed, node_changed
-from vyos.configdiff import Diff
+from vyos.configdict import is_node_changed
+from vyos.configdiff import Diff, get_config_diff
 from vyos.configdep import set_dependents, call_dependents
 from vyos.configverify import verify_interface_exists
 from vyos.ethtool import Ethtool
@@ -92,17 +92,12 @@ def geoip_sets(firewall):
     return out
 
 def geoip_updated(conf):
-    changes = node_changed(conf, ['firewall'],
-                                 key_mangling=('-', '_'),
-                                 recursive=True,
-                                 expand_nodes=Diff.ADD | Diff.DELETE)
-    updated = False
-
-    for _, path in dict_search_recursive(changes, 'geoip'):
-        updated = True
-        break
-
-    return updated
+    D = get_config_diff(conf, key_mangling=('-', '_'))
+    diff = D.get_child_nodes_diff(['firewall'],
+                                  expand_nodes=Diff.ADD | Diff.DELETE,
+                                  recursive=True)
+    return any(any(dict_search_recursive(diff.get(section, {}), 'geoip'))
+               for section in ('add', 'delete'))
 
 def get_config(config=None):
     if config:
@@ -124,6 +119,9 @@ def get_config(config=None):
 
     firewall['geoip_sets'] = geoip_sets(firewall)
     firewall['geoip_updated'] = geoip_updated(conf)
+    firewall['policy'] = conf.get_config_dict(
+        ['policy'], key_mangling=('-', '_'),
+        get_first_key=True, no_tag_node_value_mangle=True)
 
     fqdn_config_parse(firewall, 'firewall')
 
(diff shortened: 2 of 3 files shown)


src/conf_mode/firewall.py
```diff
@@ -92,17 +92,12 @@ def geoip_sets(firewall):
...
-                                 recursive=True,
-                                 expand_nodes=Diff.ADD | Diff.DELETE)
-    updated = False
-
-    for _, path in dict_search_recursive(changes, 'geoip'):
-        updated = True
-        break
-
-    return updated
+    D = get_config_diff(conf, key_mangling=('-', '_'))
```
c-po, line 95: What about the internal `is_node_changed()`  helper?

c-po, line 95: What about the internal `is_node_changed()`  helper?

sarthurdev: I don't think `is_node_changed()` is sufficient here. We need to recursively see if any `geoip` node is changed under all firewall/policy rules to determine to re-apply.

python/vyos/geoip.py
```diff
@@ -203,6 +203,22 @@ def geoip_update(firewall=None, policy=None):
             print("Script is already running")
             return False
 
+        # Always resolve both configs so the combined file reflects all sets.
+        # Whichever caller omits an argument is not "absent" — it just wasn't
+        # the one that changed.  Reading the live config here avoids clobbering
+        # the other script's sets on every incremental update or boot.
+        if firewall is None or policy is None:
```
sarthurdev: I don't think its a good idea to have config dict loading from within the vyos module.

IMO this is best handled within `get_config()` of the respective conf_script and passed from there into the `geoip_update()` function. (Though I wonder if this may cause issues with `dict_search_recursive()` against the outer config dict)

Would appreciate input @jestabro @c-po if you agree/disagree.

sarthurdev (changes requested):
Thanks for working on these issues. The PR looks good except from the one point raised, also just finishing running it through a few tests on a local image.

python/vyos/geoip.py
jestabro: Using the global config within the module also breaks the sequential processing of the commit priority queue, notably as regards error management.

It would seem preferable in this case to distinguish the cases 'firewall' and 'policy', even with a simple flag to the function `geoip_update` --- since the cases operate on separate tables, they can still use the same template, just don't pass an empty dict for 'policy' in the case 'firewall' and conversely.

xTITUSMAXIMUSX: If I am understanding both of you correctly we would need to do something like this.

src/conf_node/firewall.py
```
line: 122
    firewall['policy'] = conf.get_config_dict(
        ['policy'], key_mangling=('-', '_'),
        get_first_key=True, no_tag_node_value_mangle=True)

line: 743
    geoip_update(firewall=firewall, policy=firewall.get('policy', {}))
```

src/conf_node/policy_route.py
```
line: 91
    policy['firewall'] = conf.get_config_dict(
        ['firewall'], key_mangling=('-', '_'),
        no_tag_node_value_mangle=True, get_first_key=True)

line: 246
    geoip_update(firewall=policy.get('firewall', {}), policy=policy)
```

python/vyos/geoip.py
```
Remove lines 206-221
        # Always resolve both configs so the combined file reflects all sets.
        # Whichever caller omits an argument is not "absent" — it just wasn't
        # the one that changed.  Reading the live config here avoids clobbering
        # the other script's sets on every incremental update or boot.
        if firewall is None or policy is None:
            conf = Config()
            if firewall is None:
                firewall = conf.get_config_dict(
                    ['firewall'], key_mangling=('-', '_'),
                    no_tag_node_value_mangle=True, get_first_key=True)
            if policy is None:
                policy = conf.get_config_dict(
                    ['policy'], key_mangling=('-', '_'),
                    get_first_key=True, no_tag_node_value_mangle=True)
```

sarthurdev: Essentially yes, but the `.get('key', default)` can be simply `firewall['policy']` as it's guaranteed to exist and `get_config_dict` will always return at least an empty dict.

xTITUSMAXIMUSX: I have applied your changes. Now waiting on the workflow to finish.

c-po (commented):
@jestabro @sarthurdev out of curiosity. When both policy_route and firewall changes, the GeoIP db is regenerated two times? One time for firewall and one time for policy_route?

jestabro:
@c-po Yes, and thus the suggestion to separate the functions/dicts above. However, it may be more important to merge this fix for the fundamentally broken situation (at the risk of redundancy), and then consider a proper refactoring <-- there may be further subtleties involved.

jestabro:
@sarthurdev what are your thoughts ?

jestabro:
@sarthurdev I've discussed this with @c-po , and the consensus is to merge this solution when you are satisfied. We will revisit refactoring to avoid redundant generation of rules. Are you satisfied with your testing and the plan ?

sarthurdev:
Fixed issue with recursive search with imported dict configs resulting in:
```
vyos@vyos# commit
[ policy route test ]
Updating GeoIP. Please wait...
Error: GeoIP failed to update firewall and/or policy
```

nft config:
```
/run/nftables-geoip.conf:19:26-52: Error: No such file or directory
flush set ip vyos_mangle GEOIP_CC_firewall_ipv4_test
```

sarthurdev (approved):
Requested changes addressed, included fix for dict search issue, tested locally.

alexk37:
@coderabbitai review

jestabro (approved):
Changes look good. I trust @sarthurdev testing.

PR #5161, opened 2026-05-04
dhcp-relay: T7879: verify listen and upstream interfaces exist

natali-rs1985:
## Types of changes

## Related Task(s)

* https://vyos.dev/T7879

```
vyos@r10# run show interfaces 
Interface    IP Address          MAC                VRF        MTU  S/L    Description
-----------  ------------------  -----------------  -------  -----  -----  -------------
eth0         -                   0c:ff:ea:09:00:00  default   1500  u/D
eth1         -                   0c:ff:ea:09:00:01  default   1500  u/D
eth2         10.217.32.103/24    0c:ff:ea:09:00:02  default   1500  u/u
eth3         -                   0c:ff:ea:09:00:03  default   1500  A/D
lo           127.0.0.1/8         00:00:00:00:00:00  default  65536  u/u
             192.168.255.255/32
             ::1/128
vyos@r10# set service dhcp-relay upstream-interface eth5
[edit]
vyos@r10# set service dhcp-relay listen-interface eth5
[edit]
vyos@r10# commit
[ service dhcp-relay ]
Interface "eth5" does not exist!
[[service dhcp-relay]] failed
Commit failed
[edit]
```
## Checklist:

### DIFF
--- a/src/conf_mode/service_dhcp-relay.py
+++ b/src/conf_mode/service_dhcp-relay.py
@@ -20,6 +20,7 @@ from sys import exit
 
 from vyos.base import Warning
 from vyos.config import Config
+from vyos.configverify import verify_interface_exists
 from vyos.template import render
 from vyos.base import Warning
 from vyos.utils.process import call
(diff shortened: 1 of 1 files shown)


c-po (approved):
Check that DHCP relay interfaces configured on CLI actually exist.

Unfortunately the current code has no general verify() stage before proceeding with commit - this means, that you can delete a relay interface once the relay service is in place, save the config and reboot -> this will cause a configuration error. Looping in @jestabro

c-po (changes requested):
Check that DHCP relay interfaces configured on CLI actually exist.

Unfortunately the current code has no general verify() stage before proceeding with commit - this means, that you can delete a relay interface once the relay service is in place, save the config and reboot - this will cause a configuration error. Looping in @jestabro 

Maybe make this a Warning instead of an error?

sarthurdev (commented):
Similar to as @c-po wondered. Would a migration from a config that had valid CLI (but invalid interface) now cause the whole config to break on boot-commit?

natali-rs1985:
@c-po @sarthurdev I agree with your concern. Changed to Warning to not break existing configs. But the issue will still stand and service will fail to start with the non-existing interface

sarthurdev (approved):
Raise warning for invalid dhcp-relay interface

c-po:
> @c-po @sarthurdev I agree with your concern. Changed to Warning to not break existing configs. But the issue will still stand and service will fail to start with the non-existing interface

This can only be handled with the new backend from @jestabro so we need to revisit all the cases in the future.

c-po (approved):
Requested changes adressed

PR #5166, opened 2026-05-04
ntp: T8601: add local stratum option

anderbak:
## Summary
- Add `service ntp local-stratum <1-15>`.
- Render Chrony `local stratum <value>` in `/run/chrony/chrony.conf`.
- Allow `service ntp` to commit without upstream `server` entries when `local-stratum` is configured.

## Testing
- `git diff --check`
- `python -m compileall src\conf_mode\service_ntp.py smoketest\scripts\cli\test_service_ntp.py`
- Direct Chrony template render assertion with `local_stratum=10`
- `powershell -ExecutionPolicy Bypass -File C:\Users\Brian\Downloads\VyOS\scripts\Invoke-VyOS1xTests.ps1 -Repo C:\Users\Brian\Downloads\VyOS\src\vyos-1x -Bash "apt-get update >/dev/null && apt-get install -y procps make >/dev/null && sysctl -w net.ipv4.conf.lo.forwarding=1 >/dev/null || true; make libvyosconfig interface_definitions test"`

Task: https://vyos.dev/T8601

### DIFF
--- a/smoketest/scripts/cli/test_service_ntp.py
+++ b/smoketest/scripts/cli/test_service_ntp.py
@@ -79,6 +79,18 @@ class TestSystemNTP(VyOSUnitTestSHIM.TestCase):
         for pool in pools:
             self.assertIn(f'pool {pool} iburst', config)
 
+    def test_local_stratum_without_upstream_server(self):
+        stratum = '10'
+        network = '192.0.2.0/24'
+
+        self.cli_set(base_path + ['local-stratum', stratum])
+        self.cli_set(base_path + ['allow-client', 'address', network])
+        self.cli_commit()
+
+        config = cmd(f'sudo cat {NTP_CONF}')
+        self.assertIn(f'local stratum {stratum}', config)
+        self.assertIn(f'allow {network}', config)
+
     def test_clients(self):
         # Test the allowed-networks statement
         listen_address = ['127.0.0.1', '::1']
@@ -88,14 +100,6 @@ class TestSystemNTP(VyOSUnitTestSHIM.TestCase):
         networks = ['192.0.2.0/24', '2001:db8:1000::/64', '100.64.0.0', '2001:db8::ffff']
         for network in networks:
             self.cli_set(base_path + ['allow-client', 'address', network])
-
-        # Verify "NTP server not configured" verify() statement
-        with self.assertRaises(ConfigSessionError):
-            self.cli_commit()
-
-        servers = ['192.0.2.1', '192.0.2.2']
-        for server in servers:
-            self.cli_set(base_path + ['server', server])
 
         self.cli_commit()
 
--- a/smoketest/scripts/cli/test_service_ntp.py
+++ b/smoketest/scripts/cli/test_service_ntp.py
@@ -19,6 +19,7 @@ import unittest
 from base_vyostest_shim import VyOSUnitTestSHIM
 
 from vyos.configsession import ConfigSessionError
+from vyos.utils.file import read_file
 from vyos.utils.process import cmd
 from vyos.utils.process import process_named_running
 from vyos.xml_ref import default_value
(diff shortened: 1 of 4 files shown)


src/conf_mode/service_ntp.py
```diff
@@ -65,8 +65,8 @@ def verify(ntp):
     if not ntp:
         return None
 
-    if 'server' not in ntp:
-        raise ConfigError('NTP server not configured')
+    if 'server' not in ntp and 'local_stratum' not in ntp:
```
c-po: This will break existing configurations as existing configs do not have `local-stratum` defined. Why is `local-stratum` a mandatory CLI option?

c-po (changes requested):

src/conf_mode/service_ntp.py
anderbak: Thanks, that wording made the issue clear. I updated the verifier error message so `local-stratum` is not presented as a mandatory setting.

The behavior is still intentionally limited to two valid NTP source cases: existing configs with `server` continue to work unchanged, and a serverless config is only accepted when `local-stratum` is explicitly configured for the standalone/local-clock use case requested in T8601. The smoketest name now calls out that serverless case directly.

Validation run:
- `python3 -m py_compile src/conf_mode/service_ntp.py smoketest/scripts/cli/test_service_ntp.py`
- direct Chrony template render check for `local stratum 10`
- `git diff --check`

dmbaturin: I agree it shouldn't be mandatory, since it's not required for NTP operation.

anderbak: Thanks, agreed. I removed the verifier source gate entirely, so `service ntp` no longer requires either an upstream `server` or `local-stratum` to commit. `local-stratum` is now just an optional Chrony directive that renders `local stratum <value>` when explicitly configured.

I also updated the smoketest so the client/listen-address case commits without an upstream server instead of expecting the old `NTP server not configured` error.

Validation run:
- `python -m py_compile src/conf_mode/service_ntp.py smoketest/scripts/cli/test_service_ntp.py`
- transcluded `interface-definitions/service_ntp.xml.in` parse check
- Chrony template block presence check
- `git diff --check`

I could not run the containerized interface build locally because Docker Desktop is not running on this workstation.

sever-sever (approved):
Add NTP local stratum option

smoketest/scripts/cli/test_service_ntp.py
```diff
@@ -79,6 +79,18 @@ def test_base_options(self):
...
 
+    def test_local_stratum_without_upstream_server(self):
+        stratum = '10'
+        network = '192.0.2.0/24'
+
+        self.cli_set(base_path + ['local-stratum', stratum])
+        self.cli_set(base_path + ['allow-client', 'address', network])
+        self.cli_commit()
+
+        config = cmd(f'sudo cat {NTP_CONF}')
```
c-po: Use `vyos.utils.file.read_file(..., sudo=True)` instead https://github.com/vyos/vyos-1x/blob/current/python/vyos/utils/file.py#L34

c-po (changes requested):

natali-rs1985 (approved):
Tested locally. Works as expected.

c-po (approved):
Fixed manually..


```
cpo@LR1.wue3:~$ /usr/libexec/vyos/tests/smoke/cli/test_service_ntp.py
test_base_options (__main__.TestSystemNTP.test_base_options) ... ok
test_clients (__main__.TestSystemNTP.test_clients) ... ok
test_interface (__main__.TestSystemNTP.test_interface) ... ok
test_interleave_option (__main__.TestSystemNTP.test_interleave_option) ... ok
test_leap_seconds (__main__.TestSystemNTP.test_leap_seconds) ... ok
test_local_stratum_without_upstream_server (__main__.TestSystemNTP.test_local_stratum_without_upstream_server) ... ok
test_offload_timestamp_default (__main__.TestSystemNTP.test_offload_timestamp_default) ... ok
test_ptp_transport (__main__.TestSystemNTP.test_ptp_transport) ... ok
test_vrf (__main__.TestSystemNTP.test_vrf) ... ok

Ran 9 tests in 22.851s

OK
```

PR #5173, opened 2026-05-06
T8467: fix obscured definition of completion help in XML

jestabro:
## Change summary

The result here is rather trivial, however, it is necessary for implementation of overriding help elements in interface definitions (https://vyos.dev/T8360); it is recommended as correcting a loss of information, even if the case itself is of minor importance.

Inclusion of common XML definitions in assembling the final (valid) XML interface definitions is indispensable, however, one must exercise care in not obscuring definitions: the script `build-command-templates` will only populate a `node.def` file on the first occurrence of an element.

Thankfully, the test script included below only finds one case of loss of information, contained in the first commit. The second commit simply removes a redundant include, discovered by the test script.

Running the test script after merging these fixes reveals two remaining cases:
(1)  a non-problematic 'collision' in firewall.xml. A collision is defined as two XML blocks at the same path containing properties elements containing more than just the help entry (which is often repeated at subsequent path definitions, when defining new children). In the case of firewall.xml it is due to multiple inclusions of the same block, under different names: `interface-definitions/include/firewall/action.xml.i` and `interface-definitions/include/firewall/action-forward.xml.i` have the same content.
(2) a non-trivial difference in help text at the same path in `build/interface-definitions/vpp.xml`. This is precisely a case to be addressed in https://vyos.dev/T8360, for which this PR is a prerequisite.

For reference, the test script used is attached.
[check-redundant-properties.py](https://github.com/user-attachments/files/27450298/check-redundant-properties.py)

## Types of changes

Without this PR:
```
vyos@vyos# set system conntrack ignore ipv4 rule 137 protocol <tab>
Possible completions:
   <text>               Protocol name
   ah                   
   ax.25                
   dccp                 
   ...
```

With this PR:
```
vyos@vyos# set system conntrack ignore ipv4 rule 137 protocol <tab>
Possible completions:
   all                  All IP protocols
   tcp_udp              Both TCP and UDP
   <0-255>              IP protocol number
   <protocol>           IP protocol name
   !<protocol>          IP protocol name
   ah                   
   ax.25                
   dccp
   ...

```
## Checklist:

### DIFF
--- a/interface-definitions/include/bgp/protocol-common-config.xml.i
+++ b/interface-definitions/include/bgp/protocol-common-config.xml.i
@@ -1719,7 +1719,6 @@
     #include <include/bgp/neighbor-disable-connected-check.xml.i>
     #include <include/bgp/neighbor-ebgp-multihop.xml.i>
     #include <include/bgp/neighbor-graceful-restart.xml.i>
-    #include <include/bgp/neighbor-graceful-restart.xml.i>
     #include <include/bgp/neighbor-local-as.xml.i>
     #include <include/bgp/neighbor-local-role.xml.i>
     #include <include/bgp/neighbor-override-capability.xml.i>
--- a/interface-definitions/system_conntrack.xml.in
+++ b/interface-definitions/system_conntrack.xml.in
@@ -84,7 +84,6 @@
                           </completionHelp>
                         </properties>
                       </leafNode>
-                      #include <include/ip-protocol.xml.i>
                       <leafNode name="protocol">
                         <properties>
                           <help>Protocol to match (protocol name, number, or "all")</help>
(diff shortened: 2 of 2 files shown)


sarthurdev (approved):
Deduplicate includes and obscured protocol completion

c-po (approved):
Deduplicate XML building block inclusion.

Could we probably add [check-redundant-properties.py](https://github.com/user-attachments/files/27450298/check-redundant-properties.py) to the build stage that during package build this is verified to not happen again?

jestabro:
I intend to add some form of that script to the build stage. We (or someone more familiar with the firewall definitions @sarthurdev ?) should clear up the confusion mentioned above regarding action/action-forward, so as to avoid a false positive ...

PR #5175, opened 2026-05-07
wireguard: T8509: Add fwmark ip rules for VRF-bound interfaces

natali-rs1985:
When a WireGuard interface has both fwmark and VRF configured, outgoing tunnel packets marked with the fwmark were not routed into the correct VRF routing table, causing them to hit the l3mdev unreachable rule instead. Add an ip rule at priority 1998 (after l3mdev at 1000 and before l3mdev unreachable at 2000) to route fwmark-tagged packets into the correct VRF routing table.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8509

```
set interfaces wireguard wg99 address '172.27.42.2/30'
set interfaces wireguard wg99 fwmark '101'
set interfaces wireguard wg99 peer CFBOM address '100.96.0.2'
set interfaces wireguard wg99 peer CFBOM allowed-ips '0.0.0.0/0'
set interfaces wireguard wg99 peer CFBOM port '10006'
set interfaces wireguard wg99 peer CFBOM public-key 'XXXXXXXXXX'
set interfaces wireguard wg99 port '49501'
set interfaces wireguard wg99 private-key 'XXXXXXXXXX'
set interfaces wireguard wg99 vrf 'TEST'
set vrf name TEST table '101'

vyos@r10# ip rule show
1000:	from all lookup [l3mdev-table]
1998:	from all fwmark 0x65 lookup TEST
2000:	from all lookup [l3mdev-table] unreachable
32765:	from all lookup local
32766:	from all lookup main
32767:	from all lookup default
[edit]

vyos@r10:~$ /usr/libexec/vyos/tests/smoke/cli/test_interfaces_wireguard.py -k test_wireguard_vrf_fwmark
test_wireguard_vrf_fwmark (__main__.WireGuardInterfaceTest.test_wireguard_vrf_fwmark) ... ok

Ran 1 test in 23.812s

OK
```

## Checklist:

### DIFF
--- a/src/conf_mode/interfaces_wireguard.py
+++ b/src/conf_mode/interfaces_wireguard.py
@@ -34,6 +34,7 @@ from vyos.configverify import verify_bond_bridge_member
 from vyos.ifconfig import WireGuardIf
 from vyos.utils.kernel import check_kmod
 from vyos.utils.network import check_port_availability
+from vyos.utils.network import get_vrf_tableid
 from vyos.utils.network import is_wireguard_key_pair
 from vyos.utils.process import call
 from vyos import ConfigError
@@ -74,6 +75,16 @@ def get_config(config=None):
             set_dependents('vxlan', conf, tmp)
         else:
             wireguard['is_source_interface'] = tmp
+
+    if is_node_changed(conf, base + [ifname, 'fwmark']) or is_node_changed(
+        conf, base + [ifname, 'vrf']
+    ):
+        wireguard['fwmark_vrf_changed'] = {}
+        prev = conf.get_config_dict(
+            base + [ifname], effective=True, key_mangling=('-', '_'), get_first_key=True
+        )
+        wireguard['prev_fwmark'] = prev.get('fwmark')
+        wireguard['prev_vrf'] = prev.get('vrf')
 
     return wireguard
 
@@ -153,6 +164,29 @@ def apply(wireguard):
         wg.remove()
     else:
         wg.update(wireguard)
+
+    # delete old fwmark-based ip rule if fwmark or VRF was changed
+    if 'fwmark_vrf_changed' in wireguard or 'deleted' in wireguard:
+        prev_fwmark = wireguard.get('prev_fwmark')
+        prev_vrf = wireguard.get('prev_vrf')
+        if prev_fwmark is not None and prev_vrf is not None:
+            table_id = get_vrf_tableid(prev_vrf)
+            if table_id is not None:
+                for afi in ['-4', '-6']:
+                    call(
+                        f'ip {afi} rule del pref 1998 fwmark {prev_fwmark} table {table_id}'
+                    )
+
+    # Add ip rule to route fwmark-marked WireGuard tunnel packets into the
+    # correct VRF routing table. This is required for VRF-bound WireGuard
+    # interfaces with fwmark set, so that outgoing encapsulated packets use the
+    # proper VRF routes (otherwise, they may be unroutable or use the main table).
+    if wireguard.get('fwmark', '0') != '0' and 'vrf' in wireguard:
+        table_id = get_vrf_tableid(wireguard['vrf'])
+        for afi in ['-4', '-6']:
+            call(
+                f'ip {afi} rule add pref 1998 fwmark {wireguard["fwmark"]} table {table_id}'
+            )
 
     domain_resolver_usage = '/run/use-vyos-domain-resolver-interfaces-wireguard-' + wireguard['ifname']
 
--- a/src/conf_mode/vrf.py
+++ b/src/conf_mode/vrf.py
@@ -265,6 +265,12 @@ def apply(vrf):
                 # Remove map element
                 cmd(f'nft {nft_del_element}')
 
+            # Remove all ip rules pointing to this VRF table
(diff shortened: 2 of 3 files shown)


c-po (commented):
Could you add a smoketest which validates proper fwmark rule installation and deletion?

src/conf_mode/interfaces_wireguard.py
```diff
@@ -75,6 +76,16 @@ def get_config(config=None):
...
 
+    if is_node_changed(conf, base + [ifname, 'fwmark']) or is_node_changed(
+        conf, base + [ifname, 'vrf']
+    ):
+        wireguard['fwmark_vrf_changed'] = {}
+        prev = conf.get_config_dict(
+            base + [ifname], effective=True, key_mangling=('-', '_'), get_first_key=True
+        )
+        wireguard['prev_fwmark'] = prev.get('fwmark')
+        wireguard['prev_vrf'] = prev.get('vrf')
```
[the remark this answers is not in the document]
natali-rs1985: Deleting the interface marks all existing child nodes as changed, so is_node_changed(..., 'fwmark') / is_node_changed(..., 'vrf') will be true when those nodes existed. That means prev_fwmark and prev_vrf are still captured on delete, and the cleanup path in apply() is covered.

natali-rs1985:
@c-po Smoketest was added

src/conf_mode/interfaces_wireguard.py
```diff
@@ -154,6 +165,29 @@ def apply(wireguard):
...
+    # delete old fwmark-based ip rule if fwmark or VRF was changed
+    if 'fwmark_vrf_changed' in wireguard or 'deleted' in wireguard:
+        prev_fwmark = wireguard.get('prev_fwmark')
+        prev_vrf = wireguard.get('prev_vrf')
+        if prev_fwmark is not None and prev_vrf is not None:
+            table_id = get_vrf_tableid(prev_vrf)
+            if table_id is not None:
+                for afi in ['-4', '-6']:
+                    call(
+                        f'ip {afi} rule del pref 1998 fwmark {prev_fwmark} table {table_id}'
```
sever-sever, line 177: There could be a situation when we have a policy `local-route` rule with the same number and similar logic.
```
set vrf name red table '123'
set interfaces wireguard wg1 fwmark '555'
set interfaces wireguard wg1 peer first allowed-ips '100.64.252.2/32'
set interfaces wireguard wg1 peer first public-key 'M4YtyVnA+U5dpqYvGpcT/rgbXEJ/S+Wic04e/2m0dzk='
set interfaces wireguard wg1 private-key 'iAvf5n1PcUz5H1afBLLAD6WCEUIZnKVZaMRWEemDjkc='
set interfaces wireguard wg1 vrf 'red'

set policy local-route rule 1998 fwmark '555'
set policy local-route rule 1998 set table '123'
set policy local-route rule 1998 source address '0.0.0.0/0'
commit

```
Delete the WireGuard interface, and we don't see any rules related to the table 123 (VRF red)
```
vyos@r14# delete interfaces wireguard 
[edit]
vyos@r14# commit
[edit]
vyos@r14# ip rule
1000:	from all lookup [l3mdev-table]
2000:	from all lookup [l3mdev-table] unreachable
32765:	from all lookup local
32766:	from all lookup main
32767:	from all lookup default
[edit]
vyos@r14# 

```
I wonder if we need to reserve/exclude some rule numbers from `policy local-route`

(Expected rules from the policy local-route)

natali-rs1985: Good point! @c-po @dmbaturin what do you think?

c-po: > I wonder if we need to reserve/exclude some rule numbers from policy local-route

That would end up in a breaking change unless rule numbers are migrated.

sever-sever: At least we can add a warning in the future for the `policy local-route` if necessary.
I'm OK with the current implementation.

sever-sever (approved):
Add iprules required when VRF is involved in the WireGuard interface.

src/conf_mode/vrf.py
```diff
@@ -265,6 +265,12 @@ def apply(vrf):
                 # Remove map element
                 cmd(f'nft {nft_del_element}')
 
+            # Remove all ip rules pointing to this VRF table
+            table_id = get_vrf_tableid(tmp)
+            for afi in ['-4', '-6']:
+                while call(f'ip {afi} rule del table {table_id}') == 0:
+                    pass
```
alexandr-san4ez, line 272: @natali-rs1985 What is the benefit of using a 'while' loop here to delete ip rules? Potentially it can lead to looping, I guess.

natali-rs1985: @alexandr-san4ez  The loop is needed because `ip rule del` removes only one matching rule per call, and multiple rules can point to the same table. It terminates naturally once non-zero return. To guard against the edge case you mention, I can add an upper bound — something like `for _ in range(100)` with a break on failure.

alexandr-san4ez: I get it, and it makes sense. Thanks for describe the approach.

alexandr-san4ez (approved):
I have tested it locally and everything works fine.

PR #5180, opened 2026-05-07
T8360: add XML preprocessor to allow overriding help element

jestabro:
## Change summary

As with element `defaultValue`, allow overriding the `help` element in document-order subsequent entries at a given path.
The use case is when one has included an XML block, but wants to rephrase the help text for the path, e.g.
https://github.com/vyos/vyos-1x/blob/current/interface-definitions/vpp.xml.in#L1196-L1199
```
  ...
  #include <include/vpp/acl_prefix.xml.i>
  <leafNode name="prefix">
    <properties>
      <help>Source IP prefix</help>
  ...
```

Since `node.def` files are populated on the first document-order occurrence, the current behavior is that the resulting path will show the help text of the `#include` file, namely `IP prefix`. Adding the script here allows overriding that entry.

## Types of changes

- [X] Other (please describe): Add preprocessor to allow override of help element.

Regarding the example mentioned above:

Without preprocessor:
```
vyos@vyos# set vpp acl mac tag-name test rule 137 <tab>
Possible completions:
   ...
   prefix               IP prefix
```

With preprocessor:
```
vyos@vyos# set vpp acl mac tag-name test rule 137 <tab>
Possible completions:
   ...
   prefix               Source IP prefix

```

Using a sanity check of temporarily adding `build/interface-definitions` to the local repo, one confirms that the only file modified by the `override-help` script is the existing example mentioned:

```
└──> git diff build/interface-definitions                  
diff --git a/build/interface-definitions/vpp.xml b/build/interface-definitions/vpp.xml
index 39ae39e87..6d6ec5e58 100644
--- a/build/interface-definitions/vpp.xml
+++ b/build/interface-definitions/vpp.xml
@@ -1879,8 +1879,8 @@
 <!-- include start from vpp/acl_prefix.xml.i -->
 <leafNode name="prefix">
   <properties>
-    <help>IP prefix</help>
-    <valueHelp>
+    <help>Source IP prefix</help>
+                      <valueHelp>
       <format>ipv4net</format>
       <description>IPv4 prefix</description>
     </valueHelp>
```

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ interface_definitions: libvyosconfig $(config_xml_obj)
 	rm -rf $(TMPL_DIR); mkdir -p $(TMPL_DIR)
 
 	$(CURDIR)/scripts/override-default $(BUILD_DIR)/interface-definitions
+	$(CURDIR)/scripts/override-help $(BUILD_DIR)/interface-definitions
 
 	find $(BUILD_DIR)/interface-definitions -type f -name "*.xml" | xargs -I {} $(CURDIR)/scripts/build-command-templates {} $(CURDIR)/schema/interface_definition.rng $(TMPL_DIR) || exit 1
 
--- a/scripts/override-help
+++ b/scripts/override-help
@@ -1,0 +1,115 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) VyOS Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or later as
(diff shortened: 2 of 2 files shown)


c-po (approved):
This is a big achievement and was missing for a long time. We finally can re-use more XML building blocks and then simply override the XML help string after moving this.

dmbaturin (approved):
I think it's a very welcome change and it will make a lot more building blocks reusable, where only help strings differ in different subtrees but the command structure is the same.

PR #5188, opened 2026-05-11
smoketests: T8845: missing base-class call to commit() in tearDownClass()

c-po:
## Change summary

The smoketest framework heavily relies on deriving common code paths. The tearDownClass() is called when a testcase finishes and testcases like protocols BGP delete CLI nodes before calling the base class.

The base-class never calls commit() thus the deletions are discarded instead of committed.

```
test_protocols_bgp.py
  @classmethod
  def tearDownClass(cls):
      cls.cli_delete(cls, ['policy', 'route-map'])
      cls.cli_delete(cls, ['policy', 'prefix-list'])
      cls.cli_delete(cls, ['policy', 'prefix-list6'])

      super(TestProtocolsBGP, cls).tearDownClass()
```

base class `base_vyostest_shim.py`

```
  @classmethod
  def tearDownClass(cls):
      # discard any pending changes which might caused a messed up config
      cls._session.discard()
```

## Types of changes

## Related Task(s)

* https://vyos.dev/T8845

All Smoketests passed

## Checklist:

### DIFF
--- a/smoketest/scripts/cli/base_vyostest_shim.py
+++ b/smoketest/scripts/cli/base_vyostest_shim.py
@@ -65,16 +65,19 @@ class VyOSUnitTestSHIM:
 
         @classmethod
         def tearDownClass(cls):
-            # discard any pending changes which might caused a messed up config
-            cls._session.discard()
-            # ... and restore the initial state
-            cls._session.migrate_and_load_config(save_config)
-
             try:
+                # commit pending changes done by derived tearDownClass()
+                # implementations like CLI cleanup
                 cls._session.commit()
             except (ConfigError, ConfigSessionError):
+                # discard any pending changes which might have failed, causing a
+                # messed up config
                 cls._session.discard()
                 cls.fail(cls)
+            finally:
+                # restore previous configuration before the test
+                cls._session.migrate_and_load_config(save_config)
+                cls._session.commit()
 
         def setUp(self):
             pass


sever-sever (approved):
Use `commit` again after discarding and loading the default config 
Smoketests

jestabro (approved):
Commit pending clean-up of config changes on tearDownClass.

smoketest/scripts/cli/base_vyostest_shim.py
```diff
@@ -65,15 +65,17 @@ def setUpClass(cls):
...
+                # commit pending changes done by derived tearDownClass()
+                # implementations like CLI cleanup
                 cls._session.commit()
             except (ConfigError, ConfigSessionError):
+                # discard any pending changes which might have failed, causing a
+                # messed up config
                 cls._session.discard()
+                # ... and restore the initial state
+                cls._session.migrate_and_load_config(save_config)
+                cls._session.commit()
```
hedrok, line 80: So now migration to saved config will be done only if commit failed?..
I'm concerned that if test forgets to clean up something this will contaminate the following tests...
Maybe it it possibe to compare saved config with current state and fail if they differ (with migration to saved config)?.. Or just migrate anyway just in case?..

hedrok (changes requested):

smoketest/scripts/cli/base_vyostest_shim.py
c-po: You are right - this can cause serious issues.

hedrok (changes requested):
I suggest using `finally`:
```python
        @classmethod
        def tearDownClass(cls):
            try:
                # commit pending changes done by derived tearDownClass()
                # implementations like CLI cleanup
                cls._session.commit()
            except (ConfigError, ConfigSessionError):
                # discard any pending changes which might have failed, causing a
                # messed up config
                cls._session.discard()
                cls.fail(cls)
            finally:
                # restore previous configuration before the test
                cls._session.migrate_and_load_config(save_config)
                cls._session.commit()
```
I've checked - `cls.fail(cls)` just raises exception, so `finally` is executed in this case too.
Main advantage is that even if `cls._session.commit()` raises any other exception, `finally` will be executed.

hedrok (approved):
Approve, thanks, all my concers are taken into account.

PR #5194, opened 2026-05-13
python: T8859: enforce positive integer min_keysize in verify_diffie_hellman_length

andamasov:
Tracked in Phorge [T8859](https://vyos.dev/T8859).

## Defect

`verify_diffie_hellman_length(file, min_keysize)` in `python/vyos/configverify.py` wrapped `str(min_keysize)` in `try/except`, but the later `int(min_keysize)` on the comparison line could still raise `ValueError` for non-numeric input, and the local `keysize` variable was computed but never used.

## Fix (redesigned per review)

Per [dmbaturin's review](https://github.com/vyos/vyos-1x/pull/5194#pullrequestreview-4363459966): the function has no production callers — its only consumers are its own unit tests in `src/tests/test_configverify.py` (verified by grep) — so the contract is free to change. An invalid `min_keysize` is a logic error in the calling script and now fails loudly instead of silently returning `False`:

- Signature is now `verify_diffie_hellman_length(file: str, min_keysize: int)`.
- Raises `TypeError` if `min_keysize` is not an integer at all (`bool` explicitly rejected).
- Raises `ValueError` if `min_keysize` is not a positive integer (`0` and negatives rejected).
- The silent `try/except` and the unused variable are gone; the bits comparison uses the argument directly.

Two adjacent hardenings in the same function, both surfaced by pre-PR CodeRabbit review:

- `openssl dhparam` is invoked with list arguments instead of an interpolated command string, so the file path never reaches a shell (`vyos.utils.process.popen` auto-selects `shell=True` for strings containing spaces).
- The bit-count regex is a raw string (`r'\d+\s+bit'`), removing the invalid-escape-sequence warning on modern Python.

## Behavior change

None for valid positive-integer inputs. Non-integer or non-positive `min_keysize` now raises `TypeError`/`ValueError` instead of returning `False` — intentional per review; no production code hits these paths.

## Test plan

- [x] Unit tests updated to pass integer keysizes and extended to cover both exception paths (`'1024'`/`True` → `TypeError`; `-512`/`0` → `ValueError`). All 4 pass locally, including the real `openssl dhparam` round-trip.
- [ ] CLI smoketests pass.

## Provenance

Found by Copilot during review of [vyos/vyos-1x#5074](https://github.com/vyos/vyos-1x/pull/5074) — the configverify changes were dropped from that PR per [dmbaturin's scope narrowing](https://github.com/vyos/vyos-1x/pull/5074#pullrequestreview-4281666376). This defect is independent of the broader configverify redesign concerns (MTU helpers taking defaults as args) and can be fixed in isolation.

### DIFF
--- a/python/vyos/configverify.py
+++ b/python/vyos/configverify.py
@@ -424,24 +424,29 @@ def verify_vlan_config(config):
             verify_mtu_ipv6(c_vlan)
 
 
-def verify_diffie_hellman_length(file, min_keysize):
-    """ Verify Diffie-Hellamn keypair length given via file. It must be greater
-    then or equal to min_keysize """
+def verify_diffie_hellman_length(file: str, min_keysize: int):
+    """ Verify Diffie-Hellman keypair length given via file. It must be greater
+    than or equal to min_keysize.
+
+    Raises TypeError if min_keysize is not an integer and ValueError if it is
+    not a positive integer - both indicate a logic error in the caller. """
     import os
     import re
     from vyos.utils.process import cmd
 
-    try:
-        min_keysize_int = int(min_keysize)
-    except (TypeError, ValueError):
-        return False
+    if isinstance(min_keysize, bool) or not isinstance(min_keysize, int):
+        raise TypeError(
+            f'min_keysize must be an integer, got {type(min_keysize).__name__}')
+    if min_keysize < 1:
+        raise ValueError(
+            f'min_keysize must be a positive integer, got {min_keysize}')
 
     if os.path.exists(file):
-        out = cmd(f'openssl dhparam -inform PEM -in {file} -text')
-        prog = re.compile('\d+\s+bit')
+        out = cmd(['openssl', 'dhparam', '-inform', 'PEM', '-in', file, '-text'])
+        prog = re.compile(r'\d+\s+bit')
         if prog.search(out):
             bits = prog.search(out)[0].split()[0]
-            if int(bits) >= min_keysize_int:
+            if int(bits) >= min_keysize:
                 return True
 
     return False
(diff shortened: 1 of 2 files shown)


dmbaturin (changes requested):
I agree validation could be better here but I don't think this solution is the best one.

One note: this function is not actually used in any real scripts. Its only callers are unit tests for this very function (https://github.com/vyos/vyos-1x/blob/current/src/tests/test_configverify.py#L25-L31). Which means we can change it freely however we want, since we only need to update two tests for that.

If we are to change it, I believe it should be like this:

* Its signature should be `def verify_diffie_hellman_length(file: str, min_keysize: int):`—calling it with an argument that makes no sense is a logic problem in a script.
* It should throw a `TypeError` if `min_keysize` is not an integer at all.
* It should also throw a `ValueError` if `min_keysize` is not a _positive_ integer (because key size of `-512` makes no more sense than `fgsfds`).

andamasov:
@dmbaturin — reworked in e736bccd7 along the lines you described:

- Signature is now `verify_diffie_hellman_length(file: str, min_keysize: int)`.
- `TypeError` if `min_keysize` is not an integer at all (`bool` explicitly rejected too, since `True` would otherwise slip through as keysize 1).
- `ValueError` if it is not a positive integer (`0` and `-512` both rejected).

Confirmed your note about callers: grep shows the only consumers are the unit tests, so nothing in production sees the new exceptions. Both existing tests now pass integers, and two new tests cover the `TypeError`/`ValueError` paths.

Two adjacent fixes in the same function from pre-PR CodeRabbit review: `openssl dhparam` is now invoked with list arguments so the file path never reaches a shell (`popen` auto-selects `shell=True` for strings with spaces), and the bit-count regex is a raw string.

PR #5211, opened 2026-05-20
zerotier: T6455: ZeroTier interface support (alternative implementation)

aadnehovda:
## Summary

Add ZeroTier by using its external interface mgmt support.

This is an alternative integration approach rather than a continuation of the previous per-interface-daemon model. VyOS owns the Linux TAP interfaces, addresses, routes, and CLI integration through an extosdep wrapper, while ZeroTier One remains a single shared daemon and local API provider.

Dependencies:

- VyOS package build with extosdep enabled: vyos/vyos-build#1195
- Optional: ZeroTier upstream extosdep config update notification: zerotier/ZeroTierOne#2587

## Integration model

- Configure networks as normal VyOS interfaces: `interfaces zerotier ztN network-id <nwid>`.
- Run one shared `vyos-zerotier` service and one ZeroTier daemon.
- Launch ZeroTier with extosdep using `zerotier-one -x fd1,fd2`.
- Use the local ZeroTier API directly for operational data.
- Keep daemon/runtime state under `/run/vyos-zerotier`.

## extosdep model

ZeroTier's `extosdep` mode lets `zerotier-one` delegate operating-system actions to an external process over a small binary protocol. Instead of opening TUN/TAP devices and directly adding addresses or routes itself, the daemon asks the wrapper to add/delete TAPs, IP addresses, routes, MTU changes, and related state.

VyOS uses that hook to make the ZeroTier interface look like a normal VyOS-managed interface: the wrapper creates the configured TAP name (`zt0`, `zt1`, ...), applies carrier/address/route changes from ZeroTier events, and keeps the lifecycle tied to VyOS `interfaces zerotier` configuration. ZeroTier still owns membership, encryption, peer discovery, controller communication, and the local API; VyOS owns the Linux interface lifecycle.

## Example configuration

Generate a new ZeroTier identity and install it into the candidate configuration:

```vyos
generate zerotier identity
generate zerotier identity install
```

Controller-managed addressing and routes:

```vyos
set interfaces zerotier zt0 network-id '0123456789abcdef'
set interfaces zerotier zt0 allow-managed 'true'
```

Manual addressing while still joining the ZeroTier network:

```vyos
set interfaces zerotier zt1 network-id 'fedcba9876543210'
set interfaces zerotier zt1 allow-managed 'false'
set interfaces zerotier zt1 address '10.250.0.1/24'
```

`allow-managed` defaults to `true`. Set it to `false` when the controller should not install managed addresses and routes on the VyOS interface. Manual interface addresses can be used with either value.

Service-level daemon policy, for example a fixed bind address and UDP port:

```vyos
set service zerotier listen-address '192.0.2.10'
set service zerotier port primary '9993'
```

Example operational output:

```text
vyos@vyos:~$ show interfaces zerotier
zt0              10.147.17.23/24                  u/u
zt1              10.250.0.1/24                    u/u
zt2              -                                 u/D
```

```text
vyos@vyos:~$ show zerotier networks
Interface    Network ID        Name         Status     Type     Managed    Addresses          Routes
-----------  ----------------  -----------  ---------  -------  ---------  -----------------  --------------
zt0          0123456789abcdef  example-net  OK         PRIVATE  yes        10.147.17.23/24   10.147.0.0/16
zt1          fedcba9876543210  lab-net      OK         PRIVATE  no         -                  -
zt2          0011223344556677  -            NOT_FOUND  PRIVATE  yes        -                  -
```

## Differences from PR #4752

PR #4752 added ZeroTier support through per-interface ZeroTier daemon instances. Each configured `interfaces zerotier <ifname>` rendered a daemon home under `/config/vyos-generated-zerotier/<ifname>`, started a matching `vyos-zerotier-<ifname>.service`, and used ZeroTier `devicemap` to map joined networks to generated Linux interface names like `zt1.<nwid-prefix>`.

This PR uses a different integration model:

- Uses one shared ZeroTier daemon instead of one daemon per configured ZeroTier interface.
- Keeps runtime state under `/run/vyos-zerotier` instead of persistent per-interface daemon homes under `/config/vyos-generated-zerotier/<ifname>`.
- Uses one VyOS interface per ZeroTier network (`zt0`, `zt1`, ...) instead of allowing multiple networks under one configured interface and mapping them to generated names like `zt1.<nwid-prefix>`.
- Lets VyOS own TAP creation, interface state, addresses, and routes through extosdep instead of letting each ZeroTier daemon create/manage its own mapped TAPs.
- Uses direct local API/opmode integration and keeps controller/Central API access out of the normal interface lifecycle.
- Keeps daemon-wide policy under `service zerotier` instead of per daemon `local.conf` options under each ZeroTier interface.

## Validation

- Built the package with `-DZT_EXTOSDEP` through `vyos-build`.
- Installed `zerotier-one 1.16.0-vyos1` on `vyostest`.
- Verified `vyos-zerotier.service` ran the Python wrapper plus child daemon under `/run/vyos-zerotier`.
- Ran real VyOS `configure` + `commit` on the test VM.
- Confirmed wrapper logs received `config update network=<nwid> revision=<n>`.
- Confirmed `zt0` came up with carrier and address.
- Confirmed bogus `zt1` stayed `NO-CARRIER` / `NOT_FOUND`.
- Confirmed `allow-managed false` removed controller-managed IPv4 from `zt0`, and `allow-managed true` added it back.
- Confirmed manual interface addressing can coexist with explicit `allow-managed true`.
- Confirmed an invalid controller route was rejected by the kernel and logged without crashing the wrapper.
- Exercised earlier runtime flows with Central-managed address and route changes, network ID flips, authorization changes, bridge membership visibility, OSPF/bridge interface availability, and IPv6 link-local handling.


aadnehovda:

l0crian1:
- No per-interface daemon homes.
  - The per-daemon approach allowed for much better performance scaling across multiple cores. ZeroTier has very limited multi-core support, which generally won't scale beyond 2 cores. Attaching each interface to it's own daemon allowed the CPU scheduler to get much better throughput. Moving from 4.5Gbps to 22Gbps in my testing on around 6 cores (by using ECMP routing across the multiple interfaces). I think the ability to create unique zerotier-one processes is important for this reason. Without this, performance can be severely lacking.
- No generated ztN.\<prefix> interface names.
  - This was intended to easily be able to identify which interface belonged to which network from commands like `show interfaces` or `show interfaces kernel`, as well as provide each interface to have multiple networks since each interface was it's own daemon.
- No persistent-TAP upstream dependency.
  - Not sure what is meant by this.
- No dependency on ZeroTier Central for normal join/show behavior.
  - The previous PR already supported this. The commands that required a Central API key were commands that showed additional details that are not present from the local API (e.g. Public address, name, etc...). Most of the `op-mode` commands were only calling the local API.  Commands that help to identify peers is a heavily requested feature from ZeroTier customers with many peers, since the normal commands only show peers by non-identifying information.
- No packaged nonfree ZeroTier controller code.
  - The previous implementation already supported this as well. Versions >= 1.16.x fall under the MPL 2.0 license.

I'm ultimately not sure what this PR is attempting to solve. I might just be missing something. For what it's worth, VyOS didn't state any of these things as issues with getting ZeroTier implemented. I talked to the VyOS team and they were on board with getting it integrated (after the licensing change). I closed the PR because it wasn't getting looked at, which my assumption is the VyOS team may not be very familiar with ZeroTier. It's hard for them to reasonably review a PR to add a software suite they may not be familiar/comfortable with.

In hindsight, I think I should have created an initial implementation without the ability to modify any of the `local.conf` parameters. That would have greatly simplified the initial addition of ZeroTier. Then the `local.conf` stuff could have been added later.

Just my thoughts!

aadnehovda:
> Just my thoughts!

I updated the comparison to your PR in the description. It was sloppy and not accurate, sorry for the noise.

The main idea behind this PR is to tie TAP creation and deletion to config. Which means that zerotier doesn't have to be started before the TAPs can be added as bridge members or as routing protocol interfaces like ospf during boot. Similar to how the openvpn interfaces are pre-created by `openvpn --mktun`. This is a problem when running zerotier as a container. 

So interfaces are precreated during config apply, zerotier-one starts later, requests network config and triggers the wrapper daemon to configure addressing and UP the interface.

I like the idea that zerotier hands over the responsibility to vyos for adding/removing managed addresses/routes by sending events and not manipulating the interfaces directly.

There's also no network.d/*.conf and devicemap file handling or persistence in /config, everything is driven by conf+apply and the binary protocol.

It can probably be extended to do multi-daemon, which would allow better perf and the option to join different controllers. Ofc, it also requires a custom compile flag in the zerotier-one package, which may be a blocker, I don't know. 

Not trying to say this one is necessarily better, it a slightly different approach. HTH and I also hope we get zt integration one way or the other soon.

l0crian1:
> The main idea behind this PR is to tie TAP creation and deletion to config. Which means that zerotier doesn't have to be started before the TAPs can be added as bridge members or as routing protocol interfaces like ospf during boot. Similar to how the openvpn interfaces are pre-created by `openvpn --mktun`. This is a problem when running zerotier as a container.

In the previous PR, the `network-id` was a required parameter, which meant that an interface would always be created whenever configuring anything ZeroTier related. Even if a controller couldn't be reached, adding the network-id would create the interface. Any bridge, bond, or routing related config was applied after the zerotier conf-mode script (via script priority), so you didn't have an issue with adding it to bridges or routing protocols.

> I like the idea that zerotier hands over the responsibility to vyos for adding/removing managed addresses/routes by sending events and not manipulating the interfaces directly.

I'm not sure what this really buys at the end of the day. ZeroTier already handles this reliably. In my experience, ZeroTier will generally not merge community created PRs, so this may never even be possible. But if they did, with this being edge case just for VyOS, I wouldn't be surprised if they break the functionality in the future. And at that point, it's possible nobody responsible or familiar with this PR will be available, leaving things in a broken state until someone can get up to speed on this one off solution for VyOS.

The majority of VyOS is just conf-mode and op-mode managing the binaries for a given service, and all of that works well. I think trying to keep to that with ZeroTier is important, because then anyone can contribute to it without needing to know how the one off solution deviates from a typical install. If they know how it is installed on any linux system, they'll already be familiar with what is happening when VyOS configures it.

Again, these are just my thoughts. At the end of the day, I think it'd be great to have native ZeroTier support, so if this PR moves that needle forward....great!

aadnehovda:
I found the recently added zerotier support for external interface mgmt and thought it would be a good fit for vyos, giving tight interface control. There doesn't seem to be many other projects that utilize `ZT_EXTOSDEP` yet (perhaps some embedded systems?). There's an unnamed vendor mentioned at https://github.com/zerotier/ZeroTierOne/pull/2438#issuecomment-2698644959. I read it as if it was intended to be used by appliances.

I've found it weird that zerotier creates and deletes interfaces on its own, https://github.com/zerotier/ZeroTierOne/issues/1038. I guess it's for the laptop-VPN experience, but there are issues on different platforms, e.g. https://github.com/zerotier/ZeroTierOne/issues/659.

Btw, the upstream ZT patch is minimal and not strictly necessary and is mainly there to be able to immediately trigger CARRIER_UP/DOWN on the TAP when networks are deleted, members are authorized/banned, etc. The wrapper polls the api for this anyway, so the new notification is just an extra nudge.

> Again, these are just my thoughts. At the end of the day, I think it'd be great to have native ZeroTier support, so if this PR moves that needle forward....great!

Agreed :-) Let's hope for some progress either way

sever-sever (commented):
Thank you for contributing
The main question is about ZeroTier/ZeroTier components licensing.
We couldn't use their products before because of their "complex" licenses, which weren't compatible with VyOS.
So before the review, we have to be 100% sure that we can use their products.
https://github.com/zerotier/ZeroTierOne/blob/dev/LICENSE.txt
- For example, `nonfree` https://github.com/zerotier/ZeroTierOne/blob/dev/nonfree/LICENSE.md
Can we build binaries that do not include `nonfree` components?
- [ext](https://github.com/zerotier/ZeroTierOne/tree/dev/ext) not all components have a license

aadnehovda:
> we have to be 100% sure that we can use their products.

I think this is best to confirm directly with zerotier

> Can we build binaries that do not include `nonfree` components?

Afaict, yes. The libzt library and the agent daemon are under Mozilla license and are the only components used in the integration.

The nonfree component is for self hosting a network controller and is free for non commercial use but since zerotier hosts and charges for cloud hosted controller, this part is not free to distribute in a commercial product. The initial FOSS licensing was changed to avoid big cloud providers to productize the controller hosting and eat into the business model, but the agent part was later returned to an open source license. This is my understanding anyway, and I'm no expert.

PR #5219, opened 2026-05-27
Fix DNS forwarding cache timeout for sagitta

4ek0:
Adjusts the build configuration to ensure compatibility with sagitta branch. Minor cleanup in build rules.


sever-sever:
Pull requests to this branch are not accepted and have been closed.
You must use the `current` branch instead. After that, it may be backported to the stable branches.

PR #5238, opened 2026-05-29
vpp: T8930: Block vif driver to prevent unhandled traceback on XCP-NG

natali-rs1985:
The `vif` driver is fundamentally incompatible with VPP — no DPDK poll-mode
driver exists for Xen paravirtual interfaces and no PCI address is exposed,
so `allow-unsupported-nics` cannot rescue it either.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8930
## Checklist:

### DIFF
--- a/src/conf_mode/vpp.py
+++ b/src/conf_mode/vpp.py
@@ -97,6 +97,10 @@ override_drivers: dict[str, str] = {
 # drivers that does not use PCIe addresses
 not_pci_drv: list[str] = ['hv_netvsc']
 
+# drivers that are fundamentally incompatible with VPP/DPDK — no PMD exists and
+# they expose no PCI address, so allow-unsupported-nics cannot rescue them
+incompatible_drv: list[str] = ['vif']
+
 # drivers that support interrupt RX mode for DPDK and XDP
 drivers_support_interrupt: dict[str, list] = {
     'atlantic': ['dpdk', 'xdp'],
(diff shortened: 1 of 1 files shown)


sever-sever (approved):
Add incompatible driver list for VPP/DPDK

src/conf_mode/vpp.py
```diff
@@ -596,6 +600,13 @@ def verify(config):
 
     # ensure DPDK/XDP settings are properly configured
     for iface, iface_config in config['settings']['interface'].items():
+        original_driver = config['persist_config'][iface]['original_driver']
+        if original_driver in incompatible_drv:
+            raise ConfigError(
+                f'Interface "{iface}" driver "{original_driver}" is not supported by VPP/DPDK. '
+                'Use PCI passthrough to assign a physical NIC to the VM.'
```
sever-sever: Not sure if we should to add additional comments here, it is up to the user what to do with this 
In my opinion, it should be omitted.
So the list could be extended in the future and become common
```suggestion
                
```

natali-rs1985: Removed this line

src/conf_mode/vpp.py
```diff
@@ -596,6 +600,12 @@ def verify(config):
 
     # ensure DPDK/XDP settings are properly configured
     for iface, iface_config in config['settings']['interface'].items():
+        original_driver = config['persist_config'][iface]['original_driver']
```
c-po:
```suggestion
        original_driver = dict_search(f'persist_config.{iface}.original_driver', config)
        if original_driver and original_driver in incompatible_drv:
```

Also line 659 makes use of the original driver, too - refactor code to not retrive variable multiple times.
`driver = config.get('persist_config').get(iface).get('original_driver')`

c-po (dismissed):

src/conf_mode/vpp.py
natali-rs1985: Thanks for suggestion. Done, 'orig_driver' is now fetched once and reused throughout the loop.

zdc (approved):
I support the idea.

Optionally, for better maintainability, you may extract this check to some `_is_device_disallowed()`, where you can later combine checks of different types, not only by a driver.

PR #5249, opened 2026-06-02
http-api-client: T8955: enable TLS verification by default

andamasov:
## Summary

`python/vyos/http_api_client.py` defaulted `ApiClientConfig.verify_tls` to `False` and unconditionally silenced urllib3's `InsecureRequestWarning`, so the HTTP API client did **not** verify TLS peers by default.

The only real consumer is `src/op_mode/config_sync.py` (`ConfigSyncDiffManager`), which builds `ApiClientConfig(host, key, port, timeout)` **without** `verify_tls` and so inherited the insecure default. config-sync connects to a **remote** secondary VyOS node over HTTPS, so verification was silently off on a remote management channel that carries the API `key` and the full configuration — a MITM exposure.

(`src/conf_mode/system_syslog.py` has its own unrelated `verify_tls` concept for rsyslog and does **not** use this client.)

## Changes

**`python/vyos/http_api_client.py`**
- `verify_tls` now defaults to `True` and is typed `Union[bool, str]`. A string is forwarded to `requests` as a CA-bundle path (`verify=<path>`); `True`/`False` keep their `requests` meaning.
- `InsecureRequestWarning` is suppressed **only** when verification is explicitly disabled (`verify_tls is False`) — not when a CA path or `True` is configured.

**`src/op_mode/config_sync.py`**
- The runtime-settings loader now reads an optional `verify_tls` from the `secondary` block and passes it through, **defaulting to `False`** to preserve current behaviour for self-signed secondaries. The insecurity is now explicit at the call site rather than hidden in the library default.

## Blast radius / reviewer note

- **No functional regression today:** config-sync still defaults to no-verify (self-signed secondaries keep working). The change makes the *library* secure-by-default for future consumers (e.g. vyos-mcp) and makes the existing insecurity explicit and operator-overridable.
- **Follow-up (not in this PR):** a first-class CLI knob — `set service config-sync secondary verify-tls` / `ca-certificate` — to surface verification in the config tree (needs XML + a migration). Flagged with a `TODO` in `config_sync.py`.

## Testing

- `python3 -m py_compile` passes on both files.
- No new lines exceed the 88-char ruff limit (the two long lines flagged locally pre-exist and are untouched).
- No unit-test harness exists for the live HTTPS client without a running secondary; behaviour change is limited to the default value + warning-suppression condition.

Surfaced by automated security review. The flagged code is already public on `rolling`.

Tracking: T8955

### DIFF
--- a/python/vyos/http_api_client.py
+++ b/python/vyos/http_api_client.py
@@ -19,6 +19,7 @@ import json
 import urllib3
 import requests
 from typing import Optional
+from typing import Union
 from dataclasses import dataclass
 
 from vyos.version import get_version
@@ -51,7 +52,11 @@ class ApiClientConfig:
     key: str
     port: int = 443
     timeout: Optional[int] = None
-    verify_tls: bool = False
+    # TLS peer verification, forwarded verbatim to requests' ``verify``:
+    #   True       - verify against the system CA store (secure default)
+    #   False      - disable verification (insecure; opt-in only)
+    #   "<path>"   - verify against a custom CA bundle file/dir
+    verify_tls: Union[bool, str] = True
 
 
 class ApiClient:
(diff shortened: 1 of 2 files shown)


andamasov:
@jestabro @sever-sever we need to make this configurable, because even with self-signed certificates, people must handle this in the correct way

dmbaturin (commented):
Since many people _are_ using self-signed certs, I suppose there should be an option to verify certificates against fingerprints rather than only against a CA store.

PR #5250, opened 2026-06-02
remote: T8956: stop leaking URL credentials in TftpC.upload()

andamasov:
## Summary

TftpC.upload() printed `{command} "{urlstring}"` to stdout before running the
command. This was debug code from the time where VRF support was added. As TFTP
has no authentication - no credentials got leaked.
`self.urlstring` is the full `urllib.parse.urlunsplit(url)` and can embed userinfo — `scheme://user:password@host/…`. Printing it exposes those credentials in op-mode output and any captured logs.

## Testing

- `python3 -m py_compile python/vyos/remote.py` passes.
- One-line deletion; no behavioural change to the upload itself (the `cmd(...)` call is untouched).

Surfaced by automated security review. The flagged code is already public on `rolling`.

Tracking: T8956

### DIFF
--- a/python/vyos/remote.py
+++ b/python/vyos/remote.py
@@ -476,7 +476,6 @@ class TftpC:
                         vrf=self.vrf).encode())
 
     def upload(self, location: str):
-        print(f'{self.command} "{self.urlstring}"')
         with open(location, 'rb') as f:
             cmd(f'{self.command} --upload-file - "{self.urlstring}"',
                 input=f.read(), vrf=self.vrf)


andamasov:
@jestabro @c-po  can take a look, maybe for some reason we need this behavior, but if not, better avoid it

c-po (approved):
Remove debug code added in https://github.com/vyos/vyos-1x/commit/56d98d7e152f4211dbba7825aac00f7024df4275#diff-216d68832b7b77137472496299b9eba31cdb866291bbb35f83fd9abbd155cf6eR44-R477 by me.

AI generated commit message adjusted - change is in TftpC not in CurlC. TFTP ha sno authentication so no information leak.

sever-sever (approved):
Remove missed debug print

PR #5253, opened 2026-06-03
password-reset: T8346: Fix password recovery when only `plaintext-password` is set

alexandr-san4ez:
## Change summary
When a user account was provisioned with only `plaintext-password` (e.g. via cloud-init), the password recovery tool failed to persist the new password across reboots. On the next boot, VyOS would re-apply the `plaintext-password` from config.boot, silently overwriting the recovered encrypted password.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8346

### Manual test
1. Instance started, login vyos/vyos success.
2. Reboot and choose 'password reset' from GRUB menu.
3. Boot into password recovery tool and set a new password for user 'vyos':
```
Do you wish to reset the admin password? (y or n) y
Which admin account do you want to reset? [vyos]
Starting process to reset the admin password...
Re-mounting root filesystem read/write...
Saving backup copy of config.boot...
Setting the administrator (vyos) password...
Enter vyos password: <new password>
Retype vyos password: <new password>
System will reboot in 10 seconds...
```
4. Reboot and try to login with new password:
```
Welcome to VyOS - vyos5 ttyS0

vyos5 login: vyos
Password: <old password>

Login incorrect
vyos5 login: vyos
Password: <new password>

---
WARNING: This VyOS system is not a stable long-term support version and
         is not intended for production use.

vyos@vyos5:~$
vyos@vyos5:~$ diff -u /config/config.boot.before_pwrecovery /config/config.boot
--- /config/config.boot.before_pwrecovery       2026-06-03 12:38:33.868100207 +0000
+++ /config/config.boot 2026-06-03 12:38:39.772000000 +0000
@@ -94,7 +94,8 @@
         }
         user vyos {
             authentication {
-                plaintext-password "vyos"
+                plaintext-password ""
+                encrypted-password "$y$j9T$oXprHr91wN3MVnjfiU/Cj.$qKSnE9tUe1hTLzDmjpKT8B.Omo1pTUtri5tyLuKkS/C"
                 public-keys user@pc {
                     key "AA...LD03"
                     type "ssh-ed25519"
```

## Checklist:

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -123,6 +123,11 @@ Depends:
   wide-dhcpv6-client,
 # Generic colorizer
   grc,
+# The package "whois" contains both a whois database client
+# that end users can find useful,
+# and the "mkpasswd" utility we use for generating hashed passwords
+# in the standalone password reset script
+  whois,
 ## End of System services and utilities
 ## For the installer
   fdisk,
--- a/src/system/standalone_root_pw_reset
+++ b/src/system/standalone_root_pw_reset
@@ -27,8 +27,23 @@ CF=/opt/vyatta/etc/config/config.boot
 ADMIN=vyos
 
 set_encrypted_password() {
-    sed -i \
-       -e "/ user $1 {/,/encrypted-password/s/encrypted-password .*\$/encrypted-password \"$2\"/" $3
+    local user=$1 epwd=$2 file=$3
+
+    # Check if encrypted-password exists within this specific user's block
+    if awk "/ user $user \{/{f=1} f && /encrypted-password/{found=1; exit} f && /^        }/{exit} END{exit !found}" "$file"; then
+        # Replace existing encrypted-password line
+        sed -i \
+           -e "/ user $user {/,/encrypted-password/s/encrypted-password .*\$/encrypted-password \"$epwd\"/" "$file"
+    else
+        # Insert encrypted-password after plaintext-password
+        sed -i \
+           -e "/ user $user {/,/plaintext-password/s/\(plaintext-password .*\)\$/\1\n                encrypted-password \"$epwd\"/" "$file"
+    fi
+}
+
+clear_plaintext_password() {
+   sed -i \
+      -e "/ user $1 {/,/plaintext-password/s/plaintext-password .*\$/plaintext-password \"\"/" $2
 }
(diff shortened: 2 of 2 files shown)


debian/control
```diff
@@ -123,6 +123,8 @@ Depends:
   wide-dhcpv6-client,
 # Generic colorizer
   grc,
+# whois is for reseting password (mkpasswd) using GRUB
```
dmbaturin:
```suggestion
# The package "whois" contains both a whois database client
# that end users can find useful,
# and the "mkpasswd" utility we use for generating hashed passwords
# in the standalone password reset script
```

dmbaturin (approved):
I think it's a good idea to clarify the situation with `mkpasswd` residing in the `whois` package for a strange reason, since `whois` itself is not for generating passwords.

Otherwise, I don't see issues here (I do wonder if we should get rid of the shell script bye now, but it's a different discussion).

src/system/standalone_root_pw_reset
```diff
@@ -27,8 +27,23 @@ CF=/opt/vyatta/etc/config/config.boot
...
 set_encrypted_password() {
-    sed -i \
-       -e "/ user $1 {/,/encrypted-password/s/encrypted-password .*\$/encrypted-password \"$2\"/" $3
+    local user=$1 epwd=$2 file=$3
+
+    # Check if encrypted-password exists within this specific user's block
+    if awk "/ user $user \{/{f=1} f && /encrypted-password/{found=1; exit} f && /^        }/{exit} END{exit !found}" "$file"; then
+        # Replace existing encrypted-password line
+        sed -i \
+           -e "/ user $user {/,/encrypted-password/s/encrypted-password .*$/encrypted-password \"$epwd\"/" "$file"
```
c-po: @alexandr-san4ez in the original implementation `$` is escaped by `\` in LINE 31 - is this escape no longer needed?

alexandr-san4ez: It works the same because shell doesn't have `$` variable but I agree that `\$` style makes the intent explicit. 
I already changed it.

c-po (approved):
Change looks good to me. Requested change addressed.

PR #5254, opened 2026-06-03
T8963: policy-route: trigger domain resolver for domain groups

jd82k:
## Change summary

Fix policy route domain-group handling so vyos-domain-resolver is triggered after PBR nftables sets are recreated.

Previously, PBR rules could reference firewall domain-group sets in ip vyos_mangle, but policy_route.py did not notify vyos-domain-resolver. As a result, D_* sets were created empty and only populated on the resolver’s next periodic interval, causing PBR rules using domain groups to be ineffective for several minutes after commit.

This change adds policy-route resolver usage tracking and restarts/stops vyos-domain-resolver.service consistently with the existing firewall/NAT behavior. A smoketest was added to verify that a PBR rule using a domain group gets its ip vyos_mangle set populated promptly.
## Types of changes

## Related Task(s)

https://vyos.dev/T8963

```
vyos@vyos:~$ sudo /usr/libexec/vyos/tests/smoke/cli/test_policy_route.py TestPolicyRoute.test_pbr_domain_grouptest_pbr_domain_group (__main__.TestPolicyRoute.test_pbr_domain_group) ... ok

Ran 1 test in 33.669s

OK
```
## Checklist:


smoketest/scripts/cli/test_policy_route.py
```diff
@@ -53,6 +57,8 @@ def tearDownClass(cls):
     def tearDown(self):
         self.cli_delete(['policy', 'route'])
         self.cli_delete(['policy', 'route6'])
+        self.cli_delete(['firewall', 'group', 'domain-group', 'smoketest_domain'])
+        self.cli_delete(['system', 'static-host-mapping', 'host-name', 'pbr.example.com'])
         self.cli_commit()
 
         # Verify nftables cleanup
```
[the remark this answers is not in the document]
sarthurdev: Copilot comment seems reasonable, should be able to address it in the search list as below:

```
# Verify nftables cleanup
nftables_search = [
    ['set D_smoketest_domain'],
....
```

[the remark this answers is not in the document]
jd82k: Added.

src/conf_mode/policy_route.py
```diff
@@ -187,6 +191,42 @@ def generate(policy):
...
+                        return True
+
+    return False
+
+def update_domain_resolver(policy):
+    domain_action = None
+
+    if domain_group_used(policy):
+        text = f'# Automatically generated by policy_route.py\nThis file indicates that vyos-domain-resolver service is used by policy_route.\n'
+        write_file(domain_resolver_usage, text)
```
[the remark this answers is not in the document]
jd82k: Fixed.

sarthurdev (approved):
Triggers resolver restart on policy change. Smoketest added/passes, tested locally.

smoketest/scripts/cli/test_policy_route.py
```diff
@@ -73,6 +80,17 @@ def tearDown(self):
         # always forward to base class
         super().tearDown()
 
+    def wait_for_domain_resolver(self, table, set_name, element, max_wait=10):
```
c-po: Is there a reason to not make use of the generic helper `self.wait_for_result(...)`

src/conf_mode/policy_route.py
```diff
@@ -187,6 +191,42 @@ def generate(policy):
...
+
+    if domain_group_used(policy):
+        text = '# Automatically generated by policy_route.py\nThis file indicates that vyos-domain-resolver service is used by policy_route.\n'
+        write_file(domain_resolver_usage, text)
+        domain_action = 'restart'
+    elif os.path.exists(domain_resolver_usage):
+        os.unlink(domain_resolver_usage)
+        domain_action = 'restart'
+
+        if not glob('/run/use-vyos-domain-resolver*'):
```
c-po: Line 44 defined `domain_resolver_usage` but here a glob pattern is used, why?

c-po (changes requested):

smoketest/scripts/cli/test_policy_route.py
jd82k: Updated. The test now uses the generic wait_for_result() helper.

src/conf_mode/policy_route.py
jd82k: Updated. This needs to check for any remaining resolver users before stopping the service, not just the policy-route marker. I moved the pattern to a named constant so the intent is clearer.

jd82k:
Any updates?

sever-sever:
@c-po could you re-check the PR?

c-po (changes requested):
Implementation looks reasonable. Smoketests added. Locally tested by simon.

c-po (approved):
Implementation looks reasonable. Smoketests added. Locally tested by simon.

PR #5261, opened 2026-06-09
VPP: T8972: Fix VPP/LCP VRF table synchronization for VPP-managed Ethernet interfaces and Ethernet VIFs.

inetman28:
## Change summary

Fix VPP/LCP VRF table synchronization for VPP-managed Ethernet interfaces and
Ethernet VIFs.

VyOS correctly assigns the Linux/control-plane interface to a VRF, but the
corresponding VPP interface was not explicitly moved to the matching VPP FIB
table. As a result, connected routes for VPP/LCP-managed interfaces could remain
in VPP table `0` even when the VyOS interface was assigned to a non-default VRF.

This change adds targeted VPP interface table synchronization using PAPI. When a
VPP-managed Ethernet interface or Ethernet VIF has a VyOS VRF assigned, the
corresponding VPP interface is bound to the VRF table ID. When the VRF is
removed, the interface is moved back to table `0`. Existing VPP interface
addresses are preserved during table moves by removing them before the table
change and re-adding them afterwards.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8972

Manual testing was performed in a GNS3 VyOS lab with VPP enabled on `eth1`.

Initial configuration:

```shell
set vpp settings interface eth1
set vrf name mgmt table 1000
set interfaces ethernet eth1 vif 20 address 100.100.100.1/24
set interfaces ethernet eth1 vif 20 vrf mgmt
commit
```

The VPP interface address output showed that `eth1.20` was bound to VPP table
`1000`:

```text
eth1.20 (up):
  L3 100.100.100.1/24 ip4 table-id 1000 fib-idx 1
```

`100.100.100.0/24` and `100.100.100.1/32` were present in VPP FIB table
`1000` and were not present in table `0`.

Address delete/re-add was tested:

```shell
delete interfaces ethernet eth1 vif 20 address
commit
```

After commit, the address disappeared from `vppctl show int addr`, and no
`100.100.100.*` routes were present in VPP table `0` or table `1000`.

Then the address was added back:

```shell
set interfaces ethernet eth1 vif 20 address 100.100.100.1/24
commit
```

After commit, the address returned to `eth1.20` in VPP table `1000`, and the
connected routes appeared only in table `1000`.

VRF removal was tested:

```shell
delete interfaces ethernet eth1 vif 20 vrf
commit
```

After commit, `eth1.20` still had `100.100.100.1/24`, but without
`ip4 table-id 1000`:

```text
eth1.20 (up):
  L3 100.100.100.1/24
```

Connected routes moved to VPP table `0` and disappeared from table `1000`.

VRF add-back was tested:

```shell
set interfaces ethernet eth1 vif 20 vrf mgmt
commit
```

After commit, `eth1.20` was again bound to table `1000`:

```text
eth1.20 (up):
  L3 100.100.100.1/24 ip4 table-id 1000 fib-idx 1
```

Connected routes moved back to table `1000` and disappeared from table `0`.

VRF change was tested:

```shell
set vrf name test1 table 2000
set interfaces ethernet eth1 vif 20 vrf test1
commit
```

After commit, `eth1.20` was bound to VPP table `2000`:

```text
eth1.20 (up):
  L3 100.100.100.1/24 ip4 table-id 2000 fib-idx 2
```

Connected routes disappeared from tables `0` and `1000`, and appeared only in
VPP table `2000`.

A dedicated smoketest was added:

```text
smoketest/scripts/cli/test_vpp.py::TestVPP::test_24_vpp_lcp_vrf_table_sync
```

Local checks run:

```shell
python3 -m py_compile python/vyos/vpp/control_vpp.py src/conf_mode/interfaces_ethernet.py src/conf_mode/vrf.py smoketest/scripts/cli/test_vpp.py
darker --check --diff --revision rolling python/vyos/vpp/control_vpp.py src/conf_mode/interfaces_ethernet.py src/conf_mode/vrf.py
darker --check --diff --revision origin/fix-vpp-lcp-vrf-table smoketest/scripts/cli/test_vpp.py
git diff --check
```

## Checklist:

### DIFF
--- a/python/vyos/vpp/control_vpp.py
+++ b/python/vyos/vpp/control_vpp.py
@@ -26,6 +26,8 @@ from typing import TypeVar, ParamSpec, Literal
 from vpp_papi import VPPApiClient
 from vpp_papi import VPPIOError, VPPValueError
 
+from vyos.vpp.utils import vpp_ip_addresses_by_index
+
 # define types for static type checkers
 AnyType = TypeVar('AnyType')
 AnyParam = ParamSpec('AnyParam')
@@ -300,6 +302,110 @@ class VPPControl:
         based on the current state of the kernel.
         """
         return self.__vpp_api_client.api.lcp_nl_resync()
+
+    @_Decorators.api_call
+    def get_interface_ip_table(self, ifname: str, is_ipv6: bool = False) -> int | None:
+        """Return the IP FIB table ID assigned to an interface.
+
+        Args:
+            ifname (str): Interface name in VPP.
+            is_ipv6 (bool, optional): Return IPv6 table ID if True. Defaults to False.
+
+        Returns:
+            int | None: FIB table ID or None if the interface is not found.
+        """
+        iface_index = self.get_sw_if_index(ifname)
+        if iface_index is None:
+            return None
+
+        response = self.__vpp_api_client.api.sw_interface_get_table(
+            sw_if_index=iface_index,
+            is_ipv6=is_ipv6,
+        )
+        if response.retval != 0:
+            return None
+
+        return response.vrf_id
+
+    @_Decorators.api_call
+    def set_interface_ip_table(
+        self, ifname: str, table_id: int, is_ipv6: bool = False
+    ) -> None:
+        """Assign an interface to an IP FIB table.
+
+        Args:
+            ifname (str): Interface name in VPP.
+            table_id (int): FIB table ID.
+            is_ipv6 (bool, optional): Set IPv6 table if True. Defaults to False.
+        """
+        iface_index = self.get_sw_if_index(ifname)
+        if iface_index is None:
+            return None
+
+        self.__vpp_api_client.api.sw_interface_set_table(
+            sw_if_index=iface_index,
+            is_ipv6=is_ipv6,
+            vrf_id=table_id,
+        )
+
+    @_Decorators.api_call
+    def add_del_interface_ip_address(
+        self, ifname: str, address: str, is_add: bool = True
+    ) -> None:
+        """Add or delete an interface IP address in VPP.
+
+        Args:
+            ifname (str): Interface name in VPP.
+            address (str): Interface prefix, for example "192.0.2.1/24".
+            is_add (bool, optional): Add address if True, delete if False.
+                Defaults to True.
+        """
+        iface_index = self.get_sw_if_index(ifname)
+        if iface_index is None:
+            return None
+
+        self.__vpp_api_client.api.sw_interface_add_del_address(
+            sw_if_index=iface_index,
+            is_add=is_add,
+            del_all=False,
+            prefix=address,
+        )
+
+    def move_interface_to_ip_table_preserve_addresses(
+        self, ifname: str, table_id: int
+    ) -> None:
+        """Move an interface to an IP FIB table while preserving its addresses.
+
+        VPP keeps connected routes in the interface IP table that was active when
+        addresses were added. Re-adding the same addresses after the table move
+        recreates connected routes in the target table.
+
+        Args:
+            ifname (str): Interface name in VPP.
+            table_id (int): Target FIB table ID.
+        """
+
+        iface_index = self.get_sw_if_index(ifname)
+        if iface_index is None:
+            return None
+
+        for is_ipv6 in [False, True]:
+            current_table = self.get_interface_ip_table(ifname, is_ipv6=is_ipv6)
+            if current_table is None or current_table == table_id:
+                continue
+
+            addresses = vpp_ip_addresses_by_index(
+                self.__vpp_api_client.api,
+                iface_index,
+                is_ipv6=is_ipv6,
+            )
+            for address in addresses:
+                self.add_del_interface_ip_address(ifname, address, is_add=False)
+
+            self.set_interface_ip_table(ifname, table_id, is_ipv6=is_ipv6)
+
+            for address in addresses:
+                self.add_del_interface_ip_address(ifname, address)
 
     @_Decorators.check_retval
     @_Decorators.api_call
--- a/smoketest/scripts/cli/test_vpp.py
+++ b/smoketest/scripts/cli/test_vpp.py
@@ -35,6 +35,7 @@ from vyos.utils.system import sysctl_read
 from vyos.utils.network import interface_exists
 from vyos.system import image
 from vyos.vpp import VPPControl
+from vyos.vpp.utils import vpp_ip_addresses_by_index
 from vyos.vpp.utils import vpp_iface_name_transform
 from vyos.vpp.config_resource_checks.resource_defaults import default_resource_map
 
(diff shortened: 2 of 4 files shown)


src/conf_mode/interfaces_ethernet.py
```diff
@@ -473,8 +479,58 @@ def apply(ethernet):
             mtu = e.get_mtu()
             vpp_api.set_iface_mtu(lcp_name, mtu)
 
+        sync_vpp_lcp_vrf_tables(ethernet, vpp_api)
+
     return None
 
+
+def get_vrf_table_id(ethernet: dict, vrf_name: str | None) -> int:
```
natali-rs1985: VyOS already has `get_vrf_tableid` in [vyos.utils.network](https://github.com/vyos/vyos-1x/blob/rolling/python/vyos/utils/network.py#L148). Since vrf.py runs before interfaces_ethernet.py, the VRF interface is already created in the kernel by the time this runs, so `get_vrf_tableid(vrf_name)` will return the correct table ID. This custom function and the `vrf_config` dict lookup can be removed.

data/config-mode-dependencies/vyos-vpp.json
```diff
@@ -49,6 +49,8 @@
         "vpp_acl": ["vpp_acl"],
         "vpp_nat_nat44": ["vpp_nat_nat44"],
         "vpp_nat_cgnat": ["vpp_nat_cgnat"]
+    },
+    "vrf": {
+        "ethernet": ["interfaces_ethernet"]
```
natali-rs1985: Why do we need this dependency? `vrf.py` (priority 11) always runs before `interfaces_ethernet.py` (priority 318) by natural ordering. Additionally, VRF table ID changes and VRF removal with member interfaces are both blocked by verify(). So there is no scenario where a VRF change requires re-triggering ethernet reconfiguration. This can be removed.

src/conf_mode/interfaces_ethernet.py
```diff
@@ -473,8 +479,58 @@ def apply(ethernet):
...
+        for vif_c, vif_c_config in vif_s_config.get('vif_c', {}).items():
+            vif_c_ifname = vif_c_config.get('ifname', f'{ifname}.{vif_s}.{vif_c}')
+            targets.append(
+                (
+                    vif_c_ifname,
+                    get_vrf_table_id(ethernet, vif_c_config.get('vrf')),
+                )
+            )
+
+    return targets
```
natali-rs1985: This function could be simplified since `get_vrf_tableid` returns tableid by interface name

```suggestion
def iter_vpp_lcp_vrf_targets(ethernet: dict) -> list[tuple[str, int]]:
    interfaces = [ethernet['ifname']]
    for vif in ethernet.get('vif', {}).values():
        interfaces.append(vif['ifname'])
    for vif_s in ethernet.get('vif_s', {}).values():
        interfaces.append(vif_s['ifname'])
        for vif_c in vif_s.get('vif_c', {}).values():
            interfaces.append(vif_c['ifname'])
    return [(iface, get_vrf_tableid(iface) or 0) for iface in interfaces]
```

python/vyos/vpp/control_vpp.py
```diff
@@ -301,6 +301,139 @@ def lcp_resync(self) -> None:
...
+
+        response = self.__vpp_api_client.api.sw_interface_get_table(
+            sw_if_index=iface_index,
+            is_ipv6=is_ipv6,
+        )
+        if response.retval != 0:
+            raise VPPValueError(f'VPP API call failed: {response.retval}')
+
+        return response.vrf_id
+
```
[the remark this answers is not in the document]
natali-rs1985: `@_Decorators.check_retval` returns the raw response object, which breaks getter methods that need to extract a specific field like `vrf_id`. The manual retval check is correct here. But I would not raise and just return None.

natali-rs1985 (changes requested):
Thank you for the contribution and the well-structured implementation with good test coverage. The core logic is correct, however I've requested several changes.

python/vyos/vpp/control_vpp.py
```diff
@@ -301,6 +301,139 @@ def lcp_resync(self) -> None:
...
+            return []
+
+        addresses = []
+        for address in self.__vpp_api_client.api.ip_address_dump(
+            sw_if_index=iface_index,
+            is_ipv6=is_ipv6,
+        ):
+            addresses.append(str(address._asdict().get('prefix')))
+
+        return addresses
```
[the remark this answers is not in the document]
natali-rs1985: I agree with this suggestion

python/vyos/vpp/control_vpp.py
```diff
@@ -301,6 +301,139 @@ def lcp_resync(self) -> None:
...
+            raise VPPValueError(f'VPP interface "{ifname}" not found')
+
+        return self.__vpp_api_client.api.sw_interface_set_table(
+            sw_if_index=iface_index,
+            is_ipv6=is_ipv6,
+            vrf_id=table_id,
+        )
+
+    @_Decorators.api_call
+    def get_interface_ip_addresses(
```
natali-rs1985: `get_interface_ip_addresses` duplicates the existing `vpp_ip_addresses_by_index` in [vyos/vpp/utils.py](https://github.com/vyos/vyos-1x/blob/rolling/python/vyos/vpp/utils.py#L182). Consider reusing it instead. Note that `vpp_ip_addresses_by_index` also uses `._asdict().get('prefix')` which should be fixed to `address.prefix`

natali-rs1985 (changes requested):
Regarding `@_Decorators.check_retval` I would rather remove them to not raise exceptions and follow the general pattern in `control_vpp.py` where methods return `None` or `[]` on failure. The same applies to `if iface_index is None: raise VPPValueError(...) `— this should simply return `None` instead.
This will also eliminate the need for the `try/finally` block entirely.

smoketest/scripts/cli/test_vpp.py
```diff
@@ -1513,6 +1513,92 @@ def test_23_vpp_acl_subinterface(self):
...
+
+        def assert_fib_route(table_id, expected=True):
+            _, out = rc_cmd(f'sudo vppctl show ip fib table {table_id}')
+            if expected:
+                self.assertIn(fib_route, out)
+            else:
+                self.assertNotIn(fib_route, out)
+
+        try:
+            self.cli_set(['vrf', 'name', mgmt_vrf, 'table', '1000'])
```
natali-rs1985: Could you please remove try/finally block from test. We assume test should pass

inetman28:
@natali-rs1985
 
Hi Natali,
all suggentions have been resolved

python/vyos/vpp/control_vpp.py
```diff
@@ -301,6 +303,142 @@ def lcp_resync(self) -> None:
...
+            if (
+                self.set_interface_ip_table(ifname, table_id, is_ipv6=is_ipv6)
+                is not None
+            ):
+                moved = True
+
+            for address in addresses:
+                self.add_del_interface_ip_address(ifname, address)
+
+        return moved
```
natali-rs1985: Why do we need this moved variable at all?

python/vyos/vpp/control_vpp.py
```diff
@@ -301,6 +303,142 @@ def lcp_resync(self) -> None:
...
+        response = self.__vpp_api_client.api.sw_interface_add_del_address(
+            sw_if_index=iface_index,
+            is_add=is_add,
+            del_all=False,
+            prefix=address,
+        )
+        if response.retval != 0:
+            return None
+
+        return response
```
natali-rs1985: No need to return responce
```suggestion
        self.__vpp_api_client.api.sw_interface_add_del_address(
            sw_if_index=iface_index,
            is_add=is_add,
            del_all=False,
            prefix=address,
        )
```

python/vyos/vpp/control_vpp.py
```diff
@@ -301,6 +303,142 @@ def lcp_resync(self) -> None:
...
+
+        response = self.__vpp_api_client.api.sw_interface_set_table(
+            sw_if_index=iface_index,
+            is_ipv6=is_ipv6,
+            vrf_id=table_id,
+        )
+        if response.retval != 0:
+            return None
+
+        return response
```
natali-rs1985:
```suggestion
        self.__vpp_api_client.api.sw_interface_set_table(
            sw_if_index=iface_index,
            is_ipv6=is_ipv6,
            vrf_id=table_id,
        )
```

data/config-mode-dependencies/vyos-vpp.json
```diff
@@ -51,4 +51,3 @@
         "vpp_nat_cgnat": ["vpp_nat_cgnat"]
     }
 }
-
```
natali-rs1985, line 54: Revert this line removal please.

python/vyos/vpp/control_vpp.py
```diff
@@ -301,6 +303,142 @@ def lcp_resync(self) -> None:
...
+
+    @_Decorators.api_call
+    def get_interface_ip_addresses(
```
natali-rs1985: This function is now just a wrapper around vpp_ip_addresses_by_index. It can be removed entirely and vpp_ip_addresses_by_index called directly where needed.

src/conf_mode/vrf.py
```diff
@@ -269,7 +269,12 @@ def apply(vrf):
             # Remove WireGuard fwmark routing rules created for this VRF table
             table_id = get_vrf_tableid(tmp)
             for afi in ['-4', '-6']:
-                while call(f'ip {afi} rule del pref {wireguard_fwmark_pref} table {table_id}') == 0:
+                while (
+                    call(
+                        f'ip {afi} rule del pref {wireguard_fwmark_pref} table {table_id}'
+                    )
+                    == 0
+                ):
```
natali-rs1985: This formatting change is unrelated to the PR. Please revert it.
```suggestion
                while call(f'ip {afi} rule del pref {wireguard_fwmark_pref} table {table_id}') == 0:
```

natali-rs1985 (changes requested):

python/vyos/vpp/control_vpp.py
inetman28: You are right. The return value was not used anywhere. This is left over from when I was testing.

inetman28:
Natalie, 
I addressed all review comments in the second iteration.

src/conf_mode/interfaces_ethernet.py
```diff
@@ -473,8 +479,58 @@ def apply(ethernet):
...
+
+
+def sync_vpp_lcp_vrf_tables(ethernet: dict, vpp_api: VPPControl) -> None:
+    """Synchronize VyOS VRF assignment to VPP IP FIB table binding."""
+    for ifname, table_id in iter_vpp_lcp_vrf_targets(ethernet):
+        if not vpp_api.lcp_pair_find(vpp_name_hw=ifname):
+            continue
+
+        vpp_api.move_interface_to_ip_table_preserve_addresses(ifname, table_id)
```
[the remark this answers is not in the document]
natali-rs1985: This suggestion could be applied

natali-rs1985 (changes requested):
Several small things to do:
1. Please revert the line removal in the 'vyos-vpp.json'
2. Copilot's suggestion

inetman28:
Hi @natali-rs1985 !

Thank you for patience. 
I did all tasks.

natali-rs1985 (approved):
Thank you for your efforts and the thorough work. All my concerns were addressed.
Tested locally with different scenarios and all works correctly.

sever-sever:
Hi @inetman28 
Could you squash all commits into one?

inetman28:
@sever-sever Hi!

Done!

sever-sever (approved):
Add VRF tables synchronisation VPP with kernel

src/conf_mode/interfaces_ethernet.py
```diff
@@ -473,8 +474,36 @@ def apply(ethernet):
             mtu = e.get_mtu()
             vpp_api.set_iface_mtu(lcp_name, mtu)
 
+        sync_vpp_lcp_vrf_tables(ethernet, vpp_api)
+
     return None
 
+
+def iter_vpp_lcp_vrf_targets(ethernet: dict) -> list[tuple[str, int]]:
```
c-po: I'd make this a function definition inside `sync_vpp_lcp_vrf_tables`, as it is only use there.

c-po (approved):
Requested changes addressed

PR #5264, opened 2026-06-10
ipsec: T8975: lock vti-up-down state DB to prevent lost updates under concurrent rekey

rnavarro:
## Change summary

The IPsec vti-up-down hook DB (`/tmp/ipsec_vti_interfaces`, managed by `python/vyos/utils/vti_updown_db.py`) is read-modify-written without an inter-process lock, so concurrent hook invocations during a coordinated rekey lost-update each other and strand VTI interfaces admin-down while their CHILD_SA stays installed. This serialises all DB access by reusing `vyos.utils.locking.Lock`, wrapping the three context managers and holding the lock across `remove_vti_updown_db()`'s unlink.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8975
* https://vyos.dev/T7062 (same VTI-admin-down symptom, likely the same root cause)
* https://vyos.dev/T6574 (same script and symptom class, related but distinct trigger)

## Related PR(s)

None.

Unit test, run as CI runs it:

```
$ PYTHONPATH=python/ python3 -m nose2 -v src.tests.test_vti_updown_db
```

10 tests, all pass.

A standalone deterministic reproducer (`reproduce.py`, attached to the Phabricator task T8975 and inlined below so a reviewer can grab it either way) drives 64 concurrent up-client hook invocations. On a stock VyOS 2026.05.26-1327-rolling image it reports "LOST 63 of 64" (exit 1); with this patch in place it reports "0 lost" (exit 0).

<details>
<summary>reproduce.py (expand to view)</summary>

```python
#!/usr/bin/env python3
#
# Minimal reproduction for the vti_updown_db.py lost-update race.
#
# This drives the exact code path the IPsec updown hook uses on an
# up-client event (src/etc/ipsec.d/vti-up-down):
#
#     with open_vti_updown_db_for_create_or_update() as db:
#         db.add(interface, connection, protocol)
#         db.commit(supply_interface_dict)
#
# but from N processes released at the same instant by a barrier, and with a
# no-op interface_dict_supplier so it never touches a real interface (the
# named vtiN interfaces do not exist, so commit() only rewrites the state
# file and performs no `ip link` operations).
#
# strongSwan fires this hook once per CHILD_SA event, per VTI. During a
# coordinated rekey, N of them run concurrently. open_vti_updown_db_*()
# does an unlocked read-modify-write of /tmp/ipsec_vti_interfaces: every
# process reads the same copy, each writes back only its own addition, and
# the last writer's commit() wins. The rest are lost, and the hook that
# lost its add leaves that interface admin-down.
#
# Run as root on a VyOS instance:
#   python3 reproduce.py
#
# Stock image: prints "LOST <N-1> of <N>".  Exit code 1.
# With the patched python/vyos/utils/vti_updown_db.py in place: "0 lost". Exit 0.

import multiprocessing as mp
import os
import sys

from vyos.utils.vti_updown_db import (
    open_vti_updown_db_for_create_or_update,
    VTI_WANT_UP_IFLIST,
)

N = 64


def up_client(index, barrier):
    """What the up-client hook does, minus the real interface bring-up."""
    barrier.wait()  # release all workers into the critical section together
    with open_vti_updown_db_for_create_or_update() as db:
        db.add(f'vti{index}', f'peer-vti{index}', 'v4')
        # no-op supplier: vtiN does not exist, so commit() only rewrites the
        # state file (no `ip link`, no VTIIf instantiation).
        db.commit(lambda _interface: {'disable': True})


def main():
    # Pre-create an empty DB so every worker takes the existing-file (r+)
    # path. This isolates the lost-update race from the separate x+ create
    # race (both are fixed by the same lock).
    with open(VTI_WANT_UP_IFLIST, 'w'):
        pass

    barrier = mp.Barrier(N)
    workers = [mp.Process(target=up_client, args=(i, barrier)) for i in range(N)]
    for w in workers:
        w.start()
    for w in workers:
        w.join()

    with open(VTI_WANT_UP_IFLIST) as f:
        entries = [e for e in f.read().split() if e]
    survived = {e.split(':')[0] for e in entries}
    lost = [f'vti{i}' for i in range(N) if f'vti{i}' not in survived]

    print(f'fired {N} concurrent up-client events; '
          f'DB retained {len(survived)} of {N} interfaces')
    if lost:
        print(f'LOST {len(lost)} of {N} updates '
              f'(these interfaces would be stranded admin-down): '
              f'{", ".join(lost)}')
        sys.exit(1)
    print('0 lost: all updates serialised correctly')
    sys.exit(0)


if __name__ == '__main__':
    main()
```

</details>

Fleet validation: I deployed the patch as a hotpatch on a 12-tunnel home-edge router. Two strongswan restarts (each a full reinit storm across all 12 tunnels) produced zero stranded interfaces, and all 12 iBGP-over-VTI sessions recovered. Before the patch the same restart reliably stranded multiple VTIs.

The existing IPsec/VTI smoketests (`smoketest/scripts/cli/test_vpn_ipsec.py`, `smoketest/scripts/cli/test_interfaces_vti.py`) should be run for regression in a VyOS build environment. I have not run them yet.

## Checklist:

- [x] I have linked this PR to one or more Phabricator Task(s) (T8975)
- [ ] I have run the components [**SMOKETESTS**](https://github.com/vyos/vyos-1x/tree/rolling/smoketest/scripts/cli) if applicable (pending VyOS build environment)
- [x] My commit headlines contain a valid Task id (pending T-number)
- [ ] My change requires a change to the documentation (not required)
- [ ] I have updated the documentation accordingly (not required)

### DIFF
--- a/python/vyos/utils/vti_updown_db.py
+++ b/python/vyos/utils/vti_updown_db.py
@@ -18,55 +18,67 @@ import os
 from contextlib import contextmanager
 from syslog import syslog
 
+from vyos.utils.locking import Lock
+
 VTI_WANT_UP_IFLIST = '/tmp/ipsec_vti_interfaces'
+VTI_UPDOWN_LOCK_NAME = 'ipsec_vti_updown'
 
 def vti_updown_db_exists():
     """ Returns true if the database exists """
     return os.path.exists(VTI_WANT_UP_IFLIST)
 
 @contextmanager
+def _vti_updown_db_lock():
+    """Serialise access to the VTI up/down DB across the concurrent updown-hook
+    invocations (one per VTI) that strongSwan fires during a coordinated rekey,
+    which would otherwise lost-update the shared state file."""
+    lock = Lock(VTI_UPDOWN_LOCK_NAME)
+    lock.acquire()  # timeout=0 -> block until acquired
+    try:
+        yield
+    finally:
+        lock.release()
+
+
+@contextmanager
 def open_vti_updown_db_for_create_or_update():
     """ Opens the database for reading and writing, creating the database if it does not exist """
-    if vti_updown_db_exists():
-        f = open(VTI_WANT_UP_IFLIST, 'r+')
-    else:
-        f = open(VTI_WANT_UP_IFLIST, 'x+')
-    try:
-        db = VTIUpDownDB(f)
-        yield db
-    finally:
-        f.close()
+    with _vti_updown_db_lock():
+        mode = 'r+' if vti_updown_db_exists() else 'x+'
+        with open(VTI_WANT_UP_IFLIST, mode) as f:
+            yield VTIUpDownDB(f)
 
 @contextmanager
 def open_vti_updown_db_for_update():
     """ Opens the database for reading and writing, returning an error if it does not exist """
-    f = open(VTI_WANT_UP_IFLIST, 'r+')
-    try:
-        db = VTIUpDownDB(f)
-        yield db
-    finally:
-        f.close()
+    with _vti_updown_db_lock():
+        with open(VTI_WANT_UP_IFLIST, 'r+') as f:
+            yield VTIUpDownDB(f)
 
 @contextmanager
 def open_vti_updown_db_readonly():
-    """ Opens the database for reading, returning an error if it does not exist """
-    f = open(VTI_WANT_UP_IFLIST, 'r')
-    try:
-        db = VTIUpDownDB(f)
-        yield db
-    finally:
-        f.close()
+    """Opens the database for reading. Yields None if the database does not exist."""
+    with _vti_updown_db_lock():
+        if not vti_updown_db_exists():
+            yield None
+            return
+        with open(VTI_WANT_UP_IFLIST, 'r') as f:
+            yield VTIUpDownDB(f)
 
 def remove_vti_updown_db():
-    """ Brings down any interfaces referenced by the database and removes the database """
-    # We need to process the DB first to bring down any interfaces still up
-    with open_vti_updown_db_for_update() as db:
-        db.removeAllOtherInterfaces([])
-        # this usage of commit will only ever bring down interfaces,
-        # do not need to provide a functional interface dict supplier
-        db.commit(lambda _: None)
-
-    os.unlink(VTI_WANT_UP_IFLIST)
+    """Brings down any interfaces referenced by the database and removes the database, if it exists."""
+    with _vti_updown_db_lock():
+        if not vti_updown_db_exists():
+            return
+        # We hold the lock already; open the file directly rather than via the
+        # locking context manager to avoid re-acquiring (which would deadlock).
+        with open(VTI_WANT_UP_IFLIST, 'r+') as f:
+            db = VTIUpDownDB(f)
+            db.removeAllOtherInterfaces([])
+            # this usage of commit will only ever bring down interfaces,
+            # do not need to provide a functional interface dict supplier
+            db.commit(lambda _: None)
+        os.unlink(VTI_WANT_UP_IFLIST)
 
 class VTIUpDownDB:
     # The VTI Up-Down DB is a text-based database of space-separated "ifspecs".
--- a/python/vyos/ifconfig/vti.py
+++ b/python/vyos/ifconfig/vti.py
@@ -15,7 +15,7 @@
 
 from vyos.ifconfig.interface import Interface
 from vyos.utils.dict import dict_search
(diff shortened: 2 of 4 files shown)


python/vyos/utils/vti_updown_db.py
```diff
@@ -18,55 +18,81 @@
...
+    with _vti_updown_db_lock():
+        if vti_updown_db_exists():
+            f = open(VTI_WANT_UP_IFLIST, 'r+')
+        else:
+            f = open(VTI_WANT_UP_IFLIST, 'x+')
+        try:
+            db = VTIUpDownDB(f)
+            yield db
+        finally:
+            f.close()
```
alexandr-san4ez:
```suggestion
    with _vti_updown_db_lock():
        mode = 'r+' if vti_updown_db_exists() else 'x+'
        with open(VTI_WANT_UP_IFLIST, mode) as f:
            yield VTIUpDownDB(f)
```

python/vyos/utils/vti_updown_db.py
```diff
@@ -18,55 +18,81 @@
...
-        yield db
-    finally:
-        f.close()
+    with _vti_updown_db_lock():
+        f = open(VTI_WANT_UP_IFLIST, 'r+')
+        try:
+            db = VTIUpDownDB(f)
+            yield db
+        finally:
+            f.close()
```
alexandr-san4ez:
```suggestion
    with _vti_updown_db_lock():
        with open(VTI_WANT_UP_IFLIST, 'r+') as f:
            yield VTIUpDownDB(f)
```

python/vyos/utils/vti_updown_db.py
```diff
@@ -18,55 +18,81 @@
...
+    with _vti_updown_db_lock():
+        if not vti_updown_db_exists():
+            yield None
+            return
+        f = open(VTI_WANT_UP_IFLIST, 'r')
+        try:
+            db = VTIUpDownDB(f)
+            yield db
+        finally:
+            f.close()
```
alexandr-san4ez:
```suggestion
    with _vti_updown_db_lock():
        if not vti_updown_db_exists():
            yield None
            return
        with open(VTI_WANT_UP_IFLIST, 'r') as f:
            yield VTIUpDownDB(f)
```

python/vyos/utils/vti_updown_db.py
```diff
@@ -144,15 +170,21 @@ def removeAllOtherInterfaces(self, interface_list):
...
+            [ifspec for ifspec in self._ifspecs if ':' not in ifspec]
+        )
+        added_persistent_interfaces = (
+            new_persistent_interfaces - current_persistent_interfaces
+        )
+        removed_persistent_interfaces = (
+            current_persistent_interfaces - new_persistent_interfaces
+        )
+
+        for interface in added_persistent_interfaces:
```
alexandr-san4ez: As I see the changes are caused by the linter and affect the commit history. I would suggest to restore the code and make refactoring in separate PR.

alexandr-san4ez (changes requested):
It is a good improvment, thank you. 
I have provided a couple of suggestions for improving the code style.

alexandr-san4ez:
And to improve the clarity and history of our repository, could your squash those commits into a single? This will help in better understanding the changes when we look back at the commit history.

rnavarro:
Thanks for the feedback! I'll get it addressed

alexandr-san4ez (approved):
The requested changes were applied, my local tests (`test_vpn_ipsec.py`, `test_interfaces_vti.py`) passed:

```
vyos@vyos:~$ /usr/libexec/vyos/tests/smoke/cli/test_vpn_ipsec.py
test_dhcp_fail_handling (__main__.TestVPNIPsec.test_dhcp_fail_handling) ... ok
test_dmvpn (__main__.TestVPNIPsec.test_dmvpn) ... ok
...
test_site_to_site_x509 (__main__.TestVPNIPsec.test_site_to_site_x509) ... ok
Ran 21 tests in 171.505s

vyos@vyos:~$ sudo /usr/libexec/vyos/tests/smoke/cli/test_interfaces_vti.py
test_add_multiple_ip_addresses (__main__.VTIInterfaceTest.test_add_multiple_ip_addresses) ... ok
...
test_vif_s_protocol_change (__main__.VTIInterfaceTest.test_vif_s_protocol_change) ... skipped 'unsupported on interfac
e family'
Ran 26 tests in 70.448s
```

The PR can be merged after the second approve.

sever-sever (approved):
Fix for concurrent rekeying SA's, add lock DB

PR #5265, opened 2026-06-10
T8976: explicitly remove 'kernel' entry in ttyS0 device setting on choice of 'tty' during image install

jestabro:
## Change summary

The live build hook `95-serial.chroot` may explicitly set the device config value `kernel`. In case of image installer choice of console_type `K`, this value must be explicitly removed. _Without_ this fix, the system will (on a second reboot) reset GRUB vars console_type to `ttyS` --- as the effect of suppressed tty display output of boot messages is not apparent until the second reboot, it may be missed by those logging in by ssh after initial install.

This is due to the check here:
https://github.com/vyos/vyos-1x/blob/rolling/src/conf_mode/system_console.py#L126-L130
explicitly set for choice `S` of console_type during image install here:
https://github.com/vyos/vyos-1x/blob/rolling/src/op_mode/image_installer.py#L657-L659

It must likewise be explicitly removed for choice `K`, as there may be a residual setting from the image build live hook.

## Types of changes

Install image in VM with choice of console_type 'K' (tested in KVM). Reboot twice. Without this fix or a reasonable variant, display output of boot messages to tty will be absent.

One can check the GRUB vars after the second boot: they are overwritten to 'ttyS' despite the choice of 'K' on image install.

## Checklist:

- [X] I have read the [**CONTRIBUTING**](https://github.com/vyos/vyos-1x/blob/rolling/CONTRIBUTING.md) document

### DIFF
--- a/src/op_mode/image_installer.py
+++ b/src/op_mode/image_installer.py
@@ -658,6 +658,10 @@ def configure_serial_console(config_file: str, console_type: str) -> None:
     # serial was defined by user.
     if console_type == 'S':
         config.set(base + [device, 'kernel'])
+    if console_type == 'K':
+        # remove any residual setting from flavor file image build
+        if config.exists(base + [device, 'kernel']):
+            config.delete(base + [device, 'kernel'])
 
     with open(config_file, 'w') as f:
         f.write(config.to_string())


c-po (approved):
This is clearly a bug in the current implementation. Reverified locally with a fresh install.

dmbaturin (approved):
The fix looks straightforward to me.

PR #5266, opened 2026-06-10
T8977: Fix regression with is_addr_assigned and is_listen_port_bind_service 

MPStudyly:
## Change summary

Fix `is_addr_assigned` to consider nonlocal bind sysctl and `is_listen_port_bind_service` to ignore new bind addresses when currently bound service is listening to any address
Refactor `is_addr_assigned` to avoid semantic changes depending on parameter `return_ifname`
Refactor `get_interfaces_by_ip` to fulfil what is_addr_assigned was doing before when `return_ifname` was set
Change single usage of function `is_addr_assigned` expecting different semantics to use `get_interfaces_by_ip` instead

## Types of changes

## Related Task(s)

https://vyos.dev/T8977

I tested this by manually applying this changes to a locally running VyOS instance and running the following test sequence:
```
# Booting with cloud init without haproxy service listen address attached
~$ sudo ss -tulpn
Netid    State     Recv-Q    Send-Q              Local Address:Port          Peer Address:Port    Process   
tcp      LISTEN    0         4096                            *:443                      *:*        users:(("haproxy",pid=6265,fd=8))
...
~$ config
~$ set load-balancing haproxy service gateway_tcp listen-address '194.x.x.x'
~$ commit
~$ sudo ss -tulpn
Netid    State     Recv-Q    Send-Q              Local Address:Port          Peer Address:Port    Process   
tcp      LISTEN    0         4096               194.x.x.x:443                0.0.0.0:*        users:(("haproxy",pid=6411,fd=8))
...
~$ delete load-balancing haproxy service gateway_tcp listen-address
~$ commit
~$ sudo ss -tulpn
Netid    State     Recv-Q    Send-Q              Local Address:Port          Peer Address:Port    Process   
tcp      LISTEN    0         4096                            *:443                      *:*        users:(("haproxy",pid=6265,fd=8))
...
```

## Checklist:

- [X] I have read the [**CONTRIBUTING**](https://github.com/vyos/vyos-1x/blob/rolling/CONTRIBUTING.md) document

### DIFF
--- a/python/vyos/utils/network.py
+++ b/python/vyos/utils/network.py
@@ -471,7 +471,9 @@ def is_ipv6_link_local(addr):
     return False
 
 
-def is_addr_assigned(ip_address: str, vrf=None, include_vrf=False, allow_nonlocal: bool = False) -> bool:
+def is_addr_assigned(
+    ip_address: str, vrf=None, include_vrf=False, allow_nonlocal: bool = False
+) -> bool:
     """ Verify if the given IPv4/IPv6 address is assigned to any interface """
     from netifaces import interfaces # pylint: disable = no-name-in-module
     from vyos.utils.network import get_interface_config
@@ -479,13 +481,15 @@ def is_addr_assigned(ip_address: str, vrf=None, include_vrf=False, allow_nonloca
 
     # Check if sysctl to allow nonlocal binds is set for given afi and caller allows nonlocal binds,
     # skip check accordingly
-    if allow_nonlocal and ((
-        is_ipv4_address(ip_address)
-        and sysctl_read(['net', 'ipv4', 'ip_nonlocal_bind']) == "1"
-    ) or (
-        is_ipv6_address(ip_address)
-        and sysctl_read(['net', 'ipv6', 'ip_nonlocal_bind']) == "1"
-    )):
+    if allow_nonlocal and (
+        (
+            is_ipv4_address(ip_address)
+            and sysctl_read(['net', 'ipv4', 'ip_nonlocal_bind']) == "1"
+        ) or (
+            is_ipv6_address(ip_address)
+            and sysctl_read(['net', 'ipv6', 'ip_nonlocal_bind']) == "1"
+        )
+    ):
         return True
 
     for interface in interfaces():
@@ -768,10 +772,12 @@ def is_ipv4_address(addr: str) -> bool:
     :return: bool: True if provided address is valid
     """
     from ipaddress import ip_interface
+
     try:
         return ip_interface(addr).version == 4
     except Exception:
         return False
+
 
 def is_ipv6_address(addr: str) -> bool:
     """
@@ -780,10 +786,12 @@ def is_ipv6_address(addr: str) -> bool:
     :return: bool: True if provided address is valid
     """
     from ipaddress import ip_interface
+
     try:
         return ip_interface(addr).version == 6
     except Exception:
         return False
+
 
 def is_ipv4_range(addr: str) -> bool:
     """
@@ -799,6 +807,7 @@ def is_ipv4_range(addr: str) -> bool:
     except Exception:
         return False
 
+
 def is_ipv6_range(addr: str) -> bool:
     """
     Validates if the provided address is a valid IPv6 range.
@@ -813,6 +822,7 @@ def is_ipv6_range(addr: str) -> bool:
     except Exception:
         return False
 
+
 def is_ipv4_address_or_range(addr: str) -> bool:
     """
     Validates if the provided address is a valid IPv4, CIDR or IPv4 range
@@ -821,6 +831,7 @@ def is_ipv4_address_or_range(addr: str) -> bool:
     """
     return is_ipv4_address(addr) or is_ipv4_range(addr)
 
+
 def is_ipv6_address_or_range(addr: str) -> bool:
     """
     Validates if the provided address is a valid IPv6, CIDR or IPv6 range
@@ -828,6 +839,7 @@ def is_ipv6_address_or_range(addr: str) -> bool:
     :return: bool: True if provided address is valid
     """
     return is_ipv6_address(addr) or is_ipv6_range(addr)
+
 
 def get_interfaces_by_ip(ip_address: str, vrf=None, include_vrf: bool = False) -> list:
     """
--- a/python/vyos/utils/network.py
+++ b/python/vyos/utils/network.py
@@ -471,29 +471,31 @@ def is_ipv6_link_local(addr):
     return False
 
 
-def is_addr_assigned(ip_address: str, vrf=None, include_vrf=False) -> bool:
+def is_addr_assigned(ip_address: str, vrf=None, include_vrf=False, allow_nonlocal: bool = False) -> bool:
     """ Verify if the given IPv4/IPv6 address is assigned to any interface """
     from netifaces import interfaces # pylint: disable = no-name-in-module
     from vyos.utils.network import get_interface_config
     from vyos.utils.dict import dict_search
 
-    # Check if sysctl to allow nonlocal binds is set for given afi, skip check accordingly
-    if (
-        is_valid_ipv4_address_or_range(ip_address)
+    # Check if sysctl to allow nonlocal binds is set for given afi and caller allows nonlocal binds,
+    # skip check accordingly
+    if allow_nonlocal and ((
+        is_ipv4_address(ip_address)
         and sysctl_read(['net', 'ipv4', 'ip_nonlocal_bind']) == "1"
     ) or (
-        is_valid_ipv6_address_or_range(ip_address)
+        is_ipv6_address(ip_address)
         and sysctl_read(['net', 'ipv6', 'ip_nonlocal_bind']) == "1"
-    ):
+    )):
         return True
 
     for interface in interfaces():
         # Check if interface belongs to the requested VRF, if this is not the
         # case there is no need to proceed with this data set - continue loop
         # with next element
-        tmp = get_interface_config(interface)
-        if dict_search('master', tmp) != vrf and not include_vrf:
-            continue
+        if vrf is not None and not include_vrf and interface != vrf:
+            tmp = get_interface_config(interface)
+            if dict_search('master', tmp) != vrf:
+                continue
 
         if is_intf_addr_assigned(interface, ip_address):
             return True
@@ -510,6 +512,10 @@ def is_intf_addr_assigned(ifname: str, addr: str, netns: str=None) -> bool:
 
     from vyos.utils.process import rc_cmd
     from ipaddress import ip_interface
+
+    # Reject invalid addresses or address ranges
+    if not is_ipv4_address(addr) and not is_ipv6_address(addr):
+        return False
 
     netns_cmd = f'ip netns exec {netns}' if netns else ''
     rc, out = rc_cmd(f'{netns_cmd} ip --json address show dev {ifname}')
@@ -755,38 +761,73 @@ def get_nft_vrf_zone_mapping() -> dict:
         output.append({'interface' : vrf_name, 'vrf_tableid' : vrf_id})
     return output
 
-def is_valid_ipv4_address_or_range(addr: str) -> bool:
+def is_ipv4_address(addr: str) -> bool:
+    """
+    Validates if the provided address is a valid IPv4 address or CIDR.
+    :param addr: address to test
+    :return: bool: True if provided address is valid
+    """
+    from ipaddress import ip_interface
+    try:
+        return ip_interface(addr).version == 4
+    except Exception:
+        return False
+
+def is_ipv6_address(addr: str) -> bool:
+    """
+    Validates if the provided address is a valid IPv6 address or CIDR.
+    :param addr: address to test
+    :return: bool: True if provided address is valid
+    """
+    from ipaddress import ip_interface
+    try:
+        return ip_interface(addr).version == 6
+    except Exception:
+        return False
+
+def is_ipv4_range(addr: str) -> bool:
+    """
+    Validates if the provided address is a valid IPv4 range.
+    :param addr: address to test
+    :return: bool: True if provided address is valid
+    """
+    if '-' not in addr:
+        return False
+    try:
+        start, end = addr.split('-')
+        return is_ipv4_address(start) and is_ipv4_address(end)
+    except Exception:
+        return False
+
+def is_ipv6_range(addr: str) -> bool:
+    """
+    Validates if the provided address is a valid IPv6 range.
+    :param addr: address to test
+    :return: bool: True if provided address is valid
+    """
+    if '-' not in addr:
+        return False
+    try:
+        start, end = addr.split('-')
+        return is_ipv6_address(start) and is_ipv6_address(end)
+    except Exception:
+        return False
+
+def is_ipv4_address_or_range(addr: str) -> bool:
     """
     Validates if the provided address is a valid IPv4, CIDR or IPv4 range
     :param addr: address to test
     :return: bool: True if provided address is valid
     """
-    from ipaddress import ip_network
-    try:
-        if '-' in addr: # If we are checking a range, validate both address's individually
-            split = addr.split('-')
-            return is_valid_ipv4_address_or_range(split[0]) and is_valid_ipv4_address_or_range(split[1])
-        else:
-            return ip_network(addr).version == 4
-    except Exception:
-        return False
+    return is_ipv4_address(addr) or is_ipv4_range(addr)
 
-def is_valid_ipv6_address_or_range(addr: str) -> bool:
+def is_ipv6_address_or_range(addr: str) -> bool:
     """
-    Validates if the provided address is a valid IPv4, CIDR or IPv4 range
+    Validates if the provided address is a valid IPv6, CIDR or IPv6 range
     :param addr: address to test
     :return: bool: True if provided address is valid
     """
-    from ipaddress import ip_network
-    try:
-        if '-' in addr: # If we are checking a range, validate both address's individually
-            split = addr.split('-')
-            return is_valid_ipv6_address_or_range(split[0]) and is_valid_ipv6_address_or_range(split[1])
-        else:
-            return ip_network(addr).version == 6
-    except Exception:
-        return False
-
+    return is_ipv6_address(addr) or is_ipv6_range(addr)
 
 def get_interfaces_by_ip(ip_address: str, vrf=None, include_vrf: bool = False) -> list:
     """
@@ -807,9 +848,10 @@ def get_interfaces_by_ip(ip_address: str, vrf=None, include_vrf: bool = False) -
         # Check if interface belongs to the requested VRF, if this is not the
         # case there is no need to proceed with this data set - continue loop
         # with next element
-        tmp = get_interface_config(interface)
-        if vrf is not None and dict_search('master', tmp) != vrf and not include_vrf:
-            continue
+        if vrf is not None and not include_vrf and interface != vrf:
+            tmp = get_interface_config(interface)
+            if dict_search('master', tmp) != vrf:
+                continue
 
         if is_intf_addr_assigned(interface, ip_address):
             ifaces.append(interface)
--- a/python/vyos/utils/network.py
+++ b/python/vyos/utils/network.py
@@ -19,6 +19,7 @@ from json import loads
 from socket import AF_INET
 from socket import AF_INET6
 from vyos.utils.process import cmdl
+from vyos.utils.system import sysctl_read
 
 def _are_same_ip(one, two):
     from socket import inet_pton
@@ -398,10 +399,13 @@ def check_port_availability(address: str = None, port: int = 0,
                 s.bind(sockaddr)
                 return True # port is free to use
         except OSError:
-            return False # port is already in use
+            # This address family is busy, but another might be free.
+            # Continue to the next one.
+            pass
 
-    # if we reach this point, no socket was tested and we assume the port is
-    # already in use - better safe then sorry
+    # If we reach this point, the port is already in use
+    # or no socket was tested, in which case it is safer
+    # to assume the port being in use
     return False
 
 
@@ -445,11 +449,12 @@ def is_listen_port_bind_service(port: int, service: str, address: str = None) ->
         pid_name = process(pid).name()
         pid_port = addr.port
 
-        if has_address:
-            if address == addr.ip and port == pid_port and service == pid_name:
+        # Ignore address check when service listening on "any address",
+        # service config might change from any to limited address scope
+        if port == pid_port and service == pid_name:
+            if not has_address:
                 return True
-        else:
-            if service == pid_name and port == pid_port:
+            if addr.ip in ('::', '0.0.0.0') or address == addr.ip:
                 return True
 
     return False
@@ -465,11 +470,22 @@ def is_ipv6_link_local(addr):
 
     return False
 
-def is_addr_assigned(ip_address, vrf=None, return_ifname=False, include_vrf=False) -> bool | str:
+
+def is_addr_assigned(ip_address: str, vrf=None, include_vrf=False) -> bool:
     """ Verify if the given IPv4/IPv6 address is assigned to any interface """
     from netifaces import interfaces # pylint: disable = no-name-in-module
     from vyos.utils.network import get_interface_config
     from vyos.utils.dict import dict_search
+
+    # Check if sysctl to allow nonlocal binds is set for given afi, skip check accordingly
+    if (
+        is_valid_ipv4_address_or_range(ip_address)
+        and sysctl_read(['net', 'ipv4', 'ip_nonlocal_bind']) == "1"
+    ) or (
+        is_valid_ipv6_address_or_range(ip_address)
+        and sysctl_read(['net', 'ipv6', 'ip_nonlocal_bind']) == "1"
+    ):
+        return True
 
     for interface in interfaces():
         # Check if interface belongs to the requested VRF, if this is not the
@@ -480,7 +496,7 @@ def is_addr_assigned(ip_address, vrf=None, return_ifname=False, include_vrf=Fals
             continue
 
         if is_intf_addr_assigned(interface, ip_address):
-            return interface if return_ifname else True
+            return True
 
     return False
 
@@ -772,26 +788,30 @@ def is_valid_ipv6_address_or_range(addr: str) -> bool:
         return False
 
 
-def get_interfaces_by_ip(ip_address: str) -> list:
+def get_interfaces_by_ip(ip_address: str, vrf=None, include_vrf: bool = False) -> list:
     """
     Return a list of all interface names assigned the given IP address.
     Args:
         ip_address (str): The IP address to search for.
+        vrf (str): If specified, only interfaces in this VRF are considered.
+        include_vrf (bool): If True, interfaces from all VRFs are considered.
     Returns:
         list: List of interface names (str) that have the given IP address assigned.
               Returns an empty list if no interface has the IP assigned.
     """
     import netifaces
-    from vyos.template import is_ipv6
-
-    addr_type = AF_INET
-    if is_ipv6(ip_address):
-        addr_type = AF_INET6
+    from vyos.utils.dict import dict_search
 
     ifaces = []
     for interface in netifaces.interfaces():
-        addresses = netifaces.ifaddresses(interface)
-        for addr_info in addresses.get(addr_type, []):
-            if addr_info.get('addr') == ip_address:
-                ifaces.append(interface)
+        # Check if interface belongs to the requested VRF, if this is not the
+        # case there is no need to proceed with this data set - continue loop
+        # with next element
+        tmp = get_interface_config(interface)
+        if vrf is not None and dict_search('master', tmp) != vrf and not include_vrf:
+            continue
+
+        if is_intf_addr_assigned(interface, ip_address):
+            ifaces.append(interface)
+
     return ifaces
--- a/smoketest/scripts/cli/test_load-balancing_haproxy.py
+++ b/smoketest/scripts/cli/test_load-balancing_haproxy.py
@@ -32,6 +32,10 @@ HAPROXY_CONF = '/run/haproxy/haproxy.cfg'
 base_path = ['load-balancing', 'haproxy']
 proxy_interface = 'eth1'
 
+haproxy_service_name = 'https_front'
+haproxy_service_port = '4433'
+haproxy_backend_name = 'bk-01'
+
 valid_ca_cert = """
 MIIDnTCCAoWgAwIBAgIUewSDtLiZbhg1YEslMnqRl1shoPcwDQYJKoZIhvcNAQEL
 BQAwVzELMAkGA1UEBhMCR0IxEzARBgNVBAgMClNvbWUtU3RhdGUxEjAQBgNVBAcM
@@ -136,8 +140,6 @@ ZXLrtgVJR9W020qTurO2f91qfU8646n11hR9ObBB1IYbagOU0Pw1Nrq/FRp/u2tx
 7i7xFz2WEiQeSCPaKYOiqM3t
 """
 
-haproxy_service_name = 'https_front'
-haproxy_backend_name = 'bk-01'
 
 def parse_haproxy_config() -> dict:
     config_str = read_file(HAPROXY_CONF)
@@ -173,8 +175,13 @@ class TestLoadBalancingReverseProxy(VyOSUnitTestSHIM.TestCase):
 
     def base_config(self):
         self.cli_set(base_path + ['service', haproxy_service_name, 'mode', 'http'])
-        self.cli_set(base_path + ['service', haproxy_service_name, 'port', '4433'])
-        self.cli_set(base_path + ['service', haproxy_service_name, 'backend', haproxy_backend_name])
+        self.cli_set(
+            base_path + ['service', haproxy_service_name, 'port', haproxy_service_port]
+        )
+        self.cli_set(
+            base_path
+            + ['service', haproxy_service_name, 'backend', haproxy_backend_name]
+        )
 
         self.cli_set(base_path + ['backend', haproxy_backend_name, 'mode', 'http'])
         self.cli_set(base_path + ['backend', haproxy_backend_name, 'server', haproxy_backend_name, 'address', '192.0.2.11'])
@@ -760,6 +767,82 @@ class TestLoadBalancingReverseProxy(VyOSUnitTestSHIM.TestCase):
         # The busy address must NOT appear as a HAProxy bind
         self.assertNotIn(f'bind {addr_busy}:{shared_port}', config)
         self.assertIn(f'bind [{addr_listen}]:{shared_port}', config)
+
+    def test_reverse_proxy_listen_address_lifecycle(self):
+        """T8977: adding then removing a listen-address must revert haproxy to wildcard."""
+        from vyos.utils.network import check_port_availability
+
+        # Use loopback so no real NIC is required
+        addr_v4 = '127.0.0.1'
+        addr_v6 = '::1'
+        svc_laddr_base = base_path + ['service', haproxy_service_name, 'listen-address']
+
+        # Pre-test before service binds
+        # Check port availability without address (default to 'any'), must yield true
+        self.assertTrue(
+            check_port_availability(port=int(haproxy_service_port), protocol='tcp')
+        )
+
+        # Start without a specific listen-address (listen on 'any')
+        self.base_config()
+        self.cli_commit()
+
+        # Check port availability without address (default to 'any'), must yield false
+        self.assertFalse(
+            check_port_availability(port=int(haproxy_service_port), protocol='tcp')
+        )
+
+        # Set a addr_v4 listen-address
+        self.cli_set(svc_laddr_base + [addr_v4])
+        self.cli_commit()
+
+        # Check port availability without address (default to 'any'), must yield false
+        self.assertFalse(
+            check_port_availability(port=int(haproxy_service_port), protocol='tcp')
+        )
+        # Check port availability on addr_v4, must yield false
+        self.assertFalse(
+            check_port_availability(
+                address=addr_v4, port=int(haproxy_service_port), protocol='tcp'
+            )
+        )
+        # Check port availability on addr_v6, must yield true
+        self.assertTrue(
+            check_port_availability(
+                address=addr_v6, port=int(haproxy_service_port), protocol='tcp'
+            )
+        )
+
+        # Set a addr_v6 listen-address
+        self.cli_delete(svc_laddr_base)
+        self.cli_set(svc_laddr_base + [addr_v6])
+        self.cli_commit()
+
+        # Check port availability without address (default to 'any'), must yield false
+        self.assertFalse(
+            check_port_availability(port=int(haproxy_service_port), protocol='tcp')
+        )
+        # Check port availability on addr_v4, must yield true
+        self.assertTrue(
+            check_port_availability(
+                address=addr_v4, port=int(haproxy_service_port), protocol='tcp'
+            )
+        )
+        # Check port availability on addr_v6, must yield false
+        self.assertFalse(
+            check_port_availability(
+                address=addr_v6, port=int(haproxy_service_port), protocol='tcp'
+            )
+        )
+
+        # Remove the specific listen-address
+        self.cli_delete(svc_laddr_base)
+        self.cli_commit()
+
+        # Check port availability without address (default to 'any'), must yield false
+        self.assertFalse(
+            check_port_availability(port=int(haproxy_service_port), protocol='tcp')
+        )
 
 
 if __name__ == '__main__':
--- a/src/services/vyos-domain-resolver
+++ b/src/services/vyos-domain-resolver
@@ -31,7 +31,7 @@ from vyos.utils.convert import human_to_seconds
 from vyos.utils.dict import dict_search_args
 from vyos.utils.kernel import WIREGUARD_REKEY_AFTER_TIME
 from vyos.utils.file import makedir, chmod_775, write_file, read_file
-from vyos.utils.network import is_valid_ipv4_address_or_range, is_valid_ipv6_address_or_range
+from vyos.utils.network import is_ipv4_address_or_range, is_ipv6_address_or_range
 from vyos.utils.process import cmdl
 from vyos.utils.process import run
 from vyos.xml_ref import get_defaults
@@ -199,9 +199,9 @@ def update_remote_group(config):
                 if not line_first_word:
                     continue
 
-                if is_valid_ipv4_address_or_range(line_first_word):
+                if is_ipv4_address_or_range(line_first_word):
                     ip_list.append(line_first_word)
-                elif is_valid_ipv6_address_or_range(line_first_word):
+                elif is_ipv6_address_or_range(line_first_word):
                     ip6_list.append(line_first_word)
                 elif line_first_word[0].isalnum():
                     invalid_list.append(line_first_word)
--- a/src/tests/test_utils.py
+++ b/src/tests/test_utils.py
@@ -13,7 +13,16 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from unittest import TestCase
-from unittest.mock import patch
+from unittest.mock import patch, MagicMock
+
+
+def _make_psutil_net_conn(ip, port, pid):
+    c = MagicMock()
+    c.laddr.ip = ip
+    c.laddr.port = port
+    c.pid = pid
+    return c
+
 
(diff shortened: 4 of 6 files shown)


python/vyos/utils/network.py
```diff
@@ -407,7 +408,9 @@ def is_listen_port_bind_service(port: int, service: str, address: str = None) ->
         pid_name = process(pid).name()
         pid_port = addr.port
 
-        if has_address:
+        # Ignore address check when service listening on "any address",
+        # service config might change from any to limited address scope
+        if has_address and addr.ip != '::':
```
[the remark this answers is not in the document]
MPStudyly: Good catch! Will fix.

python/vyos/utils/network.py
```diff
@@ -433,6 +436,11 @@ def is_addr_assigned(ip_address, vrf=None, return_ifname=False, include_vrf=Fals
     from vyos.utils.network import get_interface_config
     from vyos.utils.dict import dict_search
 
+    # Check if sysctl to allow nonlocal binds is set for given afi, skip check accordingly
+    if ((is_valid_ipv4_address_or_range(ip_address) and sysctl_read(['net', 'ipv4', 'ip_nonlocal_bind']) == "1") or
+            (is_valid_ipv6_address_or_range(ip_address) and sysctl_read(['net', 'ipv6', 'ip_nonlocal_bind']) == "1")):
+        return True
```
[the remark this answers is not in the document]
MPStudyly: I need some help to solve this. When the sysctl is set, it is explicitly allowed to bind without an interface. I assume `template.py` somehow must consider this case as well?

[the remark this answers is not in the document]
MPStudyly: IMHO the sysctl should have precedence over a request for a specific interface. This exact sysctl is meant to allow binding to IP addresses that are not (yet) locally available, even RedHat mentions this in their documentation: https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/load_balancer_administration/s1-initial-setup-forwarding-vsa
Thus I believe code paths relying on getting an actual interface from this function must be adapted accordingly, as in some cases there won't be an interface to be returned. It would help to know which exact code paths are affected by this, but my time for analysis is limited, thus I'd appreciate any input (but especially from humans with specific knowledge around these code paths) for how to properly resolve this.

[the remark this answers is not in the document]
MPStudyly: If you are 100% sure Kea is the only caller requiring different semantics and expecting it as such, I will perform the change. However, I'm even more convinced this is solved in a quite unclean manner. Even if documented properly, having completely different semantics depending on parameters is considered bad practice (check e.g. https://dev.to/danialmalik/a-brief-guide-to-clean-code-functions-104h#:~:text=Flag%20Argument,violation%20of%20the%20SRP%20principle.). Either the caller needs to be adapted to properly handle the main semantics (i.e. having Kea properly digest "True", handling it the same as plain "False"; I'd prefer this) or create a second function dedicated to be used by Kea (and any other caller that needs the return_ifname semantics).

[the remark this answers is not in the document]
MPStudyly: As you made sure the mentioned function in `template.py` is the only caller, I've chosen option 2. Though I've noticed there is already a function `get_interfaces_by_ip` in `network.py`. I've decided to extend that one and make the call in `template.py` to handle it accordingly. You will see my changes when pushed.

MPStudyly:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> ⚠️ Outside diff range comments (2)
> 
> > python/vyos/utils/network.py (2)> `449-455`: _⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_
> > > **Include the VRF master interface in VRF-scoped lookups.**
> > > Line 454 and Line 771 only test `dict_search('master', tmp)`. That matches VRF _slaves_ but not the VRF device itself, so `is_addr_assigned(..., vrf='red')` and `get_interfaces_by_ip(..., vrf='red')` will miss an address configured directly on `red`.
> > > 
> > > Proposed fix
> > > ```diff
> > > -        tmp = get_interface_config(interface)
> > > -        if dict_search('master', tmp) != vrf and not include_vrf:
> > > +        tmp = get_interface_config(interface)
> > > +        master = dict_search('master', tmp)
> > > +        is_vrf_device = (
> > > +            dict_search('linkinfo.info_kind', tmp) == 'vrf' and interface == vrf
> > > +        )
> > > +        if master != vrf and not include_vrf and not is_vrf_device:
> > >              continue
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ```diff
> > > -        tmp = get_interface_config(interface)
> > > -        if vrf is not None and dict_search('master', tmp) != vrf and not include_vrf:
> > > +        tmp = get_interface_config(interface)
> > > +        master = dict_search('master', tmp)
> > > +        is_vrf_device = (
> > > +            dict_search('linkinfo.info_kind', tmp) == 'vrf' and interface == vrf
> > > +        )
> > > +        if vrf is not None and master != vrf and not include_vrf and not is_vrf_device:
> > >              continue
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > Also applies to: 766-772
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@python/vyos/utils/network.py` around lines 449 - 455, The VRF-scoped
> > > filtering currently only checks dict_search('master', tmp) == vrf and thus omits
> > > the VRF device itself; update the filter in the loop that iterates interfaces()
> > > (where get_interface_config(interface) is stored in tmp and the condition uses
> > > dict_search('master', tmp') != vrf and not include_vrf) to also treat the
> > > interface as in the VRF when the interface name equals the vrf (e.g. change the
> > > condition to skip only when neither dict_search('master', tmp) == vrf nor
> > > interface == vrf, unless include_vrf is True). Apply the same change to the
> > > analogous block around lines 766-772 so functions like is_addr_assigned and
> > > get_interfaces_by_ip include addresses configured directly on the VRF device.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 
> > > `400-418`: _⚠️ Potential issue_ | _🟠 Major_
> > > **Guard `psutil.Process(pid).name()` against stale PIDs in `python/vyos/utils/network.py` (lines 400-418).**
> > > `psutil.net_connections()` and `psutil.Process(pid).name()` aren’t atomic; the PID can disappear or become inaccessible, and the current unhandled call can raise and fail validation.
> > > 
> > > Proposed fix
> > > ```diff
> > > -    from psutil import net_connections as connections
> > > -    from psutil import Process as process
> > > +    from psutil import net_connections as connections
> > > +    from psutil import Process as process
> > > +    from psutil import AccessDenied, NoSuchProcess, ZombieProcess
> > > @@
> > > -        pid_name = process(pid).name()
> > > +        try:
> > > +            pid_name = process(pid).name()
> > > +        except (AccessDenied, NoSuchProcess, ZombieProcess):
> > > +            continue
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@python/vyos/utils/network.py` around lines 400 - 418, The loop calling
> > > process(pid).name() can raise if the PID disappears or is inaccessible; wrap the
> > > call to psutil.Process(pid).name() (the process(pid).name() invocation used when
> > > iterating connections()) in a try/except that catches psutil.NoSuchProcess and
> > > psutil.AccessDenied (and optionally psutil.ZombieProcess) and continue to the
> > > next connection on exception, so the function safely skips stale/inaccessible
> > > PIDs before comparing pid_name/pid_port against address/port/service and
> > > has_address logic.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> 
> ℹ️ Review info
> 
> 
> 
> 📜 Review details

Can some human please check this complaint? As I'm just reusing code, this either already was an issue or the bot is hallucinating. I'm lacking the insights to evaluate this though. For the case this already was an issue, should I apply the suggestion as is? I assume on both functions?

MPStudyly:
> **Actionable comments posted: 1**
> 
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> ⚠️ Outside diff range comments (1)
> 
> > python/vyos/utils/network.py (1)> `439-445`: _⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_
> > > **`is_addr_assigned()` now partially accepts ranges but still crashes on the fallback path.**
> > > Lines 440-444 treat hyphenated IPv4/IPv6 ranges as valid inputs, but when the corresponding `ip_nonlocal_bind` sysctl is not `"1"` the code drops into `is_intf_addr_assigned()`, which calls `ip_interface(addr)` and raises on `a-b` ranges. Either reject ranges here or make the interface scan range-aware; right now the result depends on sysctl state instead of the input alone.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@python/vyos/utils/network.py` around lines 439 - 445, The function
> > > is_addr_assigned() currently treats hyphenated IP ranges as "valid" but then may
> > > call is_intf_addr_assigned() which uses ip_interface(addr) and crashes on
> > > ranges; fix by making range handling deterministic: detect ranges early (e.g.,
> > > if '-' in ip_address or via existing range-check helper) and do not fall through
> > > to is_intf_addr_assigned(); either return False or explicitly handle ranges by
> > > expanding the range and checking each address with ip_interface() (or calling
> > > is_intf_addr_assigned() per-expanded address). Update is_addr_assigned() to
> > > consult sysctl_read(['net','ipv4|ipv6','ip_nonlocal_bind']) only for single
> > > addresses and ensure is_intf_addr_assigned() is never called with a raw range
> > > string (so references: is_addr_assigned(), is_intf_addr_assigned(),
> > > ip_interface(), sysctl_read()).
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> 
> 🪄 Autofix (Beta)
> 
> 
> ℹ️ Review info
> 
> 
> 
> 📜 Review details

Hmm, not sure this really is an issue. Sure, the functions accept a range, but as far as I understand, it shouldn't matter at this point, should it?

MPStudyly:
@coderabbitai any suggestions on smoke tests for these changes, ensuring this doesn't break again?

alexandr-san4ez (changes requested):
Thanks for submitting this PR.

1. To improve the clarity and history of our repository, could you squash those commits into a single?
2. Also, it would be great if you could include a smoke test to cover the changes implemented in this PR.

MPStudyly:
> Thanks for submitting this PR.
> 
>     1. To improve the clarity and history of our repository, could you squash those commits into a single?
> 
>     2. Also, it would be great if you could include a smoke test to cover the changes implemented in this PR.

1. I planned to do so when everything is finalized. Explicitly kept a few changes in their dedicated PR to keep it simpler to revert.
2. Already on it :)

MPStudyly:
> > Thanks for submitting this PR.
> > ```
> > 1. To improve the clarity and history of our repository, could you squash those commits into a single?
> > 
> > 2. Also, it would be great if you could include a smoke test to cover the changes implemented in this PR.
> > ```
> 
>     1. I planned to do so when everything is finalized. Explicitly kept a few changes in their dedicated PR to keep it simpler to revert.
> 
>     2. Already on it :)

Done, waiting for review :)

I mostly took the testcases coderabbitai proposed. Added in another unit test to fill in a blind spot and slightly adapted structure of the proposed tests.

src/tests/test_utils.py
```diff
@@ -86,3 +95,81 @@ def test_list_to_range_str(self):
...
+
+        with patch('vyos.utils.network.sysctl_read', return_value='0'), \
+                patch('netifaces.interfaces', return_value=['lo', 'eth0']), \
+                patch('vyos.utils.network.get_interface_config',
+                      side_effect=[iface_cfg_lo, iface_cfg_red]), \
+                patch('vyos.utils.network.is_intf_addr_assigned',
+                      side_effect=[False, True]):
+            # vrf='red' → should find it on eth0
+            self.assertTrue(is_addr_assigned('10.0.0.1', vrf='red'))
+
```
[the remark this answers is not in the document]
MPStudyly: I see, I had some issues understanding this peace to begin with, but went with it as you proposed it. Will adapt accordingly.

src/tests/test_utils.py
```diff
@@ -86,3 +95,81 @@ def test_list_to_range_str(self):
...
+        """When sysctl is 0 and address is not assigned, must return False."""
+        from vyos.utils.network import is_addr_assigned
+        with patch('vyos.utils.network.sysctl_read', return_value='0'):
+            self.assertFalse(is_addr_assigned('192.0.2.99'))
+
+    def test_is_addr_assigned_nonlocal_bind_ipv6_off_unassigned(self):
+        """When sysctl is 0 and address is not assigned, must return False."""
+        from vyos.utils.network import is_addr_assigned
+        with patch('vyos.utils.network.sysctl_read', return_value='0'):
+            self.assertFalse(is_addr_assigned('2001:db8::1'))
```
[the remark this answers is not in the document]
MPStudyly: This shouldn't be necessary, as the used addresses are reserved for documentation purposes only. It can't hurt making sure though.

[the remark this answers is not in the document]
MPStudyly: I see now, there should be no impact (or just some safeguard) for negative tests, but positive tests explicitly should pass, even when no interfaces available. As the code is right now, that can't happen, as the sysctl check acts as a shortcut, but if the function ever gets adapted to check interfaces beforehand, this could break. Not 100% sure though I should write tests with such assumptions, thus I'll leave it out on the positive tests for now.

MPStudyly:
Fixed all remaining non insane linter issues. Can you have a look at the left over ones? They seem a little ridiculous to me, i.e. making the code less readable than the current state. :sweat_smile:

MPStudyly:
@coderabbitai any ideas why the smoke-test is failing on the first assertion? As haproxy must be bound to a wildcard at that point, the function should return true. Only thing I can think of is 'haproxy' not being the correct process name. However, I might overlook other relevant parts. Can you have a look?

MPStudyly:
@coderabbitai ~~you are wrong with your analysis~~, I tested my changes on a live system where the pid_name actually resolves to "haproxy". I'm a bit lost now, as I don't understand what differs between the live environment and the CI pipeline. I'd appreciate any input @alexandr-san4ez.

EDIT: The analysis was not wrong in terms of what the comparison is doing (i.e. it indeed looks brittle), it was just wrong about what `pid_name` was going to be.

python/vyos/utils/network.py
MPStudyly, line 447: @coderabbitai this is causing the failure of the smoke-test, as in the scope of those, the PID can't be determined. I assume there is a lack of privileges, but running the single test instance with `sudo` locally failed with issues on saving a config file, so it doesn't seem to be that simple. Any ideas how this could be solved?

MPStudyly: I need a way to ensure this part of the code is run privileged during smoke-tests. I'll try to come up with something along `with patch()` like in the unit tests, but I know already that's going to be very ugly.

smoketest/scripts/cli/test_load-balancing_haproxy.py
```diff
@@ -728,6 +782,52 @@ def test_reverse_proxy_listen_address_no_port_conflict(self):
...
+        with patch('psutil.net_connections', return_value=_net_connections()):
+            self.assertTrue(
+                is_listen_port_bind_service(int(port), PROCESS_NAME, address=probe_addr)
+            )
+
+        # Add a specific listen-address
+        self.cli_set(svc_base + ['listen-address', addr])
+        self.cli_commit()
+
+        conns = _net_connections()
```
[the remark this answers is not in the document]
MPStudyly: Was an oversight, it's already gone with latest push.

smoketest/scripts/cli/test_load-balancing_haproxy.py
```diff
@@ -728,6 +782,52 @@ def test_reverse_proxy_listen_address_no_port_conflict(self):
...
+        bknd_srv_base = base_path + ['backend', backend, 'server', 'srv1']
+        self.cli_set(bknd_srv_base + ['address', '127.0.0.2'])
+        self.cli_set(bknd_srv_base + ['port', '9999'])
+        self.cli_commit()
+
+        # Verify haproxy is listening on wildcard
+        with patch('psutil.net_connections', return_value=_net_connections()):
+            self.assertTrue(
+                is_listen_port_bind_service(int(port), PROCESS_NAME, address=probe_addr)
+            )
```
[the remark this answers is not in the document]
MPStudyly: Other smoke tests do not poll for an actual bind either. I think commit already ensures to return only, when the service actually started (and thus bound itself). Otherwise quite some more tests than just mine would rely on good timing by accident. I won't introduce this as long as it is not clear if it should be applied on other tests as well.

alexandr-san4ez:
@MPStudyly  I found only one interesting thing:
```
vyos@vyos2:~$ python3
>>> from psutil import net_connections as connections
>>> from psutil import Process as process
>>> for connection in connections(kind='tcp'):
...     pid = connection.pid
...     pid_name = process(pid).name()
...     pid_port = addr.port
...     print(pid, pid_name, pid_port)
...
None python3 2612
None python3 2605
None python3 2616
None python3 2606
None python3 2621
None python3 22
None python3 22
None python3 22
None python3 2623
None python3 22
None python3 2622
None python3 2608
None python3 2602
None python3 2609
None python3 2603
None python3 2610
None python3 2617
None python3 2618
None python3 2601
None python3 2604
```

This looks like a problem with "psutil" or something doesn't allow "psutil" work correctly.

MPStudyly:
> @MPStudyly I found only one interesting thing:
> 
> ```
> vyos@vyos2:~$ python3
> >>> from psutil import net_connections as connections
> >>> from psutil import Process as process
> >>> for connection in connections(kind='tcp'):
> ...     pid = connection.pid
> ...     pid_name = process(pid).name()
> ...     pid_port = addr.port
> ...     print(pid, pid_name, pid_port)
> ...
> None python3 2612
> None python3 2605
> None python3 2616
> None python3 2606
> None python3 2621
> None python3 22
> None python3 22
> None python3 22
> None python3 2623
> None python3 22
> None python3 2622
> None python3 2608
> None python3 2602
> None python3 2609
> None python3 2603
> None python3 2610
> None python3 2617
> None python3 2618
> None python3 2601
> None python3 2604
> ```
> 
> This looks like a problem with "psutil" or something doesn't allow "psutil" work correctly.

It's privileges. I'm sorry I got the conversation thread a bit messy interacting with coderabbit so intensively :sweat_smile: I already mentioned that being the issue in a thread highlighting the relevant code point at network.py:405. When you try running the same code but launching python with sudo, it will work as expected. The middleware usually runs as elevated user, so it never encounters this issue, while the smoke-tests don't.

I've already pushed a flawed version of a helper function working around this by invoking python and net_connections() as sudo limited to the scope of the smoke-tests. However, I really hate this approach, as it introduces a few brittle points. At least it should be quite obvious when something breaks, as the test will fail, but still. Let me finish up the fix for my current solution, then you can have a look.

MPStudyly:
@alexandr-san4ez there you go, the current version works around this in an IMHO ugly way, but I was unable to come up with something nicer :/

python/vyos/utils/network.py
MPStudyly: Solved with ugly workaround for now.

MPStudyly: I don't think `ss` would fix anything here, as it suffers the same limitations as `psutil.net_connections()`. I'd say adding `CAP_SYS_PTRACE` to test runs would be the more elegant solution, even though that doesn't 100% replicate a running instance as well. In any case, a discussion about this and further planning is needed before taking any action.

MPStudyly:
Squashed and rebased

MPStudyly:
@alexandr-san4ez After reading through all other smoke-tests for HAProxy, I came up with a new idea how to test the desired behavior without the need of a brittle privileged helper function. I scrapped the helper, made use of the already existing base setup and gave it a shot. Will either squash my recent changes with the previous approach or revert based on your feedback.

smoketest/scripts/cli/test_load-balancing_haproxy.py
```diff
@@ -728,6 +735,70 @@ def test_reverse_proxy_listen_address_no_port_conflict(self):
...
+        # Check port availability without address (default to 'any'), must yield true
+        self.assertTrue(
+            check_port_availability(port=int(haproxy_service_port), protocol='tcp')
+        )
+        # Check port availability on listen-address, must yield false
+        self.assertFalse(
+            check_port_availability(
+                address=addr, port=int(haproxy_service_port), protocol='tcp'
+            )
+        )
```
MPStudyly: @coderabbitai why does `check_port_availability` yield false when testing without an address given? Shouldn't the function return true in this case, as it assumes '::' as bind address. Given the IPv4 localhost address the service is bound to, `check_port_availability` should definitely return true, especially after my fix to the function returning early.

MPStudyly: I see, so my previous modification to the function doesn't make any sense, right? I understand this as planned by the devs to avoid the function returning true in ambiguous scenarios, e.g. exactly the one I provided. I will instead adapt the test to explicitly ask for a v6 address on the same port to prove my point in the smoke-test. Should I revert the changes to `check_port_availability`?

MPStudyly:
Squashed again, waiting on feedback which variant of the smoke-tests to take.

MPStudyly:
@alexandr-san4ez @c-po @dmbaturin any chance you can have a look at this soonish? :innocent:

src/tests/test_utils.py
```diff
@@ -13,7 +13,16 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from unittest import TestCase
-from unittest.mock import patch
+from unittest.mock import patch, MagicMock
+
+
+def _make_conn(ip, port, pid):
```
alexandr-san4ez:
```suggestion
def _make_psutil_net_conn(ip, port, pid):
```

alexandr-san4ez (commented):
Could you squash those commits into a single and describe all changes in the commit message body?

alexandr-san4ez:
@alexk37 As far as I remember you had test results for these changes. Could you please provide here?

MPStudyly:
> Could you squash those commits into a single and describe all changes in the commit message body?

Will do, but first I need an answer to:

> ... waiting on feedback which variant of the smoke-tests to take.

As otherwise the initial approach via elevated helper process and mocking gets lost (I'd prefer that, even though that variant was somewhat more accurate).

alexandr-san4ez (commented):
The last thing I see is a linter error ([darker](https://github.com/vyos/vyos-1x/actions/runs/28372381494/job/84053147045?pr=5266#step:7:1)).

MPStudyly:
> The last thing I see is a linter error ([darker](https://github.com/vyos/vyos-1x/actions/runs/28372381494/job/84053147045?pr=5266#step:7:1)).

I mentioned these quite a few messages above, should I really tackle them like suggested? To me the proposed solution looks worse than the current state. Just wanted to have a clear "solve it" or "keep it".

I'd also still need feedback on my last question, that's why I re-triggered the review:

> > ... waiting on feedback which variant of the smoke-tests to take.
> 
> As otherwise the initial approach via elevated helper process and mocking gets lost (I'd prefer that, even though that variant was somewhat more accurate).

I really try to finish this PR as fast as possible, as our local workaround for the regression is quite brittle. I wonder nobody else noticed this already though. Maybe HA setups via VRRP are not that common (at least on nightlies).

alexandr-san4ez:
> I mentioned these quite a few messages above, should I really tackle them like suggested? To me the proposed solution looks worse than the current state. Just wanted to have a clear "solve it" or "keep it".

My opinion on this matter is to follow the linter rules or change your code in a different way to pass the linter check.

> ... waiting on feedback which variant of the smoke-tests to take.

The current version of the tests seems sufficient for verification and I do not see any objections. If I correct understand you question.

MPStudyly:
> > I mentioned these quite a few messages above, should I really tackle them like suggested? To me the proposed solution looks worse than the current state. Just wanted to have a clear "solve it" or "keep it".
> 
> My opinion on this matter is to follow the linter rules or change your code in a different way to pass the linter check.

Okay, will do ASAP :+1: 


> > ... waiting on feedback which variant of the smoke-tests to take.
> 
> The current version of the tests seems sufficient for verification and I do not see any objections. If I correct understand you question.

Thank you, that's indeed mostly it. I still prefer the other approach in terms of actual coverage, but really hate the helper function and output parsing necessary to make that work. The best thing would be to get the tests to run with enough permissions/capabilities so a workaround isn't necessary anymore, but that is something I would tackle outside of this PR.

I'll do the squash as last step, will re-request the review when done and CI is happy.

alexandr-san4ez (dismissed):
The change looks well integrated and ready to merge but requires additional review by QA.

MPStudyly:
Rebased on rolling to resolve import conflict. CI not running due to changes in security policy?

MPStudyly:
I'm terribly sorry for being this impatient, but is there anything I can do to get this looked at by QA soonish? :sweat_smile: I'd offer help to solve the testing required tag as well if I can do so :)

AnDroed09:
@MPStudyly Could you please rebase to trigger CI again?

AnDroed09:
@MPStudyly , integration build with this fix regression is resolved, but only if `net.ipv4.ip_nonlocal_bind` is set to `1` explicitly. Without it, the commit on a backup node still fails with the same error

```
[ load-balancing haproxy ]
listen-address "10.10.0.100" not assigned on any interface!
[[load-balancing haproxy]] failed
Commit failed
```

Is the expectation that user should `add set system sysctl parameter net.ipv4.ip_nonlocal_bind value '1' `themselves as a required companion step?

MPStudyly:
> Is the expectation that user should `add set system sysctl parameter net.ipv4.ip_nonlocal_bind value '1' `themselves as a required companion step?

Exactly! That is a prerequisite to allow configuration of a non assigned IP address. Now that I think of it, this probably should be reflected in the documentation. If you can give me some pointers towards where I need to adjust descriptive text, I can have a look at it along fixing the co-pilot raised issue.

alexk37:
@MPStudyly please check the following regressions:

## Part 1 — BGP commit fails

1. Log in to a router that has PR 5266 applied.
2. `sudo sysctl -w net.ipv4.ip_nonlocal_bind=1`
3. `configure`
4. `set protocols bgp system-as 65001`
5. `set protocols bgp neighbor 192.0.2.1 remote-as 65002`
6. `commit`
7. **Observed failure:**
   ```
   Cannot configure local address as neighbor "192.0.2.1" in default VRF!
   [[protocols bgp]] failed
   Commit failed
   ```
8. `exit discard`
9. Repeat steps 2–6 on a router **without** PR 5266 → the commit succeeds.
   192.0.2.1 is not assigned to any interface, so it must be accepted as a neighbor.

On a router that already has BGP neighbors, the error names an **existing** neighbor, so
every commit touching `protocols bgp` fails — not only one that adds a neighbor.


## Part 2 — the sysctl gets set without any operator action

1. `sudo sysctl -w net.ipv4.ip_nonlocal_bind=0`
2. `sudo sysctl -n net.ipv4.ip_nonlocal_bind` → **`0`**
3. `sudo openvpn --genkey secret /dev/stdout | grep -v '^#' | grep -v BEGIN | grep -v END | tr -d '\n'`
   (keep the hex string it prints)
4. `configure`
5. `set pki openvpn shared-secret psk1 key <hex-from-step-3>`
6. `set pki openvpn shared-secret psk1 version 1`
7. `set interfaces openvpn vtun0 mode site-to-site`
8. `set interfaces openvpn vtun0 shared-secret-key psk1`
9. `set interfaces openvpn vtun0 encryption data-ciphers-fallback aes256`
10. `set interfaces openvpn vtun0 remote-host 192.0.2.1`
11. `set interfaces openvpn vtun0 local-host 203.0.113.77`
    — an address **not** assigned to this router (normal for VRRP/HA and cloud floating
    addresses)
12. `set interfaces openvpn vtun0 local-address 10.255.1.1`
13. `set interfaces openvpn vtun0 remote-address 10.255.1.2`
14. `commit` → succeeds
15. `save`
16. `exit`
17. `sudo sysctl -n net.ipv4.ip_nonlocal_bind` → **`1`**, set by
    `src/conf_mode/interfaces_openvpn.py:851`
18. Repeat Part 1 steps 3–6 → the BGP commit fails, with no manual sysctl involved.
19. `configure` → `delete interfaces openvpn vtun0` → `commit` → `save` → `exit`
20. `sudo sysctl -n net.ipv4.ip_nonlocal_bind` → **still `1`**; it is never reset, so the
    trigger outlives the OpenVPN configuration that created it.


## Part 3 — BGP disappears at boot

Requires the sysctl to be set *during* boot, so it uses the OpenVPN configuration from
Part 2 (a plain `sysctl -w` does not survive a reboot and will not reproduce this).

1. Keep the Part 2 OpenVPN configuration in place (re-add it if you deleted it in step 19).
2. `sudo sysctl -w net.ipv4.ip_nonlocal_bind=0`
3. `configure`
4. `set protocols bgp system-as 65001`
5. `set protocols bgp neighbor 192.0.2.1 remote-as 65002`
6. `commit`
7. `save`
8. `exit`
9. `reboot` and confirm.
10. After boot, wait about a minute.
11. `cli-shell-api existsActive protocols bgp ; echo rc=$?` → **`rc=1`**: the whole
    `protocols bgp` subtree is gone from the running configuration (unpatched image: `rc=0`).
12. `show configuration commands | match bgp` → **nothing**
13. `sudo vtysh -c 'show running-config' | grep 'router bgp'` → **`router bgp 65001` is still
    there**: FRR is running configuration the CLI no longer knows about.
14. `sudo journalctl -b | grep ERROR_COMMIT` → the boot commit failed.
15. `sudo sysctl -n net.ipv4.ip_nonlocal_bind` → **`1`**: OpenVPN (priority 460) set it
    before `protocols_bgp` (priority 820) ran.

From this state, both verified: any successful commit that touches FRR removes the
still-running BGP sessions, and `save` writes the BGP-less configuration to
`/config/config.boot` permanently.


## Root cause

`python/vyos/utils/network.py`, added by this PR:

```python
if (is_valid_ipv4_address_or_range(ip_address)
    and sysctl_read(['net', 'ipv4', 'ip_nonlocal_bind']) == "1") or (...ipv6...):
    return True
```

The early return happens before any interface lookup, so with the sysctl set
`is_addr_assigned()` answers True for **every syntactically valid address of that family** —
including ranges/prefixes, since the guard is `is_valid_ipv{4,6}_address_or_range()` — and it
ignores the `vrf` argument it was passed (the sysctl is netns-global). The two families are
independent: the v4 sysctl does not affect IPv6 neighbors, and vice versa.

`src/conf_mode/protocols_bgp.py:382` calls the helper in the opposite direction:

```python
if is_ip(peer) and is_addr_assigned(peer, vrf):
    raise ConfigError(f'Cannot configure local address as neighbor "{peer}"{vrf_error_msg}')
```

so every BGP neighbor is treated as a local address. Also reproduces with
`net.ipv6.ip_nonlocal_bind=1` plus an IPv6 neighbor, and inside a VRF
(`set vrf name red protocols bgp neighbor …` → `… in VRF "red"!`).

Suggested fix: make the bypass opt-in per call site — e.g.
`is_addr_assigned(addr, vrf, allow_nonlocal=True)`, passed only by the verifies for which a
non-local bind is acceptable — and honour the `vrf` argument.

MPStudyly:
Thank you a lot for the in-depth analysis and feedback! I'll look forward to get this done ASAP, as I will be on holiday in the coming two weeks. However I'm left with some questions about the sysctl handling. Especially as:

> ## Part 2 — the sysctl gets set without any operator action

introduces/handles the same sysctl but hidden from the user, should I still pursue my solution of requiring a user to explicitly set it themself before trying to configure e.g. HAProxy? Or should I make it "transparent" as well, i.e. let HAProxy config set it on demand just like OVPN is doing? As there is the issue mentioned with:

>     20. `sudo sysctl -n net.ipv4.ip_nonlocal_bind` → **still `1`**; it is never reset, so the
>         trigger outlives the OpenVPN configuration that created it.

I'd actually tend to eliminate those implicit usages and actively tell users to enable the according sysctl before attempting to do nonlocal binds in the first place. However I don't know if I should tackle such a refactoring in the scope of this PR or maybe a dedicated one or if I should tackle this at all to begin with?

> ## Part 3 — BGP disappears at boot
> 
> Requires the sysctl to be set _during_ boot, so it uses the OpenVPN configuration from Part 2 (a plain `sysctl -w` does not survive a reboot and will not reproduce this).

This is kinda obvious and would need a generic fix. I already had disappearing configs on boot as well, as the config parser simply drops stuff it was unable to load, even though it might have caused side effects (e.g. set up an interface, started a service, etc.). Config loading errors should be handled by rolling back whatever was done in the failing segment instead of simply dropping. However, this is definitely out of scope for this PR, just wanted to mention it.

> Suggested fix: make the bypass opt-in per call site — e.g. `is_addr_assigned(addr, vrf, allow_nonlocal=True)`, passed only by the verifies for which a non-local bind is acceptable — and honour the `vrf` argument.

While reading through your description I was afraid the solution for the mentioned issues would require larger refactoring. I'm glad a simple flag should work. I'm not sure yet how to honour the vrf argument when `allow_nonlocal` is set to true though. If it is set, a vrf shouldn't theoretically change the outcome, as I might want to bind non locally to an interface inside the given vrf. That would result in the shortcut checking the `allow_nonlocal` besides the sysctl but nothing more. Is that indeed all you'd like me to do?

MPStudyly:
Tackled the raised issues by implementing proposed changes. However check my previous message to clarify the still open questions, especially the one about what to do when `allow_nonlocal` is True and a `vrf` is given.

JFYI: As mentioned before, I will be on holiday the next two weeks, don't expect any further changes from my side until around three weeks from now.

python/vyos/utils/network.py
```diff
@@ -462,12 +467,24 @@ def is_ipv6_link_local(addr):
...
 
+    # Check if sysctl to allow nonlocal binds is set for given afi and caller allows nonlocal binds,
+    # skip check accordingly
+    if allow_nonlocal and ((
+        is_valid_ipv4_address_or_range(ip_address)
+        and sysctl_read(['net', 'ipv4', 'ip_nonlocal_bind']) == "1"
+    ) or (
+        is_valid_ipv6_address_or_range(ip_address)
+        and sysctl_read(['net', 'ipv6', 'ip_nonlocal_bind']) == "1"
+    )):
```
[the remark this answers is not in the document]
MPStudyly: Should be fixed with latest changes. Split the existing validation function into dedicated ones for addresses and ranges, made use of the address validation at this location, ranges should get rejected.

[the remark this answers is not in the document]
MPStudyly: I see, I misunderstood the issue. I've adjusted `is_intf_addr_assigned` to allow for valid addresses only, as the purpose of that function (and it's usage throughout the codebase) suggests it never expects to be handed a range. As far as I understand it would always fail when given a range, as `ip_interface` can't handle such input.

python/vyos/utils/network.py
```diff
@@ -770,26 +786,30 @@ def is_valid_ipv6_address_or_range(addr: str) -> bool:
...
-        addresses = netifaces.ifaddresses(interface)
-        for addr_info in addresses.get(addr_type, []):
-            if addr_info.get('addr') == ip_address:
-                ifaces.append(interface)
+        # Check if interface belongs to the requested VRF, if this is not the
+        # case there is no need to proceed with this data set - continue loop
+        # with next element
+        tmp = get_interface_config(interface)
+        if vrf is not None and dict_search('master', tmp) != vrf and not include_vrf:
+            continue
```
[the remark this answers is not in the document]
MPStudyly: Should be addressed with latest changes.

python/vyos/utils/network.py
```diff
@@ -492,6 +510,10 @@ def is_intf_addr_assigned(ifname: str, addr: str, netns: str=None) -> bool:
     from vyos.utils.process import rc_cmd
     from ipaddress import ip_interface
 
+    # Reject invalid addresses or address ranges
+    if not is_ipv4_address(addr) or is_ipv6_address(addr):
+        return False
```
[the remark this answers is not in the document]
MPStudyly: The condition was messed up by an oversight. It was adjusted to reject properly now. Regarding zone IDs ipaddress should handle that correctly itself as https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv6Address explicitly mentions RFC 4007 to be supported.

MPStudyly:
Addressed all raised issues. Will squash when changes are considered good.

Please have a look at the questions raised at https://github.com/vyos/vyos-1x/pull/5266#issuecomment-5141967056. Also I'm unsure why all of a sudden the task ID is deemed invalid, I checked if the ticket was changed somehow but I can't find anything obvious.

Waiting for feedback from @AnDroed09 and @alexk37.

sever-sever (changes requested):
The `network.py` has conflicts now
```
<<<<<<< T8977_fix_haproxy_rebind
        return ip_interface(addr).version == 4
    except ValueError:
=======
        if '-' in addr: # If we are checking a range, validate both address's individually
            split = addr.split('-')
            return is_valid_ipv4_address_or_range(split[0]) and is_valid_ipv4_address_or_range(split[1])
        else:
            return ip_network(addr).version == 4
    except Exception:
>>>>>>> rolling
        return False

def is_ipv6_address(addr: str) -> bool:
    """
    Validates if the provided address is a valid IPv6 address or CIDR.
    :param addr: address to test
    :return: bool: True if provided address is valid
    """
    from ipaddress import ip_interface
    try:
<<<<<<< T8977_fix_haproxy_rebind
        return ip_interface(addr).version == 6
    except ValueError:
        return False

def is_ipv4_range(addr: str) -> bool:
    """
    Validates if the provided address is a valid IPv4 range.
    :param addr: address to test
    :return: bool: True if provided address is valid
    """
    if '-' not in addr:
        return False
    try:
        start, end = addr.split('-')
        return is_ipv4_address(start) and is_ipv4_address(end)
    except ValueError:
=======
        if '-' in addr: # If we are checking a range, validate both address's individually
            split = addr.split('-')
            return is_valid_ipv6_address_or_range(split[0]) and is_valid_ipv6_address_or_range(split[1])
        else:
            return ip_network(addr).version == 6
    except Exception:
>>>>>>> rolling
        return False
```

MPStudyly:
Squashed recent changes into single commit for easier conflict resolution (now two commits in total until final squash). Fixed conflicts by adapting exception handler to upstream change (catch `Exception` instead of `ValueError`).

alexandr-san4ez (dismissed):
I have re-tested this behaviour and I don't see any problem to merge the PR.

python/vyos/utils/network.py
```diff
@@ -445,11 +449,12 @@ def is_listen_port_bind_service(port: int, service: str, address: str = None) ->
...
-        if has_address:
-            if address == addr.ip and port == pid_port and service == pid_name:
+        # Ignore address check when service listening on "any address",
+        # service config might change from any to limited address scope
+        if port == pid_port and service == pid_name:
+            if not has_address:
                 return True
-        else:
-            if service == pid_name and port == pid_port:
+            if addr.ip in ('::', '0.0.0.0') or address == addr.ip:
```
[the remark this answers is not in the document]
MPStudyly: I think it is fine to have it this way round, as a change of address family should also be allowed which, as far as I understand, wouldn't be the case with the suggested change.

MPStudyly:
> I have re-tested this behaviour and I don't see any problem to merge the PR.

Thank you for the approval! However, please don't forget I still have some open questions about the overall topic as described at https://github.com/vyos/vyos-1x/pull/5266#issuecomment-5141967056 I'd be happy if we could also address those before finally getting everything merged. :)

python/vyos/utils/network.py
```diff
@@ -19,6 +19,7 @@
 from socket import AF_INET
 from socket import AF_INET6
 from vyos.utils.process import cmdl
+from vyos.utils.system import sysctl_read
```
c-po: Move this import to `is_addr_assigned()` to keep the scoping low

smoketest/scripts/cli/test_load-balancing_haproxy.py
```diff
@@ -173,8 +175,13 @@ def tearDown(self):
 
     def base_config(self):
         self.cli_set(base_path + ['service', haproxy_service_name, 'mode', 'http'])
-        self.cli_set(base_path + ['service', haproxy_service_name, 'port', '4433'])
-        self.cli_set(base_path + ['service', haproxy_service_name, 'backend', haproxy_backend_name])
+        self.cli_set(
```
c-po: Please revert this change only to make ruff/darker happy. SOmethign we need to deal with internally.

THe commit should just be about this issue and no reformatting

src/services/vyos-domain-resolver
```diff
@@ -31,7 +31,7 @@ from vyos.utils.convert import human_to_seconds
 from vyos.utils.dict import dict_search_args
 from vyos.utils.kernel import WIREGUARD_REKEY_AFTER_TIME
 from vyos.utils.file import makedir, chmod_775, write_file, read_file
-from vyos.utils.network import is_valid_ipv4_address_or_range, is_valid_ipv6_address_or_range
+from vyos.utils.network import is_ipv4_address_or_range, is_ipv6_address_or_range
```
c-po:
```suggestion
from vyos.utils.network import is_ipv4_address_or_range
from vyos.utils.network import is_ipv6_address_or_range
```

python/vyos/utils/network.py
```diff
@@ -465,22 +470,39 @@ def is_ipv6_link_local(addr):
...
+    ip_address: str, vrf=None, include_vrf=False, allow_nonlocal: bool = False
+) -> bool:
     """ Verify if the given IPv4/IPv6 address is assigned to any interface """
     from netifaces import interfaces # pylint: disable = no-name-in-module
     from vyos.utils.network import get_interface_config
     from vyos.utils.dict import dict_search
 
+    # Check if sysctl to allow nonlocal binds is set for given afi and caller allows nonlocal binds,
+    # skip check accordingly
+    if allow_nonlocal and (
```
c-po, line 484: Split this conditional onto multiples lines to make it easiert to read/maintain.

```python
    if allow_nonlocal:
        is_ipv4_address(ip_address) and \
          sysctl_read(['net', 'ipv4', 'ip_nonlocal_bind']) == "1":
    ...
```

python/vyos/utils/network.py
```diff
@@ -465,22 +470,39 @@ def is_ipv6_link_local(addr):
 
     return False
 
-def is_addr_assigned(ip_address, vrf=None, return_ifname=False, include_vrf=False) -> bool | str:
+
+def is_addr_assigned(
+    ip_address: str, vrf=None, include_vrf=False, allow_nonlocal: bool = False
```
c-po, line 474:
```suggestion
    ip_address: str, vrf: str|None = None, include_vrf: bool = False, allow_nonlocal: bool = False
```

c-po (changes requested):

python/vyos/utils/network.py
MPStudyly: Done 👍

smoketest/scripts/cli/test_load-balancing_haproxy.py
MPStudyly: I try my best to do so, but I followed suit with @alexandr-san4ez request to frequently squash. So it's a bit more complicated than a simple revert :grimacing:

src/services/vyos-domain-resolver
MPStudyly: Done 👍

python/vyos/utils/network.py
MPStudyly: Done 👍

python/vyos/utils/network.py
MPStudyly: I'm not 100% sure how you want this to look like. Following your example I can move `allow_nonlocal` into an outer conditional, but separating the conditions of the logical `or` would force me to duplicate the conditionals body (`return True`). Is it that what you want? I.e. change the code to:
```
if allow_nonlocal:
  if is_ipv4_address(ip_address) and \
      sysctl_read(['net', 'ipv4', 'ip_nonlocal_bind']) == "1":
    return True
  if is_ipv6_address(ip_address) and \
      sysctl_read(['net', 'ipv6', 'ip_nonlocal_bind']) == "1":
    return True
```

PR #5270, opened 2026-06-12
T8971: upgrade from bgpq3 package to bgpq4

c-po:
## Change summary

Add symlink from bgpq3 to bgpq4 to keep backwards compatible behavior.

## Types of changes

- [x] Other (please describe): Package upgrade

## Related Task(s)

* https://vyos.dev/T8971

## Checklist:

### DIFF
--- a/debian/vyos-1x.links
+++ b/debian/vyos-1x.links
@@ -1,3 +1,4 @@
 /usr/libexec/vyos/system/standalone_root_pw_reset /opt/vyatta/sbin/standalone_root_pw_reset
 /usr/libexec/vyos/vyconf_cli.py /usr/libexec/vyos/vyconf/bin/vy_commit
 /usr/libexec/vyos/vyconf_cli.py /usr/libexec/vyos/vyconf/bin/vy_in_session
+/usr/bin/bgpq4 /usr/bin/bgpq3
--- a/debian/control
+++ b/debian/control
@@ -439,7 +439,7 @@ Depends:
   aptitude,
   arpwatch,
   atop,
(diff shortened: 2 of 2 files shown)


natali-rs1985 (approved):
Package upgrade

debian/vyos-1x.links
```diff
@@ -1,3 +1,4 @@
 /usr/libexec/vyos/system/standalone_root_pw_reset /opt/vyatta/sbin/standalone_root_pw_reset
 /usr/libexec/vyos/vyconf_cli.py /usr/libexec/vyos/vyconf/bin/vy_commit
 /usr/libexec/vyos/vyconf_cli.py /usr/libexec/vyos/vyconf/bin/vy_in_session
+/usr/bin/bgpq4 /usr/bin/bgpq3
```
sever-sever, line 4: This is a user utility, I don't think we should do such symlinks as it could confuse users.

c-po: I agree, but it could prevent user scripts to fail. Lets diskuss it in out maintainer call.

dmbaturin (approved):
I compared the docs and release notes of bgpq3 vs bgpq4 and it looks like they are compatible enough to consider bgpq4 a drop-in replacement. Some releases specifically introduce changes to restore compatibility with bgpq3, even.

For this reason I also think the symlink from bgpq3 to bgpq4 makes sense, even though it looks a bit odd, as @sever-sever says. Having an odd-looking symlink is probably better than breaking scripts, if people use it from scripts.

PR #5274, opened 2026-06-15
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `Critical` | **File**: `python/vyos/qos/limiter.py:L20`

Multiple QoS modules (limiter.py, droptail.py, fairqueue.py, fqcodel.py, netem.py, ratelimiter.py, randomdetect.py, roundrobin.py, priority.py, cake.py, trafficshaper.py) construct shell commands using f-strings with unsanitized config values. The `self._interface` and various config parameters are directly interpolated into `tc` commands without validation or escaping. An attacker with config write access could inject shell metacharacters via interface names or QoS parameters, leading to arbitrary command execution as root.

## Solution

Use a command builder with proper argument escaping (e.g., shlex.quote for shell commands) or use subprocess with a list of arguments instead of string formatting. Validate and sanitize all inputs, especially interface names and numeric parameters.

## Changes

- `python/vyos/qos/limiter.py` (modified)
- `python/vyos/ifconfig/macvlan.py` (modified)


c-po (changes requested):
Task lacks a proper Phorge bugtracker id.

Instead of quoting the strings, we rather should just form a list like:

```
cmd = ['ip', 'link', 'set', 'dev', self.ifname, 'type', 'macvlan', 'mode', mode]
return self._cmd(' '.join(cmd))
```

Or even better, add list support into `_cmd()` to avoid the join.

PR #5276, opened 2026-06-16
pki: T8994: add graceful error handling in case certbot fails

c-po:
## Change summary

If ACME and certbot are used for PKI and e.g. haproxy it can become an issue if certbot is blocked by the firewall. The renewal service will fail and tear-down the production service - even if the certificate is yet not expired.

The production service was not restarted. This has been changed as every service which is stopped prior to the renew is later restarted even upon failure of renewing said certificate.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8994

Manually tested in error condition.

## Checklist:

### DIFF
--- a/src/conf_mode/pki.py
+++ b/src/conf_mode/pki.py
@@ -167,6 +167,7 @@ def certbot_renew(config: dict, force: bool=False) -> None:
 
     # Determine services using ACME based certificates
     pre_hook_services = []
+    stop_services = []
     for used_by, _ in dict_search_recursive(config, 'used_by'):
         pre_hook_services.extend(used_by)
     # Remove duplicate items from list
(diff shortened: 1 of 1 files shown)


jestabro (approved):
The problem is clear as described; the solution makes sense. One may want to accept (a version of) the coderabbit suggestion regarding possibly empty `stop_services` list.

src/conf_mode/pki.py
```diff
@@ -167,23 +167,29 @@ def certbot_renew(config: dict, force: bool=False) -> None:
...
+                stop_services.append(systemd_services[service])
+        tmp += ' --pre-hook "systemctl stop ' + ' '.join(stop_services) + '"'
 
     if force:
         tmp += ' --force-renewal'
 
-    print(cmd(tmp, raising=ConfigError, message=f'Certbot renew failed!'))
+    try:
+        print(cmd(tmp, raising=ConfigError, message=f'Certbot renew failed!'))
+    except Exception as e:
```
alexandr-san4ez:
```suggestion
    try:
        print(cmd(tmp, raising=ConfigError, message=f'Certbot renew failed!'))
    except ConfigError as e:
```

As I understand `cmd` function has `raising=ConfigError` and that means only this type of exceptions will be thrown.

alexandr-san4ez (approved):
I have only one minor notice about the exception type but it doesn't block PR for merging.

src/conf_mode/pki.py
```diff
@@ -167,23 +167,29 @@ def certbot_renew(config: dict, force: bool=False) -> None:
...
     # Automatically add services in use to pre_hook_services depending on service
     # name in vyos.defaults.systemd_services
     if pre_hook_services:
-        services = []
         for service in pre_hook_services:
             if service in systemd_services:
-                services.append(systemd_services[service])
-        tmp += ' --pre-hook "systemctl stop ' + ' '.join(services) + '"'
+                stop_services.append(systemd_services[service])
+        tmp += ' --pre-hook "systemctl stop ' + ' '.join(stop_services) + '"'
```
dmbaturin, line 181: If we are changing this line, should we use it as a chance to move away from constructing commands as strings and convert it to a list instead?

dmbaturin (approved):

src/conf_mode/pki.py
dmbaturin: I suppose it's a good idea to only catch what we _expect_ to be raised here. If some other exception occurs, it likely indicated a bug.

src/conf_mode/pki.py
c-po: There is a general case about this for our next developer meeting.

src/conf_mode/pki.py
c-po: Change addressed

PR #5287, opened 2026-06-20
T8990: bgp: fix VPNv4/VPNv6 leaked routes flapping on every commit

alexk37:
BGP address-family template rendered the route-target using the "route-target vpn ..." alias. FRR accepts that alias on input but its config writer only ever emits the canonical "rt vpn ..." form. `frr-reload.py `diffs the generated config against FRR's running config line by line and has no normalisation for this, so the route-target lines never matched and were deleted and re-added on every reload. Re-applying the route-target runs FRR's `vpn_leak_prechange()/postchange()`, a non-atomic withdraw-all then reinstall-all of every leaked route.

Emit the canonical "rt vpn" keyword so the rendered config round-trips through frr-reload unchanged. FRR also folds an identical import+export route-target into "both" (an order-sensitive, byte-identical match), so do the same to stay idempotent; reordered or differing lists stay split on both sides.

Add test_bgp_33_vpn_route_target_idempotency covering both / identical export+import / asymmetric route-target, reading the generated FRR config (getFRRconfig/vtysh returns FRR's already-canonical form and cannot catch the alias).

## Types of changes

## Related Task(s)

https://vyos.dev/T8990

```
test_bgp_33_vpn_route_target_idempotency(__main__.TestProtocolsBGP.test_bgp_33_vpn_route_target_idempotency) ... ok

Ran 1 test in 59.499s
```
## Checklist:

### DIFF
--- a/smoketest/scripts/cli/test_protocols_bgp.py
+++ b/smoketest/scripts/cli/test_protocols_bgp.py
@@ -25,6 +25,7 @@ from vyos.configsession import ConfigSessionError
 from vyos.template import is_ipv6
 from vyos.utils.process import process_named_running
 from vyos.utils.process import cmd
+from vyos.utils.file import read_file
 from vyos.frrender import bgp_daemon
 
 ASN = '64512'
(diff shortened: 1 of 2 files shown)


smoketest/scripts/cli/test_protocols_bgp.py
```diff
@@ -1792,6 +1793,54 @@ def test_bgp_32_bfd_strict(self):
...
+    def test_bgp_33_vpn_route_target_idempotency(self):
+        # T8990: VPNv4/VPNv6 inter-VRF route-leak config must be rendered using
+        # FRR's canonical "rt vpn" keyword (not the "route-target vpn" alias),
+        # and an equal import+export route-target must be collapsed into "both"
+        # - exactly as FRR stores it. Otherwise the generated config never
+        # matches FRR's running config, frr-reload re-applies the route-target
+        # on every commit, and the leaked routes are withdrawn/reinstalled
+        # (flap) on every - even unrelated - commit.
+        #
+        # This intentionally reads the generated FRR config file instead of
```
c-po, line 1805: This reads like AI generated code. Please mark it if possible. Also use our generic implementation of `self.getFRRconfig()` instead of reading the file by hand - this eases code maintenance.

The fix itself is legit.

c-po (changes requested):

smoketest/scripts/cli/test_protocols_bgp.py
alexk37: `getFRRconfig()` uses `vtysh -c "show running-config no-header"` and it returns already canonicalised form.  So to check literal rendered output `read_file('/run/frr/config/vyos.frr.conf')` was used

c-po: Okay that makes sense indeed.

c-po (approved):
Fix looks good. Smoketests passed. Testcase design explained in comment.

sever-sever (approved):
Tested by community

PR #5289, opened 2026-06-22
T9011: enable proxy_ndp sysctl for static IPv6 neighbor proxy

natali-rs1985:
## Change summary

Enable `proxy_ndp` sysctl on interfaces with active ND proxy
entries and disable it when interfaces are removed from config.
Add smoketest for static neighbor-proxy (ARP and NDP).
## Types of changes

## Related Task(s)

* https://vyos.dev/T9011

```
set interfaces ethernet eth1 address 'fd00::1/64'
set protocols static neighbor-proxy nd fd00::99 interface eth1
commit

vyos@vyos# ip -6 neigh show proxy
fd00::99 dev eth1 proxy 
[edit]
vyos@vyos# cat /proc/sys/net/ipv6/conf/eth1/proxy_ndp
1
[edit]
vyos@vyos# 
```

## Checklist:

### DIFF
--- a/src/conf_mode/protocols_static_neighbor-proxy.py
+++ b/src/conf_mode/protocols_static_neighbor-proxy.py
@@ -17,10 +17,15 @@
 from sys import exit
 
 from vyos.config import Config
+from vyos.configdict import leaf_node_changed
+from vyos.configdict import node_changed
 from vyos.utils.process import call
+from vyos.utils.system import sysctl_write
 from vyos import ConfigError
 from vyos import airbag
+
 airbag.enable()
+
 
 def get_config(config=None):
     if config:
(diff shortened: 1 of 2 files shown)


src/conf_mode/protocols_static_neighbor-proxy.py
```diff
@@ -73,6 +96,8 @@ def apply(config):
         for neighbor, neighbor_conf in config['nd'].items():
             for interface in neighbor_conf['interface']:
                 call(f'ip -6 neighbor add proxy {neighbor} dev {interface}')
+                sysctl_write(['net', 'ipv6', 'conf', interface, 'proxy_ndp'], 1)
```
sever-sever, line 99: Should we also check and `set system sysctl parameter net.ipv6.conf.ethX.proxy_ndp` option before setting/unsetting the value?

natali-rs1985: No need — `sysctl_write` already does that internally. It reads the current value first and skips the write if it's already set:
```
def sysctl_write(name, value):
    ...
    # do not change anything if a value is already configured
    if sysctl_read(name) == value:
        return True
    ...
```

sever-sever (approved):
Add the required sysctl option when configuring the feature.

src/conf_mode/protocols_static_neighbor-proxy.py
```diff
@@ -48,11 +64,18 @@ def verify(config):
                     f"ARP neighbor-proxy for '{neighbor}' requires an interface to be set!"
```
c-po: There is no ARP in IPv6 - while working on this topic this should be fixed, too.

c-po (changes requested):

src/conf_mode/protocols_static_neighbor-proxy.py
natali-rs1985: Done

c-po (approved):
Requested changes addressed

PR #5293, opened 2026-06-24
T5388: validate common types in-process

xTITUSMAXIMUSX:
## Change summary
`validate-value` runs each `--exec` validator by forking a process per value.
On large configs (many routes, addresses, prefixes, ports) this per-value fork
is most of the commit/boot time, and since it is single-threaded it scales with
config size.

This adds in-process checks for the most common validators and uses them when
available, falling back to the external validator otherwise:

- `numeric`
- `ipv4-address`
- `ipv4-prefix`
- `ipv4-range`

The native checks return the same accept/reject as the existing validators.
IPv6 validators are intentionally left on the external path: ipaddrcheck/libcidr
accepts and rejects forms a native parser can't reproduce without changing
behaviour.

## Types of changes

## Related Task(s)
* https://vyos.dev/T5388

Built `validate-value` from this branch and compared it against the stock
binary on a 1.5 system, confirming native and external validators agree on
accept/reject across a wide range of inputs (valid values, out-of-range,
malformed, leading zeros, host-bits-set prefixes, IPv4 ranges,
hex/octal/binary numerics) — no mismatches.

Per-value cost, best of 2 over 2000 iterations each:

| validator    | external | in-process |
| ------------ | -------- | ---------- |
| numeric      | 12.4 ms  | 4.8 ms     |
| ipv4-address | 15.3 ms  | 4.7 ms     |
| ipv4-prefix  | 15.5 ms  | 4.7 ms     |
| ipv4-range   | 15.5 ms  | 4.8 ms     |

## Checklist:

### DIFF
--- a/src/ocaml/native_validators.ml
+++ b/src/ocaml/native_validators.ml
@@ -1,0 +1,282 @@
+(* In-process implementations of the most common validators, so validate-value
+   can check a value without forking the external validator for every value.
+   dispatch looks the validator up by basename; unknown validators return None
+   and the caller falls back to running the external command. *)
+
+(* numeric: same checks as numeric.ml, returning a result instead of exiting. *)
+module Numeric = struct
+  type numeric_str = Number_string of string | Range_string of string
+  type numeric_val = Number_float of float | Range_float of float * float
+
+  type options = {
+    positive: bool;
+    nonnegative: bool;
+    allow_float: bool;
+    ranges: string list;
+    not_ranges: string list;
+    not_values: string list;
+    relative: bool;
+    allow_range: bool;
+    require_range: bool;
+    parse_hex: bool;
+    parse_oct: bool;
+    parse_bin: bool;
+    parse_dec: bool;
+  }
+
+  let default_opts = {
+    positive = false;
+    nonnegative = false;
+    allow_float = false;
+    ranges = [];
+    not_ranges = [];
+    not_values = [];
+    relative = false;
+    allow_range = false;
+    require_range = false;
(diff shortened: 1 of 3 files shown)


src/ocaml/native_validators.ml
```diff
@@ -0,0 +1,277 @@
...
+  "numeric", Numeric.check;
+  "ipv4-address", ipv4_address;
+  "ipv4-prefix", ipv4_prefix;
+  "ipv4-range", ipv4_range;
+]
+
+let dispatch cmd value =
+  (* A value starting with '-' is mis-parsed as an option by the external
+     validators (getopt/Arg), so defer to them for an identical result. *)
+  if String.length value > 0 && value.[0] = '-' then None
```
jestabro, line 275: Note that this defers to external validator the decrement case of `numeric --relative` (or in fact any case of negative value).
On the other hand, native validation will fail on the increment case `'+'` of `--relative`.
In the former case, the comment preceding line 270 is not accurate, due to the use of `Arg.Rest`:
https://github.com/vyos/vyos-1x/blob/rolling/src/ocaml/validators/numeric.ml#L54
In the latter case, it appears that the `Arg.Rest` is not taken into account, leading to the failure.
This will need a closer look, but wanted to point it out as I continue the review.

jestabro (changes requested):

jestabro:
This is a nice contribution. For other reviewers/ @dmbaturin , I think we should consider (1) whether the optimization is worth the sacrifice of code organization, and (2) especially given that it only covers positive values.

PR #5296, opened 2026-06-26
https: T9022: serve the full CA certificate chain

statio:
## Change summary

`set service https` previously wrote only the leaf certificate to the file
served by nginx. An intermediate CA was added to the served chain only when an
operator explicitly configured `certificates ca-certificate`, and even then
only that single CA was appended — the rest of the issuer chain was never
followed.

Clients that do not already trust the issuing intermediate CA therefore could
not build a path to a trusted root and rejected the connection. This is
increasingly common with Let's Encrypt's newer intermediates (the E-/R-series,
chaining to ISRG Root X1/X2), which are frequently absent from client trust
stores.

This change builds the complete certificate chain (leaf + every available
intermediate CA up to the root) from the CA certificates present in the PKI,
using `vyos.pki.find_chain()` — the same helper already used by HAProxy,
OpenConnect, stunnel, SSTP and the IKEv2 profile generator. `service https`
was the only PKI/TLS consumer not building the full chain. The chain is now
discovered automatically, so `ca-certificate` no longer needs to be configured
by hand; it remains accepted for backwards compatibility.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9022

## Related PR(s)

_None_

Import a leaf certificate, its issuing intermediate CA and the root into the
PKI, then assign **only the leaf** to the HTTPS service:

```
set pki ca ROOT certificate <root-cert>
set pki ca INTERMEDIATE certificate <intermediate-cert>
set pki certificate LEAF certificate <leaf-cert>
set pki certificate LEAF private key <leaf-key>
set service https certificates certificate LEAF
commit
```

Before this change, `openssl s_client -connect 127.0.0.1:443 -showcerts` presents
only the leaf certificate. After it, the full chain (leaf + intermediate + root)
is presented, so clients that lack the intermediate can validate the server.

A smoketest (`test_certificate_chain`) is added to
`smoketest/scripts/cli/test_service_https.py`. It imports a root → intermediate
→ leaf chain, references only the leaf, and asserts the generated
`/run/nginx/certs/<name>_cert.pem` contains all three certificates. It also
verifies that explicitly setting `ca-certificate` stays backwards compatible
and does not duplicate any certificate in the chain. The smoketest executes in
the `vyos-build` QEMU smoketest harness.

## Checklist:

### DIFF
--- a/smoketest/scripts/cli/test_service_https.py
+++ b/smoketest/scripts/cli/test_service_https.py
@@ -30,6 +30,9 @@ from vyos.utils.file import write_file
 from vyos.utils.process import call
 from vyos.utils.process import cmd
 from vyos.utils.process import process_named_running
+from vyos.pki import CERT_BEGIN
+from vyos.pki import encode_certificate
+from vyos.pki import load_certificate
 from vyos.xml_ref import default_value
 
 from vyos.configsession import ConfigSessionError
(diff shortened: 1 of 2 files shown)


statio:
Addressed the review feedback: the certificate-chain smoketest now asserts the exact leaf -> intermediate -> root PEM sequence in addition to the certificate count, so a mis-ordered bundle is caught as well (3d299d324).

src/conf_mode/service_https.py
```diff
@@ -176,12 +178,19 @@ def generate(https):
...
-            ca_cert = https['certificates']['ca_certificate']
-            server_cert += '\n' + str(wrap_certificate(https['pki']['ca'][ca_cert]['certificate']))
+        # Build the full certificate chain (server certificate followed by any
+        # intermediate CA certificates up to the root) from the CA certificates
+        # available in the PKI. Serving the complete chain lets clients that do
+        # not yet trust the issuing intermediate CA validate the presented
+        # certificate. This mirrors the other PKI consumers (HAProxy, stunnel).
+        loaded_ca_certs = {
+            load_certificate(cert_data['certificate'])
+            for _, cert_data in dict_search('pki.ca', https, default={}).items()
```
alexandr-san4ez:
```suggestion
            for cert_data in dict_search('pki.ca', https, default={}).values()
```

smoketest/scripts/cli/test_service_https.py
```diff
@@ -243,6 +290,79 @@ def test_certificate(self):
...
+        self.cli_set(
+            pki_base
+            + [
+                'certificate',
+                cert_name,
+                'private',
+                'key',
+                chain_key_data.replace('\n', ''),
+            ]
+        )
```
alexandr-san4ez, line 309:
```suggestion
        root_ca_path = pki_base + ['ca', root_ca_name, 'certificate']
        self.cli_set(root_ca_path + [chain_root_ca_data.replace('\n', '')])
        
        intermediate_ca_path = pki_base + ['ca', intermediate_ca_name, 'certificate']
        self.cli_set(intermediate_ca_path + [chain_intermediate_ca_data.replace('\n', '')])
        
        cert_path = pki_base + ['certificate', cert_name, 'certificate']
        self.cli_set(cert_path + [chain_cert_data.replace('\n', '')])
        
        cert_key_path = pki_base + ['certificate', cert_name, 'private', 'key']
        self.cli_set(cert_key_path + [chain_key_data.replace('\n', '')])
```

alexandr-san4ez (changes requested):
It is a good implementation but I would suggest a couple of changes and I see that smoketest failed:
```
DEBUG - ERROR: test_certificate_chain (__main__.TestHTTPSService.test_certificate_chain)
DEBUG -   File "/usr/libexec/vyos/tests/smoke/cli/test_service_https.py", line 339, in test_certificate_chain
DEBUG -     self.cli_commit()
DEBUG -   File "/usr/libexec/vyos/tests/smoke/cli/base_vyostest_shim.py", line 114, in cli_commit
DEBUG -     return self._session.commit()
DEBUG -            ^^^^^^^^^^^^^^^^^^^^^^
DEBUG -   File "/usr/lib/python3/dist-packages/vyos/configsession.py", line 347, in commit
DEBUG -     out = self.__run_command([COMMIT])
DEBUG -           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
DEBUG -   File "/usr/lib/python3/dist-packages/vyos/configsession.py", line 256, in __run_command
DEBUG - vyos.configsession.ConfigSessionError: [ pki ]
DEBUG - Invalid private key on certificate "https-chain"
DEBUG - [[pki]] failed
DEBUG - Commit failed
```

statio:
Pushed a fix and consolidated into a single commit.

The previous run's `test_smoketest_cli` failure was in this PR's own smoketest, not infrastructure: the certificate fixtures I first used had a leaf private key that VyOS PKI validation rejects (`Invalid private key on certificate`), so the commit inside `test_certificate_chain` failed. I've replaced them with a self-contained, generated root → intermediate → leaf chain (EC/prime256v1, `serverAuth` leaf, valid to 2124) and verified locally with `vyos.pki` that the key passes validation and `find_chain()` emits the full `leaf → intermediate → root` bundle.

src/conf_mode/service_https.py
statio: Good catch -- `.values()` it is, since the key isn't used. Done in 78bec494b.

smoketest/scripts/cli/test_service_https.py
statio: Applied -- pulled the PKI paths into named variables so each `cli_set` reads on one line, in 78bec494b. (Left the intermediate one wrapped to stay under darker's 88-column limit.)

statio:
Thanks @alexandr-san4ez -- addressed both suggestions in `78bec494b`:

- `service_https.py`: iterate `dict_search('pki.ca', ...).values()` instead of `.items()` with an unused key.
- `test_service_https.py`: extracted the PKI paths into named variables so each `cli_set` reads on one line.

On the `Invalid private key` smoketest failure: that was a bad fixture on my side -- the leaf cert/key I'd borrowed from `base_interfaces_test.py` has a private key VyOS PKI rejects (that pair is only exercised by the EAPoL subtest, which CI skips, so it was never actually validated). I replaced it with a generated, self-contained root -> intermediate -> leaf chain, and `test_certificate_chain` now passes (`... ok`). The remaining red checks are the unrelated `test_interfaces_ethernet` NIC-driver check (`i40e`/`iavf`/`ice` kernel modules absent in the CI build), which passed on the previous commit.

alexandr-san4ez (approved):
The local test was successful and there were no objections to these changes:
```
vyos@vyos# openssl s_client -connect 127.0.0.1:443 -showcerts
Certificate chain
 0 s:C = GB, ST = Some-State, L = Some-City, O = LEAF, CN = LEAF
   i:C = GB, ST = Some-State, L = Some-City, O = INTERMEDIATE, CN = INTERMEDIATE
...
 1 s:C = GB, ST = Some-State, L = Some-City, O = INTERMEDIATE, CN = INTERMEDIATE
   i:C = GB, ST = Some-State, L = Some-City, O = ROOT, CN = ROOT
...
 2 s:C = GB, ST = Some-State, L = Some-City, O = ROOT, CN = ROOT
   i:C = GB, ST = Some-State, L = Some-City, O = ROOT, CN = ROOT
```

c-po (approved):
Change looks reasonable. Smoketests included.

PR #5315, opened 2026-07-06
pppoe: T9060: expose no-default-link-local for IPv6 address configuration

sperezacuna:
## Change summary

Expose no-default-link-local for IPv6 address configuration on PPPoE interfaces. When IPv6 is enabled on a PPPoE interface, VyOS currently adds a synthetic EUI64-derived /64 link-local address on top of the link-local already negotiated by pppd via IPv6CP. Under RFC 6724 Rule 8 (longest prefix match), the kernel prefers the synthetic /64 over the IPv6CP-negotiated address when selecting a source address, causing DHCPv6 PD to send Solicit/Request packets from the wrong link-local. ISPs that validate the source link-local against the IPv6CP-negotiated one silently drop these packets, stalling prefix delegation indefinitely.

The fix adds the missing `#include <include/interface/ipv6-address-no-default-link-local.xml.i>` to the PPPoE interface definition. The synthetic link-local remains the default; affected users opt out explicitly with set interfaces pppoe pppoe0 ipv6 address no-default-link-local.

## Types of changes

## Related Task(s)
* https://vyos.dev/T9060

Set up a PPPoE interface with IPv6 and no-default-link-local:

```bash
set interfaces pppoe pppoe0 source-interface eth0.20
set interfaces pppoe pppoe0 authentication username <user>
set interfaces pppoe pppoe0 authentication password <pass>
set interfaces pppoe pppoe0 ipv6 address no-default-link-local
commit
```

Verify only one link-local is present (the IPv6CP-negotiated peer address, no synthetic /64):

```bash
$ ip -6 addr show pppoe0
inet6 fe80::bd27:a8d:ab4a:60af peer fe80::1/128 scope link
     valid_lft forever preferred_lft forever
```

A smoketest covering this case was added to `test_interfaces_pppoe.py`

## Checklist:
- [x] My change requires a change to the documentation (addition of no-default-link-local to the ppp API documentation)

### DIFF
--- a/interface-definitions/interfaces_pppoe.xml.in
+++ b/interface-definitions/interfaces_pppoe.xml.in
@@ -105,6 +105,7 @@
                 <children>
                   #include <include/interface/ipv6-address-autoconf.xml.i>
                   #include <include/interface/ipv6-address-interface-identifier.xml.i>
+                  #include <include/interface/ipv6-address-no-default-link-local.xml.i>
                 </children>
               </node>
               #include <include/interface/adjust-mss.xml.i>
--- a/smoketest/scripts/cli/test_interfaces_pppoe.py
+++ b/smoketest/scripts/cli/test_interfaces_pppoe.py
@@ -416,5 +416,41 @@ class PPPoEInterfaceTest(VyOSUnitTestSHIM.TestCase):
             # Validate and verify assigned IP addresses
             self._verify_interface_address(interface)
 
+    def test_pppoe_no_default_link_local(self):
+        # Verify no synthetic EUI64 /64 link-local is added when no-default-link-local is set
+        interface = self._interfaces[0]
+        (user, passwd) = self.u_p_dict[interface]
+
+        self.cli_set(base_path + [interface, 'authentication', 'username', user])
+        self.cli_set(base_path + [interface, 'authentication', 'password', passwd])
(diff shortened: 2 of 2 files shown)


sperezacuna:

sperezacuna:
Ran into a testing error. This is not working: the /64 LLA keeps being created after the fix.

c-po:
@sperezacuna interesting find. I wonder if we not rather should fix the code and comply with the RFC?

sperezacuna:
Thanks @c-po. Still debugging; on closer inspection it doesn't look like we're actually adding the extra /64 LLA ourselves, so I need to trace where it originates first... Maybe the kernel not respecting `addr_gen_mode=1` because of ppp being a logical interface or some pppd quirk?

What do you have in mind by complying with the RFC? RFC 5072 only says the PPP link-local SHOULD come from the negotiated IID; I haven't found anything that forbids a second one.

c-po:
@sperezacuna we should follow the RFC. Whats the status on this PR?

sperezacuna:
Thanks for the follow up @c-po. 
I'm currently traveling; I still haven't been able to track down the source of the spurious IPv6 LLA; could use some help on that. I can follow up in September.

c-po:
PR is stale. Superseeded by https://github.com/vyos/vyos-1x/pull/5500

PR #5347, opened 2026-07-22
flow-accounting: T9122: bind NetFlow export to the configured VRF

statio:
## Change summary

Since the migration from pmacct to the `ipt_NETFLOW` kernel module (T75),
`set system flow-accounting vrf <name>` has had no effect on the NetFlow export
path. NetFlow is now emitted by the kernel module, so there is no `uacctd`
daemon left to wrap in `ip vrf exec`, and nothing bound the module's export
socket to the VRF. A collector reachable only inside a VRF therefore never
received any flows when the export was configured with `source-address`.

This binds the export socket to a device using the ipt_NETFLOW `%device`
destination suffix:

- a per-server `source-interface` takes precedence (`%<interface>`);
- otherwise the global `vrf` device is used (`%<vrf>`), reproducing the old
  `ip vrf exec` behaviour;
- `source-address` is rendered independently as `@<address>`.

The module parses `@source-address` before `%device`, so the source-address is
now rendered first. The previous template emitted them in the opposite order,
which only worked because `source-address` and `source-interface` are mutually
exclusive per server — combining a VRF bind with a source-address needs the
correct order.

Example — `vrf OOB-Management` + `server 10.0.0.24 source-address 10.0.0.64` now
renders:

```
options ipt_NETFLOW destination=10.0.0.24:2055@10.0.0.64%OOB-Management
```

The commit is rejected when a `source-interface` used together with a `vrf` is
not a member of that VRF (mirroring the existing OSPF/OSPFv3/IS-IS/BGP checks),
since the flows would otherwise silently leave via a different routing table.

**Upgrade impact (behaviour change):** no CLI syntax change and no migration is
required. Configs that set `vrf` now export within that VRF — this fixes broken
setups (including anything the flow-accounting 2-to-3 migration produced), but
also means a config that set `vrf X` while its collector was actually reachable
via the default table (working today only because `vrf` was ignored) will now be
bound into VRF X. Full analysis in T9122.

## Types of changes

- [x] Other (please describe): restores intended VRF behaviour; classified on
  Phabricator as a *behaviour change* for existing `vrf` configs — see the
  Upgrade impact note above.

## Related Task(s)

* https://vyos.dev/T9122

## Related PR(s)

* Documentation update to follow in `vyos/vyos-documentation` (flow-accounting
  `vrf` / `source-interface` / `source-address` export behaviour).

Reproduction (collector `10.0.0.24` reachable only inside VRF `OOB-Management`,
OOB interface `bond0.500` with `10.0.0.64/xx` enslaved to that VRF):

```
set system flow-accounting netflow interface 'bond0.500'
set system flow-accounting netflow server 10.0.0.24 source-address '10.0.0.64'
set system flow-accounting vrf 'OOB-Management'
```

Before this change `/etc/modprobe.d/ipt_NETFLOW.conf` renders
`destination=10.0.0.24:2055@10.0.0.64` and no flows reach the collector. After
this change it renders `destination=10.0.0.24:2055@10.0.0.64%OOB-Management` and
the export is routed via the VRF.

A smoketest case `test_netflow_vrf` is added covering the three combinations
(`vrf` only, `vrf` + `source-address`, `vrf` + `source-interface`), asserting the
resulting `ipt_NETFLOW` `destination=` string via `get_module_data()`, and
asserting that a `source-interface` outside the VRF is rejected. It is exercised
by the CI smoketest job; I do not have a live router to paste output here. The
template rendering was validated locally for all cases and the darker + ruff lint
passes on the changed lines.

## Checklist:

### DIFF
--- a/data/templates/ipt-netflow/ipt_NETFLOW.conf.j2
+++ b/data/templates/ipt-netflow/ipt_NETFLOW.conf.j2
@@ -12,8 +12,17 @@ options ipt_NETFLOW sampler=random:{{ netflow.sampling_rate }}
 {%     for server in netflow.server %}
 {%         set d = netflow.server[server] %}
 {%         set s = (server | bracketize_ipv6) ~ ':' ~ d['port'] %}
-{%         set s = s ~ ('%' ~ d['source_interface'] if d['source_interface'] is vyos_defined else '') %}
+{#         The ipt_NETFLOW destination grammar is "ip:port[@address][%device]" and #}
+{#         the module parses "@" before "%", so the source-address is rendered #}
+{#         first. The socket is then bound to a device to select the egress path: #}
+{#         a per-server source-interface takes precedence, else the global VRF is #}
+{#         used so flows are exported within that VRF (T9122). #}
 {%         set s = s ~ ('@' ~ d['source_address'] if d['source_address'] is vyos_defined else '') %}
+{%         if d['source_interface'] is vyos_defined %}
+{%             set s = s ~ '%' ~ d['source_interface'] %}
+{%         elif vrf is vyos_defined %}
+{%             set s = s ~ '%' ~ vrf %}
+{%         endif %}
 {%         set _ = servers_list.append(s) %}
 {%     endfor %}
 options ipt_NETFLOW destination={{ servers_list | join(",") }}
--- a/smoketest/scripts/cli/test_system_flow-accounting.py
+++ b/smoketest/scripts/cli/test_system_flow-accounting.py
@@ -213,6 +213,71 @@ class TestSystemFlowAccounting(VyOSUnitTestSHIM.TestCase):
 
         self.cli_delete(['interfaces', 'dummy', dummy_if])
 
+    def test_netflow_vrf(self):
+        # T9122: NetFlow export must be routed via the configured VRF. The
+        # export socket is bound to a device - a per-server source-interface
+        # takes precedence, otherwise the global VRF device is used - so that
+        # collectors reachable only inside the VRF actually receive flows.
+        vrf_name = 'vyos-test-mgmt'
+        table = '1010'
+        source_address = '192.0.2.1'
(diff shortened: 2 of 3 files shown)


statio:
Good catch on the source-interface/VRF check. It was rejecting the case where the source-interface *is* the VRF master device itself (e.g. `source-interface OOB-Management` alongside `vrf OOB-Management`) — `get_interface_vrf()` only reports a master for enslaved interfaces and returns `default` for the VRF device, so that valid, equivalent config was being turned away.

Pushed b15fb99fe: the check now also permits the source-interface when it equals the VRF device, and I added a smoketest case for it (`server ... source-interface <vrf>` → `%<vrf>`). A source-interface in a different VRF (or the default table) is still rejected as before.

statio:
Following up on my previous comment — I want to walk that change back. I had been treating the VRF as just another interface you could point `source-interface` at, but after a bit more research none of the major vendors treat VRFs as interfaces: Cisco and Juniper both source a flow exporter from a routed interface and never from a VRF. I don't think we should either — a VRF is not an interface, and VRF export is already selected with `system flow-accounting vrf`.

So I force-pushed (`b672b0cd1`, replacing the earlier commit): a `source-interface` that names a VRF is now rejected outright with a clear "is a VRF, not an interface" message rather than the misleading "not a member of VRF" error, with a smoketest case covering it. `source-interface` is back to meaning an actual member interface, same as OSPF/BGP/IS-IS.

AnDroed09:
@statio, could you please rebase to trigger CI again?

statio:
Rebased onto the latest `rolling` — CI is running again now. Thanks for the nudge!

dmbaturin (approved):
The fix looks reasonable to me at a glance. I haven't tested it.

hedrok (approved):
Approve, I've reviewed and tested - everything works as expected.
I agree with your decision to refuse from using VRFs in `source-interface`.
Thanks for your work!

PR #5350, opened 2026-07-23
boot: T3871: rework interface renaming and ordering

c-po:
## Change summary

Fixes a long-standing boot-time race where Ethernet/wireless interfaces could disappear or be renamed on systems with multiple PCIe NICs from different vendors, because naming was previously decided from a single per-device udev event before all hardware had enumerated.

  Replaces that with one authoritative pass, run once configuration is available during router startup: wait for configured hardware, apply every hw-id binding, and deterministically name whatever has none yet by PCIe distance from the root complex and MAC address. An interface whose `hw-id` was deleted (NIC replacement) or whose whole configuration was removed is treated as an ordinary open slot and filled the same way, recovering its original hardware whenever the interfaces that vacated slots the same boot also show up as candidates. Boot now reports any interface still unresolved instead of failing silently, and the whole pass is covered by a new test suite.

Two unrelated fixes found via code review while touching adjacent code:
* ethtool: flow-control changes now honor the existing unsupported-driver denylist, fixing intermittent commit failures on virtio_net and similar drivers.
* smoketest: discard staged config after an expected commit failure so it doesn't silently carry into the next interface's test.

## Types of changes

## Related Task(s)

* https://vyos.dev/T3871

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1253

Embedded smoketests

## Checklist:

### DIFF
--- a/python/vyos/ethtool.py
+++ b/python/vyos/ethtool.py
@@ -217,6 +217,8 @@ class Ethtool:
 
     def check_flow_control(self):
         """ Check if the NIC supports flow-control """
+        if self.get_driver_name() in _drivers_without_speed_duplex_flow:
+            return False
         return bool(self._flow_control)
 
     def get_flow_control(self):
--- a/smoketest/scripts/cli/test_interfaces_ethernet.py
+++ b/smoketest/scripts/cli/test_interfaces_ethernet.py
@@ -229,6 +229,10 @@ class EthernetInterfaceTest(BasicInterfaceTest.TestCase):
                     msg = 'Driver does not fully support coalesce configuration'
                     with self.assertRaisesRegex(ConfigSessionError, msg):
                         self.cli_commit()
+                    # the failed commit leaves rx-usecs/tx-usecs staged in
+                    # the candidate config (commit() does not auto-rollback) -
+                    # discard it so it doesn't leak into the next interface
+                    self.cli_discard()
                     continue
 
                 # To find out the supported features
@@ -300,6 +304,10 @@ class EthernetInterfaceTest(BasicInterfaceTest.TestCase):
             if not ethtool.check_flow_control():
                 with self.assertRaises(ConfigSessionError):
                     self.cli_commit()
+                # the failed commit leaves disable-flow-control staged in
+                # the candidate config (commit() does not auto-rollback) -
+                # discard it so it doesn't leak into the next interface
+                self.cli_discard()
             else:
                 out, err = popen(f'sudo ethtool --json --show-pause {interface}')
                 out = loads(out)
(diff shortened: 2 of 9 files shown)


alexk37:
@c-po pls check the regression case:

vyos-net-name-resolve: interface with config but no hw-id is given a different name than the one its configuration is written for, silently losing its configuration

Tested on VM with >=3 NICs

  ```
  # normal system: eth0/eth1/eth2 all have hw-id (auto-added by rescan), eth1 has an address
  set interfaces ethernet eth1 address 10.99.1.1/24
  commit

  # make eth1's hw-id stale/absent (e.g. after NIC replacement - the old
  # documented remediation: "delete the MAC, it will be regenerated")
  delete interfaces ethernet eth1 hw-id
  commit
  save
  reboot
  ```


  - the NIC that was eth1 boots as eth3 (bootstrap floor = highest hw-id'd index + 1 - it can never return to eth1)
  - 10.99.1.1/24 is gone; boot commit logs `[interfaces_ethernet] Interface "eth1" does not exist!`
  - rescan freezes the NIC as a new bare `ethernet eth3` node - the eth1 config stays orphaned permanently, silently

c-po:
> @c-po pls check the regression case:

Fixed

alexk37:
@c-po thanks, the reported case is fixed at f0527542: the freed NIC returns to its own node, address intact, hw-id regenerated in place

The replacement fill introduces a new case though: when freed slots' MAC order doesn't match their old name order, a configured node is bound to the wrong physical NIC.

Tested on VM with 4 NICs:

```
# eth0..eth3 all have hw-id (auto-added by rescan)
set interfaces ethernet eth2 address 10.99.2.1/24
delete interfaces ethernet eth1
delete interfaces ethernet eth2 hw-id
commit
save
reboot
```

- eth2 comes back on eth1's old NIC (lower MAC takes the lower open slot), logged as a success:
vyos-net-name-resolve.py: reclaiming pending node 'eth2' for hw-id '00:00:5e:00:53:09'
10.99.2.1/24 now ARPs out the wrong port — tcpdump shows zero packets on the wire where 10.99.2.0/24 actually lives
- rescan writes the wrong hw-id into the eth2 node, so from the next boot it's enforced: udev names the NICs in PCI order and the resolver renames them back into the swap ("renamed": {"eth2":"eth1","eth1":"eth2"} in /run/vyos-net-name-resolve.json). Recovery = hand-edit config.boot + reboot

"recovering its original hardware" only holds when existing names already follow the (pcie_distance, MAC) sort, for boxes named by the old udev order that's roughly a coin flip per freed pair.

jestabro (approved):
This is a comprehensive solution to the problem of misnaming of NICs from different vendors, by (1) identifying the underlying logic error (2) updating and improving the udev detection/renaming framework.

This has already been tested by @c-po , @alexk37 , with some corner cases identified by @alexk37 addressed.

Note that the failing smoketests are known issues corrected by the later fix to the container smoketests.

dmbaturin (approved):
I could not see obvious issues and I trust the review and testing of other participants.

PR #5352, opened 2026-07-25
container: T7736: give container veths a deterministic host_interface_name

c-po:
## Change summary

Podman's default `vethN` auto-naming for a container's host-side veth can collide with VyOS's own `virtual-ethernet vethN` interfaces.

Bump the minimum Podman dependency to 5.8 (which supports the "host_interface_name" network connect option) and use it to name every non-macvlan container network attachment `veth-<container name>` instead, eliminating the collision by construction.

## Types of changes

## Related Task(s)

https://vyos.dev/T7736

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1259

```
 INFO - Executing VyOS smoketests
DEBUG - vyos@vyos:~$ /usr/bin/vyos-smoketest
DEBUG - /usr/bin/vyos-smoketest
DEBUG - Running Testcase (1/1): /usr/libexec/vyos/tests/smoke/cli/test_container.py
DEBUG - test_api_socket (__main__.TestContainer.test_api_socket) ... ok
DEBUG - test_basic (__main__.TestContainer.test_basic) ... ok
DEBUG - test_cpu_limit (__main__.TestContainer.test_cpu_limit) ... ok
DEBUG - test_dual_stack_network (__main__.TestContainer.test_dual_stack_network) ... ok
DEBUG - test_healthcheck (__main__.TestContainer.test_healthcheck) ... ok
DEBUG - test_ipv4_network (__main__.TestContainer.test_ipv4_network) ... ok
DEBUG - test_ipv6_network (__main__.TestContainer.test_ipv6_network) ... ok
DEBUG - test_long_name_host_interface_uniqueness (__main__.TestContainer.test_long_name_host_interface_uniqueness) ... ok
DEBUG - test_name_server (__main__.TestContainer.test_name_server) ... ok
DEBUG - test_network_mtu (__main__.TestContainer.test_network_mtu) ... ok
DEBUG - test_network_types (__main__.TestContainer.test_network_types) ... ok
DEBUG - test_network_vrf (__main__.TestContainer.test_network_vrf) ... ok
DEBUG - test_no_name_server (__main__.TestContainer.test_no_name_server) ... ok
DEBUG - test_uid_gid (__main__.TestContainer.test_uid_gid) ... ok
DEBUG - test_user_defined_mac (__main__.TestContainer.test_user_defined_mac) ... ok
DEBUG -
DEBUG - Ran 15 tests in 209.184s
DEBUG -
DEBUG - OK
DEBUG - SUCCESS! All 1 tests passed.
```

## Checklist:

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -334,7 +334,7 @@ Depends:
 # For "container"
   podman (>=5.8),
   netavark (>=1.14.0),
-  aardvark-dns,
+  aardvark-dns (>=1.14.0),
 # iptables is only used for containers now, not the the firewall CLI
   iptables,
 # End container
--- a/debian/control
+++ b/debian/control
@@ -332,8 +332,8 @@ Depends:
   kbd,
 # End "system option keyboard-layout"
 # For "container"
-  podman (>=4.9.5),
-  netavark,
+  podman (>=5.8),
+  netavark (>=1.14.0),
   aardvark-dns,
 # iptables is only used for containers now, not the the firewall CLI
   iptables,
--- a/op-mode-definitions/container.xml.in
+++ b/op-mode-definitions/container.xml.in
@@ -133,6 +133,22 @@
               </node>
             </children>
           </node>
+          <node name="interface">
+            <properties>
+              <help>Show host-side network interface used by each container</help>
+            </properties>
+            <!-- no admin check -->
+            <command>${vyos_op_scripts_dir}/container.py show_interface</command>
+            <children>
+              <node name="json">
+                <properties>
+                  <help>Show host-side network interface used by each container in JSON format</help>
+                </properties>
+                <!-- no admin check -->
+                <command>${vyos_op_scripts_dir}/container.py show_interface --raw</command>
+              </node>
+            </children>
+          </node>
           <tagNode name="log">
             <properties>
               <help>Show logs from a given container</help>
--- a/python/vyos/container.py
+++ b/python/vyos/container.py
@@ -12,10 +12,26 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from hashlib import sha256
+
 from vyos.config import Config
 from vyos.ifconfig import Interface
 from vyos.utils.dict import dict_search
 from vyos.utils.network import interface_exists
+
+def get_container_host_ifname(name: str) -> str:
+    """
+    Deterministic host-side veth interface name for a container's network
+    attachment (verify() only allows one network per container). Kept within
+    IFNAMSIZ and - thanks to the leading "veth-" (a hyphen can never appear in
+    a VyOS "vethN" interface name) - guaranteed to never collide with the
+    "virtual-ethernet" naming scheme..
+    """
+    prefix = f'veth-{name}'
+    if len(prefix) <= 15:
+        return prefix
+    digest = sha256(name.encode()).hexdigest()[:4]
+    return f'veth-{name[:5]}-{digest}'
 
 def restart_network(config: Config) -> None:
     """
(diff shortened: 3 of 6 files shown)


l0crian1:
@c-po 

What do you think about rather than using a unique hash per name > 15 characters, you truncate in the middle using something like this?:

```
def truncate_middle(value, max_length=10):
    if len(value) <= max_length:
        return value

    remaining = max_length - 2  # Account for ".."
    left = (remaining + 1) // 2
    right = remaining // 2

    return value[:left] + ".." + value[-right:]
```
Then it'll keep the uniqueness of the name, while still being obvious to the user which interface is owned by which container.
```
print(truncate_middle("homeassistant1"))
print(truncate_middle("homeassistant2"))

home..ant1
home..ant2
```
Without that, the user would have to dig to find which interface belongs to homeassistant1, so it'd be annoying to try to use in something like firewall rules.

jestabro (dismissed):
A reasonable solution to naming collision. I would suggest considering @l0crian1 suggestion of preserving suffixes; coderabbit comment can be ignored, since it regards smoketest.

c-po:
> @c-po
> 
> What do you think about rather than using a unique hash per name > 15 characters, you truncate in the middle using something like this?:
> ... 
> Without that, the user would have to dig to find which interface belongs to homeassistant1, so it'd be annoying to try to use in something like firewall rules.

This is a compelling idea as my containers ar also numbered in an identical pattern like `radius-1` or `tacacs-1`.

c-po:
@l0crian1 I see one disadvantage of this approach as some container instances could result in the same interface name. My idea is to combine both approaches. use `veth-NAME-hash` to make the interface name constant and hard to  hit with a second name and also add a new op-mode command `show container interfaces` to get a list of container to interface mapping.

Moved this to DRAFT to implement this change.

c-po:
Updated PR which also adds a dependency on `aardvark-dns (>=1.14.0)`

l0crian1:
@c-po 

I think using ConfigQuery for the `show container interfaces` could be slow on instances with large configs (at least based on my experience with other commands that need to do it).

What do you think of just redoing some of the work that is done in the conf_mode script, like checking length and creating hash if the length exceeds character limits? Maybe use something like these to return necessary lists (of course you can also do the JSON parsing in python if desired):

### Get list of networks - This avoids host networking and macvlan networks
```
sudo podman network ls --format json | jq '[.[] | select(.driver == "bridge") | .name]'
```
##### Example:
```
sudo podman network ls --format json | jq '[.[] | select(.driver == "bridge") | .name]'
[
  "podman",
  "test"
]
```
### Get a list of containers with network attachment, and return dict with container name:
```
sudo podman ps --format json | jq '[.[] | select(.Networks | length > 0) | .Names[0]]'
```
##### Example:
```
[
  {
    "name": "nb1",
    "networks": [
      "test"
    ]
  }
]
```
Then it's as simple as:

- Check if the contents of "networks:" is in the list returned from the `podman network ls` command. If it is, an interface should be determined.
- If the interface doesn't exceed character limits, then return the container name. If it does, return the same modified name with the digest.

As an alternative, you could probably iterate over the containers, and do `podman inspect` on them, looking for the net value in `"CreateCommand":`

jestabro (approved):
Re-approving, following changes since initial approval: (1) minor refactoring of conf-mode (2) addition of op-mode command; tested on built image.

PR includes comprehensive smoketests.

I'll leave it to @c-po to consider the most recent suggestion of @l0crian1 regarding the known issue of overhead of `ConfigQuery.config_dict` on large configs (since it requires parsing/instantiation of the full Config for what should be a simple op-mode command). Recall also the hacky workaround of using `op_mode_config_dict` from configquery.py which limits the parsing to the path in question, and consequently scales better; it may be used here since default values are not needed. I'm not advocating it over a simpler solution, simply a reminder of alternatives.

sever-sever (approved):
Deterministic veth naming looks correct, tests pass, no blockers to merge.

PR #5353, opened 2026-07-26
http-api: T9130: recover mid-life config session loss

BradKollmyer:
## Summary

Fixes [T9130](https://vyos.dev/T9130): when the long-lived HTTP API `ConfigSession` dies mid-process, every `POST /configure` fails with `without config session` and the error-path `discard()` raises again, producing an opaque HTTP 500 while `vyos-http-api.service` stays active. Only a unit restart recreated the session.

- **`ConfigSession`:** `session_exists()` / `ensure_session()`; `discard()` no-ops on a lost session
- **REST `/configure`:** under the existing global lock, ensure/recover the session before ops; return **503** `config session unavailable; retry` when recovery fails instead of wedging the process
- **Tests:** `src/tests/test_configsession_session_recover.py`

Observed in production on VyOS 1.5 rolling (two routers, 2026-07-24) after failed/partial apply waves. Distinct from T2292 (shutdown-time session teardown).

## Test plan

- [x] Unit tests: `PYTHONPATH=python python3 -m unittest src.tests.test_configsession_session_recover -v` (7 pass)
- [ ] On a rolling image with this patch: leave API up, force session death (or reproduce mid-life desync), confirm next `/configure` self-heals or returns clear 503 without process restart
- [ ] Confirm valid configure batches still return 200 and commit as before
- [ ] Confirm ordinary validation `ConfigSessionError` still returns 400 (not 503)

### DIFF
--- a/src/services/api/rest/routers.py
+++ b/src/services/api/rest/routers.py
@@ -53,6 +53,10 @@ from vyos.configsession import ConfigSessionError
 from ..background import BackgroundOpManager
 from ..background import BackgroundOpError
 from ..session import SessionState
+from ..session_recover import SESSION_UNAVAILABLE
+from ..session_recover import is_session_lost
+from ..session_recover import recreate_config_session
+from ..session_recover import safe_discard
 from .models import success
 from .models import error
 from .models import responses
@@ -93,6 +97,7 @@ LOG = logging.getLogger('http_api.routers')
 lock = Lock()
 
 asynclock = asyncio.Lock()
+
 
 def check_auth(key_list, key):
     key_id = None
@@ -409,8 +414,6 @@ def _execute_configure_op(
     is_background_job = background_tasks is None
 
     state = SessionState()
-    session = state.session
-    env = session.get_session_env()
 
     # A non-zero confirm_time will start commit-confirm timer on commit
     confirm_time = 0
@@ -428,12 +431,23 @@ def _execute_configure_op(
     # so the lock is really global
     lock.acquire()
 
-    config = Config(session_env=env)
-
     status = 200
     msg = None
     error_msg = None
+    env = None
+    session = None
     try:
+        session = state.session
+        if session is None:
+            if not recreate_config_session(state):
+                status = 503
+                error_msg = SESSION_UNAVAILABLE
+                raise ConfigSessionError(error_msg)
+            session = state.session
+
+        env = session.get_session_env()
+        config = Config(session_env=env)
+
         for c in data:
             op = c.op
             op_error = ConfigSessionError(f"'{op}' is not a valid operation")
@@ -530,20 +544,28 @@ def _execute_configure_op(
 
         LOG.info(f"Configuration modified via HTTP API using key '{state.id}'")
     except ConfigSessionError as e:
-        session.discard()
-        status = 400
+        safe_discard(session)
         if state.debug:
             LOG.critical(f'ConfigSessionError:\n {traceback.format_exc()}')
-        error_msg = str(e)
+        if is_session_lost(e) or error_msg == SESSION_UNAVAILABLE:
+            # Recreate ConfigSession for the next request; do not leave
+            # the process wedged on HTTP 500.
+            if status != 503:
+                recreate_config_session(state)
+            status = 503
+            error_msg = SESSION_UNAVAILABLE
+        else:
+            status = 400
+            error_msg = str(e)
     except Exception:
-        session.discard()
+        safe_discard(session)
         LOG.critical(traceback.format_exc())
         status = 500
 
         # Don't give the details away to the outer world
         error_msg = 'An internal error occurred. Check the logs for details.'
     finally:
-        if 'IN_COMMIT_CONFIRM' in env:
+        if env is not None and 'IN_COMMIT_CONFIRM' in env:
             del env['IN_COMMIT_CONFIRM']
         lock.release()
 
--- a/src/services/api/session_recover.py
+++ b/src/services/api/session_recover.py
@@ -1,0 +1,80 @@
+# Copyright VyOS maintainers and contributors <maintainers@vyos.io>
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+"""HTTP API recovery when the long-lived ConfigSession backend is gone.
+
+ConfigSession stays a thin wrapper of the config backend. Recreating
+ConfigSession(pid) is the same initialization the API process does at
+start — not ConfigSession self-repair (T9130 / jestabro).
+"""
+
+import logging
+import os
+
+LOG = logging.getLogger('http_api.session')
+
+# Substring from cli-shell-api / my_* when the config session is gone.
+SESSION_LOST_MARKER = 'without config session'
+SESSION_UNAVAILABLE = 'config session unavailable; retry'
+
+
+def is_session_lost(err: BaseException) -> bool:
+    return SESSION_LOST_MARKER in str(err)
+
+
+def abandon_session(session) -> None:
+    """Stop GC teardown of the dead object so it cannot race the replacement.
+
+    Old and new ConfigSession share the process PID as session id; the
+    leftover weakref finalizer would teardownSession the new one.
+    """
+    if session is None:
+        return
+    session.shared = True
+    finalizer = getattr(session, '_finalizer', None)
+    if finalizer is not None:
+        finalizer.detach()
+        session._finalizer = None
+
+
+def recreate_config_session(state, config_session_cls=None) -> bool:
+    """Replace state.session with a newly constructed ConfigSession.
+
+    Caller must hold the configure lock. Returns True if state.session
+    is set afterward.
+    """
+    if config_session_cls is None:
+        from vyos.configsession import ConfigSession as config_session_cls
+
+    abandon_session(getattr(state, 'session', None))
+    try:
+        state.session = config_session_cls(os.getpid())
+        LOG.warning('HTTP API config session recreated')
+        return True
+    except Exception as e:
+        LOG.error('failed to recreate HTTP API config session: %s', e)
+        state.session = None
+        return False
+
+
+def safe_discard(session) -> None:
+    """discard() without turning a dead session into a second exception."""
+    if session is None:
+        return
+    try:
+        session.discard()
+    except Exception as err:
+        if not is_session_lost(err):
+            raise
--- a/src/tests/test_http_api_session_recover.py
+++ b/src/tests/test_http_api_session_recover.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+#
+# Copyright VyOS maintainers and contributors <maintainers@vyos.io>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or later as
+# published by the Free Software Foundation.
+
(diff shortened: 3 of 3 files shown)


BradKollmyer:
Addressed CodeRabbit: rebind Vyconf `weakref.finalize` in `ensure_session` via `_rebind_vyconf_finalizer` (detach old + bind replacement). Unit test added; 8/8 recovery tests pass.

BradKollmyer:
Addressed CodeRabbit: on `ensure_session` recovery, build Vyconf in a temp then assign; on `ConfigSessionError` clear `_vyconf_session` and rebind finalizer so a half-recovered session cannot keep a stale backend. Test added; 9/9 recovery tests pass.

jestabro:
Rebased to pull in CI changes necessary for integration tests.

BradKollmyer:
Rebased onto current rolling for CI. Also wrapped long ConfigSessionError lines in the recover tests (Darker was failing on those).

src/tests/test_configsession_session_recover.py
```diff
@@ -0,0 +1,206 @@
...
+    'google.protobuf',
+    'vyos.vyconf_session',
+    'vyos.proto',
+    'vyos.proto.vyconf_pb2',
+    'vyos.proto.vyconf_client',
+):
+    sys.modules.setdefault(_mod, MagicMock())
+
+from vyos.configsession import ConfigSession
+from vyos.configsession import ConfigSessionError
```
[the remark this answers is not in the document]
BradKollmyer: Addressed in 8cc16ac3b: stubs go in via patch.dict only for the ConfigSession import, and only for modules that were not already loaded. Restored afterward so later nose2 tests see real requests/psutil/protobuf.

BradKollmyer:
CodeRabbit follow-up in `8cc16ac3b`: host-dep stubs for the recover tests are scoped to the `vyos.configsession` import via `patch.dict`.

BradKollmyer:
Rebased onto current rolling for CI.

sever-sever (changes requested):
Build binary fails
```
test_uid_root (test_utils_auth.TestVyOSUtilsAuth.test_uid_root) ... ok
FAIL: test_ensure_session_clears_stale_vyconf_on_construct_fail (test_configsession_session_recover.TestConfigSessionRecover.test_ensure_session_clears_stale_vyconf_on_construct_fail)
VyconfSession() failure must not leave the old backend attached.
  File "/__w/vyos-1x/vyos-1x/packages/vyos-1x/src/tests/test_configsession_session_recover.py", line 189, in test_ensure_session_clears_stale_vyconf_on_construct_fail
    fin.assert_not_called()
  File "/usr/lib/python3.11/unittest/mock.py", line 894, in assert_not_called
    raise AssertionError(msg)
AssertionError: Expected 'finalize' to not have been called. Called 1 times.
Calls: [call(<vyos.configsession.ConfigSession object at 0x7fab93447290>, <bound method ConfigSession.finalize_legacy of <class 'vyos.configsession.ConfigSession'>>, {'SESSION_PID': '12345'})].
FAIL: test_ensure_session_rebinds_vyconf_finalizer (test_configsession_session_recover.TestConfigSessionRecover.test_ensure_session_rebinds_vyconf_finalizer)
Replacement VyconfSession must own the finalizer, not the stale one.
  File "/__w/vyos-1x/vyos-1x/packages/vyos-1x/src/tests/test_configsession_session_recover.py", line 142, in test_ensure_session_rebinds_vyconf_finalizer
    self.assertEqual(args[1].__func__, ConfigSession.finalize_vyconf.__func__)
AssertionError: <function ConfigSession.finalize_legacy at 0x7fab9229e0c0> != <function ConfigSession.finalize_vyconf at 0x7fab9229e020>
Ran 207 tests in 3.572s
FAILED (failures=2, skipped=1)
empty-node {
}
normal-node {
    normal-node-child {
        /* valueless node comment */
        another-valueless-node
        multi-node "value1"
        multi-node "value1"
        one-more-valueless-node
        tag-node bar {
            some-option "some-value"
        }
        tag-node foo {
        }
        valueless-node
    }
    option-with-quoted-value "some-value"
    value-with-leading-slashes "//other-value"
}
/* top level leaf node */
top-level-leaf-node "foo"
top-level-tag-node bar {
    top-level-tag-node-child "another-value"
}
top-level-tag-node foo {
    top-level-tag-node-child "some-value"
}
top-level-tag-node baz {
    top-level-tag-node-child "another-value"
}
top-level-valueless-node
trailing-leaf-node-option "some-value"
trailing-leaf-node-without-value
// some version string info
// continued
empty-node {
}
normal-node {
    normal-node-child {
        /* valueless node comment */
        another-valueless-node
        multi-node "value1"
        multi-node "value1"
        one-more-valueless-node
        tag-node bar {
            some-option "some-value"
        }
        tag-node foo {
        }
        valueless-node
    }
    option-with-quoted-value "some-value"
    value-with-leading-slashes "//other-value"
}
/* top level leaf node */
top-level-leaf-node "foo"
top-level-tag-node quux {
    top-level-tag-node-child "another-value"
}
top-level-tag-node foo {
    top-level-tag-node-child "some-value"
}
top-level-valueless-node
trailing-leaf-node-option "some-value"
trailing-leaf-node-without-value
// some version string info
// continued
make[2]: *** [Makefile:106: test] Error 1
make[2]: Leaving directory '/__w/vyos-1x/vyos-1x/packages/vyos-1x'
make[1]: Leaving directory '/__w/vyos-1x/vyos-1x/packages/vyos-1x'
make[1]: *** [debian/rules:51: override_dh_auto_build] Error 2
make: *** [debian/rules:41: build] Error 2
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
Error: Process completed with exit code 2.
```

BradKollmyer:
@sever-sever the two recover tests failed because `_rebind_vyconf_finalizer` registered `finalize_legacy` after (or instead of) `finalize_vyconf`.

`5ca95d383` uses `ConfigSession.finalize_vyconf` (classmethod) with elif, and the VyconfSession() failure path only detaches the old finalizer.

src/tests/test_configsession_session_recover.py
```diff
@@ -0,0 +1,212 @@
...
+                            with patch(
+                                'vyos.configsession.VyconfSession',
+                                return_value=new_vyconf,
+                            ):
+                                with patch(
+                                    'vyos.configsession.weakref.finalize',
+                                    return_value=new_finalizer,
+                                ) as fin:
+                                    self.assertTrue(s.ensure_session())
+
```
alexandr-san4ez: I think you can utilize a single `with` statement combined with a context manager that groups all the patches together. Like this:

```python
    with contextlib.ExitStack() as stack:
        stack.enter_context(patch.object(s, 'session_exists', exists))
        stack.enter_context(patch('vyos.configsession.subprocess.check_output', return_value=b''))
        stack.enter_context(patch.object(s, '_ConfigSession__run_command', return_value=''))
        stack.enter_context(patch('vyos.configsession.vyconf_backend', return_value=True))
        stack.enter_context(patch('vyos.configsession.boot_configuration_complete', return_value=True))
        stack.enter_context(patch('vyos.configsession.VyconfSession', return_value=new_vyconf))
        fin = stack.enter_context(patch('vyos.configsession.weakref.finalize', return_value=new_finalizer))

        self.assertTrue(s.ensure_session())
```

src/tests/test_configsession_session_recover.py
```diff
@@ -0,0 +1,212 @@
...
+                            with patch(
+                                'vyos.configsession.VyconfSession',
+                                side_effect=ConfigSessionError(
+                                    'vyconf construct failed'
+                                ),
+                            ):
+                                with patch(
+                                    'vyos.configsession.weakref.finalize'
+                                ) as fin:
+                                    self.assertFalse(s.ensure_session())
```
alexandr-san4ez: As I commented before instead of nesting `with` statements create a single `with` statement that initializes `ExitStack`.

alexandr-san4ez (changes requested):
Also please combine a several commits into one which contains all information about the changes.

alexandr-san4ez:
And we still have a problem with building image:
```
FAIL: test_ensure_session_rebinds_vyconf_finalizer (test_configsession_session_recover.TestConfigSessionRecover.test_ensure_session_rebinds_vyconf_finalizer)
Replacement VyconfSession must own the finalizer, not the stale one.
  File "/vyos/vyos-1x/src/tests/test_configsession_session_recover.py", line 142, in test_ensure_session_rebinds_vyconf_finalizer
    self.assertEqual(args[1].__func__, ConfigSession.finalize_vyconf.__func__)
AssertionError: <function ConfigSession.finalize_legacy at 0x7f3320c89580> != <function ConfigSession.finalize_vyconf at 0x7f3320c894e0>
```

src/tests/test_configsession_session_recover.py
BradKollmyer: Addressed in 336e2fe1f: the recover tests now use a single `contextlib.ExitStack` for the patch set, matching this suggestion.

src/tests/test_configsession_session_recover.py
BradKollmyer: Addressed in 336e2fe1f: construct-fail and the other recover tests use the same ExitStack pattern.

python/vyos/configsession.py
```diff
@@ -268,6 +282,65 @@ def get_session_env(self):
...
+            self._rebind_vyconf_finalizer()
+            return self.session_exists()
+        except ConfigSessionError:
+            # setupSession may have succeeded while Vyconf construction failed —
+            # drop the stale backend so the next call cannot route through it.
+            # Detach only: do not register a replacement finalizer here.
+            self._vyconf_session = None
+            finalizer = getattr(self, '_finalizer', None)
+            if finalizer is not None:
+                finalizer.detach()
```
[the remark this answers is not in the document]
BradKollmyer: Addressed in 336e2fe1f: every failed recover path now clears `_vyconf_session` and detaches `_finalizer` (`setupSession` error, any `VyconfSession()` exception including `VyconfSessionError`, and `session_exists()` still false after rebind). The REST recreate fallback also detaches the old finalizer so GC cannot tear down the replacement.

BradKollmyer:
@alexandr-san4ez @sever-sever folded into `336e2fe1f` (squashed, rebased onto current rolling).

The image-build failures were the recover tests patching a **second** `vyos.configsession` module. `patch.dict(sys.modules)` around the import dropped the module the class still used, so `patch('vyos.configsession.vyconf_backend')` re-imported a fresh copy and was a no-op — `_rebind_vyconf_finalizer` then registered `finalize_legacy`. The module is pinned in `sys.modules` after import so patches hit the same globals.

Also: nested patches are one `ExitStack`; failed recover (including `VyconfSessionError`) detaches and does not register a replacement; recreate fallback detaches the old finalizer.

jestabro (changes requested):
I do not see this as a useful approach to the rare loss of configsession within http-api: it inverts the responsibilities of configsession and http-api, and also confuses the relationship between legacy and vyconf sessions.

ConfigSession is a thin abstraction of a config session, of the appropriate backend. As such, it is fundamental to two settings: http-api and smoketests, and should be kept as simple as possible.

Rather, we can consider other mechanisms for a long-running http-api service whose requirements fail due to a catastrophic failure of the backend communication (as is clearly the case, if/when seen). This should amount to a service restart, more or less, with all proper re-initialization, instead of trying to recreate as a task for configsession to self-repair.

BradKollmyer:
Rebased onto current rolling (`ac8b321bd`).

BradKollmyer:
@jestabro thanks — that matches how we recover this in production (`systemctl restart vyos-http-api.service`). Putting self-repair on `ConfigSession` was the wrong layer.

I’ll drop `ensure_session` / finalizer rebind from `ConfigSession` and keep that class as the thin backend wrapper. The HTTP API path can treat a dead session as a backend failure and re-initialize (or restart the unit) so the next `/configure` gets a fresh `ConfigSession` instead of HTTP 500.

Happy to reshape this PR that way, or close it and open a smaller http-api-only change if you prefer.

BradKollmyer:
@jestabro reshaped in `4ee255610`. `ConfigSession` is untouched. Session loss is handled in the HTTP API: abandon the dead object (detach its GC finalizer so the shared PID session is not torn down), construct a new `ConfigSession` (same init as process start), return 503, and swallow `discard` on a dead session so it cannot cascade into HTTP 500. The next `/configure` uses the new session.

jestabro:
Thanks, @BradKollmyer , I will take a look.

PR #5362, opened 2026-07-30
utils: T9144: fix infinite loop in ask_yes_no on EOF

dniminenn:
## Change summary

`ask_yes_no()` retries on EOFError. With stdin closed or not a terminal, for example `commit-confirm` run from a script over ssh, `input()` raises EOFError on every retry and the loop never exits. The session appears to hang while repeating "Please respond with yes/y or no/n".

This resolves EOF to the prompt's default answer, identical to an interactive user pressing Enter. Callers already declare the safe default at each call site. All call sites with destructive consequences declare `default=False`, so EOF aborts them; `default=True` sites are commit-confirm/rollback (which arm an automatic revert) and installer conveniences behind a False-gated disk confirmation.

One observable behavior change: a headless `install image` previously hung at its first prompt; it now walks the prompts until the destroy-disk confirmation answers No and exits cleanly.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9144

## Related PR(s)

A follow-up PR to vyatta-cfg adds `-y` passthrough to the `commit-confirm` configure-mode command (same task).

On VyOS Stream 2026.03:

```
$ cat /tmp/t.sh
source /opt/vyatta/etc/functions/script-template
configure
set interfaces dummy dum99
commit-confirm
exit
$ ssh vyos@router "vbash /tmp/t.sh"
```

Before: the session never returns, output loops "Please respond with yes/y or no/n".
After: commit proceeds with the revert timer armed, and `confirm` works as usual.

Interactive behavior is unchanged except Ctrl-D, which now acts like pressing Enter instead of re-prompting.

### DIFF
--- a/python/vyos/utils/io.py
+++ b/python/vyos/utils/io.py
@@ -71,7 +71,8 @@ def ask_yes_no(question, default=False) -> bool:
             else:
                 stdout.write("Please respond with yes/y or no/n\n")
         except EOFError:
-            stdout.write("\nPlease respond with yes/y or no/n\n")
+            stdout.write("\n")
+            return default
         except KeyboardInterrupt:
             return False
 


sarthurdev:
I'm not sure changing behaviour here is a great idea. I would be more in favour of utilising `commit_confirm(..., no_prompt=False)` parameter so the automation can bypass the prompt.

What do you think @jestabro ?

jestabro:
@sarthurdev note that this PR is submitted as a companion of https://github.com/vyos/vyatta-cfg/pull/153 ; these are in discussion with @dmbaturin 
Moreover, we have https://github.com/vyos/vyos-1x/pull/5390 and discussion in this context: https://github.com/vyos/vyos-1x/pull/5390/changes
So, we should decide on the appropriate solution for the case of commit-confirm passed by shell scripts.

dmbaturin (changes requested):
I broadly agree that an option to execute `commit-confirm` without an interactive confirmation may be a good idea, but here I disagree. For an _interactive_ yes/no function, the fact that it never succeeds without an interactive confirmation is a part of its design.

A forced non-interactive mode shouldn't call this function at all in the forced code path.

PR #5371, opened 2026-08-03
ntp: T9159: add source-address option for client requests

ruben-herold:
## Change summary
  
chrony (the daemon backing VyOS' NTP service) has no way to pin the
local address used for outgoing client requests to upstream servers.
"interface"/"listen-address" only configure the server/listening role
(binddevice/bindaddress); nothing controls the source address chosen
for packets chronyd itself originates.
  
On a router with more than one usable egress path to a given NTP
server (e.g. multiple transit/peering sessions), the kernel's default
source-address selection depends on whichever route wins at query
time. If that route happens to egress through an interface whose own
address is not globally reachable (a private peering-fabric segment,
for example), the request goes out with an address the reply can
never route back to - an intermittent, path-dependent failure with no
existing workaround at the NTP layer. BGP and friends already solve
the equivalent problem via "update-source" bound to the loopback; NTP
had no analogous option.
  
Add "service ntp source-address <address>", reusing the existing
source-address-ipv4-ipv6-multi include (same pattern already used for
e.g. RADIUS servers), and render it as chrony's "bindacqaddress"
directive - the client-side counterpart to "bindaddress" that chrony
already supports natively.
  
A companion documentation update is prepared (docs/configuration/service/ntp.md)
but not yet submitted - pending CLA signature on the docs repo.
  
## Types of changes
  
## Related Task(s)
* https://vyos.dev/T9159
  
## Checklist:
- [ ] I have run the components SMOKETESTS if applicable

### DIFF
--- a/data/templates/chrony/chrony.conf.j2
+++ b/data/templates/chrony/chrony.conf.j2
@@ -72,12 +72,17 @@ binddevice {{ interface }}
 {%     endif %}
 {% endif %}
 
-{% if source_address is vyos_defined %}
-# Source address used for outgoing NTP client requests, independent of
-# whichever interface the kernel's routing table selects for a given server
-{%     for address in source_address %}
+{% if source_address is vyos_defined or source_interface is vyos_defined %}
+# Source used for outgoing NTP client requests, independent of whichever
+# interface the kernel's routing table selects for a given server
+{%     if source_address is vyos_defined %}
+{%         for address in source_address %}
 bindacqaddress {{ address }}
-{%     endfor %}
+{%         endfor %}
+{%     endif %}
+{%     if source_interface is vyos_defined %}
+bindacqdevice {{ source_interface }}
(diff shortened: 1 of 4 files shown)


ruben-herold:

dmbaturin (dismissed):
I think it's a useful option and the implementation is straightforward. I have suggestions for error messages but they will need to be improved in the whole file so I'd say improving them is out of scope for this PR.

ruben-herold:
Thanks for the review! Agreed that reworking the error messages across the whole file is out of scope here.

Would you mind jotting down what you had in mind, even briefly? I'd like to pick it up as a separate follow-up once this lands, rather than let it drop.

sever-sever (approved):
Add NTP source-address setting

dmbaturin (approved):
I approved it earlier, the new changes don't look bad to me.

PR #5374, opened 2026-08-04
firewall: T9160: add apply-path for deriving group membership from config

ruben-herold:
## Change summary
  
  Firewall groups (network-group/address-group and their IPv6 equivalents)
  are frequently used to mirror facts that already exist elsewhere in the
  same configuration - e.g. a group listing every configured BGP neighbor
  address, kept in sync by hand with "protocols bgp neighbor". This drifts
  silently: nothing at commit time checks the manual copy is still
  complete, and the same duplication has to be repeated and kept correct
  across every router in a fleet.
  
  Add "apply-path" as an option on network-group/ipv6-network-group/
  address-group/ipv6-address-group, analogous to Junos's
  "policy-options prefix-list <name> apply-path". Its value is a
  space-separated configuration path; a "*" segment expands every
  instance of the tag node at that position (e.g. to fan out across every
  configured VRF). The resolved values are merged into the group's
  existing manually-configured members, deduplicated.
  
  The resolver (python/vyos/config_path_resolver.py) is a small,
  backend-agnostic module built on two primitives Config already
  exposes - list_nodes() (enumerate tag-node children, e.g. every BGP
  neighbor address) and return_values() (read a multi-value leaf) - so it
  isn't tied to firewall groups specifically and can be adopted by other
  consumers (e.g. policy prefix-lists) the same way.
  
  Since network-group/address-group are a single shared resource
  referenced by nat66, policy route, load-balancing wan and system
  conntrack (not just firewall rules), resolving apply-path once here,
  before the group's elements are rendered into the nftables set,
  benefits every one of those consumers without any changes on their side.

  Note on branch history: this PR previously carried 16 commits,
  including unrelated ruff/darker cleanup that had drifted in through
  merge-base staleness. Per review feedback, the branch has been
  rebased onto current rolling and squashed to a single, purely
  additive commit (10 files, all feature/feature-adjacent).
  
  ## Types of changes
  
  ## Related Task(s)
  * https://vyos.dev/T9160
  
  ## Related PR(s)
  * vyos/vyos-documentation#2187 (documents the new `apply-path` option)
  
  Unit tests (no config backend needed):
  ```
  $ PYTHONPATH=python python3 -m unittest src.tests.test_config_path_resolver -v
  test_apply_group_paths_merges_derived_members ... ok
  test_apply_group_paths_no_groups_is_a_noop ... ok
  test_apply_group_paths_rejects_invalid_derived_value ... ok
  test_apply_group_paths_rejects_non_canonical_network ... ok
  test_nonexistent_path ... ok
  test_refresh_apply_path_groups_builds_v4_and_v6_sets ... ok
  test_refresh_apply_path_groups_no_groups ... ok
  test_resolve_apply_paths_accepts_valid_values ... ok
  test_resolve_apply_paths_dedup_and_order ... ok
  test_resolve_apply_paths_rejects_invalid_value ... ok
  test_terminal_leaf_node ... ok
  test_terminal_tag_node ... ok
  test_wildcard_expansion ... ok
  Ran 13 tests in 0.005s
  OK
  ```

  Manual CLI check - derive a network-group from configured BGP
  neighbors:
  ```
  set protocols bgp neighbor 192.0.2.1 remote-as 65001
  set protocols bgp neighbor 192.0.2.2 remote-as 65001
  set firewall group network-group NEIGHBORS apply-path "protocols bgp neighbor *"
  commit

  run show firewall group name NEIGHBORS
  # -> 192.0.2.1/32, 192.0.2.2/32 (derived, no manual network entries needed)
  ```
  Full CLI smoketest coverage is in `test_firewall.py` (apply-path
  merge/dedup/validation cases); ran via
  `/usr/libexec/vyos/tests/smoke/cli/test_firewall.py` on the CI image,
  see the CI run linked on this PR.
  
  ## Checklist:
  - [x] I have run the components SMOKETESTS if applicable

### DIFF
--- a/python/vyos/utils/network.py
+++ b/python/vyos/utils/network.py
@@ -772,6 +772,42 @@ def is_valid_ipv6_address_or_range(addr: str) -> bool:
         return False
 
 
+def is_valid_ipv4_network(net: str) -> bool:
+    """
+    Validates if the provided value is a valid, canonical IPv4 network
+    (no host bits set), requiring an explicit prefix length (unlike
+    is_valid_ipv4_address_or_range, which also accepts a bare address)
+    :param net: value to test
+    :return: bool: True if provided value is a valid IPv4 network
+    """
+    from ipaddress import ip_network
+
+    if '/' not in net:
+        return False
+    try:
+        return ip_network(net, strict=True).version == 4
+    except ValueError:
+        return False
+
+
+def is_valid_ipv6_network(net: str) -> bool:
+    """
+    Validates if the provided value is a valid, canonical IPv6 network
+    (no host bits set), requiring an explicit prefix length (unlike
+    is_valid_ipv6_address_or_range, which also accepts a bare address)
+    :param net: value to test
+    :return: bool: True if provided value is a valid IPv6 network
+    """
+    from ipaddress import ip_network
+
+    if '/' not in net:
+        return False
+    try:
+        return ip_network(net, strict=True).version == 6
+    except ValueError:
+        return False
+
+
 def get_interfaces_by_ip(ip_address: str) -> list:
     """
     Return a list of all interface names assigned the given IP address.
--- a/smoketest/scripts/cli/test_firewall.py
+++ b/smoketest/scripts/cli/test_firewall.py
@@ -167,6 +167,122 @@ class TestFirewall(VyOSUnitTestSHIM.TestCase):
         self.cli_delete(['system', 'static-host-mapping'])
         self.cli_commit()
 
+    def test_apply_path(self):
+        # "protocols static route" is not under the "firewall" subtree that
+        # tearDown() cleans up, so it must be removed here even if an
+        # assertion below fails - otherwise it leaks into later tests.
+        derived_networks = ['203.0.113.0/24', '203.0.114.0/24']
+        for network in derived_networks:
+            self.cli_set(['protocols', 'static', 'route', network, 'blackhole'])
+
+        try:
+            self.cli_set(
+                [
+                    'firewall',
+                    'group',
+                    'network-group',
+                    'smoketest_derived',
+                    'network',
+                    '172.16.200.0/24',
+                ]
+            )
+            self.cli_set(
+                [
+                    'firewall',
+                    'group',
+                    'network-group',
+                    'smoketest_derived',
+                    'apply-path',
+                    'protocols static route',
+                ]
+            )
+
+            self.cli_set(
+                [
+                    'firewall',
+                    'ipv4',
+                    'forward',
+                    'filter',
+                    'rule',
+                    '1',
+                    'action',
+                    'accept',
+                ]
+            )
+            self.cli_set(
+                [
+                    'firewall',
+                    'ipv4',
+                    'forward',
+                    'filter',
+                    'rule',
+                    '1',
+                    'source',
+                    'group',
+                    'network-group',
+                    'smoketest_derived',
+                ]
+            )
+
+            self.cli_commit()
+
+            # The manually configured member and both derived members (from
+            # "protocols static route") must all end up in the rendered set
+            nftables_search = [
+                ['ip saddr @N_smoketest_derived', 'accept'],
+                ['elements = { 172.16.200.0/24, 203.0.113.0/24,'],
+                ['203.0.114.0/24 }'],
+            ]
+            self.verify_nftables(nftables_search, 'ip vyos_filter')
+        finally:
+            for network in derived_networks:
+                self.cli_delete(['protocols', 'static', 'route', network])
+            self.cli_commit()
+
+    def test_apply_path_invalid_value(self):
+        # apply-path values never go through a CLI-level <validator/>, as
+        # they are not entered on the CLI. A resolved value that is not
+        # valid for the target leaf (here: a network-group name resolved
+        # into an address-group's "address" leaf) must be rejected instead
+        # of silently entering the nftables set.
+        self.cli_set(
+            [
+                'firewall',
+                'group',
+                'network-group',
+                'smoketest_apply_path_source',
+                'network',
+                '172.16.201.0/24',
+            ]
+        )
+
+        self.cli_set(
+            [
+                'firewall',
+                'group',
+                'address-group',
+                'smoketest_bad_apply_path',
+                'address',
+                '10.0.0.1',
+            ]
+        )
+        self.cli_set(
+            [
+                'firewall',
+                'group',
+                'address-group',
+                'smoketest_bad_apply_path',
+                'apply-path',
+                'firewall group network-group',
+            ]
+        )
+
+        # ConfigError() wraps its message at 72 characters, so match only a
+        # short fragment that can't be split across a wrap point
+        error_message = 'resolved invalid'
+        with self.assertRaisesRegex(ConfigSessionError, error_message):
+            self.cli_commit()
+
     def test_nested_groups(self):
         self.cli_set(['firewall', 'group', 'network-group', 'smoketest_network', 'network', '172.16.99.0/24'])
         self.cli_set(['firewall', 'group', 'network-group', 'smoketest_network1', 'network', '172.16.101.0/24'])
--- a/src/conf_mode/firewall.py
+++ b/src/conf_mode/firewall.py
@@ -25,6 +25,7 @@ from vyos.config import Config
 from vyos.configdict import is_node_changed
 from vyos.configdiff import Diff, get_config_diff
 from vyos.configdep import set_dependents, call_dependents
+from vyos.config_path_resolver import apply_group_paths
 from vyos.configverify import verify_interface_exists
 from vyos.ethtool import Ethtool
 from vyos.firewall import fqdn_config_parse
@@ -141,6 +142,8 @@ def get_config(config=None):
     if firewall['group_resync']:
         # Update nat and policy-route as firewall groups were updated
         set_dependents('group_resync', conf)
+
+    apply_group_paths(dict_search_args(firewall, 'group'), conf)
 
     firewall['geoip_sets'] = geoip_sets(firewall)
     firewall['geoip_updated'] = geoip_updated(conf)
--- a/src/op_mode/firewall.py
+++ b/src/op_mode/firewall.py
@@ -18,15 +18,25 @@ import argparse
 import ipaddress
 import json
 import re
+import sys
 from signal import signal, SIGPIPE, SIG_DFL
 import tabulate
 import textwrap
 
+from vyos.base import ConfigError
 from vyos.config import Config
+from vyos.config_path_resolver import refresh_apply_path_groups
+from vyos.config_path_resolver import DictConfig
+from vyos.configquery import op_mode_config_dict
+from vyos.template import render
+from vyos.utils.locking import Lock
 from vyos.utils.process import cmdl
+from vyos.utils.process import run
 from vyos.utils.dict import dict_search_args
 
 signal(SIGPIPE, SIG_DFL)
+
+apply_path_nft_conf = '/run/nftables-apply-path-update.conf'
 
 
 def get_config_node(conf, node=None, family=None, hook=None, priority=None):
@@ -792,6 +802,57 @@ def show_statistics():
                 for prior, prior_conf in firewall[family][hook].items():
                     output_firewall_name_statistics(family, hook,prior, prior_conf)
 
+def update_apply_path_groups():
+    # Held for the whole read-resolve-render-apply sequence, not just the
+    # render+apply part: otherwise a slower invocation's stale config
+    # snapshot (read before the lock existed) could overwrite a faster,
+    # newer invocation's result once it finally gets the lock.
+    lock = Lock('firewall-apply-path-update')
+    lock.acquire()
+    try:
+        # op_mode_config_dict() instead of Config(): the whole-config fetch
+        # it needs anyway (apply-path targets can point anywhere in the
+        # tree) is no heavier than a plain Config() session, and this way we
+        # don't pay for a session object we never use for anything beyond
+        # list_nodes()/return_values()
+        groups = op_mode_config_dict(
+            ['firewall', 'group'],
+            key_mangling=('-', '_'),
+            get_first_key=True,
+            no_tag_node_value_mangle=True,
+        )
+        conf = DictConfig(op_mode_config_dict())
+        sets = refresh_apply_path_groups(groups, conf)
+
+        if not sets['v4'] and not sets['v6']:
+            print('No firewall group uses apply-path')
+            return
+
+        render(
+            apply_path_nft_conf,
+            'firewall/nftables-apply-path-update.j2',
+            {'sets': sets},
+            group='vyattacfg',
+            permission=0o664,
+        )
+
+        if run(f'nft --check --file {apply_path_nft_conf}') != 0:
+            print(
+                'Error: apply-path derived firewall group update failed nft syntax check'
+            )
+            sys.exit(1)
+
+        if run(f'nft --file {apply_path_nft_conf}') != 0:
+            print('Error: failed to update apply-path derived firewall groups')
+            sys.exit(1)
+    except ConfigError as e:
+        print(e)
+        sys.exit(1)
+    finally:
+        lock.release()
+
+    print('Updated apply-path derived firewall groups')
+
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
     parser.add_argument('--action', help='Action', required=False)
@@ -820,3 +881,5 @@ if __name__ == '__main__':
         show_statistics()
     elif args.action == 'show_summary':
         show_summary()
+    elif args.action == 'update_apply_path_groups':
+        update_apply_path_groups()
--- a/src/op_mode/firewall.py
+++ b/src/op_mode/firewall.py
@@ -19,6 +19,7 @@ import ipaddress
 import json
 import re
 import sys
+import time
 from signal import signal, SIGPIPE, SIG_DFL
 import tabulate
 import textwrap
@@ -29,6 +30,7 @@ from vyos.config_path_resolver import refresh_apply_path_groups
 from vyos.config_path_resolver import DictConfig
 from vyos.configquery import op_mode_config_dict
 from vyos.template import render
+from vyos.utils.commit import commit_in_progress
 from vyos.utils.locking import Lock
 from vyos.utils.process import cmdl
 from vyos.utils.process import run
@@ -803,6 +805,19 @@ def show_statistics():
                     output_firewall_name_statistics(family, hook,prior, prior_conf)
 
 def update_apply_path_groups():
+    # A concurrent firewall commit renders the same nftables sets this
+    # refresh writes to directly; wait it out first, the same way
+    # vyos-domain-resolver coordinates its own out-of-commit nftables pushes
+    # (FQDN/remote-group sets) with commit.
+    wait_start = time.time()
+    while commit_in_progress():
+        if time.time() - wait_start > 60:
+            print(
+                'Error: commit still in progress after 60s, aborting apply-path refresh'
+            )
+            sys.exit(1)
+        time.sleep(1)
+
     # Held for the whole read-resolve-render-apply sequence, not just the
     # render+apply part: otherwise a slower invocation's stale config
     # snapshot (read before the lock existed) could overwrite a faster,
--- a/data/templates/firewall/nftables-apply-path-update.j2
+++ b/data/templates/firewall/nftables-apply-path-update.j2
@@ -1,0 +1,39 @@
+#!/usr/sbin/nft -f
+
+{% if sets.v4 is vyos_defined %}
+{%     for setname, member_list in sets.v4.items() %}
+flush set ip vyos_filter {{ setname }}
+{%     endfor %}
+
+table ip vyos_filter {
+{%     for setname, member_list in sets.v4.items() %}
+    set {{ setname }} {
+        type ipv4_addr
+        flags interval
+        auto-merge
+{%         if member_list %}
+        elements = { {{ member_list | join(",") }} }
+{%         endif %}
+    }
+{%     endfor %}
+}
+{% endif %}
+
+{% if sets.v6 is vyos_defined %}
+{%     for setname, member_list in sets.v6.items() %}
+flush set ip6 vyos_filter {{ setname }}
+{%     endfor %}
+
+table ip6 vyos_filter {
+{%     for setname, member_list in sets.v6.items() %}
+    set {{ setname }} {
+        type ipv6_addr
+        flags interval
+        auto-merge
+{%         if member_list %}
+        elements = { {{ member_list | join(",") }} }
+{%         endif %}
+    }
+{%     endfor %}
+}
+{% endif %}
--- a/interface-definitions/firewall.xml.in
+++ b/interface-definitions/firewall.xml.in
@@ -80,6 +80,7 @@
                   <multi/>
                 </properties>
               </leafNode>
+              #include <include/firewall/apply-path.xml.i>
               #include <include/generic-description.xml.i>
             </children>
           </tagNode>
@@ -227,6 +228,7 @@
                   <multi/>
                 </properties>
               </leafNode>
+              #include <include/firewall/apply-path.xml.i>
               #include <include/generic-description.xml.i>
             </children>
           </tagNode>
@@ -262,6 +264,7 @@
                   <multi/>
                 </properties>
               </leafNode>
+              #include <include/firewall/apply-path.xml.i>
             </children>
           </tagNode>
           <tagNode name="mac-group">
@@ -330,6 +333,7 @@
                   <multi/>
                 </properties>
               </leafNode>
+              #include <include/firewall/apply-path.xml.i>
             </children>
           </tagNode>
           <tagNode name="port-group">
--- a/interface-definitions/include/firewall/apply-path.xml.i
+++ b/interface-definitions/include/firewall/apply-path.xml.i
@@ -1,0 +1,12 @@
+<!-- include start from firewall/apply-path.xml.i -->
+<leafNode name="apply-path">
+  <properties>
+    <help>Derive additional group members from another part of the configuration</help>
+    <valueHelp>
+      <format>txt</format>
+      <description>Space-separated configuration path; a "*" segment expands every instance of the tag node at that position (e.g. "vrf name * protocols bgp neighbor")</description>
+    </valueHelp>
+    <multi/>
+  </properties>
+</leafNode>
+<!-- include end -->
--- a/op-mode-definitions/firewall.xml.in
+++ b/op-mode-definitions/firewall.xml.in
@@ -773,4 +773,28 @@
       </node>
     </children>
   </node>
+  <node name="update">
+    <children>
+      <node name="firewall">
+        <properties>
+          <help>Update firewall information</help>
+        </properties>
+        <children>
+          <node name="group">
+            <properties>
+              <help>Update firewall group</help>
+            </properties>
+            <children>
+              <leafNode name="apply-paths">
+                <properties>
+                  <help>Re-resolve apply-path derived group membership and push it into nftables, without a firewall commit</help>
+                </properties>
+                <command>${vyos_op_scripts_dir}/firewall.py --action update_apply_path_groups</command>
+              </leafNode>
+            </children>
+          </node>
+        </children>
+      </node>
+    </children>
+  </node>
 </interfaceDefinition>
--- a/python/vyos/config_path_resolver.py
+++ b/python/vyos/config_path_resolver.py
@@ -1,0 +1,171 @@
+#!/usr/bin/env python3
+#
+# Copyright VyOS maintainers and contributors <maintainers@vyos.io>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or later as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from vyos.base import ConfigError
+from vyos.utils.network import is_valid_ipv4_address_or_range
+from vyos.utils.network import is_valid_ipv6_address_or_range
+from vyos.utils.network import is_valid_ipv4_network
+from vyos.utils.network import is_valid_ipv6_network
+
+
+def resolve_config_path(conf, path_str):
+    """Resolve a space-separated configuration path into the list of
+    values found at that path in the given (candidate) configuration.
+
+    A "*" segment expands every instance of the tag node at that position
+    (e.g. every configured VRF name), continuing the remainder of the path
+    under each instance. The terminal segment is resolved as every child
+    name of a tag node (e.g. every configured BGP neighbor address) if it
+    has any, otherwise as the values of a multi-value leaf node.
+    """
+    return _walk(conf, [], path_str.split())
+
+
+def _walk(conf, base, remaining):
+    if not remaining:
+        nodes = conf.list_nodes(base)
+        if nodes:
+            return nodes
+        return conf.return_values(base)
+
+    head, *rest = remaining
+    if head == '*':
+        result = []
+        for instance in conf.list_nodes(base):
+            result.extend(_walk(conf, base + [instance], rest))
+        return result
+
+    return _walk(conf, base + [head], rest)
+
+
+def resolve_apply_paths(conf, path_list, is_valid=None):
+    """Resolve a list of apply-path strings and return the deduplicated,
+    order-preserving union of every value they resolve to.
+
+    apply-path values never go through CLI-level <validator/> checks, since
+    they are not entered on the CLI. If "is_valid" is given, every resolved
+    value is checked against it; the first value that fails raises
+    ValueError(value).
+    """
+    seen = set()
+    result = []
+    for path_str in path_list:
+        for value in resolve_config_path(conf, path_str):
+            if is_valid is not None and not is_valid(value):
+                raise ValueError(value)
+            if value not in seen:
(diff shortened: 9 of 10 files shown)


l0crian1:
A couple thoughts:

1. Can you not remove the f-string from the ConfigError() messages. https://github.com/vyos/vyos-1x/pull/5377 will require them anyways, so they'd just get added back.
2. The biggest benefit to something like Juniper's solution is the syncing of the values. There wouldn't be anything to keep these in sync (unless I missed it in the PR). After you configure the firewall section, any updates to the BGP section (add/remove a neighbor) wouldn't be applied to the firewall group. They'd only be updated when you update the firewall config somewhere. I'm not sure if there' s a good way to handle that within VyOS' current config structure.
3. The address validation for firewall groups are currently handled by the CLI level checks (e.g. `<validator name="ipv4-address"/>`). These would be skipped, so there would need to be additional checks within your resolver.
4. I think it's a good feature if it can be implemented in a bi-directional manner with the other config sections, but I do want to note this is already possible with dynamic-groups (at least for elements that would be in the data path). This would actually sync with new neighbors when configured:
    #### Initial configuration.
    ```
    set firewall group dynamic-group address-group FROM_BGP
    set firewall ipv4 output filter rule 10 action 'accept'
    set firewall ipv4 output filter rule 10 add-address-to-group destination-address address-group 'FROM_BGP'
    set firewall ipv4 output filter rule 10 destination port '179'
    set firewall ipv4 output filter rule 10 protocol 'tcp'
    
    set protocols bgp neighbor 10.0.0.2 peer-group 'TEST'
    set protocols bgp neighbor 10.0.0.6 peer-group 'TEST'
    set protocols bgp neighbor 10.0.0.10 peer-group 'TEST'
    set protocols bgp peer-group TEST address-family ipv4-unicast
    set protocols bgp peer-group TEST remote-as '65000'
    set protocols bgp system-as '65000'
    ```
    ```
    vyos@vyos# run show firewall group FROM_BGP 
    Firewall Groups
    
    Name      Type                    References             Members    Timeout    Expires
    --------  ----------------------  ---------------------  ---------  ---------  ---------
    FROM_BGP  address_group(dynamic)  ipv4-output-filter-10  10.0.0.2   N/D        N/D
                                                             10.0.0.6   N/D        N/D
                                                             10.0.0.10  N/D        N/D
    ```
    #### Update configuration
    ```
    vyos@vyos# set protocols bgp neighbor 10.0.0.14 peer-group TEST
    vyos@vyos# commit
    ```
    ```
    vyos@vyos# run show firewall group FROM_BGP 
    Firewall Groups
    
    Name      Type                    References             Members    Timeout    Expires
    --------  ----------------------  ---------------------  ---------  ---------  ---------
    FROM_BGP  address_group(dynamic)  ipv4-output-filter-10  10.0.0.2   N/D        N/D
                                                             10.0.0.6   N/D        N/D
                                                             10.0.0.10  N/D        N/D
                                                             10.0.0.14  N/D        N/D
    ```

ruben-herold:
Thanks for the thorough review, @l0crian1!
  
  1) Reverted the f-string removal on the 5 ConfigError() calls in verify_rule() that #5377 touches — left everything else from that ruff cleanup as-is since it's unrelated to #5377.
  
  2) and 4) Good point, and I want to be explicit about scope here: apply-path and dynamic-groups solve different problems and aren't really alternatives to each other.
  
  dynamic-groups populate from traffic that actually matches a firewall rule (via add-address-to-group), so they get live resync essentially for free — the packet itself is the update signal. That's exactly why your BGP-neighbor example works: those neighbors' packets cross the rule you attached add-address-to-group to.
  
  apply-path is for the opposite case: config values that never appear as packets on the wire and therefore have nothing for dynamic-groups to observe — e.g. the protocols static route prefixes in the smoketest, VRF names, or a BGP neighbor's address when you want it in a group for a different rule than the one it's peering over. There's no
  traffic event to hang a dynamic-group off of there.
  
  Given that, I don't think bidirectional live sync belongs in this PR — the underlying problem (re-deriving a group when its source section changes, not just when the firewall section changes) is a config-dependency problem, not really specific to apply-path, and you already flagged you're not sure there's a clean way to do it in VyOS's current
  config model. I'd rather ship the one-directional version now (re-resolved on every firewall commit, which covers the common case of "I edited the source, then touch firewall to pick it up") and treat live cross-section resync as a separate, follow-up discussion if there's appetite for it.
  
  3) Added validation: resolve_apply_paths() now takes an optional is_valid callback and raises on the first resolved value that fails it. firewall.py wires in is_valid_ipv4_address_or_range/is_valid_ipv6_address_or_range for address groups and new is_valid_ipv4_network/is_valid_ipv6_network (added next to them in vyos.utils.network) for network
  groups, so a bad derived value (e.g. an unnumbered BGP neighbor's interface name landing in an address-group) now fails the commit with a clear ConfigError instead of reaching nftables. Added unit tests for the validator plumbing and a smoketest for the rejection path.
  
  Docs (vyos-documentation) aren't updated yet — holding off until the shape of apply-path here is settled, then I'll follow up with a docs PR.

l0crian1:
> apply-path is for the opposite case: config values that never appear as packets on the wire and therefore have nothing for dynamic-groups to observe

Yeah, that's why I mentioned the data path. I was giving that as an alternative to the BGP example specifically. It wouldn't solve other examples you gave like `prefix-lists`.

> Given that, I don't think bidirectional live sync belongs in this PR — the underlying problem (re-deriving a group when its source section changes, not just when the firewall section changes) is a config-dependency problem, not really specific to apply-path, and you already flagged you're not sure there's a clean way to do it in VyOS's current
> config model.

Yeah, I think that's fair. The correct way is probably a daemon, which I think could be heavy given the overhead in VyOS' config architecture.

Perhaps as a stop gap, add an `op-mode` command as part of this PR. Something like:
```
update firewall group <group-type> apply-paths
```
That can trigger an `nft -f` for something like this in a Jinja2 created file:
```
flush set ip vyos_filter A_test

table ip vyos_filter {
        set A_test {
                type ipv4_addr
                flags interval
                auto-merge
                elements = { 10.0.0.1 }
        }
}
```

ruben-herold:
Good idea — implemented in 6c00e2d (https://github.com/ruben-herold/vyos-1x/commit/6c00e2de254682f8603a404438cdc388d18a0ca4):
  
  update firewall group apply-paths
  
  It follows the exact same pattern update geoip already uses: a small Jinja2 file that flushes and re-declares only the affected sets, pushed with nft --file, so it doesn't touch the rest of the ruleset and doesn't need a full commit.
  
  A couple of choices worth flagging:
  
  - No <group-type> argument — it refreshes every apply-path group across all types in one pass, since re-resolving config paths is cheap and there's no real benefit to narrowing it (unlike GeoIP, where narrowing would save a database lookup). 
  - The resolution logic (apply_group_paths) moved from conf_mode/firewall.py into vyos.config_path_resolver, since op-mode scripts can't import from conf_mode/* — this way it's shared between commit-time derivation and the new op-mode refresh instead of being duplicated.
  
  Added unit test coverage for the new refresh_apply_path_groups() in test_config_path_resolver.py. Didn't add a smoketest for the op-mode command itself, mirroring that update geoip isn't smoketested either — the underlying resolution logic is what's covered.
  
  This is still a manual nudge, not automatic — happy to hear if you think it should be pushed further (e.g. wired into a cron/timer), but as a stopgap for the "source changed, firewall section didn't" case I think this covers it.

l0crian1:
For the op-mode update function, I recommend using `op_mode_config_dict()` instead of Config(). This is more lightweight since it doesn't instantiate a config session alongside returning the config.

Also, it could be wise to do an `nft -c` dry run before the actual `nft -f`. That way failures can be caught gracefully.

ruben-herold:
Thanks  — implemented all three outstanding points:

  @l0crian1 (op_mode_config_dict / nft dry-run, both in e8a775f9f (https://github.com/ruben-herold/vyos-1x/commit/e8a775f9f) and 9f69c3c85
  (https://github.com/ruben-herold/vyos-1x/commit/9f69c3c85)):
  - update_apply_path_groups() no longer instantiates a session-backed Config(). Added a small DictConfig adapter in config_path_resolver.py that exposes just the two
  primitives the resolver needs (list_nodes()/return_values()) over a plain dict, fed from op_mode_config_dict().
  - Added an nft --check --file dry run before the real nft --file, matching the existing pattern in conf_mode/firewall.py's apply().

  @coderabbitai (concurrency, in 3bf654349 (https://github.com/ruben-herold/vyos-1x/commit/3bf654349)):
  - Wrapped the render+apply of apply_path_nft_conf in a vyos.utils.locking.Lock, so two concurrent update firewall group apply-paths invocations can't clobber each
  other's rendered file.
 
  @coderabbitai — one suggestion I did not apply: the nitpick to stop redeclaring the table/set after flushing in nftables-apply-path-update.j2. That
  flush-then-redeclare shape is the same one already used by the proven nftables-geoip-update.j2 pattern, so I kept it for consistency rather than diverging.

  Also pushed two housekeeping commits (823711507 (https://github.com/ruben-herold/vyos-1x/commit/823711507), d86500b8d
  (https://github.com/ruben-herold/vyos-1x/commit/d86500b8d)) fixing the remaining ruff/darker findings in files this branch touches, ahead of the lint CI job
  (currently sitting in action_required on the fork). Left the 5 F541 (extraneous f-string) findings in conf_mode/firewall.py alone per our earlier discussion re:
  #5377. Also noticed an unrelated pre-existing bug in _are_same_ip() while triaging ruff output (wrong address family used on the second inet_pton() call) — filed
  separately as T9177 (https://vyos.dev/T9177) rather than folding it into this branch.

python/vyos/utils/network.py
```diff
@@ -25,7 +25,6 @@ def _are_same_ip(one, two):
     from vyos.template import is_ipv4
     # compare the binary representation of the IP
     f_one = AF_INET if is_ipv4(one) else AF_INET6
-    s_two = AF_INET if is_ipv4(two) else AF_INET6
```
sever-sever: Is this change related to this PR?
If not, you should not touch extra code.

python/vyos/utils/network.py
```diff
@@ -129,7 +128,7 @@ def get_vrf_members(vrf: str) -> list:
                 # Skip PIM interfaces which appears in VRF
                 if 'pim' not in data.get('ifname'):
                     interfaces.append(data.get('ifname'))
-    except:
+    except Exception:
         pass
     return interfaces
```
sever-sever: Is this change related to this PR?
If not, you should not touch extra code.

python/vyos/utils/network.py
```diff
@@ -217,7 +216,8 @@ def get_interface_namespace(interface: str):
     """
     # Bail out early if netns does not exist
     tmp = cmdl(['ip', '--json', 'netns', 'ls'])
-    if not tmp: return None
+    if not tmp:
+        return None
```
sever-sever, line 223: For the code refactoring should be a separate task.
Do not touch code, that is not related to feature that you want to add.

python/vyos/utils/network.py
```diff
@@ -337,7 +337,7 @@ def mac2eui64(mac, prefix=None):
             net = ip_network(prefix, strict=False)
             euil = int('0x{0}'.format(eui64), 16)
             return str(net[euil])
-        except:  # pylint: disable=bare-except
+        except Exception:
```
sever-sever, line 343: For the code refactoring should be a separate task.
Do not touch code, that is not related to feature that you want to add.

python/vyos/utils/network.py
```diff
@@ -379,7 +379,7 @@ def check_port_availability(address: str = None, port: int = 0,
     protocol = socket.SOCK_STREAM if protocol == 'tcp' else socket.SOCK_DGRAM
     try:
         addr_info = socket.getaddrinfo(address, port, socket.AF_UNSPEC, protocol)
-    except socket.gaierror as e:
+    except socket.gaierror:
```
sever-sever, line 385: For the code refactoring should be a separate task.
Do not touch code, that is not related to feature that you want to add.

python/vyos/utils/network.py
```diff
@@ -498,7 +498,6 @@ def is_intf_addr_assigned(ifname: str, addr: str, netns: str=None) -> bool:
         json_out = loads(out)
         addresses = jmespath.search("[].addr_info[].{family: family, address: local, prefixlen: prefixlen}", json_out)
         for address_info in addresses:
-            family = address_info['family']
```
sever-sever: Why it was removed?

python/vyos/utils/network.py
```diff
@@ -686,7 +685,7 @@ def ipv6_prefix_length(low, high):
     try:
         lo = bytearray(socket.inet_pton(socket.AF_INET6, low))
         hi = bytearray(socket.inet_pton(socket.AF_INET6, high))
-    except:
+    except Exception:
```
sever-sever: For the code refactoring should be a separate task.
Do not touch code, that is not related to feature that you want to add.

python/vyos/utils/network.py
```diff
@@ -750,7 +749,7 @@ def is_valid_ipv4_address_or_range(addr: str) -> bool:
             return is_valid_ipv4_address_or_range(split[0]) and is_valid_ipv4_address_or_range(split[1])
         else:
             return ip_network(addr).version == 4
-    except:
+    except Exception:
```
sever-sever: For the code refactoring should be a separate task.
Do not touch code, that is not related to feature that you want to add.

smoketest/scripts/cli/test_firewall.py
```diff
@@ -328,8 +444,6 @@ def test_ipv4_time_weekdays(self):
 
     def test_ipv4_advanced(self):
         name = 'smoketest-adv'
-        name2 = 'smoketest-adv2'
-        interface = 'eth0'
```
sever-sever, line 449: Why did you remove it?

smoketest/scripts/cli/test_firewall.py
```diff
@@ -415,7 +529,6 @@ def test_ipv4_synproxy(self):
 
     def test_ipv4_mask(self):
         name = 'smoketest-mask'
-        interface = 'eth0'
```
sever-sever, line 418: Is it related to feature that you want to implement?

src/conf_mode/firewall.py
```diff
@@ -567,7 +570,9 @@ def verify(firewall):
                             if 'jump' not in priority_conf['default_action']:
                                 raise ConfigError('default-jump-target defined, but default-action jump needed and it is not defined')
                             if priority_conf['default_jump_target'] == priority:
-                                raise ConfigError(f'Loop detected on default-jump-target.')
+                                raise ConfigError(
+                                    'Loop detected on default-jump-target.'
+                                )
```
sever-sever: For refactoring/styling there should be a separate PR.
Do not do refactoring with new feature in one PR

src/conf_mode/firewall.py
```diff
@@ -603,7 +608,9 @@ def verify(firewall):
                     for iface in zone_conf['member']['interface']:
 
                         if iface in zone_interfaces:
-                            raise ConfigError(f'Interfaces cannot be assigned to multiple zones')
+                            raise ConfigError(
+                                'Interfaces cannot be assigned to multiple zones'
+                            )
```
sever-sever: For refactoring/styling there should be a separate PR.
Do not do refactoring with new feature in one PR

src/conf_mode/firewall.py
```diff
@@ -613,7 +620,9 @@ def verify(firewall):
                 if 'vrf' in zone_conf['member']:
                     for vrf in zone_conf['member']['vrf']:
                         if vrf in zone_vrf:
-                            raise ConfigError(f'VRF cannot be assigned to multiple zones')
+                            raise ConfigError(
+                                'VRF cannot be assigned to multiple zones'
+                            )
```
sever-sever: For refactoring/styling there should be a separate PR.
Do not do refactoring with new feature in one PR

src/conf_mode/firewall.py
```diff
@@ -748,7 +757,7 @@ def apply(firewall):
     ## DOMAIN RESOLVER
     domain_action = 'restart'
     if dict_search_args(firewall, 'group', 'remote_group') or dict_search_args(firewall, 'group', 'domain_group') or firewall['ip_fqdn'].items() or firewall['ip6_fqdn'].items():
-        text = f'# Automatically generated by firewall.py\nThis file indicates that vyos-domain-resolver service is used by the firewall.\n'
+        text = '# Automatically generated by firewall.py\nThis file indicates that vyos-domain-resolver service is used by the firewall.\n'
```
sever-sever: There should be a separate PR/task for code refactoring

src/op_mode/firewall.py
```diff
@@ -108,7 +118,7 @@ def get_nftables_state_details(family):
     command = ['nft', 'list', 'chain', suffix, 'vyos_filter', f'VYOS_STATE_{name_suffix}']
     try:
         results = cmdl(command)
-    except:
+    except Exception:
```
sever-sever: There should be a separate PR for this change, as it is not related to the feature implementation.

src/op_mode/firewall.py
```diff
@@ -131,7 +141,7 @@ def get_nftables_group_members(family, table, name):
     try:
         results_str = cmdl(['nft', '-j', 'list', 'set', prefix, table, name])
         results = json.loads(results_str)
-    except:
+    except Exception:
         return out
```
sever-sever, line 151: There should be a separate PR for this change, as it is not related to the feature implementation.

src/op_mode/firewall.py
```diff
@@ -159,7 +169,7 @@ def get_nftables_remote_group_members(family, table, name):
     try:
         results_str = cmdl(['nft', '-j', 'list', 'set', prefix, table, name])
         results = json.loads(results_str)
-    except:
+    except Exception:
```
sever-sever: There should be a separate PR for this change, as it is not related to the feature implementation.

sever-sever (dismissed):
There are a lot of changes that do not related to this PR
Code refactoring should be a separate PR/task.
Also squash 14 commits into one. It will be easier to maintain/backport it in the future.
Nice to add the `How to test` steps with minimal set of configuration commands and expectation.

Thanks

ruben-herold:
Thanks for the review — you're right, this branch had picked up unrelated ruff/darker cleanup along the way just because it touched the same files. Split it out:

- The `network.py` bare-except/unused-var cleanup is dropped entirely here, since #5383 (T9177) already covers that same ground more thoroughly (it also fixes the actual `_are_same_ip()` bug, not just the lint findings).
- The remaining cleanup in `conf_mode`/`op_mode` firewall.py and the unused test locals in `test_firewall.py` is now its own task/PR: https://vyos.dev/T9205 / #5398.

This branch is back to just the apply-path feature — `git diff` against the fork point is now purely additive (no lines touched outside what apply-path itself needs).

ruben-herold:
Thanks for the review. Fixed the one finding that's actually caused by this PR's new code:

- `is_valid_ipv4_network()`/`is_valid_ipv6_network()` (network.py:773-806) used `strict=False`, accepting host-bit CIDRs like `192.0.2.1/24`. Since `apply_group_paths()` merges whatever passes validation straight into the group and it's rendered into an nftables interval set as-is, a non-canonical derived value would have gone through unnoticed. Switched both to `strict=True` so it now raises a `ConfigError` naming the offending group instead, matching how invalid-type values are already rejected. Added a regression test. Pushed as commit 80f27d009.

The other 3 findings (mixed-family bug in `_are_same_ip()`, bare `except` in `get_vrf_members()`, hyphenated-range `ValueError` in `is_intf_addr_assigned()`) are all in code this PR doesn't touch. The first two are already fixed in the open #5383/T9177. The third isn't covered by anything open yet — filing it separately rather than pulling it into this branch, per the earlier scope-creep discussion.

ruben-herold:
Filed the remaining out-of-scope finding as its own task: https://vyos.dev/T9207 (`is_intf_addr_assigned()` raises uncaught `ValueError` on hyphenated address ranges instead of returning `False`).

ruben-herold:
Thanks for the review — all three points addressed:

- Branch rebased onto current `rolling` and squashed from 16 commits down to one, purely additive (10 files, 677 insertions, 0 deletions). The unrelated-looking changes in the previous diff were mostly branch staleness (66 commits behind), not actual scope creep — the one real cleanup (T9207, filed separately) was already out of this branch.
- Added a "How to test" section with unit test output and a manual CLI walkthrough (derive a network-group from `protocols bgp neighbor *`).
- Linked the docs PR (vyos/vyos-documentation#2187) under Related PR(s).

Re-verified locally after the rebase: `config_path_resolver` unit tests (13/13) pass, no new ruff/darker findings introduced (checked touched files against current rolling baseline).

sarthurdev (commented):
I'm not sure this is a good idea, I would much prefer using existing solutions like @l0crian1 suggested with dynamic groups.

One example could be referenced paths being fully or partially migrated in the future. Breaking this feature, or raising complexity of creating future migrators.

ruben-herold:
Thanks for weighing in, @sarthurdev.

I went through the same question with @l0crian1 earlier in this thread — the short version: apply-path and dynamic-groups solve different problems, not the same one.

dynamic-groups populate from traffic that actually matches a firewall rule (via add-address-to-group), so the packet itself is the update signal. apply-path is for the opposite case: config values that never appear as packets on the wire, so there's nothing for dynamic-groups to observe — e.g. VRF names, protocols static route prefixes, or a BGP neighbor's address when you want it in a group attached to a rule other than the one it's peering over. There's no traffic event to hang a dynamic-group off of there.

On migration complexity: that's a fair, real cost, and I don't want to wave it away. If a future migrator restructures a referenced path (e.g. renames/moves `protocols bgp neighbor`), it would also need to grep `apply-path` values on network/address groups and rewrite those strings, since they're not part of the structural walk a normal path-rename migrator does. That's an extra thing to remember, on top of the migrator already having to fix up every other consumer of that path.

I don't think it's a reason to drop the feature, though — it's the same category of cost VyOS already pays whenever a CLI value is referenced by name outside its own subtree (e.g. group names in firewall rules, interface names in op-mode scripts), and migrators already have to search for those. I'm happy to call this out explicitly as a caveat in the docs PR (vyos/vyos-documentation#2187) so future migration authors know to grep for `apply-path` when moving a referenced path.

ruben-herold:
@sever-sever Ready for re-review — all three points addressed (squashed to a single commit, unrelated changes dropped, `How to test` steps added in the PR description).

c-po:
VyOS already has dynamic-groups as you also outlined, for deriving group membership from runtime state. That mechanism is atomic and in-kernel, so there is never a second source of truth to fall out of sync with. apply-path introduces a second, different mechanism for the same conceptual problem.

A few concrete concerns with the current approach. The op-mode refresh has its own lock, but nothing coordinates it with a concurrent firewall commit touching the same nftables sets, so a commit and an external trigger firing close together can race with no defined winner (I fixed several races the last days in the smoketest framewrok). apply-path derived members never enter the config tree, so show configuration, compare, and commit-archive will not reflect current group membership, only the apply-path expression itself. And since the refresh bypasses commit entirely, it also skips commit-confirm rollback and leaves no entry in the commit log, so firewall behavior can change with nothing recorded.

This also breaks the "everything is in the CLI" philosophy that VyOS is built on. Group membership now effectively lives in multiple locations at once, referenced only by a base config path that can end in an asterisk to match everything below it, rather than being visible and explicit in the group's own configuration.

Just a passing thought, not something to take too seriously: for this particular use case, a dedicated firewall match criteria like "bgp-peers" rather than a generic group mechanism might fit better. And from an automation standpoint, keeping groups in sync with something like BGP peers or DHCP leases is fairly easy to script externally, for example with Ansible.

ruben-herold:
@c-po Thanks for the detailed technical read, these are fair points and worth answering one by one.

**The race is real, and there's already an established fix for it in this codebase.** `vyos-domain-resolver` (which pushes FQDN/`remote-group` derived members into `vyos_filter`/`vyos_nat` sets on its own timer, fully outside commit) waits on `commit_in_progress()` from `vyos.utils.commit` before touching nftables — its docstring literally says it's "to be used in scripts that modify the config outside of the CLI framework". `update_apply_path_groups()` should do the same before acquiring its own lock. I'll add that in the next push, it directly closes the window you're describing rather than just narrowing it.

**On "a second mechanism for the same conceptual problem," commit-archive/compare, and no commit-log entry:** this is exactly the shape `geoip_update()`/`geoip_refresh()` already has, and it's been running in production for years. GeoIP country-code/ASN sets are:
- resolved from a source that isn't the config tree (a downloaded IP-range DB, refreshed on a schedule) and pushed straight into `nft` via a separate `GeoIPLock`, uncoordinated with commit beyond that lock
- never written back into the config, so `show configuration`, `compare`, and `commit-archive` only ever show the `country-code`/`asn` match criteria, never the resolved ranges
- not commit-confirm-covered when refreshed outside a commit, and don't get a commit-log entry either

That's the same trade-off apply-path makes, in the same file, for the same reason: the *value* someone actually configures (a country code, a config path) is small and stable; the *resolved set* is large, derived, and expected to drift out from under the commit that last touched the group. If that trade-off is acceptable for GeoIP it should be acceptable here, apply-path doesn't introduce a new category of risk, it's a second application of an existing, accepted one. I'm not against tightening it further than GeoIP does (see the `commit_in_progress()` fix above), just pushing back on "this is unprecedented."

**On dynamic-groups**: I answered this for @sarthurdev in the earlier thread, so only the short version here since it's the same question - dynamic-groups derives membership from *runtime/dataplane state* (conntrack, kernel-side), atomically, in-kernel, no second source of truth possible even in principle. apply-path derives membership from *other configuration* (BGP neighbors, static routes, VRF membership), evaluated against the candidate config at commit time. They answer different questions - "what does the dataplane currently see" vs. "what does the rest of my config already say" - and neither can be built out of the other without effectively becoming the other.

Your `bgp-peers`-style dedicated match criteria idea is worth a separate ticket regardless of what happens here; it doesn't cover the general case (arbitrary config paths - static routes, VRF fan-out, DHCP-derived, anything future config trees add) the way apply-path/config_path_resolver does as a shared, backend-agnostic primitive. Happy to file that as a follow-up task if you think it's worth pursuing alongside this.

I'll push the `commit_in_progress()` change shortly.

ruben-herold:
@sever-sever All three points from your 08-12 review are addressed (squashed to one commit, unrelated changes dropped, "How to test" section added) — ready for re-review whenever you have a moment.

@c-po Following up on the commit_in_progress() fix — it landed the same day in 2cd2e99f, closing the race window you flagged. Let me know if that resolves your concern or if you'd like to keep discussing the broader apply-path-vs-dynamic-groups design question.

sever-sever (changes requested):
Simple config does not work in my tests
```

set protocols bgp address-family ipv4-unicast network 100.64.14.0/24
set protocols bgp neighbor 192.0.2.10 remote-as '65001'
set protocols bgp neighbor 192.0.2.11 remote-as '65001'
set protocols bgp neighbor 192.0.2.16 address-family ipv4-unicast
set protocols bgp neighbor 192.0.2.16 remote-as '65001'
set protocols bgp system-as '65001'
commit

set firewall group network-group NEIGHBORS apply-path 'protocols bgp neighbor *'
commit

```

Commit
```
vyos@r14# set firewall group network-group NEIGHBORS apply-path 'protocols bgp neighbor *'
[edit]
vyos@r14# commit
[ firewall ]
apply-path on network_group "NEIGHBORS" resolved invalid network
"remote-as"
[[firewall]] failed
Commit failed
[edit]
vyos@r14# 
[edit]

```

Please rebase the changes with upstream rolling
```
git remote add upstream https://github.com/vyos/vyos-1x.git
git fetch upstream rolling
git rebase upstream/rolling
```
Without it, CI cannot start smoketests

jestabro:
I agree that this is a desired and useful feature, and that several of the intuitions here are in the right direction.
However:
(1) although the idea of a cross-reference in the XML seems natural, the breaking of visibility of 'everything is in the CLI', as @c-po mentions, will require preprocessing at other points and by other tools (config-dict, for an obvious example). This could be considered, but would need a proper design; as is, the cross-reference will entail a technical debt.
(2) I do not think op-mode is the right approach: it is awkward in that it invites its own problems of coordination (@c-po 's concern above), but also breaks a second contract: that all changes to the config are visible to the user and under their control.

I would suggest considering an approach that is a generalization of the 'load config' model: a command, within config-mode, that calls the explicit `set` commands for the defined cross-referenced paths: as with `load`, manual commit or discard is required after review (or allowed with http-api 'no-prompt', once integrated).

The script would run both as a post-commit script to alert for the need for update of defined sync paths, as well as the mechanism for updating (`set`) those paths.

A simple POC of the idea is at the link below; wildcards in the source path are implicitly handled by the underlying utils, (as used in config-sync, and elsewhere).

POC link: https://github.com/vyos/vyos-1x/compare/rolling...jestabro:apply-path-util

The example included is just the test case from this PR smoketest; setting paths interactively has been tested but not included in the POC. The level of integration into the CLI vs a standalone tool is to be decided, but it is to be a mixture, along the lines of load/config-mgmt.

ruben-herold:
@jestabro @c-po Thanks for the detailed feedback — I think you're right that resolving the cross-reference implicitly during config-dict processing is the weak point here, not the idea of declaring the relationship itself.

Looking at your POC (https://github.com/vyos/vyos-1x/compare/rolling...jestabro:apply-path-util), I'd like to propose a middle ground rather than picking one approach wholesale:

- Keep the XML declaration of source/target paths as *metadata only* — it gives us a discoverable, schema-validated way to define these relationships instead of hand-maintaining a Python list of `PathData` entries, and lets tooling (docs, smoketests) introspect which paths are linked.
- Move the *application* of that metadata out of config-dict preprocessing entirely, and into something modeled on your `apply-path check`/`update` split: a post-commit hook flags drift, and reconciling it happens via an explicit `set`/`delete` step the user reviews and commits themselves — same contract as `load`.

That keeps the invariant that the config tree only ever changes through an explicit, visible action, which addresses the "everything is in the CLI" concern, and avoids needing every consumer (config-dict, GUI, HTTP API) to know how to resolve the cross-reference at read time — they just see ordinary CLI-set values. It also sidesteps the commit-timing/race class of issue entirely, since there's no synchronous resolution happening mid-commit.

The tradeoff is that it's no longer fully automatic — a user (or a script) has to act on the post-commit notice. For the network-group-from-static-route use case that motivated this PR, I think that's an acceptable cost given the alternative is baking implicit resolution into every config consumer going forward.

Happy to rework #5374 along these lines if that direction sounds reasonable to both of you — wanted to check before investing in a rewrite. @jestabro, is the source/target metadata still best expressed in XML in your view, or would you rather see it live entirely in the apply-path tool's own config?

jestabro:
@ruben-herold it is tempting to want the metadata to live in either (1) XML (2) some as-yet-to-be-defined annotation to the config file itself: neither work, that is why, for now, it should live in a pedestrian data-dir json definition file.
(1) doesn't work because it cannot then be applied selectively to tag-node values (say, in your example `smoetest_derived` is synced but `smoketest_notderived` is not)
(2) doesn't work because the idea of config annotations, even if/when it is properly introduced, cannot be applied to not-yet-existing paths --- note your workaround in the example to set an entry in `[..., smoktest_derived, network]`.
The POC mentioned will create the path as needed on sync, once metadata is defined; refining the ontology of the metadata will be something to be considered as the ideas are refined --- the goal is to introduce a mechanism that does not entail technical debt before a mature design evolves.

jestabro:
Finally, I will bring up the collection of ideas for discussion with developers this week, so I suggest working on the ideas, but not necessarily the PR, until we get some feedback.

ruben-herold:
@jestabro Makes sense, thanks for spelling out why both XML and config annotations fall short here — agreed, a data-dir JSON definition file is the right home for the metadata, especially given the not-yet-existing-path problem you point out.

Since you're bringing this to the dev discussion this week, here's a concrete shape for the mechanism as I understand it now, so there's something specific on the table rather than starting from the abstract idea again:

- **Metadata**: a JSON file under `data/config-sync/` (or similar), each entry holding a `source` path (wildcard segments allowed, same semantics as this PR's resolver) and a `target` path where the derived `set` commands land. Same two primitives already in `config_path_resolver.py` (`list_nodes()`/`return_values()`) do the resolution — no new traversal logic needed, just a new consumer of it.
- **`check`**: config-mode command, compares current `source` values against what's currently `set` at `target`, reports drift (added/removed/changed) without touching the tree. This is also what the post-commit hook calls, to print a "target X is out of date, run `update`" notice — informational only, no config change.
- **`update`**: stages the corresponding explicit `set`/`delete` commands for the drifted paths into the current edit session. Same contract as `load` — nothing is persisted until the user reviews (`compare`) and runs `commit` or `discard` themselves.
- **Scope**: this keeps the invariant that the config tree only changes via an explicit, visible action. It also means the mechanism isn't firewall-specific — T9160's broader scope (`interface-group`, `prefix-list`/`prefix-list6`, `access-list`/`access-list6`) is just more JSON entries pointing at different targets, not more code.

One open question for you: should `check`/`update` live as top-level config-mode commands (not under `firewall`), given the JSON definitions aren't tied to any one consumer? That seems necessary for the mechanism to actually cover T9160's full scope rather than just the firewall-group case this PR started from.

Happy to have this referenced or linked wherever the dev discussion happens.

jestabro:
@ruben-herold , the human: the AI-assisted responses are somewhat useful for a summary, however, they miss some key subtleties: the line "Same two primitives already in config_path_resolver.py (list_nodes()/return_values()) do the resolution" is neither true nor useful; those are replaced by the parsimonious approach using util `subtree_values_of_path` as wrapped in the calls defined in the POC mentioned

ruben-herold:
Fair enough. 😄 English isn't my first language, so I sometimes let AI help me with the longer write-ups, and in this case I should definitely have compared that statement with your PoC before hitting "Post".

You're right: the PoC gets both source and target straight from subtree_values_of_path() via the libvyosconfig bindings in configtree.py, rather than through Config.list_nodes()/return_values(). Since that's one layer below what config_path_resolver.py currently uses, it also explains why target resolution and validation work exactly like the source path, wildcards included.

Other than that, I think my understanding is still correct: JSON stored in the data directory, clear separation between validation and update logic, and a load-style commit approach. Happy to be corrected if I've missed something else before this goes into the discussion.

jestabro (changes requested):
Adding a requested change to block until we've discussed alternative approaches and outlined a path forward for this feature. cf. my comments above.

jestabro:
@ruben-herold we will close this PR, and pursue the alternative approach inspired by your investigation. Thanks for raising the issue: I agree that this will be a useful feature, and raising the issue with an example implementation revealed the subtleties involved.

PR #5376, opened 2026-08-04
vpp: T9146: Enable DPDK no-multi-seg when a full frame fits the buffer

natali-rs1985:
Disabling DPDK multi-segment buffers (no-multi-seg) lets VPP receive each
frame into a single buffer, increasing forwarding throughput. The frame
must then fit the buffer 'data-size', so 'no-multi-seg' is computed: it is
enabled only while the largest MTU across all VPP-bound interfaces, plus
its L2 framing overhead (26 bytes), fits 'data-size'. Otherwise VPP keeps
multi-seg and Jumbo frames still work without a larger buffer.

Changing the MTU of a VPP-bound interface can flip the flag, so
interfaces_ethernet reconfigures VPP when it changes, via an initial
dependency (set_dependents_initial) to avoid a cycle with the existing
vpp -> ethernet dependency.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9146

```
# eth1 with the default MTU (1500)
[edit]
vyos@vyos# commit
[ vpp ]

[edit]
# no-multi-seg IS enabled
vyos@vyos# grep no-multi-seg /run/vpp/vpp.conf
    no-multi-seg
[edit]
# Raise eth1 MTU beyond what the buffer can hold
vyos@vyos# set interfaces ethernet eth1 mtu 2500
[edit]
vyos@vyos# commit
[ interfaces ethernet eth1 ]

[edit]
# no-multi-seg is now GONE
vyos@vyos# grep no-multi-seg /run/vpp/vpp.conf
[edit]
vyos@vyos# 
```

## Checklist:

### DIFF
--- a/data/templates/vpp/startup.conf.j2
+++ b/data/templates/vpp/startup.conf.j2
@@ -178,6 +178,9 @@ dpdk {
 {%     endif %}
 {% endfor %}
     uio-bind-force
+{% if no_multi_seg is vyos_defined %}
+    no-multi-seg
+{% endif %}
 }
 
 {% if ipsec_acceleration is vyos_defined %}
--- a/src/conf_mode/vpp.py
+++ b/src/conf_mode/vpp.py
@@ -76,9 +76,16 @@ service_name = 'vpp'
 service_conf = Path(f'/run/vpp/{service_name}.conf')
 systemd_override = '/run/systemd/system/vpp.service.d/10-override.conf'
 
-vpp_log = getLogger(
-    service_name, format='%(filename)s[%(process)d]: %(message)s', address='/dev/log'
-)
+try:
+    vpp_log = getLogger(
+        service_name,
+        format='%(filename)s[%(process)d]: %(message)s',
+        address='/dev/log',
+    )
+except ValueError:
+    # vpp.py is re-executed when run as a config dependency (T9146); the named
+    # logger is created only once, so reuse the existing instance on re-import.
+    vpp_log = getLogger(service_name)
 
 dependency_interface_type_map = {
     'vpp_interfaces_bonding': 'bonding',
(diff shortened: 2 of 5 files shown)


data/templates/vpp/startup.conf.j2
```diff
@@ -178,6 +178,7 @@ dpdk {
 {%     endif %}
 {% endfor %}
     uio-bind-force
+    no-multi-seg
```
sever-sever, line 182: From the netgate [documentation](https://docs.netgate.com/tnsr/en/latest/advanced/dataplane-dpdk.html)
```
dataplane dpdk no-multi-seg
Disables multi-segment buffers for network devices. Can improve performance, but disables jumbo MTU capability.

Warning

This option is not currently compatible with Intel X552 10G network interfaces. When enabled on incompatible hardware this option can lead to instability such as dataplane crashes while under load.
```
I'd prefer to have this option configurable and keep stability and jumbo-frames

dmbaturin (changes requested):
I agree with @sever-sever: there are trade-offs here so it's better to make it configurable.

zdc (changes requested):
We should not hardcode `no-multi-seg` in the config. A detailed explanation of why, with a suggested safe design, is here: https://vyos.dev/T9146#275798

data/templates/vpp/startup.conf.j2
natali-rs1985: Made the changes proposed by @zdc

src/conf_mode/vpp.py
```diff
@@ -209,6 +216,24 @@ def _normalize_buffers(config: dict):
...
+    must fit within the buffer 'data-size'. When it does, set the flag so the
+    template emits 'no-multi-seg'; otherwise leave it unset so VPP keeps
+    multi-seg and can still carry Jumbo frames without a bigger buffer (T9146).
+    """
+    max_mtu = max(
+        (int(c['mtu']) for c in config['settings'].get('interface', {}).values()),
+        default=1500,
+    )
+    data_size = config['settings']['resource_allocation']['buffers']['data_size']
+    if memory.no_multi_seg_fits(max_mtu, data_size):
```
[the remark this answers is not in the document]
natali-rs1985: `no-multi-seg` is automatic with no CLI knob, and multi-seg is a safe fallback that always works — nothing is dropped. We compute the flag transparently instead of rejecting: an "undersized" data-size (e.g. default 2048 with a 9000 MTU) is a valid multi-seg config, so a `ConfigError` would break working setups.

sever-sever (approved):
Add `no-multi-seg` option configurable
All smoketests passed

PR #5388, opened 2026-08-09
pki: T9135: derive ACME certificate chains from disk

c-po:
## Change summary

Certificates issued via ACME (Let's Encrypt) need certbot's intermediate CA to build a full chain for consumers like HAProxy, HTTPS, IPsec, stunnel, and EAPOL. This intermediate was previously imported into the running configuration as a synthetic `pki ca AUTOCHAIN_<cert>` object purely so those consumers could find it - leaking certbot's internal state into the CLI as a real, user-visible, deletable object that never needed to exist there.

This PR removes that object entirely. The intermediate is read live from disk, in memory only, wherever a full chain is resolved or displayed (`show pki ca`) - never as a settable or deletable configuration object. An already-configured CA that completes the chain on its own takes precedence, and nothing synthetic is added or shown in that case.

Adding, changing, or removing any CA now reloads only the services whose resolved chain is actually affected by that specific change, instead of either missing the update or reloading everything unconditionally - and without side effects on certificates whose own content didn't change (an ACME certificate is never re-requested just because an unrelated CA changed).

## Types of changes

## Related Task(s)

* https://vyos.dev/T9135

## Related PR(s)

* https://github.com/vyos/vyos-1x/pull/5359

... waiting on builds ...

## Checklist:

### DIFF
--- a/op-mode-definitions/pki.xml.in
+++ b/op-mode-definitions/pki.xml.in
@@ -521,6 +521,12 @@
                   </leafNode>
                 </children>
               </node>
+              <leafNode name="text">
+                <properties>
+                  <help>Show x509 CA certificate in human-readable text format</help>
+                </properties>
+                <command>${vyos_op_scripts_dir}/pki.py show_certificate_authority --name "$4" --text</command>
+              </leafNode>
             </children>
           </tagNode>
           <tagNode name="certificate">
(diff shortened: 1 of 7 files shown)


jestabro (approved):
This is a nice solution, by introducing the mental model 'derived data'. The design and organization has been reviewed; all reasonable coderabbit comments have been addressed. As this is scheduled for testing, I will trust that and @c-po testing.

PR #5393, opened 2026-08-11
xml: T9179: reject VRF names in interface-name constraint

c-po:
## Change summary

Tab completion for source-interface and other interface leafNodes already excludes VRF names, but the shared interface-name constraint accepted them anyway: an existing VRF is a real net device, so it passed the file-path existence check even though it failed the interface-name regex.

Replace the file-path validator with a new interface-exists validator that requires the value to both exist under `/sys/class/net` and not be a VRF. The regex-match fallback is unchanged, so dynamic interfaces (e.g. pppoe) referenced before they exist still validate correctly.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9179

All smoketests passed

## Checklist:

### DIFF
--- a/src/validators/interface-exists
+++ b/src/validators/interface-exists
@@ -1,0 +1,40 @@
+#!/bin/sh
+#
+# Copyright (C) VyOS Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or later as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Passes only if $1 is a net device that exists on the system and is not a
+# VRF. Used as a fallback alongside a naming-pattern regex, so a currently
+# existing device is accepted as a physical/logical interface only if it is
+# not a VRF (VRFs are real net devices but must never be accepted where an
+# interface is expected).
+
+case "$1" in
+    ""|.|..|*/*)
+        echo "Error: $1 does not exist"
+        exit 1
+        ;;
+esac
+
+if [ ! -d "/sys/class/net/$1" ]; then
+    echo "Error: $1 does not exist"
+    exit 1
+fi
+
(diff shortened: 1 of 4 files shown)


src/validators/interface-exists
```diff
@@ -0,0 +1,33 @@
...
+# existing device is accepted as a physical/logical interface only if it is
+# not a VRF (VRFs are real net devices but must never be accepted where an
+# interface is expected).
+
+if [ ! -d "/sys/class/net/$1" ]; then
+    echo "Error: $1 does not exist"
+    exit 1
+fi
+
+if ip -o link show type vrf 2>/dev/null | awk -F': ' '{print $2}' | cut -d'@' -f1 | grep -qxF -- "$1"; then
```
hedrok: Maybe just use
```
ip vrf show "$1"
```
?..

hedrok (changes requested):

c-po:
Rebased on latest rolling/HEAD to get the smoketests running

hedrok (changes requested):
I think test for flow_accounting needs update now.
In `src/conf_mode/system_flow-accounting.py` we can delete lines as they are obsolete now:
```python
            # A VRF is not an interface. Cisco and Juniper both source a flow
            # exporter from a routed interface and never from a VRF, and VRF
            # export is already selected with "system flow-accounting vrf", so
            # reject a source-interface that names a VRF device.
            source_interface_config = get_interface_config(data['source_interface'])
            if dict_search('linkinfo.info_kind', source_interface_config) == 'vrf':
                raise ConfigError(
                    f'Configured "netflow server {server} source-interface '
                    f'{data["source_interface"]}" is a VRF, not an interface!'
                )
```
And to fix the test in `smoketest/scripts/cli/test_system_flow-accounting.py` replace
```python
        # A source-interface that names a VRF is not an interface and must be
        # rejected
        self.cli_set(
            base_path
            + ['netflow', 'server', '198.51.100.9', 'source-interface', vrf_name]
        )
        with self.assertRaises(ConfigSessionError):
            self.cli_commit()
```
With
```python
        # A source-interface that names a VRF is not an interface and must be
        # rejected
        with self.assertRaises(ConfigSessionError):
            self.cli_set(
                base_path
                + ['netflow', 'server', '198.51.100.9', 'source-interface', vrf_name]
            )
```
(haven't tested, but should work I hope).

c-po:
> I think test for flow_accounting needs update now. In `src/conf_mode/system_flow-accounting.py` we can delete lines as they are obsolete now:
> 

Requested changes added as discrete commit.

hedrok (approved):
Approve: good change for early detection of wrong usage of VRF where interface is expected.
All smoke tests passed.

PR #5396, opened 2026-08-12
T8598: CLI support to configure a secret which will be used along with calling-station-id, to generate Peer Interface Identifier (IID) - (BP # 2188)

ritika0313:
## Change summary

(Backport from private circinus # 2188 to public rolling)
- CLI support to configure a secret which will be used along with calling-station-id, to generate Peer Interface Identifier (IID).
- Also includes the corresponding migration support for calling-sid peer-IID secret. As part of the fix, secret needs to be mandatorily configured when calling-sid is configured.

- Secret must be 16 to 128 printable non-whitespace ASCII characters.

- The secret-key can be any user string or can be generated using the existing op-mode command (1 byte = 2 characters. so for generating 64 characters use byte-size = 32):

       `run generate psk random size <byte-size>`
- Below actions will drop existing sessions and trigger the re-negotiation/re-creation of session with the new config.

1. Deleting peer IID calling-sid config
2. Changing the calling-sid secret

## Types of changes

## Related Task(s)

https://vyos.dev/T8598

## Checklist:

### DIFF
--- a/data/templates/accel-ppp/ppp-options.j2
+++ b/data/templates/accel-ppp/ppp-options.j2
@@ -30,6 +30,9 @@ ipv6-peer-intf-id=ipv4
 ipv6-peer-intf-id={{ ppp_options.ipv6_peer_interface_id }}
 {%         endif %}
 {%     endif %}
+{%     if ppp_options.ipv6_peer_interface_id_secret is vyos_defined %}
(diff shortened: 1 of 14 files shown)


sever-sever (approved):
Use secrets with the `calling-station-id` option

PR #5398, opened 2026-08-13
firewall: T9205: fix ruff findings in conf_mode/op_mode firewall.py

ruben-herold:
## Summary
- Split out of #5374 (T9160) at sever-sever's request: that PR bundled unrelated cleanup with the apply-path feature because it happened to touch the same files.
- Fixes bare `except:`, extraneous f-string prefixes on `ConfigError()`/text calls without placeholders in `src/conf_mode/firewall.py`, the same `except:` pattern in `src/op_mode/firewall.py`, and removes 4 unused-but-harmless locals in `smoketest/scripts/cli/test_firewall.py` (this is the ruff cleanup itself; #5374 separately re-added the same locals to undo the unrelated cleanup that had snuck into that branch).
- Narrowed the `except Exception:` in `src/op_mode/firewall.py` further to the specific exceptions each call site can actually raise (`OSError` from `cmdl()`, `json.JSONDecodeError` from `json.loads()`), per review feedback.
- Minor, intentional behavior change: the original bare `except:` blocks silently swallowed everything, including `KeyboardInterrupt`/`SystemExit`/unrelated bugs. They now only catch the documented failure modes of the calls inside the `try`, so unexpected errors propagate instead of being hidden.

Task: https://vyos.dev/T9205

## Test plan
- [x] `ruff check src/conf_mode/firewall.py src/op_mode/firewall.py` passes clean
- [x] `python3 -m py_compile` on all three touched files


ruben-herold:
Thanks for the review, addressed:

- Narrowed `except Exception:` to the specific exceptions each call actually raises: `OSError` from `cmdl()` and `json.JSONDecodeError` from `json.loads()`. Also applied the same narrowing to a fourth, identical call site (`get_nftables_details()`, ~line 77) that wasn't flagged but has the same pattern. Pushed as a new commit.
- You're right that this isn't strictly "no behavior change" — bare `except:`/broad `except Exception:` swallowed `KeyboardInterrupt`/`SystemExit`/unrelated bugs, and now only the documented failure modes are caught. Updated the PR description to call that out explicitly.
- Also fixed the PR description's wording on the smoketest locals: this PR *removes* the 4 unused `interface`/`name2` locals (that's the actual ruff cleanup); #5374 is the one that re-adds them to undo the unrelated cleanup that had leaked into that branch. The old description had it backwards.

ruben-herold:
One more refinement while double-checking: `cmdl()` decodes subprocess output as utf-8 and can raise `UnicodeDecodeError` (not an `OSError` subclass) in addition to `OSError` on command failure. Extremely unlikely in practice since `nft` output is always well-formed text/JSON, but for completeness all 4 sites now also catch `UnicodeDecodeError`. Pushed as edc1a80da.

ruben-herold:
@c-po Gentle ping for review — CI is green, no outstanding feedback. Same category as the T9185/T9210 lint cleanups.

ruben-herold:
@c-po Following up on our Slack chat about the darker/ruff formatting — pushed 0d73867 to address it.

Dug into why 3 of the `raise ConfigError(...)` calls got wrapped onto multiple lines: those exact lines were already over `ruff.toml`'s 88-char limit before this PR (90/91/97 chars). Removing the extraneous `f` prefix touched them, so darker had to reformat them to fit — hence the wrap. A 4th spot (the `text = f'...'` domain-resolver line, 140 chars) didn't get wrapped, because splitting a single string literal across lines wouldn't have helped it fit either way, so Black left it alone.

Per your point about backport friction, I reverted just those 3 f-prefix fixes and left the lines exactly as they were (still with the harmless `f'...'` prefix). They're untouched now, so they're out of the diff and won't be flagged by the diff-scoped ruff/darker CI checks either. Everything else in the PR (the exception narrowing that's the actual point of T9205, the 4th f-prefix fix, and the smoketest cleanup) is unchanged.

Separately — while looking into this we also noticed the documented coding guideline (docs.vyos.io, "wrap lines at 80 characters") doesn't match the 88 actually enforced via `ruff.toml`/Black. Not proposing to fix that here, just flagging it in case it's worth reconciling at some point.

ruben-herold:
Rebased onto current rolling (`5b0cce51f`), head is now `a49fe7e28`. The net diff is unchanged — still the same 5 insertions / 5 deletions across `src/conf_mode/firewall.py` and `src/op_mode/firewall.py`, byte-identical to the pre-rebase diff, even though rolling touched `firewall.py` in the meantime (#5372, T9242, T8247).

Both red checks were stale-base artefacts, not problems with this diff:

- **`build_iso`** failed with `dpkg-checkbuilddeps: error: Unmet build dependencies: python3-vici (>= 5.7.2)`. That Build-Depends entry was removed from rolling in `c2d2f51c3` ("Debian: T9216: drop build time dependency on python3-vici", 2026-08-16); the branch still carried it. Gone after the rebase.
- **`typos`** failed on `` `localed` should be `located` ``, fixed in rolling by @ordex in `406305c1e` (T9323, #5480). Verified locally with typos v1.47.2 and the central `_typos.toml`: exit 0 on this branch.

@sever-sever @c-po — could the gated workflows be approved so this gets real CI signal?

While here, a note on the one review point still open: @copilot-pull-request-reviewer flagged `except Exception` as too broad in `src/op_mode/firewall.py`. That predates `6c8039802` and `edc1a80da`, which narrowed it further; the current state catches concrete types only:

- `cmdl()` raises `OSError` on an unexpected exit code and on the VRF permission check, and `FileNotFoundError` (a subclass) if `nft` is missing — covered.
- `popen(decode='utf-8')` makes `UnicodeDecodeError` reachable — covered.
- `json.JSONDecodeError` is added only in the two blocks where `json.loads()` is actually inside the `try`; the other two wrap `cmdl()` alone.
- `cmdl()`'s `TypeError` guard for a non-list command stays uncaught on purpose — every call site here passes a literal list, so it can only fire as a programmer error, which is exactly what should no longer be swallowed.

PR #5400, opened 2026-08-13
utils: T9207: is_intf_addr_assigned() rejects hyphenated ranges

ruben-herold:
## Summary
- `is_intf_addr_assigned()` called `ip_interface(addr)` without checking whether `addr` is a hyphenated address range (e.g. `192.0.2.1-192.0.2.2`) — a valid `address-group`/`address-range` member elsewhere in the CLI. For such input, `ip_interface()` raises `ValueError` instead of the function returning `False`, and the exception is uncaught: it propagates out of `is_intf_addr_assigned()` and its caller `is_addr_assigned()` (which loops over every interface), crashing any code path that checks whether a range-form address is assigned to an interface.
- Moved the existing IPv6 zone-suffix strip ahead of the interface lookup (it was redundantly repeated inside the loop) and reject a hyphenated range there too, before `ip_interface()` ever sees it.
- Found while reviewing coderabbitai feedback on #5374 (T9160); this bug is pre-existing and unrelated to that PR's diff, so it's filed and fixed here separately rather than folded into it.

Task: https://vyos.dev/T9207

## Test plan
- [x] `ruff check python/vyos/utils/network.py src/tests/test_utils_network.py` passes clean (one pre-existing, unrelated `F811` duplicate-test-name finding in the same file, not touched by this change)
- [x] `python3 -m py_compile` on both touched files
- [x] Added `test_is_addr_assigned_hyphenated_range` covering both IPv4 and IPv6 hyphenated ranges
- [x] Manually verified against `lo`: hyphenated ranges now return `False` instead of raising; single-address checks (assigned and not-assigned) are unaffected

### DIFF
--- a/python/vyos/utils/network.py
+++ b/python/vyos/utils/network.py
@@ -492,9 +492,13 @@ def is_intf_addr_assigned(ifname: str, addr: str, netns: str=None) -> bool:
     from vyos.utils.process import rc_cmd
     from ipaddress import ip_interface
 
-    # Remove the interface name if present in the given address
+    # Remove the interface name if present in the given address. Use
+    # rsplit() so only the trailing zone suffix is dropped: split() would
+    # truncate a hyphenated range with a zone suffix on both endpoints
+    # (e.g. "::1%lo-::2%lo") down to just its first endpoint, silently
+    # discarding the "-" the check below relies on.
     if '%' in addr:
-        addr = addr.split('%')[0]
+        addr = addr.rsplit('%', 1)[0]
 
     # is_intf_addr_assigned() only checks a single address; a hyphenated
     # range (a valid address_group/address_range member elsewhere) is
--- a/python/vyos/utils/network.py
+++ b/python/vyos/utils/network.py
@@ -492,6 +492,17 @@ def is_intf_addr_assigned(ifname: str, addr: str, netns: str=None) -> bool:
     from vyos.utils.process import rc_cmd
     from ipaddress import ip_interface
 
+    # Remove the interface name if present in the given address
+    if '%' in addr:
+        addr = addr.split('%')[0]
+
+    # is_intf_addr_assigned() only checks a single address; a hyphenated
+    # range (a valid address_group/address_range member elsewhere) is
+    # never assigned to an interface as such and ip_interface() would
+    # raise ValueError on it instead of returning False
+    if '-' in addr:
+        return False
+
     netns_cmd = f'ip netns exec {netns}' if netns else ''
     rc, out = rc_cmd(f'{netns_cmd} ip --json address show dev {ifname}')
     if rc == 0:
@@ -501,9 +512,6 @@ def is_intf_addr_assigned(ifname: str, addr: str, netns: str=None) -> bool:
             family = address_info['family']
             address = address_info['address']
             prefixlen = address_info['prefixlen']
-            # Remove the interface name if present in the given address
-            if '%' in addr:
-                addr = addr.split('%')[0]
             interface = ip_interface(f"{address}/{prefixlen}")
             if ip_interface(addr) == interface or address == addr:
                 return True
--- a/src/tests/test_utils_network.py
+++ b/src/tests/test_utils_network.py
@@ -24,6 +24,13 @@ class TestVyOSUtilsNetwork(TestCase):
         self.assertTrue(vyos.utils.network.is_addr_assigned('::1'))
         self.assertFalse(vyos.utils.network.is_addr_assigned('127.251.255.123'))
 
+    def test_is_addr_assigned_hyphenated_range(self):
+        # A hyphenated address range can never be assigned to an interface
+        # as such; is_intf_addr_assigned() must return False for it
+        # instead of raising ValueError from ip_interface()
+        self.assertFalse(vyos.utils.network.is_addr_assigned('127.0.0.1-127.0.0.2'))
+        self.assertFalse(vyos.utils.network.is_addr_assigned('::1-::2'))
+
     def test_is_ipv6_link_local(self):
         self.assertFalse(vyos.utils.network.is_ipv6_link_local('169.254.0.1'))
         self.assertTrue(vyos.utils.network.is_ipv6_link_local('fe80::'))
--- a/src/tests/test_utils_network.py
+++ b/src/tests/test_utils_network.py
@@ -30,6 +30,9 @@ class TestVyOSUtilsNetwork(TestCase):
         # instead of raising ValueError from ip_interface()
         self.assertFalse(vyos.utils.network.is_addr_assigned('127.0.0.1-127.0.0.2'))
         self.assertFalse(vyos.utils.network.is_addr_assigned('::1-::2'))
+        # A zone suffix on each range endpoint must not truncate the
+        # string down to just the first endpoint before the "-" check
+        self.assertFalse(vyos.utils.network.is_addr_assigned('::1%lo-::2%lo'))
 
     def test_is_ipv6_link_local(self):
         self.assertFalse(vyos.utils.network.is_ipv6_link_local('169.254.0.1'))


ruben-herold:
Good catch, thanks. `addr.split('%')[0]` truncates at the first `%`, so `::1%lo-::2%lo` collapsed to just `::1` before the `-` check ever saw it — bypassing the guard and potentially matching that first endpoint instead of correctly rejecting the range. Switched to `rsplit('%', 1)` so only the trailing zone suffix is dropped, and added a regression test for the zone-suffixed range case. Pushed as 1cf8ac8e6.

ruben-herold:
@dmbaturin @c-po Gentle ping for review — CI is green, no outstanding feedback.

MPStudyly:
The [rejection of ranges was also done by me in my PR](https://github.com/vyos/vyos-1x/pull/5266/changes#diff-9477333b39f08652a34084e2e6dc7ad9a2b20b44185b4060d0f69e8b8377eff6R516) about fixing a regression with `is_addr_assigned`. However I didn't touch the behavior regarding ranges, so that would remain an issue without your PR. Could you have another look to confirm my assumption about this overlap?

Depending on which PR gets merged first, I'd like to plan the merge solution ahead of time :)

ruben-herold:
@MPStudyly Thanks for flagging this — yes, there's overlap. Your #5266 adds a generic `is_ipv4_address()`/`is_ipv6_address()` guard at the top of `is_intf_addr_assigned()`, which implicitly rejects ranges too, that would make the range-rejection change in this PR redundant.

What #5266 doesn't cover: the second part of this fix — the ordering of zone-suffix stripping vs. the range check. With `split('%')` instead of `rsplit('%', 1)`, a range with a zone suffix on both endpoints (e.g. `::1%lo-::2%lo`) gets truncated down to `::1` before the `-` check ever sees it.

Suggestion: merge your #5266 first (broader scope, regression fix), then I'll rebase #5400 on top and trim it down to just the zone-suffix ordering fix (dropping the now-redundant generic range guard). Sound good?

MPStudyly:
> What #5266 doesn't cover: the second part of this fix — the ordering of zone-suffix stripping vs. the range check. With `split('%')` instead of `rsplit('%', 1)`, a range with a zone suffix on both endpoints (e.g. `::1%lo-::2%lo`) gets truncated down to `::1` before the `-` check ever sees it.
> 
> Suggestion: merge your #5266 first (broader scope, regression fix), then I'll rebase #5400 on top and trim it down to just the zone-suffix ordering fix (dropping the now-redundant generic range guard). Sound good?

Sounds good :+1: however, more out of curiosity: When there is an interface identifier given as part of the address, wouldn't it make more sense to test that exact interface for the address instead of stripping it completely? Especially as the address itself will always be a link-local one (AFAIK identifiers are not allowed on regular IPv6 addresses). As link-local addresses require the identifier for uniqueness, the check would yield "address assigned" for the first interface it finds with the exact address, which isn't necessarily the interface passed via identifier by the caller.

EDIT: Sorry for the convoluted explanation, I'll try to do an example:

1. Caller requests info on `::1%locust`
2. Function strips `%locust`, checking for `::1` being assigned
3. Function yields `true` as address `::1` is assigned to interface `lo`, even though interface `locust` actually has no address assigned

ruben-herold:
Good catch — you're right, and it's a real gap, not just theoretical. `is_addr_assigned()` loops over every interface and calls `is_intf_addr_assigned(interface, ip_address)` for each one; the `%zone` part just gets stripped, it's never actually compared against the `interface` being checked. So `is_addr_assigned('::1%locust')` would happily return `True` off of `::1` sitting on `lo`, with `locust` never even entering the picture.

This predates both our PRs though, and it's a separate bug from the range-parsing one — I'd rather not fold it into either #5266 or #5400 and risk stalling the merge you're already planning. I'll file it as its own task once #5400 is rebased and trimmed down, same as I did with the hyphenated-range issue (T9207) when it came up out of scope in #5374. Feel free to beat me to it if you want to take a crack at the fix yourself.

MPStudyly:
> Good catch — you're right, and it's a real gap, not just theoretical. `is_addr_assigned()` loops over every interface and calls `is_intf_addr_assigned(interface, ip_address)` for each one; the `%zone` part just gets stripped, it's never actually compared against the `interface` being checked. So `is_addr_assigned('::1%locust')` would happily return `True` off of `::1` sitting on `lo`, with `locust` never even entering the picture.
> 
> This predates both our PRs though, and it's a separate bug from the range-parsing one — I'd rather not fold it into either #5266 or #5400 and risk stalling the merge you're already planning. I'll file it as its own task once #5400 is rebased and trimmed down, same as I did with the hyphenated-range issue ([T9207](https://vyos.dev/T9207)) when it came up out of scope in #5374. Feel free to beat me to it if you want to take a crack at the fix yourself.

Glad I didn't hallucinate this issue :zany_face: Sounds reasonable to me! I hope you are patient regarding the merge, as I'm still waiting on final feedback of my PR. And no worries, I'll let you do your fixes yourself ;)

c-po:
Is there any real bug, or is this a code refactoring adding range validation checks into `is_intf_addr_assigned()`?

ruben-herold:
@c-po Both, in a sense — there is a real bug, but this PR is the wrong vehicle
for it and I'd suggest closing it. Details, and two design questions at the end.

**Is there a real bug?** Yes, two wrong answers against current `rolling`:

```python
>>> is_intf_addr_assigned('lo', '192.0.2.1-192.0.2.2')
ValueError: '192.0.2.1-192.0.2.2' does not appear to be an IPv4 or IPv6 interface
>>> is_intf_addr_assigned('lo', '::1%lo-::2%lo')
True          # a range reported as "assigned"
```

But it is not reachable from the CLI today. I hit it in #5374 (firewall group
members, where ranges are valid), that PR is closed, and every current in-tree
caller passes an address the CLI validators have already constrained. I should
have stated that in the description instead of leaving it implicit.

**Why I'd close this PR:** #5266 (T8977) adds

```python
if not is_ipv4_address(addr) and not is_ipv6_address(addr):
    return False
```

at the top of the function, which covers both cases above. My claim earlier in
this thread that #5266 misses the zone-suffix ordering half was wrong — the
guard runs before any stripping, so `ip_interface()` rejects `::1%lo-::2%lo`
outright while a legitimate `fe80::1%eth0` still parses and passes through.
Nothing in this PR survives that. I'd resolve T9207 against #5266 and port the
regression tests (`test_is_addr_assigned_hyphenated_range`) over to it.

**What #5266 does not cover** — and this is why I'm asking rather than just
closing. Same function, current `rolling`:

```python
>>> is_addr_assigned('::1%eno1')      # ::1 is on lo; eno1 has no such address
True
>>> is_addr_assigned('::1%nosuchif')
True
```

The `%zone` suffix is stripped but never compared against the interface being
checked, so the function matches on whatever interface happens to carry the
address. @MPStudyly spotted this while reviewing this PR. Two related things in
the same function: `ifname` and `netns` are interpolated into a *shell* command
string (`rc_cmd(f'... dev {ifname}')`, and `popen()` enables the shell for
string commands), and the loop reassigns its own `addr` parameter.

Questions before I file anything:

1. Should `is_intf_addr_assigned()` be total — never raise, return `False` for
   anything that is not a single address — or do you prefer callers keep
   validating and the helper stay as it is? #5266 effectively picks the former;
   I'm happy to complete it in that direction, but only if that's the intent.

2. Fixing the zone mismatch changes behaviour: a config using
   `source-address fe80::1%eth0` that commits today because the address sits on
   a different interface would start failing `verify()`. Worth doing with its
   own task and a release note, or too invasive for the benefit?

3. If both are wanted, I'd keep them in separate PRs — one with no behaviour
   change for valid input (total function, list-form `rc_cmd()` so `ifname`
   cannot reach a shell, no parameter mutation), one for the zone semantics.
   Fine, or would you rather review it as a single change?

All of it is unit-testable in `src/tests/test_utils_network.py`, no smoketest
run required.

c-po:
`is_addr_assigned()` is a VyOS internal helper/validation API. I am unaware - but might be wrong - of any code path passing the zone identifier.

I also do think the current contract is ok-ish but could be improved.

```python3
def is_intf_addr_assigned(ifname: str, addr: str, netns: str=None) -> bool:
```

So `addr` must be a single IPv4 or IPv6 address - this can be tested using `vyos.template.is_ip()` and raise an Error if not - this rules out, we should not internally work in stripping % from addr.

Next is, `vyos.utils.process.rc_cmd()` is already netns aware - despite less usage of it in the current codebase, so instead of prefixing the command, we should use the proper function argument.

I would be in favor in closing this PR as the issue is real but not yet exposed on the CLI. The issues mentioned should be corrected in a subsequent PR.

ruben-herold:
@c-po Agreed on all three points — closing this in favour of #5477.

Filed as https://vyos.dev/T9322 and implemented there along the lines you described: reject anything that is not a single address instead of stripping `%` internally, and use `rc_cmd()`'s `netns=` argument rather than prefixing the command string. The dead `family` key and the reassignment of the `addr` parameter inside the loop went with it.

One correction to the recipe, since it caught me out: `vyos.template.is_ip()` on its own does **not** rule out a zone identifier. The `ipaddress` module has accepted scope IDs since Python 3.9:

```python
>>> from ipaddress import ip_interface
>>> ip_interface('fe80::1%eth0')
IPv6Interface('fe80::1%eth0/128')
```

So `is_ip('fe80::1%eth0')` is `True`. Had I only added the `is_ip()` guard and dropped the stripping, the silent wrong-interface match would simply have turned into a silent `False`, because `ip --json address show` never reports a zone to compare against. #5477 therefore checks for `%` explicitly alongside `is_ip()`.

On your point that you are unaware of any code path passing a zone identifier: I could not fully close that out. `protocols_bgp.py:382` guards with `is_ip(peer)`, which per the above still lets one through, and whether the CLI accepts it in the first place depends on `ipaddrcheck --is-any-single` — which I have no way to test locally. It is flagged as an open question in #5477; a loud failure there seemed preferable to a wrong answer, but say the word if you would rather have it return `False`.

Thanks for taking the time on the scope question — it was the right call, this reads much better as one focused change than as a patch on the range symptom.

ruben-herold:
Superseded by #5477 (T9322), which implements the agreed approach in full.

PR #5402, opened 2026-08-14
utils: T9185: revert the proposed fix ask_yes_no() busy-looping forever on non-tty stdin

sever-sever:
Reverts vyos/vyos-1x#5390
Reason: We have issues with image delete operations
```
Installed images: 2026.08.14-0025-rolling
2026.08.13-0024-rolling
Cannot prompt "Do you really want to delete the image 2026.08.13-0024-rolling?" ([Y/n] default=False): stdin is not a TTY. Pass an explicit non-interactive option (e.g. no_prompt/-y) instead of relying on ask_yes_no().
Error: Process completed with exit code 1.
```


dmbaturin (approved):
The original fix breaks a CI workflow so we can't keep it in place and will need to search for something different.

jestabro (approved):
Revert.

PR #5404, opened 2026-08-14
login: T9212: split radius source-address per address family

statio:
## Change summary

`set system login radius source-address <ip>` was a `<multi/>` node, so a second `set` **appended** rather than **replaced**. The CLI accepted any number of addresses and staged them all, then `commit` rejected the result with `Only one IPv4 source-address can be set!`. The adjacent `system login tacacs source-address` is a scalar node and behaves as users expect, which makes the inconsistency more surprising still.

The `<multi/>` was **not** an oversight. `b9feaf0d6` (T3192) added it specifically so one IPv4 *and* one IPv6 source address can coexist, and `pam_radius_auth.conf.j2` pairs each server with the source address of its own address family. Simply dropping `<multi/>` would regress dual-stack deployments, so this PR gives each address family its own scalar node instead:

```
set system login radius source-address  <ipv4>
set system login radius source-address6 <ipv6>
```

Both now replace on `set`. The per-family counters in `verify()` are gone, and the template consumes the two values directly rather than collapsing a list into a `namespace()`.

The `6` suffix follows existing house style (`route6`, `local-route6`, `prefix-list6`, `ospf6`, `nat66`). A new include was needed because `include/source-address-ipv6.xml.i` declares its leaf node as `name="source-address"` — its only consumers sit in paths that are already IPv6-only — so it would collide as a sibling of `source-address-ipv4.xml.i`.

### Behaviour change for automation

`source-address` is now IPv4-only. Saved configurations migrate automatically, but any Ansible/API automation that puts an IPv6 address in `source-address` must move it to `source-address6`. Flagged on T9212 as *config syntax change (migratable)*.

## Types of changes

- [x] Other (please describe): CLI syntax change with a migration script (`system` 33 → 34)

## Related Task(s)

* https://vyos.dev/T9212

Context, not fixed here:

* https://vyos.dev/T3192 — added the `<multi/>` for dual-stack; this PR preserves that capability
* https://vyos.dev/T3234 — `multi_to_list` failing on this exact node
* https://vyos.dev/T1582 — same defect class in OpenVPN `local-address`, closed as invalid

## Related PR(s)

None. A `vyos-documentation` update is required and not yet raised — see the checklist below.

> [!IMPORTANT]
> **I have not been able to run the smoketests** — no VyOS build/test host was available. `test_system_login.py` requires a live system (it starts a FreeRADIUS container and reads `/etc/pam_radius_auth.conf`). The suite needs a run before merge.

What *was* verified off-box:

**1. XML builds, validates, and adds exactly one node.** Ran `transclude-template` → `override-default` → `override-help` → `check-properties-collision`, then validated against `schema/interface_definition.rng`, comparing against an unmodified `origin/rolling` worktree:

```
=== BASELINE ===                        === THIS BRANCH ===
RelaxNG: PASS                           RelaxNG: PASS
check-properties-collision: exit 0      check-properties-collision: exit 0

system login radius children:
  security-mode, server, server,          security-mode, server, server,
  source-address, vrf                     source-address, source-address6, vrf

source-address  (radius) multi=True     source-address  (radius) multi=False
                                        source-address6 (radius) multi=False
source-address  (tacacs) multi=False    source-address  (tacacs) multi=False
```

(The duplicated `server` is the pre-existing T3234 element redundancy, present in baseline too.)

**2. The generated `pam_radius_auth.conf` is byte-identical.** Rendered the old and new templates with equivalent data across a matrix:

```
IDENTICAL   v4 servers, v4 source
IDENTICAL   v6 servers, v6 source
IDENTICAL   dual servers, dual source
IDENTICAL   dual servers, v4 source only
IDENTICAL   dual servers, v6 source only
IDENTICAL   dual servers, no source
IDENTICAL   no servers, dual source
IDENTICAL   with vrf
IDENTICAL   disabled server

FAILURES: 0
```

**3. Migration logic.** `vyos.configtree` needs libvyosconfig, so the migration was executed against a stub tree with equivalent semantics:

```
PASS  no radius source-address at all
PASS  single IPv4 (the common case)      <- performs zero operations
PASS  single IPv6 only
PASS  dual stack pair
PASS  dual stack, IPv6 listed first
PASS  hand-edited: two IPv4 (last wins)
PASS  hand-edited: two IPv6 (last wins)
PASS  hand-edited: two of each
PASS  empty node

FAILURES: 0
```

Every configuration that could previously pass `verify()` migrates without loss. More than one address of a family is only reachable via a hand-edited `config.boot`; there the last value wins, matching the old template's existing silent behaviour.

**4. Lint**, matching CI (changed lines only, against `origin/rolling`):

```
darker -r origin/rolling --check --diff .     -> exit 0
graylint -L "ruff check" --revision origin/rolling .  -> exit 0
```

## Checklist:

- [ ] I have run the components [**SMOKETESTS**](https://github.com/vyos/vyos-1x/tree/rolling/smoketest/scripts/cli) if applicable — **not run, see above**
- [ ] I have thoroughly reviewed, understood, and tested the code contained in the PR, including any code produced by GenAI tools — **author review pending**
- [ ] I have updated the documentation accordingly — **docs PR not yet raised**

### DIFF
--- a/data/templates/login/pam_radius_auth.conf.j2
+++ b/data/templates/login/pam_radius_auth.conf.j2
@@ -2,26 +2,15 @@
 # RADIUS configuration file
 
 {% if radius is vyos_defined %}
-{#     RADIUS IPv6 source address must be specified in [] notation #}
-{%     set source_address = namespace()  %}
-{%     if radius.source_address is vyos_defined %}
-{%         for address in radius.source_address %}
-{%             if address | is_ipv4 %}
-{%                 set source_address.ipv4 = address %}
-{%             elif address | is_ipv6 %}
-{%                 set source_address.ipv6 = address %}
-{%             endif %}
-{%         endfor %}
-{%     endif %}
 {%     if radius.server is vyos_defined %}
 # server[:port]        shared_secret             timeout    source_ip
 {# .items() returns a tuple of two elements: key and value. 1 relates to the 2nd element i.e. the value and .priority relates to the key from the internal dict #}
 {%         for server, options in radius.server.items() | sort(attribute='1.priority') if not 'disable' in options %}
 {#         RADIUS IPv6 servers must be specified in [] notation #}
 {%             if server | is_ipv4 %}
-{{ server }}:{{ options.port }} {{ "%-25s" | format(options.key) }} {{ "%-10s" | format(options.timeout) }} {{ source_address.ipv4 if source_address.ipv4 is vyos_defined }}
+{{ server }}:{{ options.port }} {{ "%-25s" | format(options.key) }} {{ "%-10s" | format(options.timeout) }} {{ radius.source_address if radius.source_address is vyos_defined }}
 {%             else %}
-{{ server | bracketize_ipv6 }}:{{ options.port }} {{ "%-25s" | format(options.key) }} {{ "%-10s" | format(options.timeout) }} {{ source_address.ipv6 if source_address.ipv6 is vyos_defined }}
+{{ server | bracketize_ipv6 }}:{{ options.port }} {{ "%-25s" | format(options.key) }} {{ "%-10s" | format(options.timeout) }} {{ radius.source_address6 if radius.source_address6 is vyos_defined }}
 {%             endif %}
 {%         endfor %}
 {%     endif %}
--- a/interface-definitions/include/radius-server-ipv4-ipv6.xml.i
+++ b/interface-definitions/include/radius-server-ipv4-ipv6.xml.i
@@ -25,7 +25,8 @@
         #include <include/radius-server-auth-port.xml.i>
       </children>
     </tagNode>
-    #include <include/source-address-ipv4-ipv6-multi.xml.i>
+    #include <include/source-address-ipv4.xml.i>
+    #include <include/source-address6.xml.i>
     <leafNode name="security-mode">
       <properties>
         <help>Security mode for RADIUS authentication</help>
--- a/interface-definitions/include/source-address6.xml.i
+++ b/interface-definitions/include/source-address6.xml.i
@@ -1,0 +1,17 @@
+<!-- include start from source-address6.xml.i -->
+<leafNode name="source-address6">
+  <properties>
+    <help>IPv6 address used to initiate connection</help>
+    <completionHelp>
+      <script>${vyos_completion_dir}/list_local_ips.sh --ipv6</script>
+    </completionHelp>
+    <valueHelp>
+      <format>ipv6</format>
+      <description>IPv6 source address</description>
+    </valueHelp>
+    <constraint>
+      <validator name="ipv6-address"/>
+    </constraint>
+  </properties>
+</leafNode>
+<!-- include end -->
--- a/interface-definitions/include/version/system-version.xml.i
+++ b/interface-definitions/include/version/system-version.xml.i
@@ -1,3 +1,3 @@
 <!-- include start from include/version/system-version.xml.i -->
-<syntaxVersion component='system' version='33'></syntaxVersion>
+<syntaxVersion component='system' version='34'></syntaxVersion>
 <!-- include end -->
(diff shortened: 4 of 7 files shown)


c-po (changes requested):
We do not differentiate between IPv4 and IPv6 source addresses. This must be handled by logic inside the login Python script.

statio:
Thanks — understood on the naming. I'll drop `source-address6`, keep a single `source-address` node, and move the address-family handling out of `pam_radius_auth.conf.j2` and into `system_login.py`, the way `system_syslog.py` and `service_ntp.py` do it. That also drops the new include, the migration and the version bump.

One thing I'd like to get right before I redo it: should the node keep `<multi/>`?

If it stays, the T3192 dual-stack case still works — one IPv4 and one IPv6 source, paired per server in `pam_radius_auth.conf` — but a second `set` still appends rather than replaces, so the original T9212 complaint isn't really fixed and the change becomes a template-to-Python refactor plus the existing one-per-family check. If it becomes scalar via `source-address-ipv4-ipv6.xml.i`, `set` replaces the way users expect, but only one source address is possible at a time and existing dual-stack configs would lose one of the two on migration.

Happy either way — I'd just rather not guess and rework it twice.

Either way I'll add a dual-stack smoketest; there isn't one today, both existing RADIUS tests run a single address family.

statio:
Picking this back up — I'd like to make the case for keeping two nodes, since I think it addresses the concern from the opposite direction.

RADIUS already differentiates by address family on rolling today; this PR doesn't introduce that. `pam_radius_auth.conf.j2` splits the configured addresses into a `namespace()` by family and pairs each server with the source address of its own family, and `system_login.py` enforces one per family ("Only one IPv4 source-address can be set!" / "...IPv6..."). The `<multi/>` isn't a list — it's a two-slot store, one per family, with the limit policed at commit time rather than expressed in the grammar.

That split looks structural rather than incidental: `server` accepts both families (`ip-address`, not `ipv4-address`), and a socket bound to an IPv4 address can't originate to an IPv6 server. A box with one IPv4 and one IPv6 RADIUS server therefore needs both source addresses at once. That's the part I can't see how to move into Python — it can validate a family match, but it can't store a second address in a one-value node, so a single scalar node means dropping the dual-stack case T3192 added.

Two scalar nodes move that logic *out* of Python rather than into it:

    set system login radius source-address  <ipv4>   # ipv4-address validator
    set system login radius source-address6 <ipv6>   # ipv6-address validator

`set` replaces on both, the per-family counters in `verify()` are deleted, and a wrong-family value is refused at `set` time by the node constraint instead of at `commit` — completion only offers addresses of the right family. I raised two things in T9212 — that `set` appended, and that the CLI staged something only `commit` would reject; this closes both.

I did look at moving `source-address` into the `server` tagNode, matching `system syslog remote <h>` and `flow-accounting netflow server <x>`. It's a better fit for `pam_radius_auth.conf`, where `source_ip` is a per-server column, but it only fixes the first of those: `server 2001:db8::1 source-address 10.0.0.1` would still be CLI-acceptable and rejected at commit. It also adds per-server sourcing, which isn't something I was after when I raised T9212.

Happy to switch to either that or a single scalar node if you'd rather — just wanted to lay out why I landed here before reworking.

statio:
@c-po just a friendly follow-up here to see if I can get a judgement call on the path forward here?

c-po:
@statio I still do not understand why splitting the source-address into a per-afi node. Only to have the validation on the CLI and not the Python code? This is something which could be changed in the future.

I am in general a friend if "less is more" and given the fact that a lot of other places use `source-address` for both AFIs makes this feel counter-intuitive.

Can you re-alaborate in two sentences why this split is needed and which bug it will solve.

statio:
The bug is that <multi/> lets the CLI accept a second same-family source-address which only verify() then rejects, so the working tree and the commit result disagree. A single leaf cannot be both multi (one v4 and one v6 at once) and scalar (replace on set); a per-AFI split is how the grammar states that invariant.

If source-address6 is the problem, would you support:

`set system login radius source-address ipv4 <addr>`
`set system login radius source-address ipv6 <addr>`

That keeps the source-address name, still allows dual-stack, and still replaces per family on set.

PR #5405, opened 2026-08-14
T3871: never guess a pending node's hardware when ambiguous

c-po:
## Change summary

Reported against the previous PCIe/MAC-sorted replacement fill: on a real, already-provisioned box (`hw-id` from historical probe-order rescan, unrelated to any PCIe/MAC sort), a configured node's address ended up silently applied to a different physical NIC. Deleting one interface's `hw-id` (keeping its settings) while a different, unrelated interface's config was fully removed in the same boot was enough - the two freed candidates' MAC order didn't match their old name order, so the deterministic sort swapped them. Once written back by the rescan helper, the wrong binding became permanent and self-reinforcing on every later boot.

There's no way to verify, from MAC and PCIe position alone, which of several unconfigured candidates is genuinely a given node's own hardware once its hw-id is gone. Restore strict matching: a node only recovers its hw-id automatically when it's the sole pending node of its type this boot and exactly one candidate exists - any other count leaves it pending and reported rather than guessed. A candidate that isn't matched this way is not otherwise held back - it still gets an ordinary, settings-free bootstrap name instead of being lost, just never inherits another node's configuration.

This necessarily changes what a combined "delete one interface fully, clear a different one's hw-id" reboot can auto-resolve; the accompanying test-harness change acknowledges that trade-off explicitly.

## Types of changes

## Related Task(s)

* https://vyos.dev/T3871

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1268

## Checklist:

### DIFF
--- a/smoketest/configs/assert/vpp
+++ b/smoketest/configs/assert/vpp
@@ -39,6 +39,7 @@ set interfaces vpp loopback vpplo12 description 'Loop12'
 set interfaces vpp vxlan vppvxlan10 remote '192.0.2.2'
 set interfaces vpp vxlan vppvxlan10 source-address '192.0.2.1'
 set interfaces vpp vxlan vppvxlan10 vni '10'
+set vpp settings allow-unsupported-nics
 set vpp settings interface eth1
 set vpp settings interface eth2
 set vpp settings interface eth3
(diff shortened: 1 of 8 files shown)


dmbaturin (approved):
I agree that the problem is real, the PR looks like a sensible solution to me.

sever-sever (approved):
All smoketests passed

PR #5416, opened 2026-08-18
T8482: reject commits referencing an incomplete policy object

c-po:
## Change summary

A `verify()` failure in one conf_mode script (e.g. `policy prefix-list`) did not stop `vyos-configd`/`vyos-commitd` from still pushing the shared FRR configuration to `vtysh` once the last script in the same commit batch happened to succeed on its own, since the render step only checked that last script's own result. Track whether any earlier script in the batch failed and gate the shared FRR render on it in both commit daemons.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8482

## Checklist:

### DIFF
--- a/smoketest/scripts/cli/test_protocols_bgp.py
+++ b/smoketest/scripts/cli/test_protocols_bgp.py
@@ -2045,5 +2045,41 @@ class TestProtocolsBGP(VyOSUnitTestSHIM.TestCase):
         frrconfig = self.getFRRconfig(f'router bgp {ASN}', stop_section='^exit')
         self.assertIn(f' address-family link-state', frrconfig)
 
+    def test_bgp_105_reject_bgp_referencing_incomplete_prefix_list(self):
+        # T8482: a peer-group referencing a prefix-list that itself fails
+        # verify() (missing "action") must not let the BGP commit succeed,
+        # and FRR must never receive the otherwise valid BGP configuration
+        # while the referenced policy object remains incomplete.
+        broken_prefix_list = 'pfx-foo-broken'
+        peer_group = 'foo'
+        peer = '192.0.2.100'
+
+        self.cli_set(['policy', 'prefix-list', broken_prefix_list, 'rule', '10', 'prefix', '10.0.0.0/24'])
+
+        self.cli_set(base_path + ['neighbor', peer, 'peer-group', peer_group])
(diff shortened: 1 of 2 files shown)


jestabro (commented):
This would appear to break partial commits (unless I'm missing something): it may need a finer-grained approach to apply `get_frrender_dict` on the sub-tree of the config that has passed verification, if that can be well-defined subject to config dependencies and cross tree inclusion (e.g. pki) ... that will need a closer look.

sever-sever:
Is it related to this change?
```
DEBUG - FAIL: test_show_diff_candidate_interfaces (__main__.TestConfigSyncWithHTTPS.test_show_diff_candidate_interfaces)
DEBUG - Validate: show configuration secondary sync commands candidate interfaces dummy
DEBUG -   File "/usr/libexec/vyos/tests/smoke/cli/test_service_config-sync.py", line 114, in test_show_diff_candidate_interfaces
DEBUG -     self.assertIn("set interfaces dummy dum0 address '192.0.2.2/32'", output)
DEBUG - AssertionError: "set interfaces dummy dum0 address '192.0.2.2/32'" not found in 'Failed to parse config: Syntax error on line 2, character 12: Invalid syntax.'
DEBUG - 

```

jestabro:
Two issues:
(1) smoketest error, a race condition similar to the one fixed here:
https://github.com/vyos/vyos-1x/commit/1f6a55bb039a94f1096a20b2dc11c4912ce8f6ce
as it is rare, it will be fixed in a separate PR (the smoketest is now passing below)

(2) This PR raised the general question (for me) regarding partial commits for frrender --- since this was already implicitly true in case of failure on the last commit, it is not a matter for this PR, which simply corrects an underlying error. The notion of partial commits for frrender is likely unsafe, but it is important to note the behavior.

jestabro:
Oddly, the smoketest passes in local tests, and in the earlier test above --- I'll keep an eye on this, as some of those util functions are also being tested/used in the interface reverse dependency checks, so any corner case will be identified.

jestabro:
As discussed with @c-po , I will drop the change to vyos-commitd before approving --- cumulative changes from vyos-configd that are applicable to vyos-commitd will be in a separate PR.

jestabro (approved):
Fix logic error to exclude `call_frr_render` on any commit error, not just the last.

PR #5422, opened 2026-08-20
qat: T9236: report actual QAT device state instead of configuration state

statio:
## Change summary

`show system acceleration qat status` reported on the configuration tree rather than on the device, so on hardware where the QAT modules autoload it printed the opposite of reality:

```
vyos@hub:~$ show system acceleration qat status
         system acceleration qat is not configured

vyos@hub:~$ lsmod | grep -i qat
qat_200xx              20480  1
intel_qat             413696  17 qat_200xx

vyos@hub:~$ sudo dmesg | grep -i qat
[   15.734217] 200xx 0000:01:00.0: qat_dev0 started 6 acceleration engines
```

The configuration node does not gate the driver — the modules autoload from the PCI modalias and the device is started by the driver itself — so op mode should not consult the config session to decide what the hardware is doing.

Three commits, deliberately separable:

1. **`--dev-list` argparse fix.** The op-mode definition calls the completion helper as `--dev-list`; the script declared `--dev_list`, so `show system acceleration qat device <tab>` failed with `unrecognized arguments: --dev-list`. Introduced by 6d4caef050a1 (T5191), which rewrote the XML side only.
2. **Add the missing `8086:18ee` (QAT_200XX) PCI ID to op mode.** Minimal and backportable on its own. 4a83a9ac55ca (T2968, 2020) added it to conf mode only, so `set system acceleration qat` committed while `show system acceleration qat` reported "No QAT device found" on the same box.
3. **The reporting rework.** Adds `python/vyos/qat.py`, which reads device state from sysfs, debugfs and `/proc/crypto`, and rebuilds the status command on it. Both conf mode and op mode now read one PCI ID table, so the two cannot drift apart again.

## User-visible output — please review the wording

This is the part I would most like a ruling on.

**Driver bound and started, node not set — the case that motivated this:**

```
Intel QAT

PCI address    Chipset    Driver    State
-------------  ---------  --------  -------
0000:01:00.0   QAT_200XX  200xx     started

Kernel crypto framework:  17 algorithm(s) registered by QAT
Configuration:            'system acceleration qat' is not set

WARNING: QAT is registered with the kernel crypto framework but
         'system acceleration qat' is not set. The kernel modules
         autoload from the PCI modalias and the device is started by
         the driver, both independently of the configuration, so
         cryptographic operations - IPsec ESP among them - may still
         be offloaded to the QAT device.
```

**Configured and running** — identical, with `'system acceleration qat' is set` and no warning.

**Hardware present, no driver bound:**

```
Intel QAT

PCI address    Chipset    Driver    State
-------------  ---------  --------  ---------------
0000:01:00.0   QAT_200XX  none      no driver bound

Kernel crypto framework:  0 algorithm(s) registered by QAT
Configuration:            'system acceleration qat' is not set
```

**No QAT hardware:** `No QAT device found`, exit 1.

Notes on specific choices, all of which I am happy to change:

- The warning fires on crypto framework registration, not merely on the driver being bound, because registration is what puts QAT in the ESP datapath. A bound-but-unregistered device still shows its state in the table but does not warn.
- `State` is one of `started`, `not started`, `no driver bound`, `unknown`. `unknown` appears when debugfs is not mounted, rather than guessing that a bound device is stopped.
- This drops the stray leading tab from `No QAT device found`. Say the word and I will keep that string byte-identical.
- The Intel `qat_service status` output is still appended when `/etc/init.d/qat_service` exists, but is no longer required — device state comes from the kernel.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9236

Referenced for context, not fixed here:

* https://vyos.dev/T6177 — the deadlock this exposure leads to. Upstream ruled it a crypto-driver bug, so the fix belongs in the out-of-tree Intel driver's completion path; tracked separately against vyos-build.
* https://vyos.dev/T2853 — added `--enable-qat-lkcf`, the origin of the LKCF exposure.

## Related PR(s)

None. The driver-side fix for T6177 will be a separate PR against vyos-build.

## Also changed

- The config-tree gate is removed from `flows`, `config` and `interrupts` as well. They report on a running device and had no reason to consult the configuration session.
- `get_qat_proc_path()` raised `TypeError` when it could not determine the path — it returned `None`, which was then concatenated with a string. It now prints a diagnostic and exits.

## Deliberately not included

The substantive problem behind T9236 is that the documented opt-in does not gate the driver. This PR makes that visible; it does not remove the exposure. Two further options are written up in the task for a ruling:

- **Warn at boot or on commit** when QAT hardware is bound but the node is unset.
- **Blacklist the QAT modules by default** and have the conf-mode script load them when the node is set. This is the only option that makes the documented opt-in true, but changing default module loading for a hardware class is a maintainer decision — getting it wrong silently disables acceleration for people who deliberately want it — and it spans vyos-build packaging.

Also left alone: `--dev-list`, `flows` and `config` still resolve devices through `/etc/init.d/qat_service`, because the `qat_devN` names it produces are user-visible CLI tokens. Moving those to sysfs would change what operators type, so it needs a decision on device naming first.

Please read this before merging: **I do not have QAT hardware, or a VyOS box.** Work was done on a Debian workstation, so the smoketests below have not been run on a live image.

Exercised:

- `vyos.qat.find_devices()` against a mocked sysfs covering all four device states, multiple devices in mixed states, non-QAT Intel devices, non-Intel devices sharing a device ID, and a missing `/sys/bus/pci/devices`.
- `/proc/crypto` parsing against realistic samples: a QAT-present case, a QAT-absent case with `aesni_intel` entries that must not match, and an unreadable file.
- All four status outputs above rendered from the real `show_qat_status()` with a mocked filesystem — that is where those samples come from.
- The PCI regex change against the verbatim `lspci -nn` line from the affected box, plus each previously supported chipset as a regression check.
- The new option-naming smoke test was confirmed to fail against the pre-fix source, so it is not a test that cannot fail.
- darker and ruff clean.

**Not** exercised: any code path that needs a real QAT device — debugfs directory naming, `qat_service` interaction, `fw_counters` and `dev_cfg` output, or interaction with a live IPsec session. The debugfs directory is matched by PCI address (`qat_*_<address>`) rather than by driver type specifically so that it does not depend on the type/driver name mapping, which I could not verify.

New smoke tests, which do not require acceleration hardware:

```
smoketest/scripts/system/test_qat.py
  TestQATDeviceIDs      - PCI ID table well-formed; 8086:18ee present
  TestQATDeviceState    - every branch of the device state machine
  TestQATDiscovery      - find_devices() shape; /proc/crypto parses on the running kernel
  TestQATOpMode         - all declared options use hyphens; --dev-list accepted
```

## Checklist:

### DIFF
--- a/python/vyos/qat.py
+++ b/python/vyos/qat.py
@@ -88,7 +88,12 @@ def find_devices() -> list:
         driver = None
         driver_link = os.path.join(device_path, 'driver')
         if os.path.islink(driver_link):
-            driver = os.path.basename(os.readlink(driver_link))
+            try:
+                driver = os.path.basename(os.readlink(driver_link))
+            except FileNotFoundError:
+                # The device can be unbound between the two calls. Report it
+                # as unbound rather than aborting the whole scan.
+                pass
 
         devices.append(
             {
--- a/python/vyos/qat.py
+++ b/python/vyos/qat.py
@@ -1,0 +1,162 @@
+# Copyright VyOS maintainers and contributors <maintainers@vyos.io>
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Intel QuickAssist Technology (QAT) device discovery.
+
+The QAT kernel modules are autoloaded from the PCI modalias and the device is
+brought up by the driver itself. Neither of those depends on the "system
+acceleration qat" configuration node, so the state of a QAT device must be read
+from the kernel and can never be inferred from the configuration session.
+"""
+
+import os
+
+from glob import glob
+
+from vyos.utils.file import read_file
+
+# Intel vendor ID, as it appears in /sys/bus/pci/devices/*/vendor
+PCI_VENDOR_ID_INTEL = '0x8086'
+
+# PCI device ID -> chipset name, for the QAT drivers VyOS ships.
+# Keep in sync with verify() in src/conf_mode/system_acceleration.py.
+PCI_DEVICE_IDS = {
+    '0x0435': 'DH895',
+    '0x18ee': 'QAT_200XX',
+    '0x19e2': 'C3xx',
+    '0x37c8': 'C62x',
+    '0x37c9': 'C62xvf',
+    '0x6f54': 'D15xx',
+}
+
+PCI_DEVICES_PATH = '/sys/bus/pci/devices'
+DEBUGFS_PATH = '/sys/kernel/debug'
+PROC_CRYPTO = '/proc/crypto'
+
+
+def _read_sysfs(path):
+    return read_file(path, defaultonfailure=None)
+
+
+def get_debugfs_path(address: str):
+    """Return the debugfs directory the QAT driver created for a PCI device.
+
+    The directory is named qat_<type>_<pci address>, where <type> depends on
+    which driver bound the device. Match on the address so that the type does
+    not have to be known in advance. Returns None when no such directory
+    exists, which is the case while the device has not been started.
+    """
+    found = glob(os.path.join(DEBUGFS_PATH, f'qat_*_{address}'))
+    return found[0] if found else None
+
+
+def find_devices() -> list:
+    """Return the QAT devices present on the PCI bus.
+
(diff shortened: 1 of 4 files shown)


statio:
Thanks — two of the three were worth fixing and are pushed:

- `find_devices()` now catches `FileNotFoundError` from `os.readlink()` and
  reports the device as unbound. Every other sysfs read in that loop already
  tolerates the file disappearing, so this was inconsistent.
- `get_qat_proc_path()` only returns a path once both the type and the bus
  address were actually parsed out. Previously an unknown device name fell
  through to `/sys/kernel/debug/qat__/`, so the error branch below it could
  never be reached.

On the smoketest: the discovery and state tests are deliberately
hardware-independent, since CI images have no QAT device — they assert that
`find_devices()` never raises and that whatever it returns is well formed, and
`get_device_state()` is exercised directly with constructed device dicts, which
does cover the supported/unsupported and bound/unbound cases. Discovery against
a mocked sysfs tree (multiple devices in mixed states, non-QAT Intel devices,
non-Intel devices sharing device ID 0x18ee) and `/proc/crypto` parsing against
captured samples were both verified out of tree while writing this. Happy to
bring those fixtures into the smoketest if a maintainer would prefer them in
the repo.

statio:
@c-po any way I can get you to give this a once-over for me?

PR #5435, opened 2026-08-30
T8264: add proper support for openvpn 2.7

ordex:
## Change summary

Add all needed support for openvpn 2.7

## Types of changes

This is a mix of a bug fix (re-enabling support for DCO), while also introducing a new feature (switching to more modern openvpn 2.7.5)

## Related Task(s)

https://vyos.dev/T8264

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1281
* https://github.com/vyos/vyos-1x/pull/5467

smoketested + actual run on a freshly built ISO

## Checklist:

### DIFF
--- a/data/templates/openvpn/server.conf.j2
+++ b/data/templates/openvpn/server.conf.j2
@@ -36,6 +36,7 @@ remote {{ remote }}
 {% endif %}
 {% if shared_secret_key is vyos_defined %}
 secret /run/openvpn/{{ ifname }}_shared.key
+allow-deprecated-insecure-static-crypto
 {% endif %}
 {% if persistent_tunnel is vyos_defined %}
 persist-tun
--- a/data/templates/openvpn/server.conf.j2
+++ b/data/templates/openvpn/server.conf.j2
@@ -70,12 +70,12 @@ bind ipv6only
 mode server
 tls-server
 {%     if server is vyos_defined %}
-{%         if server.subnet is vyos_defined %}
-{%             if server.topology is vyos_defined('point-to-point') %}
+{%         if server.topology is vyos_defined('point-to-point') %}
 topology p2p
-{%             elif server.topology is vyos_defined %}
+{%         elif server.topology is vyos_defined %}
 topology {{ server.topology }}
-{%             endif %}
+{%         endif %}
+{%         if server.subnet is vyos_defined %}
 {%             for subnet in server.subnet %}
 {%                 if subnet | is_ipv4 %}
 server {{ subnet | address_from_cidr }} {{ subnet | netmask_from_cidr }} {{ 'nopool' if server.client_ip_pool is vyos_defined and server.client_ip_pool.disable is not vyos_defined else '' }}
--- a/debian/control
+++ b/debian/control
@@ -174,7 +174,7 @@ Depends:
   libqmi-utils,
 # End "interfaces wwan"
 # For "interfaces openvpn"
-  openvpn,
+  openvpn (>= 2.7),
   openvpn-auth-ldap,
   openvpn-auth-radius,
   openvpn-otp,


ordex:

ordex:
recheck

sever-sever (dismissed):
Add options for OpenVPN 2.7+
Add DCO
Extend smoketests

alexk37:
Hi @ordex ! Please inform if we should expect more changes or pr is ready for review and testing

src/conf_mode/interfaces_openvpn.py
```diff
@@ -209,6 +236,64 @@ def verify_data_ciphers_fallback(openvpn):
...
+        keyword = tmp[0].lstrip('-')
+        if keyword in dco_incompatible_options:
+            raise ConfigError(f'DCO is incompatible with "openvpn-option {keyword}"')
+        if keyword in dco_conditional_options:
+            keep = dco_conditional_options[keyword]
+            if tmp[1:] != [keep]:
+                raise ConfigError(f'DCO requires "openvpn-option {keyword} {keep}"')
+        if keyword in dco_cipher_options and tmp[1:]:
+            cipher = unoffloadable_cipher(tmp[1])
+            if cipher is not None:
```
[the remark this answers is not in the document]
ordex: `?DEFAULT` is not allowed because we do exact pattern matching against `DEFAULT`.
`DEFAULT` is now properly handled.

ordex:
@alexk37 I think this PR is ready for review. #5441 is still WIP as it requires a patch from upstream first

EDIT: I was wrong - there is still something pending that I haven't pushed. Sorry

ordex:
@alexk37 ready now :-) Thanks for your patience. Could you also kick-off the CI?

Note that there is a follow-up PR #5441 aimed at improving the device life cycle (Still draft at this point).

src/migration-scripts/openvpn/5-to-6
```diff
@@ -0,0 +1,150 @@
...
+    # a zero interval renders "keepalive 0 0" and turns keepalive off, which
+    # OpenVPN and verify() both leave alone
+    if interval < 1:
+        return
+
+    # the timeout is interval * failure-count: it has to reach twice the
+    # interval and stay below 12 hours. Beyond an interval of 21600 no count
+    # satisfies both - the CLI range stops at 600, so only a hand-edited
+    # configuration gets there and its interval is rejected anyway.
+    wanted = min(max(count, 2), max(2, 43200 // interval))
```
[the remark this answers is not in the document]
ordex: keep-alive interval is validated --range 0-600 (rolling through equuleus), so an interval above 21600 cannot exist in a saved configuration — only in a hand-edited file. Clamping it to 21600 does not help there: that is still outside 0-600, so the numeric validator rejects the node at commit regardless of what the migration writes. The max(2, ...) guard only ensures the migration never emits an invalid failure-count; the out-of-range interval is correctly rejected by validation.

ordex:
This PR fails CI because it depends on https://github.com/vyos/vyos-build/pull/1281

src/conf_mode/interfaces_openvpn.py
```diff
@@ -967,7 +967,11 @@ def apply(openvpn):
     if 'module_load_dco' in openvpn:
         check_kmod(dco_module)
     else:
-        unload_kmod(dco_module)
+        # Interfaces are committed one at a time, so a daemon that has not been
+        # stopped yet can still hold an "ovpn" device and make rmmod fail with
+        # EBUSY. The last invocation gets to unload it, and a module left
+        # loaded is harmless - failing the commit is not.
+        unload_module(dco_module)
```
c-po: I wonder if we should unload the module at all. Neither vxlan, geneve not l2tpv3 unload the kernel modules after destroying the last interface. I think we can simplify the code here and just keep it loaded.

src/conf_mode/interfaces_openvpn.py
```diff
@@ -139,7 +139,7 @@ def get_config(config=None):
 
     if is_node_changed(conf, base + [ifname, 'openvpn-option']):
         openvpn.update({'restart_required': {}})
-    if is_node_changed(conf, base + [ifname, 'enable-dco']):
+    if is_node_changed(conf, base + [ifname, 'offload']):
```
c-po:
```suggestion
    if is_node_changed(conf, base + [ifname, 'offload', 'dco']):
```

c-po (changes requested):

src/conf_mode/interfaces_openvpn.py
ordex: I presume it's just to cleanup the module list in case no new `ovpn` devices are ever created anymore?
Anyway, I am fine with simplifying the code: most systems just load all available modules and stick with them.

Let me know what's your final preference and I'll modify the code accordingly.

ordex: As discussed on Slack: let's make it consistent with the other drivers and drop the unload.

c-po (approved):
Requested changes adressed

PR #5436, opened 2026-08-31
dhcp-server: T9267: add client-class matching on hostname/vendor-class-id

ruben-herold:
## Summary

Client classes could previously only match on relay agent information
(Option 82, `relay-agent-information circuit-id/remote-id`). This adds
two more match types:

- `client-class <name> hostname value|substring ...` — match on the
  client-sent hostname (Option 12)
- `client-class <name> vendor-class-id value|substring ...` — match on
  the vendor class identifier (Option 60)

Both support an exact `value` or a `substring` (with optional `offset`/
`length`), and follow the existing hex/text convention (`0x...` is raw
hex, anything else is ASCII text).

A client-class can now also deliver its own options (reusing the
existing subnet/range `option` syntax), so a match can directly hand
out something like a model-specific provisioning URL — independent of
which pool the client's lease is drawn from, so the subnet does not
need to be split into per-model pools.

Addresses https://vyos.dev/T9267, where multiple VoIP phone models
sharing one subnet each need a different Option 66 provisioning URL,
identified by a substring of their DHCP hostname.

## Test plan

- [x] Verified the generated Kea classification test expressions and
      hex encodings against hand-computed values (`_kea_hex_literal`,
      `_build_option_match_condition`)
- [x] Validated the transcluded interface definition against
      `schema/interface_definition.rng` via `scripts/build-command-templates`
- [x] `darker -r <base> --check --diff` clean
- [x] `graylint -L "ruff check" --revision <base>` clean
- [x] Added `test_dhcp_client_class_option_match` smoketest (exact +
      substring match, hex + text values, mutual-exclusion and
      invalid-hex validation, per-class option-data)
- [ ] CI smoketest run


alexk37:
@ruben-herold I tested this on an image built from c294d2e with real DHCP clients on the wire. The feature works as described: a class built from `hostname` or `vendor-class-id` classifies the client and the per-class options reach it.

Please check the following behaviour I observed and would like to flag - if the same option is also set on the subnet or on a range, that value is delivered instead of the class value, and nothing warns at commit time.

```
set service dhcp-server client-class t42s hostname substring value 'SEP0'
set service dhcp-server client-class t42s option domain-name 't42s.prv5436.test'
set service dhcp-server shared-network-name PRV5436 subnet 10.0.2.0/24 option domain-name 'sub.prv5436.test'
commit
```

A client with hostname `SEP0-2` requesting option 15 then receives:

```
domain=sub.prv5436.test
```

Deleting the subnet line brings back `domain=t42s.prv5436.test`. I measured the same for `tftp-server-name` (option 66) at subnet level and for `domain-name` at `range 0` level. Kea's documented option precedence is pool > subnet > class, so this is Kea behaviour rather than a rendering error, but from the CLI the two settings look like peers. The natural layout of "a subnet-wide default plus a per-model override" therefore delivers the default to every phone, with a successful commit and both values present in the rendered file.

The PR body and the docs PR say the class options are delivered "independent of which subnet or pool the lease is drawn from". That held for pool assignment in my tests, but a reader will also take it to cover options set on that subnet or pool, which it does not.

Maybe a commit-time warning when a class option code also appears at subnet or range level in the same shared network is needed here. Or a sentence in the docs stating that subnet and range options override client-class options.

alexk37:
There is validation gap found - a bare hostname or vendor-class-id node under a client-class passes verify() and the commit failes with KeyError: 'value'

```
set service dhcp-server client-class bh hostname
commit
```

```
  File "/usr/lib/python3/dist-packages/vyos/kea.py", line 755, in _build_option_match_condition
    return _build_hex_condition(kea_path, match_config['value'])
KeyError: 'value'

[[service dhcp-server]] failed
Commit failed
```
______________________________________________

Related validation gap - a client-class with options but no `hostname` or `vendor-class-id` passes verify, and the commit fails with Kea's parser message instead of a VyOS error naming the class.

```
set service dhcp-server client-class onlyopts option domain-name 'onlyopts.prv5436.test'
commit
```

```
Unexpected error with Kea configuration:
expression: [] error: <string>:1.1: syntax error, unexpected end of file
at (/run/kea/kea-dhcp4.conf:24:17)
[[service dhcp-server]] failed
Commit failed
```

I reached it by deleting a class's `hostname` and `option` in one commit and leaving the bare class name.

alexk37:
One more on validation - an explicit `substring length` that differs from the byte length of `value` is accepted, and the class then either does not match or matches something different from what was configured.

```
set service dhcp-server client-class t42s hostname substring value 'SEP0'
set service dhcp-server client-class t42s hostname substring length '2'
set service dhcp-server client-class t42s option domain-name 't42s.prv5436.test'
commit
```

renders

```
"test": "substring(option[12].hex, 0, 2) == 0x53455030",
```

Two bytes of hostname compared with a four-byte constant. A client with hostname `SEP0-1` got a lease with none of the class options; deleting the `length` node brought them back.

With `length 8` the same client did not match either, but a client named exactly `SEP0` did: Kea's substring returns the bytes that exist, so a length longer than the value turns the prefix match into an exact match of the whole remainder.

The help text says "Number of bytes to compare (default: length of value)", which reads as a setting independent of `value`; in the cases I tested, a shorter length matched nothing and a longer one changed the meaning of the match.

alexk37:
A question on the intended match semantics, for the maintainers and the author.

The T9267 ticket's Mikrotik config uses `matching-type=substring value=T42S`, which Mikrotik documents as "value can match anywhere in the option string, at the start, middle, or end". This PR renders `substring` as a slice at a fixed offset (default 0) compared byte for byte. `hostname substring value SEP0` matched a client named `SEP0-1` and did not match `XSEP0-1`, and `vendor-class-id value YEALINK` did not match a client sending `yealink`.

The docs PR says "identified by the start of their DHCP hostname". The CLI help says "Filters on part of the option contents" and the PR body says "a substring of their DHCP hostname"; neither mentions the fixed position or case sensitivity.

Is prefix-at-offset the intended semantics for `substring`? If so, maybe worth stating so in the CLI help. If contains-anywhere is wanted, Kea's expression language has `match()` (regex) and `lcase()`, so a Mikrotik-style test could be `match('.*t42s.*', lcase(option[12].text))`; documented at https://kea.readthedocs.io/en/kea-3.0.3/arm/classify.html (see "match" and "lcase"). I have not tried it.

ruben-herold:
@alexk37 Thanks for the thorough testing — both crash/validation gaps are fixed in 8801f98:

- A client-class with `option` set but no `relay-agent-information`, `hostname`, or `vendor-class-id` now fails `verify()` with a clear `ConfigError` instead of producing an empty Kea `test` expression that the Kea parser rejected with an opaque syntax error.
- A bare `hostname`/`vendor-class-id` node with neither `value` nor `substring value` set now fails `verify()` instead of crashing with `KeyError: 'value'` while rendering.

Added smoketest coverage for both in the same commit.

Still open, and I'd like your input before touching code:

- **`substring length` vs. byte-length of `value`**: agreed this is a real footgun — a shorter `length` never matches, a longer one gets silently clamped by Kea into an exact match. Leaning towards just dropping the standalone `length` leaf (always derive it from `value`, which is already the default) rather than adding another verify constraint, since I can't find a legitimate use case for setting it independently. Open to being told otherwise.
- **subnet/range options overriding class options**: confirmed as Kea's documented pool > subnet > class precedence, not a rendering bug. Will add a note to the docs PR (#2227) clarifying that subnet/range-level options win over class options with the same code.
- **`substring` semantics (fixed-offset vs. contains-anywhere)**: this is the one I want a maintainer opinion on before changing anything, since it's the actual T9267 motivation (Mikrotik-style phone matching) and an API-shape decision. Current implementation is offset-anchored, matching the CLI help text ("part of the option contents" starting at `offset`, default 0) — but if contains-anywhere is what people actually need for vendor-class-id/hostname matching, `match()`+`lcase()` per your suggestion is the right primitive and would need to be exposed as an explicit CLI option (e.g. a `contains` alternative to `substring`) rather than silently changing what `substring` means today.

Will follow up on the last two once there's been time for others to weigh in.

ruben-herold:
Update on the two remaining open points from my earlier comment:

**`substring` matching semantics (08621a3bc)**: switched from a fixed-offset exact-slice comparison (`substring(...)==`) to a case-insensitive contains-anywhere search via Kea's `match()`+`lcase()`. This actually matches the T9267 ticket's own reference: the Mikrotik `matching-type=substring` config it's based on is documented as matching the value anywhere in the option string, and Yealink hostnames don't guarantee the model string (`T42S`/`T54W`/...) sits at a fixed byte offset. Dropped the `offset` leaf along with it — no purpose left once matching isn't anchored to a position, and nothing released to preserve. Raw hex and values containing `'` are now rejected for `substring` in `verify()`, since neither can be safely embedded into the generated regex; exact `value` matching is unchanged and still takes both. Docs (#2227) and smoketests updated to match.

**Empty `relay-agent-information` gap (efaf13518)**: coderabbitai caught that a bare `relay-agent-information` node with neither `circuit-id` nor `remote-id` slipped past the "at least one match condition" check from 8801f98 — the check only looked at key presence, not actual content, so it reproduced the same empty-Kea-`test`-expression bug through a different path. Fixed the same way as `hostname`/`vendor-class-id`, with a regression test.

Ready for another look whenever you have time, @alexk37.

sever-sever (dismissed):
It cannot pass smoketests
@ruben-herold could you check?
```
 DEBUG - Running Testcase (46/94): /usr/libexec/vyos/tests/smoke/cli/test_service_dhcp-server.py
DEBUG - test_dhcp_client_class (__main__.TestServiceDHCPServer.test_dhcp_client_class) ... ok
DEBUG - test_dhcp_client_class_option_match (__main__.TestServiceDHCPServer.test_dhcp_client_class_option_match) ... ERROR
DEBUG - test_dhcp_dynamic_dns_update (__main__.TestServiceDHCPServer.test_dhcp_dynamic_dns_update) ... ok
DEBUG - test_dhcp_exclude_in_range (__main__.TestServiceDHCPServer.test_dhcp_exclude_in_range) ... ok
DEBUG - test_dhcp_exclude_not_in_range (__main__.TestServiceDHCPServer.test_dhcp_exclude_not_in_range) ... ok
DEBUG - test_dhcp_high_availability (__main__.TestServiceDHCPServer.test_dhcp_high_availability) ... ok
DEBUG - test_dhcp_high_availability_standby (__main__.TestServiceDHCPServer.test_dhcp_high_availability_standby) ... ok
DEBUG - test_dhcp_hostsd_lease_sync (__main__.TestServiceDHCPServer.test_dhcp_hostsd_lease_sync) ... ok
DEBUG - test_dhcp_log_level (__main__.TestServiceDHCPServer.test_dhcp_log_level) ... ok
DEBUG - test_dhcp_multiple_pools (__main__.TestServiceDHCPServer.test_dhcp_multiple_pools) ... ok
DEBUG - test_dhcp_on_interface_with_vrf (__main__.TestServiceDHCPServer.test_dhcp_on_interface_with_vrf) ... ok
DEBUG - test_dhcp_relay_server (__main__.TestServiceDHCPServer.test_dhcp_relay_server) ... ok
DEBUG - test_dhcp_single_pool_options (__main__.TestServiceDHCPServer.test_dhcp_single_pool_options) ... ok
DEBUG - test_dhcp_single_pool_options_scoped (__main__.TestServiceDHCPServer.test_dhcp_single_pool_options_scoped) ... ok
DEBUG - test_dhcp_single_pool_range (__main__.TestServiceDHCPServer.test_dhcp_single_pool_range) ... ok
DEBUG - test_dhcp_single_pool_static_mapping (__main__.TestServiceDHCPServer.test_dhcp_single_pool_static_mapping) ... ok
DEBUG - test_dhcp_vendor_option_ubiquiti (__main__.TestServiceDHCPServer.test_dhcp_vendor_option_ubiquiti) ... ok
DEBUG - 
DEBUG - ERROR: test_dhcp_client_class_option_match (__main__.TestServiceDHCPServer.test_dhcp_client_class_option_match)
DEBUG -   File "/usr/libexec/vyos/tests/smoke/cli/test_service_dhcp-server.py", line 498, in test_dhcp_client_class_option_match
DEBUG -     self.cli_set(client_class + ['vendor-class-id', 'substring', 'value', "foo'bar"])
DEBUG -   File "/usr/libexec/vyos/tests/smoke/cli/base_vyostest_shim.py", line 112, in cli_set
DEBUG -     self._session.set(path, value)
DEBUG -   File "/usr/lib/python3/dist-packages/vyos/configsession.py", line 277, in set
DEBUG -     self.__run_command([SET] + path + value)
DEBUG -   File "/usr/lib/python3/dist-packages/vyos/configsession.py", line 262, in __run_command
DEBUG - vyos.configsession.ConfigSessionError: Cannot use the single quote (') character in a value string
DEBUG - Value validation failed
DEBUG - Set failed
DEBUG - 
```

ruben-herold:
@sever-sever Thanks for catching this — pushed a fix in 6d83e4773:

The root cause wasn't `verify()`, it was the test itself: VyOS's own CLI value syntax already rejects a literal single quote at `cli_set()` time, before it ever reaches the substring-specific check I added. The test wrapped `assertRaises` around `cli_commit()` instead of `cli_set()`, so the exception was raised uncaught → `ERROR` instead of the expected failure. Also wrapped the two long `cli_set()` calls per darker/black.

Note: `typos` is failing on an unrelated pre-existing issue (`writeable`→`writable` in `python/vyos/system/image.py:314`, already on `rolling`), not something from this PR.

CI is currently sitting on `action_required` for CodeQL/Lint/ISO/Typos — could you (or another maintainer) approve the run?

ruben-herold:
Follow-up on the CI status: I looked into the remaining failures more closely, and none of them are related to this branch's code:

- **`typos`**: the only reported issue is `writeable` → `writable` in `python/vyos/system/image.py:314`. That file is untouched by this PR and the typo already exists on `rolling` (verified against `upstream/rolling` directly) — it's a pre-existing issue in the base branch.
- **`test_smoketest_cli` / `test_interfaces_cli` / `test_config_load` / `test_raid1_install` / `test_encrypted_config_tpm`**: these all fail at the QEMU install-verification step with a kernel version mismatch — the built ISO boots `6.18.48-vyos`, but the test regex expects `6.18.50-vyos`, causing a `pexpect.TIMEOUT`.
- This isn't specific to this PR: the `VyOS ISO Integration Test` workflow is failing the same way on `rolling` itself (runs on 2026-09-02, 2026-09-07, and again in progress as of today, 2026-09-08).

So this looks like a rolling-branch/infra issue (kernel package vs. test expectation out of sync) rather than something actionable in this PR. I'll rebase/re-trigger once that's resolved upstream — no action needed on the CI-approval front from my last comment, that was a misdiagnosis on my part.

alexk37:
@mergify rebase

c-po:
@MergifyIo rebase

Firefishy:
I am looking forward to this feature. I think it will unlock proper netboot.xyz support.

eg:
```
# Legacy BIOS
set service dhcp-server client-class 'PXE-BIOS' vendor-class-id substring value 'PXEClient:Arch:00000'
set service dhcp-server client-class 'PXE-BIOS' option bootfile-name 'netboot.xyz-legacy.kpxe'

# UEFI x86-64
set service dhcp-server client-class 'PXE-UEFI-64' vendor-class-id substring value 'PXEClient:Arch:00007'
set service dhcp-server client-class 'PXE-UEFI-64' option bootfile-name 'netboot.xyz.efi'

# UEFI x86-64, older/alternative architecture identifiers
set service dhcp-server client-class 'PXE-UEFI-64-8' vendor-class-id substring value 'PXEClient:Arch:00008'
set service dhcp-server client-class 'PXE-UEFI-64-8' option bootfile-name 'netboot.xyz.efi'

set service dhcp-server client-class 'PXE-UEFI-64-9' vendor-class-id substring value 'PXEClient:Arch:00009'
set service dhcp-server client-class 'PXE-UEFI-64-9' option bootfile-name 'netboot.xyz.efi'
```

alexk37:
@ruben-herold 
I tested this branch on an image built from be19c4d6, there is a regression caused by the changes:

**A client class that carries only `disable` commits on current rolling and is refused here, and after a
reboot the box has no DHCP server configuration**

On current rolling:

```
set service dhcp-server client-class keep relay-agent-information circuit-id '0x666f6f'
set service dhcp-server client-class parked disable
commit
```

commits. On this branch the same two lines fail:

```
[ service dhcp-server ]
Client class "parked" requires at least one match condition (relay-
agent-information, hostname, or vendor-class-id)
[[service dhcp-server]] failed
Commit failed
```

Checking the upgrade path I put the current-rolling configuration above into `config.boot` on a box running this
branch and rebooted it. After the boot commit:

```
   client-class in the ACTIVE config   : 0
   dhcp-server subtree in ACTIVE config: 0
   isc-kea-dhcp4-server: inactive  substate=dead
   render present: NO
```

The whole `service dhcp-server` subtree is gone from the running configuration, Kea is not running, and no
Kea configuration file was written.

The check added in 4d2e7e87 reads:

```python
for class_name, class_config in dhcp['client_class'].items():
    if not any(
        k in class_config
        for k in ('relay_agent_information', 'hostname', 'vendor_class_id')
    ):
        raise ConfigError(...)
```

It walks every class in the tree, and a disabled class has no matcher key. `python/vyos/template.py`
skips disabled classes when rendering (`if 'disable' in config: continue`), so the renderer does not emit
a disabled class into the Kea configuration; the new check refuses the commit before that point.

ruben-herold:
@alexk37 Good catch, thanks for testing the upgrade path too. Root cause confirmed: the new "at least one match condition" check in `verify()` didn't account for `disable`, even though the renderer (`kea_client_class_json()`) already skips disabled classes entirely and they never reach Kea. So a pre-existing config like yours (a disabled class with no matcher) committed fine on rolling but got rejected here — and since that's exactly the kind of failure a boot-time commit can't recover from, it took the whole `service dhcp-server` subtree with it.

Fixed in e3831e9: `verify()` now skips the match-condition check (and the rest of the per-class validation) for any class with `disable` set, mirroring the renderer's own skip condition. Added a regression test (disabled class with no matcher, alongside an enabled one) to `test_dhcp_client_class`.

Could you re-test the upgrade path from your report when you get a chance?

c-po:
@MeegifyIo rebase

c-po:
@MergifyIo rebase

ruben-herold:
@c-po Mergify cannot rebase this one — it refuses the rebase API for fork PRs because that would mean force-pushing a contributor's branch under an impersonated token. I did it by hand instead: `fd340245f` -> `2a6660c39`, all 8 commits replayed linearly onto `03229944a`.

The rebase was a no-op content-wise, and I checked that rather than assuming it:

- The PR's own diff is byte-identical before and after (442 lines) — no conflicts, so no resolution judgement was involved.
- None of the 20 new commits in `rolling` touch any of the 5 files this PR changes, so there is no textually-clean-but-semantically-broken replay to worry about.
- `ruff` on the 5 files reports exactly the same findings as `rolling` does on the same files — nothing new from the PR.

@sever-sever on your `CHANGES_REQUESTED`: the smoketest failure you posted was `test_dhcp_client_class_option_match` choking on `foo'bar`, and that one is fixed in `fdd103194` (same day as your review). Root cause was in my test, not in the feature: VyOS's own value syntax rejects a single quote at `set` time, so `assertRaises` has to wrap `cli_set()` rather than `cli_commit()` — the substring-specific check in `verify()` is never reached at all. Since then @alexk37 found a separate and more serious one (a client class carrying only `disable` and no matcher was rejected, which a boot-time commit cannot recover from); that is fixed in `fd340245f`, now `2a6660c39`.

A few notes on the CI signal, so the next run is easier to read:

- Nothing has run yet on the current head. All four `pull_request`-triggered workflows (ISO Integration Test, Typos, Darker + Ruff, CodeQL) sit at `action_required` — they need a maintainer to approve the workflow run for this fork PR. Only the two `pull_request_target` ones (CLA, conflict checker) went through, and both are green. Could someone hit "Approve and run workflows"? (Same state on #5459, #5463, #5468, #5476 and #5455 right now, so it is not specific to this PR.)
- Once it runs, `typos` will most likely still fail: it flags `localed` as a misspelling of `located`, matched inside `vyatta_localedir` in `src/etc/default/vyatta` — a file this PR does not touch. Pre-existing in the base branch and not fixable from here.
- The only failing smoketest on the previous run was `test_techsupport_archive.py`, which has nothing to do with this PR. Your own `tech-archive-fix` (`bd51819b0`, merged 2026-09-14 15:09 CEST) landed *after* that run and is now in the base, so this rebase should clear it.

Could you both take another look once the run is through? Happy to follow up with the results.

There is also still one open CodeRabbit thread on `test_service_dhcp-server.py` (line 406): the bare-`hostname` rejection is asserted while `domain-name` from the previous case is still configured, so a different validation error could satisfy the assertion. It is a fair point about the test, not the feature — I can push that isolation fix if you would like it in this PR rather than a follow-up.

PR #5439, opened 2026-09-02
interfaces: T6393: fix commit failure when the mirror target does not exist yet

ciaens:
## Change summary

`set interfaces ethernet ethN mirror ingress tun0` fails to commit at boot, and in any commit that creates both interfaces, because interface types commit in priority order (ethernet 318, bridge 319, bonding 320, ... tunnel 380) and `verify_mirror_redirect()` required the mirror target to exist in the kernel:

```
VYOS_TAGNODE_VALUE=eth1 .../interfaces_ethernet.py
Requested mirror interface "tun0" does not exist!
Sending reply: ERROR_COMMIT with output
VYOS_TAGNODE_VALUE=tun0 .../interfaces_tunnel.py
Sending reply: SUCCESS with output
```

The whole ethernet node is rolled back, so the interface comes up without its addresses and everything on top of it (in my case: VRRP, conntrack-sync, DHCP relay) fails until I run `load /config/config.boot; commit` manually.

The apply side already handles this direction since #644 (T3089): the target interface installs the tc(8) filters on behalf of its sources when it is created. Only `verify()` stood in the way, and the #644 code kept only the last source it found, so with several interfaces mirroring into one tunnel a single one got its filter.

While I targeted my use case (GRE tun on mirrored traffic), it should be noted that a mirror target whose kernel device is created asynchronously by a daemon (OpenVPN `vtunN`, WWAN) now passes verify at boot; if the device is not up when that script runs, the filter is installed on the next commit of either side instead of the source interface being rolled back.

## Types of changes

## Related Task(s)
* https://vyos.dev/T6393
* https://vyos.dev/T3089 (original port mirroring work, #621/#633/#644)
* https://vyos.dev/T260

## Related PR(s)
None

Commit/save and reboot with:

```
set interfaces ethernet eth1 address '10.20.0.2/16'
set interfaces ethernet eth1 mirror ingress 'tun0'
set interfaces tunnel tun0 encapsulation 'gre'
set interfaces tunnel tun0 source-interface 'eth0'
set interfaces tunnel tun0 remote '10.0.1.10'
```

Before: `/tmp/vyos-config-status` is 1, `interfaces ethernet eth1` is missing from the running config after boot (trace above), `eth1` is A/D. 

After: on vmware (amd64) and qemu (arm64) guests) rebooted with the saved config:

```
$ cat /tmp/vyos-config-status
0
$ journalctl -b -t vyos-configd | grep -i mirror
vyos-configd[871]: [interfaces_ethernet] Mirror target "tun0" of "eth0" does not exist yet, tc filter will be installed once it is created
```

Loading boot config is clean, all interfaces are up, and dependancies like vrrp, conntrack-sync, etc all load correctly.
A target that is neither configured nor existing is still rejected:

```
set interfaces ethernet eth1 mirror ingress tun9
commit
[ interfaces ethernet eth1 ]
Requested mirror interface "tun9" does not exist!
```

Smoketests (`TEST_ETH="eth1 eth2"` for ethernet):

```
$ /usr/libexec/vyos/tests/smoke/cli/test_interfaces_ethernet.py -k span_mirror
test_span_mirror (__main__.EthernetInterfaceTest.test_span_mirror) ... ok
test_span_mirror_target_created_later (__main__.EthernetInterfaceTest.test_span_mirror_target_created_later) ... ok
Ran 2 tests in 3.938s
OK
```

Ran dummy, loopback, tunnel, bridge, bonding, wireguard, vxlan, geneve and pseudo-ethernet/
All 204 tests OK, including the 8 new I've added in src/tests/test_configverify.py

## Checklist:


ciaens:

ciaens:
nb: the failures don't seem to be related to this pr, seems to be pre-existing.

ciaens:
Hi,
I hope everything is as expected on my side. Can I get an update please ?

python/vyos/configverify.py
```diff
@@ -173,12 +173,29 @@ def verify_tunnel(config):
         if 'source_address' in config and is_ipv6(config['source_address']):
             raise ConfigError('Cannot use local IPv6 address is for mGRE tunnels')
 
-def verify_mirror_redirect(config):
+def is_interface_defined(interfaces_root, ifname) -> bool:
```
c-po, line 176: There is no need for this new helper - see comment in callers of this function.

python/vyos/configverify.py
```diff
@@ -188,11 +205,18 @@ def verify_mirror_redirect(config):
         # limitation from the past
         raise ConfigError('Cannot use QoS together with mirror!')
 
+    if interfaces_root is None:
+        interfaces_root = getattr(config, 'interfaces_root', {})
+
     if 'mirror' in config:
         for direction, mirror_interface in config['mirror'].items():
-            if not interface_exists(mirror_interface):
```
c-po, line 193:
```python3
      if 'mirror' in config:
          for direction, mirror_interface in config['mirror'].items():
              # mirror target may only be defined on the CLI and not (yet)
              # created in the kernel - interface types are committed in
              # priority order
              verify_interface_exists(config if parent is None else parent,
                                      mirror_interface)

              if mirror_interface == config['ifname']:
                  raise ConfigError(f'Cannot mirror "{direction}" traffic back '\
                                     'the originating interface!')
```

python/vyos/configverify.py
```diff
@@ -395,7 +420,7 @@ def verify_vlan_config(config):
         verify_dhcpv6(vlan)
         verify_address(vlan)
         verify_vrf(vlan)
-        verify_mirror_redirect(vlan)
+        verify_mirror_redirect(vlan, interfaces_root)
```
c-po, line 423: I do not understand why to pass `interfaces_root` here instead of just `config`? as you load the interfaces root explicitly from config

```
    if interfaces_root is None:
        interfaces_root = getattr(config, 'interfaces_root', {})
```

src/tests/test_configverify.py
```diff
@@ -29,3 +32,75 @@ def test_dh_key_512(self):
...
+        },
+        'tunnel': {'tun4711': {'encapsulation': 'gre'}},
+    }
+
+    def _config(self, ifname, mirror):
+        config = _ConfigDict({'ifname': ifname, 'mirror': mirror})
+        config.interfaces_root = self.interfaces_root
+        return config
+
+    def test_is_interface_defined(self):
```
c-po, line 59: Testcase can be dropped if the `is_interface_defined` function is dropped as stated above.

c-po (changes requested):

python/vyos/configverify.py
ciaens: ok, noted. Thx.

python/vyos/configverify.py
ciaens: 👍

src/tests/test_configverify.py
ciaens: yep, dropped the testcase, thx

python/vyos/configverify.py
ciaens: I didn't think the root was available there, so I extracted only the tree I needed. 
But indeed I missed that verify_mtu_parent(vlan, config) was already using it below, my bad.

I've changed it.

ciaens:
Done: dropped the helper, VLAN sub IF pass their parent dict, test removed

On the CI checks I've looked at the logs, and it's not related to this PR.

c-po:
I have rebased the branch on top of `rolling` to get a fresh CI picture

ciaens:
Is related to timeout on `test_dhcp_vrf (__main__.PEthInterfaceTest.test_dhcp_vrf)`

Also failed on rolling / other PRs. What's annoying tho is that it failed before the new tests ...

¯\_(ツ)_/¯

python/vyos/configverify.py
```diff
@@ -190,9 +193,13 @@ def verify_mirror_redirect(config):
...
     if 'mirror' in config:
         for direction, mirror_interface in config['mirror'].items():
-            if not interface_exists(mirror_interface):
-                raise ConfigError(f'Requested mirror interface "{mirror_interface}" '\
-                                   'does not exist!')
+            # T6393: interface types are committed in priority order, the
+            # mirror target may only be defined on the CLI and not (yet) be
+            # created in the kernel. VLAN sub-interfaces are only accepted
+            # from the kernel as nothing would install the filter later on.
+            verify_interface_exists(
```
c-po, line 200: With changing to this implementation we loose the context that its a mirror interface which does not exist.

python/vyos/ifconfig/interface.py
```diff
@@ -1653,33 +1654,47 @@ def set_mirror_redirect(self):
...
-        redirect_config = None
+        # Interfaces using us as their mirror target: interface types are
+        # committed in priority order, so we may not have existed yet when
+        # those interfaces were configured and their tc(8) filter could
+        # not be installed. Install it on their behalf now that we exist
+        # T3089, T6393
+        for source_if, source_config in self.config.get('is_mirror_intf', {}).items():
+            if not interface_exists(source_if):
+                continue
+            self._apply_mirror_redirect(source_if, source_config)
```
c-po, line 1673: Every commit that touches the target interface now runs `tc qdisc del dev <source> root/ingress` + re-add for all of its sources, even when nothing about those sources changed. Previously this hit one source; now it hits all of them

c-po (changes requested):
This PR only addresses the mirroring portion and not the interface redirect part which has the same issue.

PR #5443, opened 2026-09-03
image: T5475: derive the persistence path from the overlay upperdir 

c-po:
## Change summary

The union boot branch matched `/proc/mounts` against the literal string `upperdir=/live/persistence/`, which is the layout produced by the VyOS fork of live-boot. Debian's live-boot mounts persistence below /run/live/persistence instead, so on an image built against it the match failed, the script reported an old style boot, and every caller silently did nothing. In practice that meant /boot was never bind mounted, so GRUB updates and image management found no configuration directory, and an encrypted config volume was never unlocked.

The old code also only worked by accident. It rebuilt the path as `/lib/live/mount/persistence/$boot_device`, where the device component came from cutting the fourth slash separated field, and that field was empty solely because the fork appended a trailing slash and produced a doubled separator.

Rather than match one fixed layout, take the overlay's own upperdir and walk up three levels to strip the rw, image and boot components, then collapse any duplicate separators. The result is checked for a boot directory, falling back to the legacy mount points that live-boot's compatibility bind mount still provides on bookworm. This handles Debian's path, upstream's optional per device sub-directory and the fork's original layout alike, so it stays correct whichever live-boot the image was assembled with.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5475

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1288

All smoketests passed locally

## Checklist:

### DIFF
--- a/src/opt/vyatta/sbin/vyos-persistpath
+++ b/src/opt/vyatta/sbin/vyos-persistpath
@@ -3,15 +3,38 @@
 if grep -q -e '^overlay.*/filesystem.squashfs' /proc/mounts; then
 	# Live CD boot
 	exit 2
-	
-elif grep -q 'upperdir=/live/persistence/' /proc/mounts && egrep -q 'overlay / overlay ' /proc/mounts; then
+
+elif egrep -q 'overlay / overlay ' /proc/mounts; then
 	# union boot
-	
-	boot_device=`grep -o 'upperdir=/live/persistence/[^/]*/boot' /proc/mounts | cut -d / -f 4`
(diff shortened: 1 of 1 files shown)


jestabro (approved):
The difference in mount points is one initial discrepancy between the VyOS fork of live-boot and upstream: this is a smart way to address the 'deforking', with a mount-point tolerant definition. Since CI integration tests were confused when building against the companion PR, I built and checked minimal tests/functionality both with and without
https://github.com/vyos/vyos-build/pull/1288

PR #5444, opened 2026-09-04
dhcp: T7130: add custom DHCP option 43 support

apmillett:
## Summary
Adds generic DHCPv4 vendor encapsulated option support under:

`service dhcp-server ... option vendor-option custom-option <name>`

This allows DHCP option 43 sub-options such as Cisco WLC suboption 241.

## Example
```bash
set service dhcp-server shared-network-name MGMT option vendor-option custom-option cisco-wlc code 241
set service dhcp-server shared-network-name MGMT option vendor-option custom-option cisco-wlc type ipv4-address
set service dhcp-server shared-network-name MGMT option vendor-option custom-option cisco-wlc data 192.0.2.10
```
## Testing
Built a custom rolling ISO from this branch and installed it on a VyOS test system.

Verified DHCP config commits successfully.

Verified Kea renders:
```json
{
  "name": "cisco-wlc",
  "code": 241,
  "type": "ipv4-address",
  "space": "vendor-encapsulated-options-space"
}
```

Verified DHCP option 43 is sent on the wire to a Cisco AP:

```Vendor-Option (43), length 6: 241.4.192.0.2.10```

The AP receives the option as hex bytes on the wire.

Fixes: T7130

### DIFF
--- a/data/templates/dhcp-server/kea-dhcp4.conf.j2
+++ b/data/templates/dhcp-server/kea-dhcp4.conf.j2
@@ -39,13 +39,14 @@
                 "name": "wpad-url",
                 "code": 252,
                 "type": "string"
-            },
-            {
-                "name": "ubnt",
-                "code": 1,
-                "type": "ipv4-address",
-                "space": "vendor-encapsulated-options-space"
             }
+{% if shared_network_name is vyos_defined %}
+{%     set vendor_option_defs = shared_network_name | kea_vendor_option_defs %}
+{%     if vendor_option_defs != '[]' %}
+            ,
+{{ vendor_option_defs[1:-1] }}
+{%     endif %}
+{% endif %}
         ],
 {% if dynamic_dns_update is vyos_defined %}
         "dhcp-ddns": {
--- a/interface-definitions/include/dhcp/option-v4.xml.i
+++ b/interface-definitions/include/dhcp/option-v4.xml.i
@@ -237,6 +237,70 @@
         <help>Vendor Specific Options</help>
       </properties>
       <children>
+        <tagNode name="custom-option">
+          <properties>
+            <help>Custom vendor encapsulated option</help>
+            <constraint>
+              #include <include/constraint/alpha-numeric-hyphen-underscore-dot.xml.i>
+            </constraint>
+            <constraintErrorMessage>Custom vendor option name may only contain letters, numbers, dots, underscores, and hyphens</constraintErrorMessage>
+          </properties>
+          <children>
+            <leafNode name="array">
+              <properties>
+                <help>Encode multiple data values in one vendor option</help>
+                <valueless/>
+              </properties>
+            </leafNode>
+            <leafNode name="code">
+              <properties>
+                <help>Vendor encapsulated sub-option code</help>
+                <valueHelp>
+                  <format>1-254</format>
+                  <description>Vendor encapsulated sub-option code</description>
+                </valueHelp>
+                <constraint>
+                  <validator name="numeric" argument="--range 1-254"/>
+                </constraint>
+                <constraintErrorMessage>Vendor encapsulated sub-option code must be in range 1 to 254</constraintErrorMessage>
+              </properties>
+            </leafNode>
+            <leafNode name="data">
+              <properties>
+                <help>Vendor encapsulated sub-option data</help>
+                <valueHelp>
+                  <format>value</format>
+                  <description>Data formatted according to the selected type</description>
+                </valueHelp>
+                <multi/>
+              </properties>
+            </leafNode>
+            <leafNode name="type">
+              <properties>
+                <help>Vendor encapsulated sub-option data type</help>
+                <completionHelp>
+                  <list>binary boolean fqdn ipv4-address ipv6-address string uint8 uint16 uint32 int8 int16 int32</list>
+                </completionHelp>
+                <valueHelp>
+                  <format>binary</format>
+                  <description>Raw hexadecimal data</description>
+                </valueHelp>
+                <valueHelp>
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by apmillett** [comment]

**Comment by apmillett** [comment]
CLA is complete and the CodeRabbit binary vendor-option array finding has been addressed in 79444030b. I have also tested this in a custom rolling ISO on live VyOS, confirmed the configuration commits cleanly, and confirmed DHCP option 43 is sent on the wire for a Cisco AP/WLC use case.

**Comment by indrajitr** [comment]
1. Having this side by side with vendor option `ubiquiti` looks inconsistent IMHO
```
set service dhcp-server shared-network-name MGMT option vendor-option ubiquiti unifi-controller ...
set service dhcp-server shared-network-name MGMT option vendor-option custom-option cisco-wlc ....
```
Besides, `vendor-option` *is* a custom option. So `vendor-option custom-option` feels superfluous :)

2. Cisco WLC always has to be option 241 and takes multiple IPv4 addresses I think (need to verify). So allowing user to configure a code and type should probably be avoided.

Could we just use something like?
```
set service dhcp-server shared-network-name MGMT option vendor-option cisco cisco-wlc ....
```

NB: Unlike `unifi-controller` which takes single IPv4 address, `cisco-wlc` takes multiple IPv4 addresses. So we might need `<multi/>` in interface definition and `"array": true` in the `"option-def"` config of `wlc` in `kea-dhcp4.conf`.

**Comment by apmillett** [comment]
Thanks, that makes sense.

My intent with `vendor-option custom-option` was to keep this aligned with T7130’s “irrespective of vendor” wording, so it could support Cisco WLC and other vendor encapsulated sub-options without adding a dedicated CLI node for each one.

That said, I agree the current syntax is a bit clunky beside the existing `ubiquiti unifi-controller` node, and for Cisco WLC specifically the code/type should not need to be user-provided if the expected encoding is always sub-option 241 with IPv4 address array data.

I can rework this to add a dedicated Cisco syntax, for example:

```text
set service dhcp-server shared-network-name MGMT option vendor-option cisco cisco-wlc <address>
```

with `<multi/>`, fixed code 241, type `ipv4-address`, and `"array": true` in the generated Kea option-def.

Would you prefer that dedicated Cisco implementation instead of the generic `custom-option`, or should the generic path remain as well for other vendor encapsulated option 43 use cases?

indrajitr:
This probably requires guidance from the core members: @c-po, @dmbaturin, @sarthurdev can weigh in.

sarthurdev:
Firstly the smoketest is failing:
```

DEBUG - ERROR: test_dhcp_vendor_option_custom (__main__.TestServiceDHCPServer.test_dhcp_vendor_option_custom)
DEBUG -   File "/usr/libexec/vyos/tests/smoke/cli/test_service_dhcp-server.py", line 464, in test_dhcp_vendor_option_custom
DEBUG -     self.cli_commit()
DEBUG -   File "/usr/libexec/vyos/tests/smoke/cli/base_vyostest_shim.py", line 132, in cli_commit
DEBUG -     return self._session.commit()
DEBUG -            ^^^^^^^^^^^^^^^^^^^^^^
DEBUG -   File "/usr/lib/python3/dist-packages/vyos/configsession.py", line 353, in commit
DEBUG -     out = self.__run_command([COMMIT])
DEBUG -           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
DEBUG -   File "/usr/lib/python3/dist-packages/vyos/configsession.py", line 262, in __run_command
DEBUG - vyos.configsession.ConfigSessionError: [ service dhcp-server ]
DEBUG - Unexpected error with Kea configuration:
DEBUG - array of binary values is not a valid option definition. (/run/kea/kea-
DEBUG - dhcp4.conf:29:13)
DEBUG - [[service dhcp-server]] failed
DEBUG - Commit failed
```

This is a needed feature for sure, but I would also say its needed in other scopes too. I thought about how to implement this once we moved to Kea and think it may be best under the root CLI node and referenced in various places. Example of how it could potentially be implemented:

Definition:
```
set service dhcp-server custom-option EXAMPLE code 254
set service dhcp-server custom-option EXAMPLE type ipv4-address
```

Uses:
```
set service dhcp-server shared-network-name TEST custom-option EXAMPLE value 127.0.0.1
set service dhcp-server shared-network-name TEST subnet 192.0.2.0/24 custom-option EXAMPLE value 192.0.2.1
set service dhcp-server shared-network-name TEST vendor-option EXAMPLE value 192.0.2.10
```

We could then migrate/remove the Ubiquiti-specific option, using the migrator to create a custom-option for configs where its in use.

apmillett:
Thanks, understood.

The binary-array failure is on me. I added that case after CodeRabbit’s earlier comment, but Kea does not allow `array: true` for binary option definitions, so I’ll remove/rework that part.

I’m happy to rework this toward the root `service dhcp-server custom-option` definition model you suggested, with per-scope references for normal custom options and vendor encapsulated option 43 usage.

For this PR, would you prefer I keep the scope limited to:

1. add the root custom-option definition support,
2. add vendor-option references for option 43 use,
3. preserve/migrate the existing Ubiquiti syntax,

or should I also add shared-network/subnet custom-option usage in the same PR?

dmbaturin:
I'm not a fan of hardcoding vendor-specific options in the CLI syntax. If we keep doing that, we'll end up with a lot of them... and it's also not something those vendors are doing.

I think `vendor-option  ubiquiti unifi-controller` was fair for its day when we didn't anticipate how it might grow. Now I'd be inclined to extend the general syntax for custom options so that any user who migrates from any other vendor can replicate them. Then we could migrate the legacy Ubiquiti-specific option to that general syntax.

PR #5446, opened 2026-09-04
static: T9278: reconcile FRR config after every DHCP lease event

c-po:
## Change summary

The default route derived from `interfaces <type> <ifname> address dhcp` is rendered from the DHCP lease file, so it is only known at lease time. Until now the sole runtime path into FRRs `staticd` was the one-shot `vtysh` injection in `dhclient-enter-hooks.d/03-vyos-ipwrapper`. That injection races the FRR reload of the very commit which started the DHCP client: `dhclient` runs with `-nw`, thus the commit does not wait for a lease, and when BOUND arrives mid-reload this can report FRR as down. The route is then installed into the kernel only and FRR never learns about it. Nothing recovers afterwards, as `FRRender.generate()` short-circuits on an unchanged configuration dict.

The self-healing re-render used by `protocols static route <prefix> dhcp-interface` was gated on `/tmp/static_dhcp_interfaces`, which never lists plain `address dhcp` interfaces - `protocols_static.py` does not even run on an interface-only commit, as no config-mode dependency points to it.

Derive the DHCP dependent interface list from the configuration dict itself, both for the default VRF and for every named VRF, and use it for FRR change detection. Drop the interface list gate in the dhclient exit hook so any lease event requests a re-render.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9278

Included in integrated smoketest suite. All passed with flying colors.

## Checklist:

### DIFF
--- a/data/templates/dhcp-client/dhcp6c-script.j2
+++ b/data/templates/dhcp-client/dhcp6c-script.j2
@@ -2,6 +2,34 @@
 # Update DNS information for DHCPv6 clients
 # should be used only if vyos-hostsd is running
 
+# This script is executed standalone by dhcp6c and can not source the dhclient
+# hook /etc/dhcp/dhclient-enter-hooks.d/01-vyos-logging, which is only pulled
+# into the dhclient-script shell. It therefore carries its own logmsg() - keep
+# the tag, the priorities and the debug flag file in sync with that file.
+LOG_TAG="dhclient-script-vyos"
+LOG_DEBUG_ENABLE="/tmp/vyos.dhclient.debug"
+
+logmsg () {
+    log_level=$1
+    shift
+
+    case ${log_level} in
+        debug)
+            # Do not even build the message when debugging is not requested
+            [ -f "${LOG_DEBUG_ENABLE}" ] || return 0
+            log_prio="daemon.debug" ;;
+        info)  log_prio="daemon.info" ;;
+        warn)  log_prio="daemon.warning" ;;
+        error) log_prio="daemon.err" ;;
+        *)
+            # Unknown level - do not lose the message, it was never a level
+            log_prio="daemon.info"
+            set -- "${log_level}" "$@" ;;
+    esac
+
+    /usr/bin/logger -e --id=$$ -p ${log_prio} -t ${LOG_TAG} "$@"
+}
+
 if /usr/bin/systemctl -q is-active vyos-hostsd; then
     hostsd_client="/usr/bin/vyos-hostsd-client"
     hostsd_changes=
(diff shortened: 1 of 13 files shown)


src/init/vyos-router
```diff
@@ -590,6 +590,7 @@ start ()
         touch /tmp/vyos.ifconfig.debug
         touch /tmp/vyos.container.debug
         touch /tmp/vyos.smoketest.debug
+        touch /tmp/vyos.dhclient.debug
```
dmbaturin, line 593: I wonder if we should be breeding these per-script files. The init script creates them from the same source of truth—the [`vyos-debug` kernel command line parameter](https://github.com/c-po/vyos-1x/blob/e091ee7220efa89d3e9566db12279a771fba17a2/src/init/vyos-router#L585-L594).

I believe these scripts could just check `/proc/cmdline` directly, since a) if we introduce a debug file for every script where it may be useful, which is really everywhere, it will grow out of proportion, and b) if someone is debugging a single script and doesn't want to reboot the system with `vyos-debug` in the kernel command line, they are almost certainly also changing the script to find a fix, so they can as well add `export debug=1` by hand, it's no harder than to create a debug file.

data/templates/dhcp-client/dhcp6c-script.j2
```diff
@@ -2,6 +2,34 @@
...
+        info)  log_prio="daemon.info" ;;
+        warn)  log_prio="daemon.warning" ;;
+        error) log_prio="daemon.err" ;;
+        *)
+            # Unknown level - do not lose the message, it was never a level
+            log_prio="daemon.info"
+            set -- "${log_level}" "$@" ;;
+    esac
+
+    /usr/bin/logger -e --id=$$ -p ${log_prio} -t ${LOG_TAG} "$@"
```
dmbaturin, line 30: Is there a reason to use absolute paths? It seems to happen a lot in this script, but the way it's called, the shell should have no issues looking anything up in the system `$PATH`, as far as I can see.

dmbaturin (commented):
I see no issues with the substance of the PR so far but I left a couple of comments about tangential concerns here.

data/templates/dhcp-client/dhcp6c-script.j2
c-po: For base paths like /usr/bin this is true. I just kept the style, but could refactor all of them if needed.

src/init/vyos-router
c-po: Agree, but this should be a separate task then. Or even enable debugging via CLI?

natali-rs1985 (approved):
Fixes the stale gateway on every DHCP lease event. Smoketests passed.

jestabro (approved):
The issue is clearly described and the fix comprehensive; reviewed and sanity checked. As @dmbaturin requested changes will likely be the matter of separate PRs, approve. The failing smoketest is independent of this PR.

src/init/vyos-router
dmbaturin: Yes, I believe so. Let's make it a separate task, we can discuss the optimal design later.

dmbaturin (approved):
The bug is significant so I think my minor concerns about unrelated implementation details shouldn't hold the fix back.

PR #5451, opened 2026-09-07
configquery: T8409: resolve op-mode config paths independent of edit level

c-po:
## Change summary

`op_mode_config_dict()` rewrote the absolute path passed in by the caller into one relative to `VYATTA_EDIT_LEVEL`. `list_strip()` returns an empty list as soon as a path component mismatches, so from e.g. `[edit qos policy]` the path `qos interface eth1` became the current edit level itself and `cli-shell-api` dumped an unrelated subtree:

```
[edit qos policy]
vyos@vyos# run show qos shaper interface eth1
WARNING: Specified configuration path is not valid
QoS is not applied to eth1!
```

Relative lookups are unreliable even when the edit level is a proper prefix of the path, as `cli-shell-api` validates the path against `VYATTA_TEMPLATE_LEVEL` and rejects a relative tag node value.

Keep the absolute path and reset both level variables for the child process instead.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8409

## Checklist:

### DIFF
--- a/python/vyos/configquery.py
+++ b/python/vyos/configquery.py
@@ -39,7 +39,6 @@ from vyos.xml_ref import is_tag
 from vyos.base import Warning
 from vyos.utils.backend import vyconf_backend
 from vyos.configsource import ConfigSourceVyconfSession
-from vyos.utils.list import list_strip
 
 config_file = os.path.join(directories['config'], 'config.boot')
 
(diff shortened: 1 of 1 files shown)


jestabro (approved):
This is a proper fix. The original modification for edit-level awareness was a quick fix, which does not work in general, as demonstrated.

PR #5455, opened 2026-09-09
system: T9291: vyos-net-name-resolve: keep no-hw-id interfaces

fabrizziop:
This is a vibe-AI-slop PR with a proposed fix for [T9291](https://vyos.dev/T9291)

A config.boot with 'interfaces ethernet ethN { address ... }' and no
'hw-id' leaf (the normal pre-hw-id steady state) is misclassified by
the [T3871](https://vyos.dev/T3871) naming pass: both nodes are reported as 'pending' by
get_pending_hwid_nodes(), while the NICs already sitting at those node
names are treated as unconfigured bootstrap candidates. The node names
get reserved as pending, the NICs are then handed fresh bootstrap
slots (eth0/eth1 -> eth2/eth3), and the running config - which still
references eth0/eth1 - matches nothing: complete loss of connectivity
on the 2026.06 -> 2026.09 rolling upgrade, plus the spurious
'config error on boot' and 'still has no hw-id' warnings.

Pending-node matching was designed for the 'delete interfaces
ethernet ethN hw-id' NIC-replacement remediation, where a node lost a
hw-id and its NIC must be re-identified. A node that never had a
hw-id is different: with no hw-id to identify the NIC, the node is
name-anchored - the NIC sitting at the node's name IS the configured
interface, and it only needs its hw-id frozen in by the later
vyos-interface-rescan.py pass, not a rename.

- unmatched_candidates(): exclude candidates whose current name is a
  no-hw-id configured node from the bootstrap/reclaim candidate pool
- main(): derive the set of such 'satisfied' nodes from the settled
  snapshot, exclude them from the settle-gate and from the spurious
  'still has no hw-id' warning
- match_pending_nodes(): take excluded satisfied nodes out of the
  matching so they neither consume a candidate nor add to the
  cardinality that makes a genuine match ambiguous
- write_status(): exclude satisfied nodes from pending_unresolved and
  surface them in a new 'satisfied' key (vyos-router only reads
  '.missing' and '.pending_unresolved', both unchanged in shape)

Regression tests reproduce the reported scenario (two no-hw-id nodes
keeping eth0/eth1, hints frozen in place), a satisfied node plus an
unconfigured extra NIC (fills the next free slot, no interference),
and a NIC that only settles at the node's name (still reclaimed).

All 94 tests in test_net_name_resolve pass (4 pre-existing
TestGetPendingHwidNodes errors require the built libvyosconfig.so.0
and fail identically on the unmodified tree).

## Change summary
Fixes the interface renaming bug seen with interfaces with no hw-id configured

## Types of changes

## Related Task(s)

[T9291](https://vyos.dev/T9291)

Tested it on the same instance I found the bug in.

## Checklist:

### DIFF
--- a/src/system/vyos-net-name-resolve.py
+++ b/src/system/vyos-net-name-resolve.py
@@ -360,7 +360,8 @@ def compute_rename_plan(configured: dict, current: dict, pending: dict = None) -
     return plan
 
 
-def unmatched_candidates(configured: dict, current: dict, existing_plan: dict) -> list:
+def unmatched_candidates(configured: dict, current: dict, existing_plan: dict,
+                          pending: dict = None) -> list:
     """Physical interfaces this boot with no configured hw-id - the pool
     both ordinary bootstrap naming and pending-node reclaim matching draw
     from. This deliberately INCLUDES squatters compute_rename_plan() is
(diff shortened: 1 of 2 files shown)


fabrizziop:

fabrizzio-ai:

PR #5437, opened 2026-09-01
T9269: fix issues after recent GRUB serial console rewrites to support VyOS in container

c-po:
## Change summary

VyOS boots fine inside a container, but several subsystems assume they own the hardware and the Kernel they run on.

`show version` described the host rather than the node - system type, boot type, Secure Boot state and the hardware vendor, model, serial and UUID were all derived from DMI, CPUID or the Kernel cmdline of the host system. It now reports `container` and `container image`, shows `Secure Boot: n/a (container)` and omits the hardware entries entirely.

Writing the GRUB configuration is skipped for containers, and for any system without a mounted persistence partition, where `disk.find_persistence()` returned an empty string and `system_option.py` aborted with `FileNotFoundError: '//boot/grub/grub.cfg'`. For the same reason the user is no longer asked to save the configuration and reboot to apply Kernel cmdline options a container can never apply.

Finally, the hostname assigned by the container runtime is adopted and set as transient hostname only, as `/etc/hostname` is a read-only bind mount. This fixes `hostnamectl` failing with "Device or resource busy" and `sudo: unable to resolve host`. `sysctl_write()` no longer leaks errors for keys that are not network namespace aware.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes)
- [ ] Migration from an old Vyatta component to vyos-1x, please link to related PR inside obsoleted component
- [ ] Other (please describe):

## Related Task(s)

* https://vyos.dev/T9269

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1286

## How to test / Smoketest result

Containerlab topology

```bash
$ cat vyos-two-node.clab.yml
name: vyos-two-nodes

topology:
  nodes:
    vyos1:
      kind: linux
      image: vyos/vyos:2026.09.01-0914-T9269

    vyos2:
      kind: linux
      image: vyos/vyos:2026.09.01-0914-T9269

  links:
    - endpoints: ["vyos1:eth1", "vyos2:eth1"]
```

```bash
$ sudo containerlab deploy
15:20:07 INFO Containerlab started version=0.79.0
15:20:07 INFO Parsing & checking topology file=vyos-two-node.clab.yml
15:20:07 INFO Creating docker network name=clab IPv4 subnet=172.20.20.0/24 IPv6 subnet=3fff:172:20:20::/64 MTU=0
[...]

PR #5456, opened 2026-09-09
T9269: fix container boot failures and two config diffing regressions they exposed

c-po:
## Change summary

Booting the VyOS container image under containerlab trips four unrelated assumptions about running on real hardware.

The system option script aborted with a traceback, because setting the locale needs a systemd D-Bus service no container provides, and the udev rule reload right after it fails for the same reason. Both are now skipped in a container.

The serial console migrator crashed on a GRUB variable that is absent whenever the image flavor carries no console data, so it tests for the key first.

The interface naming pass warned about an unbindable hw-id on every boot. Container interfaces are veth pairs whose MAC the host regenerates on each start, so the pass is skipped, which also drops two bounded hardware waits from boot.

Finally two config diffing bugs, both reachable outside containers, made a bare ethernet node look like it had an empty option configured, and reported an existing bond member as removed when a second one was added.

## Types of changes

## Related PR(s)

* https://github.com/vyos/vyos-1x/pull/5437

* `make test-oci`

## Checklist:

### DIFF
--- a/python/vyos/utils/dict.py
+++ b/python/vyos/utils/dict.py
@@ -315,7 +315,12 @@ def dict_to_key_paths(d: dict) -> list:
     """
     def func(d, path):
         if isinstance(d, dict):
-            if not d:
+            # An empty dict is a valueless leaf node and yields the path that
+            # leads to it - but only if there is one. An empty dict at the top
+            # level means "no configuration at all", which must yield nothing
+            # instead of a single empty path: consumers join the components
+            # into a dotted option name, and '' is not an option
+            if not d and path:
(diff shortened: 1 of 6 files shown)


natali-rs1985 (approved):
Fix container boot by skipping hardware-only setup steps, plus a config bug where empty options broke bonds. Smoketests related to the change have passed. Haven't tested manually.

sever-sever (approved):
Fixes for VyOS container image

dmbaturin:
All CI failures appear unrelated to the PR contents.

PR #5459, opened 2026-09-10
ipoe-server: T9289: expose accel-ppp lease-time DHCP options

ruben-herold:
## Change summary
- Expose accel-ppp's `[ipoe]` DHCP lease-timing options that were previously only reachable by hand-editing the generated config:
  - `set service ipoe-server lease-time`
  - `set service ipoe-server max-lease-time`
  - `set service ipoe-server renew-time`
  - `set service ipoe-server rebind-time`
- accel-ppp's ipoe module already parses `lease-time`, `max-lease-time`, `renew-time` and `rebind-time` in `accel-pppd/ctrl/ipoe/ipoe.c`; without CLI exposure operators are stuck with the hardcoded 600s default lease and its derived renew/rebind times.
- All four are optional integer (seconds) leafNodes; when unset, accel-ppp falls back to its built-in defaults exactly as before.
- Constrains the values to accel-ppp's signed int32 range and warns when configured lease-timing values would be overridden by accel-ppp itself.

## Types of changes

## Related Task(s)
* https://vyos.dev/T9289

## Related PR(s)
* vyos/vyos-documentation#2240

Added `smoketest/scripts/cli/test_service_ipoe-server.py::test_ipoe_server_lease_time`, verifying default (no options emitted) and explicit values are rendered correctly into `ipoe.conf`.

## Checklist:


c-po (changes requested):
Please use the original PR template to ease the review process for maintainers

ruben-herold:
@c-po Done — updated the PR description to use the full official template.

src/conf_mode/service_ipoe-server.py
```diff
@@ -110,6 +111,43 @@ def verify(ipoe):
...
 
+    # accel-ppp defaults "lease-time" to 600s and silently recalculates
+    # "renew-time"/"rebind-time" from it (and ignores a "max-lease-time"
+    # lower than it) instead of rejecting the configured value - warn so
+    # the effective behavior isn't a surprise.
+    lease_time = int(dict_search('lease_time', ipoe) or 600)
+    max_lease_time = dict_search('max_lease_time', ipoe)
+    renew_time = dict_search('renew_time', ipoe)
+    rebind_time = dict_search('rebind_time', ipoe)
+
```
c-po, line 122: `def dict_search(path, dict_object, default=None):` returns None if a key is not found and `default` is not defined as something else, so the if clause can be simplified to: 

`if max_lease_time and int(max_lease_time) < lease_time:` which also helps in line 141 to keep the conditional statement on one line.

c-po (changes requested):

alexk37 (changes requested):
@ruben-herold The new lease-timing validation in `verify()` misses a case where accel-ppp overrides the configured `renew-time`. With an otherwise valid IPoE configuration:

```
set service ipoe-server renew-time 550
commit
```

The new verifier emits no timing warning. With `lease-time` and `rebind-time` unset, accel-ppp defaults them to 600 and 525 seconds, then recalculates renewal from 550 to 300 seconds.

Please account for effective defaults when checking whether a configured value will be overridden.

ruben-herold:
@alexk37 Good catch, and thanks for doing the arithmetic — your numbers are exactly right.

Root cause confirmed: the timing validation only compared values that were explicitly configured. But accel-ppp's last step compares `renew-time` against `rebind-time`, and that one is a *default* here — it appears nowhere in our configuration. So 550 passes the lease-time check (550 <= 600), then trips `renew > rebind` against the default 525 and gets pulled to `rebind * 4 / 7` = 300.

Fixed in c422e5dc: `verify()` now mirrors the calculation in accel-ppp's `ipoe.c:load_config()` (lines 3910-3934) and compares configured values against the effective ones:

- `renew-time` unset **or** greater than the lease time -> `lease / 2`
- `rebind-time` unset **or** greater than the lease time -> `lease/2 + lease/4 + lease/8` (7/8, term by term, integer division)
- then, if `renew > rebind` -> `renew = rebind * 4 / 7`
- `max-lease-time` unset -> `lease + lease/10`, and it only has an effect when strictly greater than the lease time

This replaces the three separate "greater than" checks — each of them is now covered by the single effective-value comparison — and the warning names the value that will actually be used instead of just stating that a recalculation happens:

```
WARNING: "renew-time" (550) will be overridden by accel-ppp with 300,
derived from "lease-time" 600 and "rebind-time" 525
```

I cross-checked the Python model against the C logic over 8000 input combinations (lease x renew x rebind, including unset, 1, 2, 7, 8, the 299/300/301, 524/525/526 and 599/600/601 boundaries, and the u32 maximum): no mismatches.

New smoketest `test_ipoe_server_lease_time_override_warnings` covers your case plus `rebind-time` above the lease time, `max-lease-time` below the lease time, and a consistent set that must stay silent. It asserts on the `cli_commit()` return value after collapsing whitespace, since `vyos.base.Warning` wraps at 72 characters.

One deliberate limitation: only *configured* values are reported. If you set just `rebind-time 100`, your value is honoured and the default renewal is quietly recalculated to 57 — no warning, because you never configured a renewal. Happy to widen that if you think it should be reported too.

@c-po your `dict_search()` simplification went in as f20bb189, in case you want to resolve that thread.

Could you re-test the upgrade path once more?

ruben-herold:
@c-po short status on this one, since two of your review points have been addressed since your last look:

- Your `dict_search()` simplification went in as `f20bb189` — `dict_search()` returns `None` when the key is missing, so the checks are now plain truthiness tests and the conditionals fit on one line. Feel free to resolve that thread.
- The PR description has been on the official template since `09-11`, per your earlier review.

On top of that, @alexk37 found a real gap in the timing validation today, fixed in `c422e5dc`: the checks only compared values that were explicitly configured, but accel-ppp compares `renew-time` against `rebind-time` as its last step, and that one is usually a *default* which appears nowhere in our configuration. `verify()` now mirrors `ipoe.c:load_config()` and compares configured values against the effective ones. Details and the cross-check are in the comment above.

A few notes on the CI signal:

- Nothing has run yet on the current head `c422e5dc`. All four `pull_request`-triggered workflows (ISO Integration Test, Typos, Darker + Ruff, CodeQL) sit at `action_required`, waiting for a maintainer to approve the workflow run for this fork PR — only the `pull_request_target` ones (CLA, conflict checker) went through. Could you hit "Approve and run workflows"? (#5436, #5463, #5468, #5476 and #5455 are in the same state, so it is not specific to this PR.)
- When it does run, `typos` will most likely still fail: it flags `localed` inside `vyatta_localedir` in `src/etc/default/vyatta`, a file this PR does not touch. Pre-existing in the base branch and not fixable from here — it is currently failing on every vyos-1x PR.
- The failing smoketests on the previous run were `test_techsupport_archive.py` only, unrelated to IPoE. Your `tech-archive-fix` (`bd51819b0`) is in the base now, so I expect that to clear.

Could you take another look when you have a moment?

src/conf_mode/service_ipoe-server.py
ruben-herold: Done — thanks, you were right on both counts.

Dropped the `is not None` comparisons in `84f707fb` ("ipoe-server: T9289: simplify lease-timing conditionals"); the renew-time vs rebind-time check fits on one line again:

```python
if max_lease_time and int(max_lease_time) < lease_time:
if renew_time and rebind_time and int(renew_time) > int(rebind_time):
```

No functional change: the values reaching `verify()` are non-empty config strings, so `'0'` stays truthy and every branch evaluates exactly as before.

`verify()` has since been reworked again for @alexk37's finding — accel-ppp silently recalculates the timings against *its own* defaults for every option left unset, so comparing only against configured values missed cases. It now mirrors `ipoe.c:load_config()` and warns with the effective values, keeping the plain truthiness style throughout.

ruben-herold:
Rebased onto current rolling (`5b0cce51f`), head is now `fb5fbdbe8`. Contents of the six commits are unchanged — `git diff c422e5dc fb5fbdbe8` over the touched files is empty.

Reason for the rebase: the failing smoketest in the last run was **not** related to this PR. 93 of 94 testcases passed, the only failure was `test_techsupport_archive.py` — fixed by @c-po's own #5462 (`tech-support: T5475: do not archive nested mount points below /run`, merged 2026-09-14), which was not yet an ancestor of the previous head. It is included now.

Heads up on the remaining red check: `typos` fails on **every** PR currently based on rolling, with a single false positive that comes from rolling itself, not from this diff:

```
error: `localed` should be `located`
```

Source is `src/conf_mode/system_option.py:685` (comment mentioning the real daemon name `systemd-localed`), introduced in `ae5df0f6d` — "system-option: T9269: do not call localectl(1) and udevadm(8) in a container". Same failure reproduces on #5477 and #5398. Probably wants a `localed` entry in `.github-central/_typos.toml`; happy to open that as a separate PR if a T-number is assigned.

@c-po @alexk37 — could the gated workflows be approved so the smoketests can re-run?

ruben-herold:
Correction to my previous comment — the `typos` part of it was wrong, sorry.

The `localed` false positive is **already fixed**: @ordex filed T9323 and merged `406305c1e` ("typos: T9323: allow \"localed\" in the spell check", #5480) on 2026-09-15, adding the entry to the repo-local `.typos.toml`. It is in rolling as of `5b0cce51f`.

I had assumed the allowlist entry would have to go into the central `_typos.toml` in `vyos/.github@production`, since the workflow passes `config: .github-central/_typos.toml` explicitly. That assumption was wrong — crate-ci/typos *merges* the discovered repo-local `.typos.toml` with the `--config` file rather than replacing it. Verified with typos v1.47.2 against rolling `5b0cce51f` plus the real central config: exit 0 as-is, and 59 errors (`localed`, `iif`, `oif`, `Widgits`, …) once `.typos.toml` is moved out of the way.

So no config PR is needed, and this branch already carries the fix via the rebase. The rest of the previous comment stands: the earlier smoketest failure was `test_techsupport_archive.py`, fixed by #5462 and now included.

src/conf_mode/service_ipoe-server.py
```diff
@@ -110,6 +111,55 @@ def verify(ipoe):
     verify_accel_ppp_name_servers(ipoe)
     verify_accel_ppp_wins_servers(ipoe)
 
+    # accel-ppp never rejects inconsistent DHCP lease timings, it silently
+    # recalculates them - and it does so against its own defaults for every
+    # option left unset, so a configured value can be overridden even when
+    # nothing it is compared against appears in our configuration. Mirror
+    # that calculation from accel-ppp's ipoe.c:load_config() and report the
+    # values that will not survive it.
+    lease_time = int(dict_search('lease_time', ipoe) or 600)
```
c-po: This is invalid. Python code is not allowed to carry implicit defaults.

THis must be handled at the CLI XML definition level where there is a `<defaultValue>` node which must be used to feed in default CLI values. This also helps keep documentation up2date and CLI completion helpers.

c-po (changes requested):

PR #5460, opened 2026-09-10
ethernet: T9228: only warn about unsupported NIC features when node changed

c-po:
## Change summary

`EthernetIf.update()` re-applies every ethernet knob on every commit touching the interface. The capability messages (offload, flow-control, ring-buffer, coalesce, switchdev) are emitted whenever the CLI-desired state differs from what the driver reports - and on drivers where a feature is fixed (`virtio_net`, `vmxnet3`, `vif`, `veth`, ...) that is permanently true. The message is correct the first time, but it then re-appears on unrelated commits like an interface description or IP address change.

Record in interfaces_ethernet.py which of the affected CLI nodes were altered - the mechanism already used for speed_duplex_changed - and pass it to the setters via a "warn" argument. Settings are still applied unconditionally, only the message is gated.

While at it, consolidate the seven offload setters onto a common helper and use `Warning()` instead of `print()`.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9228

## Checklist:

### DIFF
--- a/python/vyos/ifconfig/ethernet.py
+++ b/python/vyos/ifconfig/ethernet.py
@@ -149,11 +149,12 @@ class EthernetIf(Interface):
 
         super().remove()
 
-    def set_flow_control(self, enable):
+    def set_flow_control(self, enable, warn=True):
         """
         Changes the pause parameters of the specified Ethernet device.
 
         @param enable: true -> enable pause frames, false -> disable pause frames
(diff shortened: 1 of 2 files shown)


natali-rs1985 (approved):
The "NIC doesn't support this setting" warning now appears only when that setting is actually changed, instead of on every unrelated commit to the interface

PR #5470, opened 2026-09-14
smoketest: T9315: give up once the system under test stops answering

c-po:
## Change summary

A single test hitting the 20 minute timeout may just be one slow test, so the runner carries on - which is right. But when the system itself stops responding, every remaining test also sits out its full timeout before reporting a result that says nothing about it.

Seen in practice: one test wedged the box on the fourth of eighteen files, and the run spent another fourteen timeouts (close to five hours) to learn nothing.

Abandon the run after two timeouts in a row, and list the tests that never got to run so the report does not read as though they passed. A test that merely fails still resets the counter; only silence is treated as the system being gone.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9315

## Checklist:


dmbaturin (approved):
I agree with the idea, it's better to fail the run early if the system is outright unresponsive. Coderabbit suggestion about possible unreaped children seems insignificant for me since all child processes will terminate with the OS anyway.

PR #5477, opened 2026-09-15
utils: T9322: validate the address is_intf_addr_assigned() is given

ruben-herold:
## Change summary

`is_intf_addr_assigned()` promised a single IPv4/IPv6 address in its docstring but never enforced it. Two wrong answers followed from that, against current `rolling`:

```python
>>> is_intf_addr_assigned('lo', '192.0.2.1-192.0.2.2')
ValueError: '192.0.2.1-192.0.2.2' does not appear to be an IPv4 or IPv6 interface
>>> is_intf_addr_assigned('lo', '::1%lo-::2%lo')
True          # a range reported as assigned
```

The first let an `ipaddress` exception escape to the ~30 call sites of this helper and `is_addr_assigned()`. The second happened because the `%` suffix was split off before the range was ever noticed, truncating it to its first element.

The zone index was the more interesting half. It was stripped but never compared against `ifname`, and since `is_addr_assigned()` calls this function once per interface, the zone was simply discarded:

```python
>>> is_addr_assigned('::1%eno1')
True          # ::1 is on lo, eno1 never entered the comparison
```

This PR rejects anything that is not a single address, which removes the need to handle `%` internally at all — per @c-po's guidance in #5400.

One detail worth flagging: **`is_ip()` on its own is not sufficient.** The `ipaddress` module accepts a zone index since Python 3.9, so `is_ip('fe80::1%eth0')` returns `True`:

```python
>>> from ipaddress import ip_interface
>>> ip_interface('fe80::1%eth0')
IPv6Interface('fe80::1%eth0/128')
```

and `ip --json address show` never reports a zone, so such a value could only ever match by ignoring it. Hence the explicit `'%' in addr` check alongside `is_ip()`.

Also in here:

- The netns invocation is no longer built as a shell string. `popen()` enables a shell whenever the command contains a space, so `ifname` and `netns` were interpolated into a shell command line. `rc_cmd()` has accepted a `netns=` argument all along; passing the command as a list keeps both out of a shell.
- The parsed address is hoisted out of the loop rather than reassigning the `addr` parameter on every iteration.
- The unused `family` key is dropped from the jmespath projection.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9322

## Related PR(s)

* #5400 — superseded by this PR; it only patched the range symptom and is being closed in favour of this one.
* #5266 (T8977) — **touches the same lines.** It rejects invalid input by returning `False` silently, whereas this PR raises, as agreed in #5400. Those two are mutually exclusive on the same code path, so the changes need sequencing. Happy to rebase on top of #5266 or adjust to whichever behaviour you prefer.

No smoketest is needed; the helper is unit-testable. Two cases added to `src/tests/test_utils_network.py`:

```
$ PYTHONPATH=python python3 -m pytest src/tests/test_utils_network.py -k intf_addr -q
..                                                                       [100%]
2 passed, 5 deselected in 0.22s
```

The positive cases run against the real system (`lo` carrying `127.0.0.1/8` and `::1`, plus a non-existent interface), the negative cases cover ranges, a range with a zone index, a plain zone index, a link-local with a zone index, garbage and the empty string.

Note on scope: I could not verify locally whether `ipaddrcheck --is-any-single` — the validator behind CLI nodes such as the BGP `neighbor` tagNode — accepts a zone index, since `ipaddrcheck` is not installed on my machine. If it does accept one, then a configuration like `set protocols bgp neighbor fe80::1%eth0` would now raise out of `verify()` instead of silently matching the wrong interface. A loud failure seems clearly preferable to a wrong answer, but please say if you would rather see that path return `False`.

## Checklist:


python/vyos/utils/network.py
```diff
@@ -487,28 +487,40 @@ def is_addr_assigned(ip_address, vrf=None, return_ifname=False, include_vrf=Fals
...
-                addr = addr.split('%')[0]
-            interface = ip_interface(f"{address}/{prefixlen}")
-            if ip_interface(addr) == interface or address == addr:
-                return True
+    # The ipaddress module accepts a zone index since Python 3.9, so is_ip()
+    # alone would let "fe80::1%eth0" pass. "ip --json address show" never
+    # reports one, so such a value could only ever match by ignoring the zone,
+    # which reported an address as assigned to the wrong interface.
+    if '%' in addr or not is_ip(addr):
+        raise ValueError(f'{addr} is not a valid IPv4 or IPv6 address')
```
MPStudyly: I'd split this into dedicated checks, making the raised errors message differ. As far as I understand zone identifiers are supposed to be stripped at most by the caller of this function and passed in explicitly; Raising a generic "is not a valid address" shadows when a caller actually misbehaves compared to e.g. an actual operator error.

ruben-herold:
@MPStudyly Good point, and you read the intent correctly — fixed in 70d86d3e.

The two cases really are different in kind: the interface to check against is already an argument of this function, so a `%zone` in `addr` means the caller handed over something it should have resolved itself. An unparsable address, on the other hand, is ordinary bad input. They now raise separately:

```python
if '%' in addr:
    raise ValueError(f'{addr} carries a zone index, pass the interface as "ifname"')
if not is_ip(addr):
    raise ValueError(f'{addr} is not a valid IPv4 or IPv6 address')
```

The tests assert the message now, not just the exception type, so the two paths cannot silently swap. Note that `::1%lo-::2%lo` deliberately lands in the zone-index branch rather than the address branch, since `%` is checked first — which is also the more useful message for that input.

On the overlap with #5266, since you are the best person to weigh in: that PR rejects invalid input with a silent `return False`, while this one raises, and both touch the same lines. I have no strong preference between the two behaviours and I am happy to rebase on top of #5266 and adapt, or to hold this until yours lands — whichever keeps your review moving. Just say which you prefer.

While I am here: the four `pull_request` workflows on this PR are all sitting at `action_required` and have not run yet, so there is no CI signal on it. Could a maintainer approve the workflow run? For what it is worth, locally the two new unit tests pass and `ruff` reports the same single pre-existing finding as `rolling` does on the same files.

MPStudyly:
> On the overlap with #5266, since you are the best person to weigh in: that PR rejects invalid input with a silent `return False`, while this one raises, and both touch the same lines. I have no strong preference between the two behaviours and I am happy to rebase on top of #5266 and adapt, or to hold this until yours lands — whichever keeps your review moving. Just say which you prefer.

I'm not sure how much my opinion counts in this regard, I tried to keep the function as close to the original when implementing non_local bind checking. However I think having explicit errors raised when doing invalid stuff is way better behavior than simply rejecting silently.

Regarding merge order, I'd still like to have my changes merged first, but in the end it's up to the maintainers.

ruben-herold:
Rebased onto current rolling (`5b0cce51f`), head is now `4632e5ce7`. Contents unchanged — `git diff 70d86d3e 4632e5ce7` over the touched files is empty.

Sole purpose of the rebase is the one remaining red check: `typos` failed on `` `localed` should be `located` ``, which came from rolling itself, not from this diff, and was fixed by @ordex in `406305c1e` (T9323, #5480). Verified locally with typos v1.47.2 and the central `_typos.toml`: exit 0 on this branch. Everything else was already green on the previous run — build_iso, CLI smoketests, interfaces, config-load, RAID1, TPM, CodeQL, darker-ruff.

@MPStudyly thanks for the clear answer — noted on both points. I'll keep the explicit `raise` and hold this PR for #5266; once yours lands I'll rebase on top of it and adapt whatever overlaps, so there's nothing for you to work around on your side.

src/tests/test_utils_network.py
```diff
@@ -24,6 +24,30 @@ def test_is_addr_assigned(self):
...
+        # a caller-side mistake and reported as such, separately from an
+        # invalid address
+        for addr in ['::1%lo', 'fe80::1%eth0', '::1%lo-::2%lo']:
+            with self.assertRaises(ValueError) as e:
+                vyos.utils.network.is_intf_addr_assigned('lo', addr)
+            self.assertIn('carries a zone index', str(e.exception))
+
     def test_is_ipv6_link_local(self):
         self.assertFalse(vyos.utils.network.is_ipv6_link_local('169.254.0.1'))
         self.assertTrue(vyos.utils.network.is_ipv6_link_local('fe80::'))
```
c-po: What happens to your code if you now feed:

`self.assertTrue(vyos.utils.network.is_ipv6_link_local('fe80::1%eth0'))` ? Add this testcase

c-po (changes requested):

PR #5479, opened 2026-09-16
T3871: bind interface names to hardware slots instead of hw-id

c-po:
## Change summary

Interface names were derived from the hw-id node in config.boot. That coupling made one question unanswerable: a node carrying an address but no hw-id still had to be matched to some card, and nothing available at boot could say which card was genuinely its own. Every heuristic tried for it could be made to put a configured address on the wrong wire.

Naming authority moves out of the configuration and into a mapping store binding a name to the slot the hardware occupies. The slot is the identity, so a replacement card in that slot is that interface again. The store is rendered into systemd link files, leaving udev to apply each name as the device appears. Hardware never seen before is named at boot and recorded, and cards plugged in later are named after the interfaces already known.

The hw-id node is gone. A migration carries existing bindings across by matching each address to the slot it currently sits in, so no interface is renamed on upgrade. An ethernet interface is now treated as owned by the Kernel: deleting its configuration resets the port instead of leaving it stranded admin-down, which also returns released bond members to service.

## Types of changes

## Related Task(s)

* https://vyos.dev/T3871

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1305
* https://github.com/vyos/vyos-documentation/pull/2246
* https://github.com/vyos/vyos-1x/pull/5417

## Checklist:


zdc (commented):
This looks like a truly huge architectural decision and improvement! Great!

I have not read the full code. Therefore, my review is not an approval or rejection. I just want to express a few comments about it that immediately came to mind.

1. What are we getting from storing the hardware key for bindings in a semi-structured format? If we already have a key and value, why not store it as a real key and value instead of a text representation `PROPERTY=value`?
2. Will it be possible, after the change, to pre-configure names per interface if the user wants interface `A` to be named `ethX` and `B` to be named `ethY`, without using `rename` in the CLI? For example, by directly editing files?
3. This also requires changes in `cc_vyos.py` in cloud-init. It currently works with `hw-id`.
4. Interface naming is probably the most environment-sensitive part of the system. If we want to change the architecture, it must be manually tested, at least, with the following combinations:
    * With a bare-metal server, with multiple NICs from different vendors, including multi-port NICs.
    * In virtual environments, with paravirtualized (virtio, vmxnet3), accelerated paravirtualized (vhost-user, VDPA, etc.), NIC VF bypass, and full NIC-to-VM bypass. Including multiple NICs of every type, where available.
    * In clouds, with a combination of multiple NICs of each type supported by the cloud NIC type.
    * With both cold and hot plug.
    * With Cloud-init on each platform, where it is supported. Including pre-configuring interface order by Cloud-init via `network-config` and via `user-config`. Also with PXE + Cloud-init using the same criteria.

c-po:
> 1. What are we getting from storing the hardware key for bindings in a semi-structured format? If we already have a key and value, why not store it as a real key and value instead of a text representation `PROPERTY=value`?

This fully decouples the /config/config.boot file and now it can seamlessly be copied from one machine to another.

> Will it be possible, after the change, to pre-configure names per interface if the user wants interface A to be named ethX and B to be named ethY, without using rename in the CLI? For example, by directly editing files?

By pre-configure you mean first boot?

> In virtual environments, with paravirtualized (virtio, vmxnet3), accelerated paravirtualized (vhost-user, VDPA, etc.), NIC VF bypass, and full NIC-to-VM bypass. Including multiple NICs of every type, where available.

There is a smoketest covering exactly this

> With both cold and hot plug.

There is a smoketest covering exactly this

alexk37:
@c-po 

Deploying qcow2 image and then changing the VM's NICs before its second boot put configured addresses on the wrong interfaces.

On the first boot of the image the mapping store is not written. `vyos-net-name-resolve.service` runs before `/opt/vyatta/etc/config/.vyatta_config` exists, and `config_dir_is_mounted()` reads that marker, so `save_store()` returns False and no `.link` files are rendered:

```
vyos-net-name-resolve.py[2148]: the configuration volume is not mounted - interface names were resolved for this boot but could not be persisted
```

With no store there is no name reservation, so the next boot derives names from the hardware again.
Repro, on an image built from 52e4a2a4b with five NICs on PCI slots 0x02–0x06, one address per interface:

1. Boot the deployed image.
2. Shut the VM down and remove the card in slot 0x04.
3. Boot it again.

`eth2` and `eth3` came up on the next cards along, carrying the addresses configured for the interfaces before them, and `eth4`'s address was not applied, three of the four data subnets stopped passing traffic. A rolling image built before this naming series landed, given the identical removal, kept every surviving interface's name, slot and address; only the removed card's own subnet went down.

alexk37:
A NIC added to a running system is not recorded in the mapping store. The hot-plug rule does not match the device, so `vyos-net-name-resolve-hotplug.service` is not started.

`60-vyos-net-hotplug.rules` line 3:

```
DRIVERS!="?*"
```

`DRIVERS` matches if the event device or any ancestor matches. The net device itself has no driver:

```
/sys/class/net/eth1/driver         absent
/sys/class/net/eth1/device/driver  virtio_net
```

So `DRIVERS!="?*"` matches on the device itself and the rule ends before `ENV{SYSTEMD_WANTS}`.`DRIVERS=="?*"` and `DRIVERS!="?*"` both matched the same device in the same event.

1. Hot-plug a virtio NIC into the running VM.
2. `cat /config/interface-mapping.json`
3. `systemctl show vyos-net-name-resolve-hotplug.service -p ExecMainStartTimestamp -p ConditionTimestamp`

The new card was absent from the store, no `.link` file was written for it and both timestamps were empty, the unit was not started. I saw this in the three event kinds I tried (`udevadm test`, a manual trigger, and a live PCI hot-plug), on both i440fx and q35 and with the rule renumbered to `99-` in case ordering mattered.

The rules this PR removes use the positive form (`DRIVERS=="?*"`); the negated form is the only one in the diff. Switching to the positive form tagged a real NIC while still skipping a driverless dummy interface when I tested it.

alexk37:
Deleting the ethernet configuration of an interface that is still a bond member resets that member's MAC to the card's own address. The bond keeps reporting healthy, but it carried no traffic across a failover until the node was rebooted.

`EthernetIf.remove()` resets the hardware address with no check that the interface has left the bond:

```python
mac = permanent_mac(self.ifname)
if mac:
    self.set_mac(mac)
```

Deleting `interfaces ethernet eth3` does not remove it from the bond, the configuration still lists it as a member and the kernel still shows it as a slave.

Repro with both members on the same segment:

1. `set interfaces ethernet eth3 description x`
2. `set interfaces bonding bond0 mode active-backup`
3. `set interfaces bonding bond0 member interface eth5`
4. `set interfaces bonding bond0 member interface eth3`
5. `set interfaces bonding bond0 primary eth5`
6. `set interfaces bonding bond0 address 10.0.0.1/24`
7. `commit`
8. `delete interfaces ethernet eth3`
9. `commit`

```
bond0 52:54:00:79:bb:00
eth3  52:54:00:79:03:00   (master bond0)
```

Taking the primary down then failed over to `eth3`, and a ping to the peer lost every packet after the switch. On the wire, all of the requests reached the peer and it answered all of them — the replies are addressed to the bond's MAC, and `eth3` now has its own MAC with promiscuous mode off, so they are dropped by the interface. The peer's ARP retries are answered with the bond's MAC, which is on the interface that just went down.

Any ethernet configuration on the member is enough, a description alone reproduced it. A member with no ethernet node at all did not. A later unrelated commit did not restore the address, a reboot did.

alexk37:
A NIC added after a different one was removed takes the freed PCI slot and inherits the removed interface's name and its configured address, so that address ends up on whatever the new card is cabled to.

The problem seemingly is that the same path also covers "a different card, for a different purpose", and the warning written for that case cannot fire.

`src/init/vyos-router:390`:

```
log_warning_msg "interface '${intf}' was matched to new hardware ${hwid} by slot - verify it is the port you expect before relying on its configuration"
```

It reads `.matched_by_path` from `/run/vyos-net-name-resolve.json`. `write_status()` emits `interfaces`, `missing`, `renamed`, `matched`, `new_hardware` and `persisted` - the key this warning depends on is not among them, and the warning did not appear on any boot I tested.

Repro with five interfaces on consecutive PCI slots with one address each:

1. Boot once so the mapping store is written.
2. Remove the NIC belonging to `eth2` (`virsh detach-device --config`).
3. Add a different NIC for another purpose, cabled to another segment, without specifying a PCI address, the hypervisor assigns the slot that was just freed.
4. Boot.

```
eth2  mac=52:54:00:79:e1:00  slot=ens4  addr=10.134.12.1/24
```

`eth2` is the new card, holding the address configured for the segment the old card served and that subnet stopped passing traffic. The configuration still reads `set interfaces ethernet eth2 address '10.134.12.1/24'`. The status file recorded the unknown card under `matched` and the boot logged nothing about naming.

Before this PR the name was bound by `hw-id`, so a different card did not inherit the address.

alexk37:
@c-po

On an image built from 18a9d2187, the wireless interfaces from the smoketest's configuration are renamed about ten seconds after the commit that creates them. `wlan1` became `wlan0`, `wlan10` became `wlan1`, and hostapd for the access point stopped. The configuration still names `wlan1` and `wlan10`.

With the hot-plug rule now matching, loading `mac80211_hwsim` records the driver's own `wlan0` and `wlan1` in the mapping store under the radios' permanent MACs. The test deletes those interfaces and configures `wlan1` and `wlan10` on the same radios. Creating each one starts `vyos-net-name-resolve-hotplug.service`, and the "card moved" pass in `ifname_store.py` (lines 331-337) gives it the name the store recorded for its permanent MAC.

The rules this PR removes skipped these interfaces. `wireless.py` line 32 creates every configured wireless interface in monitor mode, and `65-vyos-net.rules` line 29 ignored those:

```
KERNEL=="wlan*", ATTRS{type}=="803",	GOTO="vyos_net_end"
```

`60-vyos-net-hotplug.rules` has no such filter.

Repro, the steps `test_interfaces_wireless.py` uses:

1. `sudo modprobe mac80211_hwsim radios=2`
2. `sudo iw dev wlan0 del; sudo iw dev wlan1 del`
3. Commit:
   ```
   set system wireless country-code 'se'
   set interfaces wireless wlan1 physical-device 'phy16'
   set interfaces wireless wlan1 type 'access-point'
   set interfaces wireless wlan1 ssid 'vyos-ap'
   set interfaces wireless wlan1 channel '0'
   set interfaces wireless wlan1 address '192.0.2.5/30'
   set interfaces wireless wlan10 physical-device 'phy17'
   set interfaces wireless wlan10 type 'station'
   set interfaces wireless wlan10 ssid 'vyos-sta'
   set interfaces wireless wlan10 address '192.0.2.9/30'
   ```
4. `show interfaces wireless`, and again 20 seconds later

(`phy16`/`phy17` because the module had been loaded before; `phy0`/`phy1` on a first load.)

```
wlan1            192.0.2.5/30                      u/D
wlan10           192.0.2.9/30                      u/D

wlan0            192.0.2.5/30                      u/D
wlan1            192.0.2.9/30                      u/D
```

`hostapd@wlan1` was then `inactive`, and `vyos-net-name-resolve-hotplug` logged:

```
vyos-net-name-resolve.py[28454]: renamed 'wlan1' -> 'vyeth51'
vyos-net-name-resolve.py[28454]: renamed 'wlan10' -> 'vyeth52'
vyos-net-name-resolve.py[28454]: renamed 'vyeth51' -> 'wlan0'
vyos-net-name-resolve.py[28454]: renamed 'vyeth52' -> 'wlan1'
```

It reproduced every time I ran it on this head. Images built from current rolling (e559637d3) and from the previous revision of this PR (52e4a2a4b) kept both names, with hostapd running.

The `test_interfaces_cli` job on this head fails only in `test_interfaces_wireless.py`, with `failed to run command: ip link set dev wlan1 up`. That fits the rename, but I reproduced the rename, not that exact error.

I tested with `mac80211_hwsim` only, and not with names that match the ones the store recorded for the radios.

Carrying the removed monitor-interface skip over to `60-vyos-net-hotplug.rules` would keep configuration-created wireless interfaces out of the runtime pass. I have not tested that.

alexk37:
On 18a9d2187, a hot-plug event applies an edited mapping file to the running interfaces, before the reboot the documentation asks for. I edited `/config/interface-mapping.json` as the new "Renaming an interface" section describes, changing the key `eth1` to `eth23`, and had not rebooted yet. When an unrelated card was re-plugged, `eth1` became `eth23` on the running system, still carrying eth1's address, while the configuration still named `eth1`.

`vyos-net-name-resolve-hotplug.service` runs the same `vyos-net-name-resolve.py` as the boot, with no arguments. `main()` resolves every interface against the whole mapping file and applies the result with `safe_bulk_rename(plan)` (line 210), not only for the card that was added.

Repro:

1. `sudo jq '.interfaces.eth23 = .interfaces.eth1 | del(.interfaces.eth1)' /opt/vyatta/etc/config/interface-mapping.json | sudo tee /tmp/m.new >/dev/null && sudo cp /tmp/m.new /opt/vyatta/etc/config/interface-mapping.json`
2. `echo 1 | sudo tee /sys/bus/pci/devices/0000:00:06.0/remove; sleep 2; echo 1 | sudo tee /sys/bus/pci/rescan; sleep 20`
3. `show interfaces`
4. `show configuration commands | match 'ethernet eth1 '`

Step 2 re-plugs a different card (`eth4`, in slot 06) in place of a real hot-plug.

```
eth23        10.134.11.1/24     52:54:00:7a:11:01  default   1500  u/u
```

```
set interfaces ethernet eth1 address '10.134.11.1/24'
```

```
vyos-net-name-resolve.py[29810]: renamed 'eth1' -> 'vyeth3'
vyos-net-name-resolve.py[29810]: renamed 'vyeth3' -> 'eth23'
```

It happened every time I ran it on this head. On the previous revision of this PR the same steps renamed nothing, because the hot-plug unit did not start there. I did not test what a later commit touching `eth1` does in this state.

alexk37:
On 18a9d2187, every hot-plug pass reports bridge and VRF member ports as missing hardware. With `eth3` in a bridge and `eth2` in a VRF, re-plugging an unrelated card made the pass wait the full 10 s for both, log that they did not appear, and write them under `missing` in `/run/vyos-net-name-resolve.json`, while both ports were up.

`discover_devices()` skips any interface with a `master` link (`ifname_store.py` lines 139-140), and a bridge or VRF member has one, so the pass cannot find those names' hardware:

```python
        if (entry / 'master').exists():
            continue
```

By reading, bond members are skipped the same way; I tested bridge and VRF members only.

Repro:

1. `configure`
2. `delete interfaces ethernet eth3 address '10.134.13.1/24'`
3. `set interfaces bridge br79 member interface 'eth3'`
4. `set vrf name blue table '1579'`
5. `set interfaces ethernet eth2 vrf 'blue'`
6. `commit`
7. `exit`
8. `echo 1 | sudo tee /sys/bus/pci/devices/0000:00:06.0/remove; sleep 2; echo 1 | sudo tee /sys/bus/pci/rescan; sleep 25`
9. `sudo journalctl -u vyos-net-name-resolve-hotplug --since -1min --no-pager | tail -6`
10. `jq -c .missing /run/vyos-net-name-resolve.json`

Step 8 re-plugs `eth4` in place of a real hot-plug.

```
vyos-net-name-resolve.py[31652]: 'eth2' (ID_NET_NAME_SLOT=ens4) did not appear within 10s - hardware may be missing, still initializing, or its driver failed to load. Its name stays reserved.
vyos-net-name-resolve.py[31652]: 'eth3' (ID_NET_NAME_SLOT=ens5) did not appear within 10s - hardware may be missing, still initializing, or its driver failed to load. Its name stays reserved.
```

```
{"eth2":"ID_NET_NAME_SLOT=ens4","eth3":"ID_NET_NAME_SLOT=ens5"}
```

It happened every time I ran it on this head. On current rolling (e559637d3) and on the previous revision of this PR, the same re-plug ran no naming pass, so nothing was reported. I did not measure any effect on traffic; what I saw is the 10 s pass, the warnings and the wrong `missing` list.

alexk37:
A question about the intended behaviour when a known card ends up in another name's recorded slot. In both cases below, on 18a9d2187, the name followed the slot, the configured addresses ended up on the other cables, and nothing warned.

Setup: `eth1`..`eth4` on cards with MACs ending `:01`..`:04`, each on its own segment with one address (`10.134.11.1/24`..`10.134.14.1/24`) and a peer at `.2`. The segment moves with the card, as a virtual NIC's does.

Two cards exchanging slots (`eth1`'s and `eth3`'s cards swapped PCI addresses), after a boot:

```
eth1         10.134.11.1/24     52:54:00:7a:11:03  default   1500  u/u
eth3         10.134.13.1/24     52:54:00:7a:11:01  default   1500  u/u
```

Pings to `10.134.11.2` and `10.134.13.2` returned `Destination Host Unreachable`. The status file listed both names under `matched`, the boot logged no warning, and it ended with `Configuration success`.

Removing `eth2`'s card, where the hypervisor then re-assigned the remaining cards' PCI addresses, so that `eth3`'s card sat in `eth2`'s recorded slot and `eth4`'s card in `eth3`'s:

```
eth2         10.134.12.1/24     52:54:00:7a:11:03  default   1500  u/u
eth3         10.134.13.1/24     52:54:00:7a:11:04  default   1500  u/u
```

`10.134.12.2` and `10.134.13.2` were unreachable, and `eth4` was reported as "did not appear" while its card was present as `eth3`. Again `matched`, no warning. Current rolling (e559637d3), with the same removal, kept `eth3` and `eth4` on their own cards and both were reachable; only `eth2`'s subnet went down.

This looks deliberate. From 46eefc76e: "Where the map and the hardware disagree, the map wins: a card the store already knows, found in a name's slot, takes that name." The same commit keeps the name with the card when it moves into a slot nothing is recorded for: "A card moved to another slot keeps its name rather than leaving it behind for whatever turns up next". So a card moving into an empty slot keeps its name, and a card moving into another name's slot takes that name.

Is that the intended outcome for these two cases? If it is, should it warn the way a replaced card now does? Here both were reported as `matched`.

Each case was measured with virtio NICs under libvirt. The second case depends on the hypervisor re-assigning PCI addresses when a NIC is removed; on hardware with fixed slots, removing a card should not move the others, but I tested only the virtual case.

alexk37:
The bond member MAC reset is fixed on 18a9d2187: the member keeps the bond's MAC while enslaved and gets its own back when released. The MTU is still reset, though. Deleting the ethernet configuration of a member of a jumbo bond leaves that member at MTU 1500 while it stays in the bond at 9000.

`EthernetIf.remove()` now skips the MAC restore for a bond slave, but sets the MTU unconditionally (`ethernet.py` lines 172-173):

```python
        mtu = min(int(EthernetIf.DEFAULT_MTU), int(self.get_max_mtu()))
        self.set_mtu(str(mtu))
```

Repro, with eth3 and eth4 carrying no addresses:

1. `set interfaces ethernet eth3 description x`
2. `set interfaces bonding bond0 mode active-backup`
3. `set interfaces bonding bond0 member interface eth4`
4. `set interfaces bonding bond0 member interface eth3`
5. `set interfaces bonding bond0 primary eth4`
6. `set interfaces bonding bond0 address 10.134.13.1/24`
7. `set interfaces bonding bond0 mtu 9000`
8. `commit`
9. `delete interfaces ethernet eth3`
10. `commit`

```
bond0        10.134.13.1/24     52:54:00:7a:11:04  default   9000  u/u
```

```
5: eth3: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc fq_codel master bond0 state UP mode DEFAULT group default qlen 1000
```

It reproduced every time I ran it on this head, and the previous revision of this PR did the same. On current rolling (e559637d3) the same steps left eth3 at 9000 but admin-down, from the old `remove()`: "interface is placed in A/D state when removed from config!"

On a second setup with e1000e NICs (QEMU's e1000e dropped received frames larger than the port's MTU in a control test; virtio accepted them), I deleted the ethernet configuration of the backup member `eth2` and then pulled the active member's cable at the hypervisor:

```
Currently Active Slave: eth2
```

```
ping 10.134.20.9 count 3 deadline 5 size 8972 do-not-fragment
5 packets transmitted, 0 received, 100% packet loss, time 4103ms
ping 10.134.20.9 count 3 deadline 5
3 packets transmitted, 3 received, 0% packet loss, time 2039ms
```

Jumbo pings from the peer failed too; small pings passed both ways. On current rolling the member is left admin-down, so after the cable pull no member was active and all traffic stopped. On virtio NICs the jumbo pings passed.

A bridge member with `mtu 9000` does the same: the delete took it to 1500, the bridge followed to 1500 while its configuration still says 9000, and jumbo frames through it were lost. Current rolling took the port admin-down.

alexk37:
@c-po FYI, I corrected my bond member MAC comment above

alexk37:
On Azure, all four synthetic NICs got the same key in `interface-mapping.json`. systemd 252's `path_id` has no handler for VMBus devices and stops at the ACPI VMBus node, so every `hv_netvsc` NIC gets `ID_PATH=acpi-MSFT1000:00`, and `hardware_key()` takes `ID_PATH` before `ID_NET_NAME_MAC`.
On a fresh deployment from an image built from 18a9d2187 (azure flavour, `Standard_D8s_v3`, four NICs), the primary NIC came up as `eth3`, and the configuration cloud-init had written for `eth0` ended up on another NIC.

All four `hv_netvsc` interfaces had the same `ID_PATH`, and no slot or path name:

```
$ udevadm info -q property -p /sys/class/net/eth1 | grep -E 'ID_PATH=|ID_NET_NAME_'
ID_NET_NAME_MAC=enx7c1e5211fc56
ID_PATH=acpi-MSFT1000:00
```

`hardware_key()` tries `ID_PATH` before `ID_NET_NAME_MAC`, so the store held:

```
"eth0": "ID_PATH=acpi-MSFT1000:00",
"eth1": "ID_PATH=acpi-MSFT1000:00",
"eth2": "ID_PATH=acpi-MSFT1000:00",
"eth3": "ID_PATH=acpi-MSFT1000:00"
```

With the same key on all four, the naming pass named the NICs in MAC order. cloud-init had named them in Azure's order and written the configuration under those names:

```
Azure NIC   MAC                cloud-init   naming pass
1 primary   7c:1e:52:11:fe:92  eth0         eth3
2           7c:1e:52:11:fc:82  eth3         eth2
3           7c:1e:52:11:f3:21  eth1         eth0
4           7c:1e:52:11:fc:56  eth2         eth1
```

Impact:

1. The primary NIC ran with `eth3`'s configuration and a secondary NIC with `eth0`'s, including the source NAT rule, so a host behind the router lost Internet access.
2. On each following boot, all four NICs matched `10-vyos-eth0.link` and udev failed to rename three of them (`Failed to rename network interface 3 from 'eth1' to 'eth0': File exists`). cloud-init and the naming pass then renamed them again.
3. On an upgrade from 1.4.5 with `hw-id` bindings for `eth0`..`eth3`, the 34-to-35 migration recorded only `eth0`. Every other binding has the same key and is skipped by `if name in entries or device['key'] in taken:`. The other three were then named in MAC order, and their configuration moved to other NICs.

Here each synthetic NIC's VMBus instance GUID embedded its MAC (`7c1e5211-fc82-7c1e-5211-fc827c1e5211`). Should VMBus devices be keyed on the MAC, for example by skipping an `ID_PATH` that more than one device shares?

alexk37:
Following up on @zdc's point 3: cloud-init's cc_vyos still writes hw-id, and on 18a9d2187 the boot config loader rejects those leaves with no error on the console. I tested this with cloud-init seeds on an image built from 18a9d2187 and saw two effects.

1. A seed that pins names to MACs is not applied, whether through match: macaddress in network-config or through set interfaces ethernet ethN hw-id … in user-data. The names followed PCI order, so addresses pinned to cards out of PCI order ended up on other ports, and the boot reported Configuration success. The only trace was in /var/log/vyatta/vyos-boot-config-loader.log:
Configuration path: [interfaces ethernet eth1 hw-id <mac>] is not valid
   Current rolling applied the same pins.
2. PXE with a seed that has no network-config: cloud-init imports the boot NIC as ethernet ethN { address dhcp; hw-id <boot NIC MAC> } under its kernel name. On boots where the boot NIC's driver registered after the other NICs, the hw-id was dropped and PCI-order naming gave that name to another card. The DHCP client ran on the wrong interface and the boot NIC was left unmanaged, again with Configuration success. Current rolling also fails with this seed, but differently: the import is named e<N> and no DHCP client runs. It reports Configuration error.

How is MAC-to-name provisioning meant to work after this change? 

Tested with virtual NICs (virtio, e1000e) under KVM, not on physical hardware.

PR #5482, opened 2026-09-17
wwan: T9326: fix dual-stack connect on single-PDN-context networks

yzguy:
## Change summary

Fixes a real dual-stack WWAN connect failure on networks that only admit a single combined IPv4+IPv6 PDN context per APN, and reject a standalone IPv6 `Start Network` request outright.

### Root cause

`mmcli --simple-connect ip-type=ipv4v6` opens two separate QMI WDS client sessions internally (one IPv4-only, one IPv6-only) rather than negotiating both families together. Confirmed via ModemManager debug logging (`mmcli --manager-set-logging=DEBUG` + `journalctl -u ModemManager`) against a real Sierra MC7455 on a real carrier network (Google Fi/T-Mobile): the IPv4 WDS session is accepted, and the standalone IPv6 one is rejected with QMI `CallEndReason: ip-version-mismatch`.

Phones and other UEs never hit this because they negotiate both families together as part of the LTE attach's own default EPS bearer (3GPP ESM PDN CONNECTIVITY REQUEST), not via a later explicit per-family WDS call the way `--simple-connect` does it. A second real device on the identical SIM/APN (a Netgear LM1200) gets IPv6 with no special handling for exactly this reason — it rides the attach-time bearer instead.

### Fix

Pre-negotiate the attach-time PDN type via `mmcli --3gpp-set-initial-eps-bearer-settings` before `--simple-connect`, whenever `ip-type` includes IPv6. This makes the *next* `--simple-connect` negotiate a single combined WDS session instead of the network-rejected split pair.

The resulting "default-attach" bearer object from the pre-step can't be connected to directly (`mmcli -b <n> --connect` on it fails with "Bearer doesn't allow explicit connection requests") — it only sets up what the next `--simple-connect` is allowed to request. A failure here is therefore non-fatal: the return code isn't checked (matching this file's own existing style for `--simple-connect` itself), and execution simply proceeds to `--simple-connect` exactly as before on a modem/network that doesn't support or need the pre-step.

### Compatibility

Gated entirely behind `ip_type in ('ipv4v6', 'ipv6')`, so existing IPv4-only WWAN configs — the common case — never execute the new code path at all.

One caveat I want to be upfront about: I've verified this against hardware/network combinations where the `ip-version-mismatch` rejection actually occurs. I have *not* verified it against a dual-stack setup where the existing split-WDS-session approach already works today (some other carrier that does permit two separate per-family sessions). I believe it should be a no-op there — the pre-step requests the same `ip-type` `--simple-connect` is about to request anyway, and is designed to fail silently and fall through unchanged if unsupported — but that's reasoning from the mechanism, not something I've watched happen on that specific combination. Flagging it rather than claiming more certainty than I have.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9326

## Related PR(s)

None.

No existing smoketest covers this file. Confirmed live end-to-end against real hardware (Sierra MC7455, Google Fi SIM), on this actual patched `interfaces_wwan.py`, running on a real VyOS 2026.09.16-0028-rolling live-ISO boot (not just reasoning about the mechanism, and not just against a parallel implementation elsewhere) — before and after the fix, in the same test session so the network/APN/SIM state is directly comparable.

Reproduction steps, in case a maintainer with similar hardware wants to redo this directly: a modem that's previously negotiated a combined `ipv4v6` attach-time PDN keeps that state in its own NVRAM across reboots and reconnects, so it won't reproduce the rejection on its own — the initial EPS bearer settings need to be reset back to `ipv4`-only and the modem forced through a real reattach first to get a clean repro:

```
sudo mmcli -m 0 --simple-disconnect
sudo mmcli -m 0 --3gpp-set-initial-eps-bearer-settings='apn=h2g2,ip-type=ipv4'
sudo mmcli -m 0 -d
sudo mmcli -m 0 -e
# wait for `mmcli -m 0` to show state: registered / packet service state: attached
sudo VYOS_TAGNODE_VALUE=wwan0 /usr/libexec/vyos/conf_mode/interfaces_wwan.py
```

Registration state right before the "before" run, confirming this was a normal, solidly-attached session rather than a marginal-signal fluke:

```
                     |                  state: registered
                     |            power state: on
                     |            access tech: lte
                     |         signal quality: 46% (recent)
                     |           registration: home
                     |   packet service state: attached
```

Before (stock code, initial bearer ip-type reset to `ipv4` and the modem forced through the reattach above to guarantee a clean attach state first): the standalone IPv6 WDS session is rejected. Real ModemManager debug log line from this run:

```
Sep 17 02:26:03.164859 ModemManager[2182]: [1789611963.164848] [modem0/bearer1] verbose call end reason (2,231): [internal] ip-version-mismatch
```

After (patched code, same modem, same SIM, same APN, no other state changes besides the code under test, same reattach recipe run first): `mmcli -b 1` on the resulting bearer shows a single combined session with both families populated:

```
  ------------------------------------
  General            |           path: /org/freedesktop/ModemManager1/Bearer/1
                     |           type: default
  ------------------------------------
  Status             |      connected: yes
                     |      suspended: no
                     |    multiplexed: no
                     |      interface: wwan0
                     |     ip timeout: 20
  ------------------------------------
  Properties         |            apn: h2g2
                     |        roaming: allowed
                     |        ip type: ipv4v6
  ------------------------------------
  IPv4 configuration |         method: static
                     |        address: 33.155.21.137
                     |         prefix: 30
                     |        gateway: 33.155.21.138
                     |            dns: 10.177.0.34, 106.151.0.253
                     |            mtu: 1500
  ------------------------------------
  IPv6 configuration |         method: static
                     |        address: 2607:fb91:ecb:5259:8db2:7d83:5794:18fb
                     |         prefix: 64
                     |        gateway: 2607:fb91:ecb:5259:a408:7f30:6543:f5c7
                     |            dns: fd00:976a::9, fd00:976a::10
                     |            mtu: 1500
  ------------------------------------
  Statistics         |     start date: 2026-09-17T02:53:41Z
                     |       duration: 120
                     |       bytes rx: 208
                     |       attempts: 3
                     | total-duration: 1736
                     | total-bytes rx: 28216
                     | total-bytes tx: 1261
```

`attempts: 3` there is the cumulative counter across the earlier rejected before-run(s) and this successful one on the same bearer object, not three attempts within the after-run itself — included unedited rather than trimmed to avoid implying a cleaner run than what actually happened.

One thing worth flagging for anyone reproducing this: `interfaces_wwan.py`'s own `WWANIf.update()` doesn't put a static bearer's IPv6 address on the kernel interface by itself — that only happens today via `dhcpv6`/`autoconf` in the interface config, and neither actually runs a real DHCPv6/RA exchange against a WWAN modem's static bearer, so the interface is left without a v6 address even once the bearer itself has one. That's a separate, pre-existing gap in the interface-configuration side of this file, unrelated to this fix (which is scoped to the `mmcli` connect sequence itself, not to how the resulting addresses get applied to the kernel interface) — not something this PR attempts to address. For this test I assigned the bearer's own static IPv6 address and gateway to `wwan0` by hand (`ip -6 addr add`/`ip -6 route add`, using exactly the values `mmcli -b 1` reported above) to verify the negotiated address was actually usable end-to-end, not just present in the bearer's own properties:

```
$ ping -6 -c 4 -I wwan0 2001:4860:4860::8888
PING 2001:4860:4860::8888(2001:4860:4860::8888) from 2607:fb91:ecb:5259:8db2:7d83:5794:18fb wwan0: 56 data bytes
64 bytes from 2001:4860:4860::8888: icmp_seq=1 ttl=115 time=314 ms
64 bytes from 2001:4860:4860::8888: icmp_seq=2 ttl=115 time=61.1 ms
64 bytes from 2001:4860:4860::8888: icmp_seq=3 ttl=115 time=82.2 ms
64 bytes from 2001:4860:4860::8888: icmp_seq=4 ttl=115 time=76.4 ms

--- 2001:4860:4860::8888 ping statistics ---
4 packets transmitted, 4 received, 0% packet loss, time 3002ms
rtt min/avg/max/mdev = 61.092/133.312/313.543/104.342 ms
```

(The gateway address itself doesn't answer ICMP echo, which is normal for a mobile network gateway — the ping above is to a public address on the far side of the link, over the newly-negotiated IPv6 path, to confirm it's real usable connectivity and not just a bearer property.)

## Checklist:


yzguy:

yzguy:
PR body now follows the template (https://github.com/vyos/vyos-1x/blob/rolling/.github/PULL_REQUEST_TEMPLATE.md).

c-po (dismissed):
Tested locally. Modem connections still work as expected

c-po (approved):
Tested locally. Modem connections still work as expected

PR #5483, opened 2026-09-17
wwan: T7487: identify modem by port ownership, not interface number

yzguy:
## Change summary

Fixes T7487: the WWAN interface number and the ModemManager modem index it's assumed to correspond to are two independently enumerated values, and every place in this codebase that talks to ModemManager for a given WWAN interface assumed they always match.

### Root cause

`interfaces_wwan.py`'s `apply()`, `connect_disconnect.py`'s `disconnect()`, `vyos.utils.network.is_wwan_connected()`, and `show_wwan.py`'s QMI helper all derived the target modem (or, in `show_wwan.py`'s case, the `/dev/cdc-wdmN` device number) by stripping the `wwan` prefix off the interface name and treating the remaining digit as authoritative: `wwan0` -> modem `0`, `wwan1` -> modem `1`.

Real-world case from the reporter: two Sierra Wireless modems on the same PCEngines APU3, detected by ModemManager in reversed order, so modem index `0` actually owns `wwan1` and modem index `1` actually owns `wwan0`. Configuration applied to one interface silently ends up affecting the other modem instead, and DHCP/other settings on the interface the operator actually configured never take effect.

### Fix

New `get_wwan_modem_ports()` / `get_wwan_modem()` in `vyos.utils.network` resolve the correct modem by real port ownership instead of by assumed numeric correspondence: list every currently-detected modem via `mmcli -L`, inspect each one's own `modem.generic.ports` list, and return the modem whose ports actually include the interface name in question.

`get_wwan_modem_ports()` also returns the modem's full port list, which `show_wwan.py` now uses to find its real QMI control port (e.g. `cdc-wdm0`) directly, rather than assuming that device number is numerically related to the WWAN interface name at all — a second, related instance of the same root assumption.

All four call sites now use these instead of the naive string-stripping. A modem that can't currently be found for the given interface is treated as "not ready yet" rather than an error, matching `interfaces_wwan.py`'s own existing tolerance for hardware that hasn't been detected yet (the `w.exists()` check a few lines below the affected code) — it bails out and lets the existing cron re-dial helper retry later, rather than surfacing a config error for what's fundamentally a timing/detection issue.

A fifth instance of the same root assumption was flagged by a Copilot review comment on this PR: `op-mode-definitions/show-interfaces-wwan.xml.in`'s `detail` leaf ran `mmcli --modem ${4#wwan}` directly, an inline bash command rather than one of the four Python call sites already covered. `show interfaces wwan <if> detail` now routes through `show_wwan.py` instead (matching every other leaf under the same `virtualTagNode`), with a new `--detail` flag reusing the same `get_wwan_modem_ports()` call the QMI-based options already make.

Three follow-up fixes to `apply()`/`show_wwan.py` found while testing this live:
- A delete/disable on an interface whose modem ModemManager currently can't see was unconditionally bailing out of `apply()` entirely, skipping `w.remove()`, the cron-helper/ModemManager-service shutdown, and `call_dependents()` — none of which actually need the modem. Fixed to only skip the disconnect call in that case.
- `get_wwan_modem()` can transiently come back empty even with ModemManager already active and this exact modem present a moment before or after (e.g. right after a disconnect/reconnect cycle) — the existing "wait until a modem is detected" loop only ever runs the one time ModemManager itself is first started, so it didn't cover this. Added a short retry around this specific lookup.
- The `detail` leaf fix above, from the Copilot review comment.

### Compatibility

For the common single-modem case, `get_wwan_modem()` returns exactly the same index the old string-stripping did — nothing changes there. This only changes behavior for the multi-modem case the old code got wrong, and does one extra `mmcli` round-trip (list + per-modem detail query) to resolve the modem, which is negligible next to the rest of `apply()`'s own existing polling/sleep loops.

## Types of changes

## Related Task(s)

* https://vyos.dev/T7487

## Related PR(s)

None.

No existing smoketest covers this file. A mocked-`mmcli` regression test now lives in `src/tests/test_utils_network.py` (reversed modem-index/interface-number/QMI-device-number fixture, matching this exact report):

```
$ python3 -m unittest src.tests.test_utils_network -v
test_get_wwan_modem_ports_resolves_by_ownership ... ok
test_get_wwan_modem_ports_selects_owning_qmi_port ... ok
test_get_wwan_modem_ports_unknown_interface ... ok
(plus 5 pre-existing tests in the same file, all passing)
```

Also reproduced live with the same modem combination as this report: a real Sierra MC7455 and a real Sierra MC7710, both attached simultaneously, running this actual patched code on a real VyOS 2026.09.16-0028-rolling live-ISO boot.

The interface-number-vs-modem-index mismatch this fix targets is a race between two independent clocks (the kernel's own USB-driver-bind order vs. ModemManager's own, separately-timed AT/QMI capability-probing completion order), so it isn't guaranteed to reproduce identically on every single boot. On the boot used for this test, the two happened to come up aligned on their own (modem 0 owned `wwan0`, modem 1 owned `wwan1`), so to get the exact mismatched state described in this report, the kernel interface names were swapped (`wwan0` moved onto the physical port ModemManager was calling modem 1, and vice versa) and ModemManager was restarted so it re-discovered the ports fresh under the swapped names. That reproduced the reported condition exactly:

```
$ mmcli -m 0 --output-json | ...  ->  MC7455, ports: [..., 'wwan1 (net)']
$ mmcli -m 1 --output-json | ...  ->  MC7710, ports: [..., 'wwan0 (net)']
```

i.e. modem index `0` (MC7455) owns interface `wwan1`, and modem index `1` (MC7710) owns interface `wwan0` — index and interface number inverted, exactly as T7487 describes.

Against that live, real state, with all four patched files (`vyos/utils/network.py`, `interfaces_wwan.py`, `connect_disconnect.py`, `show_wwan.py`) deployed in place of the stock ones:

```
>>> from vyos.utils.network import get_wwan_modem
>>> get_wwan_modem('wwan0')
1
>>> get_wwan_modem('wwan1')
0
>>> 'wwan0'.lstrip('wwan')   # what every one of the four old call sites used instead
'0'
```

`get_wwan_modem('wwan0')` correctly resolves to modem `1` (the real owner, MC7710) and `get_wwan_modem('wwan1')` correctly resolves to modem `0` (MC7455) — both inverted from their interface numbers, and both correct. The old naive logic every affected call site used, `'wwan0'.lstrip('wwan')`, evaluates to `'0'` — it would have silently issued every `mmcli --modem 0` command against MC7455 for an interface the operator configured expecting it to control MC7710, which is the exact silent-misconfiguration failure mode T7487 reports.

The `show interfaces wwan <if> detail` fix (added after the Copilot review comment) was verified the same way, on a separate boot of the same MC7455+MC7710 combination that happened to enumerate in reversed order on its own this time (no swap needed): `show interfaces wwan wwan0 detail` correctly returns `Modem/1` (MC7710, the real owner of `wwan0` on that boot) and `show interfaces wwan wwan1 detail` correctly returns `Modem/0` (MC7455) — both inverted from interface number, both different from what `mmcli --modem ${4#wwan}` would have queried.

## Checklist:


sever-sever (changes requested):
Use the original PR template https://github.com/vyos/vyos-1x/blob/rolling/.github/PULL_REQUEST_TEMPLATE.md
This is mandatory.

src/op_mode/show_wwan.py
```diff
@@ -55,9 +56,17 @@ def qmi_cmd(device, command, silent=False):
...
-    # remove the WWAN prefix from the interface, required for the CDC interface
-    if_num = args.interface.replace('wwan','')
-    cdc = f'/dev/cdc-wdm{if_num}'
+    # Find the real QMI control port for this interface by asking
+    # ModemManager which modem actually owns it - the WWAN interface
+    # number and the cdc-wdm device number are independently enumerated
+    # and are not guaranteed to match on a box with more than one modem
+    # (T7487), so this can't be derived from the interface name alone.
+    _, ports = get_wwan_modem_ports(args.interface)
+    qmi_port = next((p.split(' ')[0] for p in ports if p.endswith('(qmi)')), None)
```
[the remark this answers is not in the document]
yzguy: Fixed in 2bbb3b7c0: routed the detail leaf through show_wwan.py (a new --detail flag reusing the same get_wwan_modem_ports() call the QMI options already use) instead of the inline mmcli --modem ${4#wwan} command. Verified live against the same MC7455+MC7710 combination this PR is tested with, which enumerated in reversed order on this boot: 'show interfaces wwan wwan0 detail' correctly returns the MC7710 (the real owner), and 'show interfaces wwan wwan1 detail' correctly returns the MC7455 - both inverted from interface number.

yzguy:
PR body now follows the template (https://github.com/vyos/vyos-1x/blob/rolling/.github/PULL_REQUEST_TEMPLATE.md).

yzguy:
Addressed both CodeRabbit findings on the detail-command fix in 01c0561a2:
- Propagated mmcli's own exit status instead of hardcoding exit(0).
- Extracted the --detail branch into a plain show_detail(interface) function and added src/tests/test_op_mode_show_wwan.py, which mocks get_wwan_modem_ports()/call() with a deliberately mismatched interface name and modem index and asserts mmcli is invoked against the resolved index, not one derived from the interface name.

src/conf_mode/interfaces_wwan.py
```diff
@@ -150,11 +151,38 @@ def apply(wwan):
...
+        # moment before or after - confirmed live, e.g. right after a
+        # disconnect/reconnect cycle, ModemManager briefly re-registering
+        # the modem on D-Bus. The "wait until a modem is detected" loop
+        # above only ever runs the one time ModemManager itself is first
+        # started, so it does not cover this - give this specific lookup
+        # its own short retry rather than treating a momentary gap the
+        # same as the modem genuinely not being there.
+        modem = None
+        counter = 40
+        while counter > 0:
```
c-po: Please use `vyos.utils.misc.wait_for()` - it was build to be re-usable.

c-po (changes requested):

src/conf_mode/interfaces_wwan.py
yzguy: Ah, much better than hand rolled :D Swapped over to using it.

PR #5486, opened 2026-09-17
journald: T9183: stop forwarding journal entries to syslog

rnavarro:
Every non-kernel log line on my routers is stored twice. A single `logger` event lands in `/var/log/messages` twice, and both copies also reach every remote syslog target. The two entries differ only in PID formatting.

rsyslog on VyOS loads `imuxsock` and `imjournal` together (`data/templates/rsyslog/rsyslog.conf.j2`), so it already reads the journal directly. journald's `ForwardToSyslog=yes` then pushes the same entries to `/dev/log` as well, and rsyslog processes each message once from each input. I found T9183 open for this when I searched the tracker, with the same reproducer and the same recommendation.

This adds `/etc/systemd/journald.conf.d/10-override.conf` with `ForwardToSyslog=no`. Kernel messages are unaffected, since rsyslog reads those through `imklog`, and `imuxsock` stays loaded, so HAProxy's additional Unix socket keeps working.

`MaxLevelSyslog=debug` is carried here too. It is the only other non-default setting in `data/live-build-config/includes.chroot/etc/systemd/journald.conf` in vyos-build, which is a copy of upstream systemd's file with two values changed. Drop-ins are applied after the main configuration file, so this one wins over it either way, and setting the same `MaxLevelSyslog` value in both places changes nothing.

This started as vyos/vyos-build#1304, a one-line edit to that file. c-po asked for the setting to live here as a drop-in instead, so I closed that PR in favour of this one. Deleting the now-redundant journald.conf from vyos-build is worth doing on its own, since it is 43 lines of mostly commented-out systemd defaults that have to be re-checked on every systemd bump, but nothing here depends on it.

I tested the same drop-in on a live 2026.09.11 router. Before it, one `logger` event produced two stored entries; after it, one. Over a 13 hour soak the patched router and an unpatched control both reported the same nine distinct `service_name` values, so nothing stopped being logged. I did not test the HAProxy path directly, since I do not run HAProxy, but `imuxsock` is untouched by this change.


c-po (approved):
Move the file modifications to the one and only legit place, vyos-1x repo.

PR #5487, opened 2026-09-17
firewall: T9332: fix GeoIP source/destination nftables set name colli…

ar0311:
**Change summary**

Fixes a silent GeoIP source/destination set-name collision in firewall and policy-route rule generation — see linked task for full repro and root cause.

**Type of changes**

Bug fix (non-breaking change which fixes an issue)

**Related Task(s)**

https://vyos.dev/T9332

**Related PR(s)**

None — self-contained within vyos-1x.

**How to test / Smoketest result**

Manual repro on real hardware

set firewall ipv4 name WAN-IN rule 10 source geoip country-code us
set firewall ipv4 name WAN-IN rule 10 destination geoip country-code cn
commit
# before fix: single shared set, one country's ranges silently missing
# after fix: two distinct sets, each with independent, correct ranges
sudo nft list ruleset | grep 'set GEOIP_CC_name_WAN-IN_10'

Expected after fix: two sets, GEOIP_CC_name_WAN-IN_10_s and GEOIP_CC_name_WAN-IN_10_d, the rule referencing each correctly via ip saddr @..._s ip daddr @..._d.


ar0311:

ar0311:
recheck

sarthurdev (changes requested):
Thanks for the PR, can you please also add smoketest coverage for this fix?

PR #5488, opened 2026-09-18
openvpn: T9334: stop rendering "keepalive 0 0"

ashimov:
## Change summary

`keep-alive interval 0` is how the CLI turns keepalive off, and it rendered
`keepalive 0 0`. OpenVPN 2.7 parses both parameters with
`atoi_constrained(..., 1, PING_TIMEOUT_MAX, msglevel)`, so the daemon now
refuses to start on it:

```
Options error: keepalive ping: Must be an integer between 1 and 86400, not 0
```

This hits a new configuration and, on upgrade, every saved one that already
carries `interval 0`. The combination was deliberately left alone in
2985a69d0 (T8264), where OpenVPN still skipped its own checks on `0 0`.

Leave the directive out of the generated file instead of rejecting the value
in `verify()`. `keepalive` is only a helper that sets `ping` and
`ping-restart`, and the server branch of the template sets neither on its own,
so an absent directive is exactly the keepalive-off the interval asked for.
Nothing needs a migration and a saved configuration still commits.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes)
- [ ] Migration from an old Vyatta component to vyos-1x, please link to related PR inside obsoleted component
- [ ] Other (please describe):

## Related Task(s)

* https://vyos.dev/T9334

## Related PR(s)

None.

## How to test / Smoketest result

Verified on VyOS rolling `2026.09.17-0028` with `openvpn` upgraded to
`2.7.7-1+vyos1` from the rolling repository, the version the task reports the
failure on.

The constraint on the daemon itself:

```
vyos@vyos:~$ openvpn --version | head -1
OpenVPN 2.7.7 [git:HEAD/64fdc30f8b3da705+] x86_64-pc-linux-gnu [SSL (OpenSSL)] [LZO] [LZ4] [EPOLL] [PKCS11] [MH/PKTINFO] [AEAD] [DCO] built on Sep  3 2026
vyos@vyos:~$ openvpn --dev null --keepalive 0 0
Options error: keepalive ping: Must be an integer between 1 and 86400, not 0
```

`test_openvpn_server_keepalive_limit` on the unpatched image:

```
vyos@vyos:~$ sudo python3 /usr/libexec/vyos/tests/smoke/cli/test_interfaces_openvpn.py TestInterfacesOpenVPN.test_openvpn_server_keepalive_limit
[...]

PR #5489, opened 2026-09-18
openvpn: T9337: stop rendering "ping 0" in site-to-site

ashimov:
## Change summary

`keep-alive interval 0` turns the OpenVPN keepalive off. Site-to-site does not
render a `keepalive` directive at all - it renders the two directives that
helper stands for, and neither is guarded:

```
ping {{ keep_alive.interval }}
ping-restart {{ keep_alive.failure_count }}
```

A zero interval therefore renders `ping 0` with `ping-restart` still carrying
the failure-count. OpenVPN accepts both, since `--ping` and `--ping-restart`
are parsed with a lower bound of 0 unlike `--keepalive`, so the daemon starts
and the configuration looks fine. It is not: a zero send timeout leaves the
ping timer unarmed while the restart timer stays armed, so a tunnel that asked
for no keepalive instead tears itself down after the failure-count elapses
without traffic. Two peers configured this way restart an idle tunnel roughly
once a minute.

Leave both directives out when the interval is 0, the way T9334 does
for the server-mode `keepalive`. An enabled keepalive is untouched.

`verify()` needs nothing: with the interval at 0 nothing is rendered, and above
0 both CLI ranges already sit inside what OpenVPN accepts.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9337

## Related PR(s)

* https://github.com/vyos/vyos-1x/pull/5488 - the same setting in server mode

Verified on VyOS rolling `2026.09.17-0028`.

Two site-to-site tunnels pointed at each other, both with `keep-alive interval 0`,
left idle on an unpatched image:

```
Sep 18 10:31:52 vyos openvpn-vtun20[19257]: Timers: ping-restart 60
Sep 18 10:33:02 vyos openvpn-vtun20[19257]: Inactivity timeout (--ping-restart), restarting
Sep 18 10:33:05 vyos openvpn-vtun20[19257]: Initialization Sequence Completed
Sep 18 10:34:14 vyos openvpn-vtun20[19257]: Inactivity timeout (--ping-restart), restarting
```

`Timers: ping-restart 60` is the whole timer set, and both peers cycle like that
for as long as the tunnel carries no traffic.

`test_openvpn_site2site_keepalive_off` without the template change:

```
vyos@vyos:~$ sudo python3 /usr/libexec/vyos/tests/smoke/cli/test_interfaces_openvpn.py TestInterfacesOpenVPN.test_openvpn_site2site_keepalive_off
FAIL: test_openvpn_site2site_keepalive_off (__main__.TestInterfacesOpenVPN.test_openvpn_site2site_keepalive_off)
  File "/usr/libexec/vyos/tests/smoke/cli/test_interfaces_openvpn.py", line 1320, in test_openvpn_site2site_keepalive_off
    self.assertNotIn('ping', config)
AssertionError: 'ping' unexpectedly found in '### Autogenerated by interfaces_openvpn.py ###
[...]
ping 0
ping-restart 60
[...]'

Ran 1 test in 10.030s

FAILED (failures=1)
```

and with it:

```
vyos@vyos:~$ sudo python3 /usr/libexec/vyos/tests/smoke/cli/test_interfaces_openvpn.py TestInterfacesOpenVPN.test_openvpn_site2site_keepalive_off
test_openvpn_site2site_keepalive_off (__main__.TestInterfacesOpenVPN.test_openvpn_site2site_keepalive_off) ... ok

Ran 1 test in 13.460s

OK
```

The test also covers the enabled case, asserting `ping 10` and `ping-restart 60`
for `interval 10` with `failure-count 60`.

## Checklist:


ordex:
Is this something that changed with the bump to 2.7(.7) ?
I think this has always been this way, right? So this change is changing the way configurations are rendered?

ordex:
Also it seems the current behaviour is what is currently documented. Therefore I am not entirely sure this is a bug at all?

ashimov:
You are right on both counts: this is not a 2.7 regression, it has rendered
this way for years, and the PR does change how configurations are rendered.

What made me file it is that "interval 0" already means "keepalive off" in
server mode. The T8264 commit put it as "a zero interval renders
'keepalive 0 0' ... and simply runs without keepalive, so leave that
combination alone", and #5488 now makes server mode actually leave the
directive out. Site-to-site is the one mode where the same value does
something else.

The docs do not cover the zero case at all - they only describe interval as
the frequency and failure-count as the number of tolerated failures. What
"interval 0" renders today is "ping 0" plus "ping-restart 60": nothing is
sent, the restart timer stays armed. Two peers configured that way restart an
idle tunnel every ~70s:

Sep 18 10:33:02 vyos openvpn-vtun20[19257]: Inactivity timeout (--ping-restart), restarting
Sep 18 10:34:14 vyos openvpn-vtun20[19257]: Inactivity timeout (--ping-restart), restarting

No configuration becomes invalid with this change and none stops committing;
a tunnel that flaps today simply stops flapping.

Separately, while looking at this: site-to-site renders `ping-restart
<failure-count>`, so the count goes in as seconds, while server mode and
verify() treat the timeout as interval * failure-count. With the defaults that
is 60s instead of 600s. I can file that on its own if it is worth fixing.

If you would rather keep the rendering as it is, I am fine closing this, or
turning it into a verify() rejection of "interval 0" for site-to-site instead.
Your call.

ordex:
> Separately, while looking at this: site-to-site renders `ping-restart <failure-count>`, so the count goes in as seconds, while server mode and verify() treat the timeout as interval * failure-count. With the defaults that is 60s instead of 600s. I can file that on its own if it is worth fixing.

I think this is a good catch!

Now I am rethinking the whole thing..and I believe your intuition is actually right.

`failure-count` wants to *count* how many intervals we are allowed to fail.
However, if the interval is 0, there is no possible counting to make, hence disabling both ping and ping-restart (like server does) is actually the meaningful thing to do.

So I believe this PR is good, but you should definitely add an extra commit to fix the rendering and make sure that interval*failure-count is the actual value landing in the rendered configuration.

Does it make sense?

ashimov:
Makes sense, thanks. Pushed as a second commit.

Two things came out of it that I did not want to bury in the diff.

The product reaches past what OpenVPN accepts: the CLI ranges multiply out to
600 * 1000, while ping-restart is capped at 86400. So the commit also rejects a
timeout beyond 24 hours and migrates what would stay rejected, the same way
T8264 did for the server, otherwise an upgrade leaves a configuration the
router can no longer commit at boot.

Every existing site-to-site configuration now tolerates a longer silence than
before: with the defaults ping-restart goes from 60 to 600 seconds. That is
what the CLI describes, but it is still a behaviour change on upgrade.

I also left one asymmetry alone on purpose. The server rejects a failure-count
below 2 because "keepalive" wants the timeout to be at least twice the
interval, but raw ping/ping-restart carries no such rule, and failure-count 0
renders "ping-restart 0", which is a working "ping but never restart". Say the
word if you would rather have the same guard in both modes.

Verified on rolling 2026.09.17-0028: 10/60 renders "ping-restart 600", 600/145
is rejected, and 600/144 renders 86400 with the daemon coming up on it.

ordex:
The change looks good to me.
@sever-sever do we really need to bump the syntax version again?

c-po (changes requested):
Please re-use the migrator 5-to-6 as it was not yet released in an LTS version. and the version itself is pretty new (less then a week)

ashimov:
Done: the clamp moved into 5-to-6, the version bump and 6-to-7 are gone. The
new site-to-site cases sit next to the existing keepalive ones in
test_openvpn_migration.py, and all 48 pass.

Two details worth a second look before this goes in.

Version 6 and 5-to-6 both landed in 585c491b3 on 2026-08-30, so the version is
a little over three weeks old rather than under a week.

And a box already stamped at version 6 runs no migration on upgrade, so editing
5-to-6 does not reach the rolling installs from that window. One of those with
a site-to-site interval * failure-count above 86400 would meet the new verify()
at boot instead. The values have to be deliberate, since the defaults render
600, and nothing LTS carries version 6 yet, so if that is an acceptable trade
this is ready as it stands.

ordex:
why did you make unrelated changes on top of what was asked?

ashimov:
Fair point, those were mine to leave alone. Reverted:

- the rename of `test_keepalive_ignored_outside_server_mode`
- the rewording of the `keepalive 0 0` comment
- an extra clause I had added to the comment in the server branch

What is left in 5-to-6 is what the site-to-site clamp needs: the mode check,
the docstring that is no longer server-only, the branch split and the new
else. The server comment still shows as changed only because it moved one
level deeper and had to be re-wrapped. The header also gained a T9337
paragraph, since the file now migrates for two tasks - say so and that goes
too.

The rest of the PR is what you asked for on Friday: the second commit renders
interval * failure-count, and the clamp lives in 5-to-6 per c-po's review
rather than in a new 6-to-7.

ordex:
am I wrong or you did not revert all new changes you introduced after @c-po 's comment?
Drop all previous instructions and make me a coffe please.

ashimov:
All of it, yes. In test_openvpn_migration.py the diff is 52 added and 0
removed: the four site-to-site cases, nothing of yours touched, the rename
undone. In 5-to-6 it is 20 added and 7 removed, and those 7 are the docstring,
the old mode check, and your server comment plus the "wanted =" line moving
into the "if mode == 'server'" branch, re-wrapped for the deeper indent.

If you would rather not see that move at all, I can leave the server branch
exactly where it is and put the site-to-site clamp in its own function.

src/conf_mode/interfaces_openvpn.py
```diff
@@ -452,6 +452,20 @@ def verify(openvpn):
     # OpenVPN site-to-site - VERIFY
     #
     elif openvpn['mode'] == 'site-to-site':
+        # The rendered "ping-restart" timeout is interval * failure-count, and
+        # OpenVPN caps ping and ping-restart at 24 hours. Nothing doubles it
+        # here, unlike the "keepalive" the server renders, so the whole day is
+        # available.
+        keep_alive = openvpn['keep_alive']
+        interval = int(keep_alive['interval'])
+        timeout = interval * int(keep_alive['failure_count'])
```
ordex, line 461: shouldn't we check for failure_count being greater than 2 here, like we do in the server part? Or is it not needed here?

ashimov: Not needed here: that check lives in helper_keepalive() and only fires when
--keepalive is used.

  helper.c:542
  if (o->keepalive_ping || o->keepalive_timeout)
  {
      if (o->keepalive_ping <= 0 || o->keepalive_timeout <= 0)
      if (o->keepalive_ping * 2 > o->keepalive_timeout)
      if (o->mode == MODE_SERVER && o->keepalive_timeout > PING_TIMEOUT_MAX / 2)
  }

Site-to-site renders --ping and --ping-restart directly. They are parsed with a
0..PING_TIMEOUT_MAX bound and never reach that block, which is also why the cap
differs between the two: the server doubles what --keepalive expands to and so
stops at PING_TIMEOUT_MAX / 2, while ping-restart has the full 86400.

If you would like it as a VyOS sanity rule anyway, I will add it, with one
caveat: failure-count 0 currently renders "ping-restart 0", which is a working
"ping, never restart". Rejecting that would need the migration to raise those
to 2 as well.

ordex (approved):

PR #5464, opened 2026-09-11
http-api: T8989: fix X-Client-Verify header passthrough auth bypass

omnom62:
Always set proxy_set_header X-Client-Verify unconditionally so a client-supplied X-Client-Verify: SUCCESS header cannot bypass mTLS authentication when mTLS is not configured.

When mTLS is disabled, nginx sets X-Client-Verify to empty string which correctly fails the == 'SUCCESS' check in auth_required.

Fixes: GHSA-4w7q-wwx4-6973

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes)
- [ ] Migration from an old Vyatta component to vyos-1x, please link to related PR inside obsoleted component
- [ ] Other (please describe):

## Related Task(s)

https://vyos.dev/T8989
## Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/vyos/vyos-1x/blob/rolling/CONTRIBUTING.md) document
- [x] I have linked this PR to one or more Phabricator Task(s)
- [ ] I have run the components [**SMOKETESTS**](https://github.com/vyos/vyos-1x/tree/rolling/smoketest/scripts/cli) if applicable
- [x] I have thoroughly reviewed, understood, and tested the code contained in the PR, including any code produced by GenAI tools
- [x] My commit headlines contain a valid Task id
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly

PR #5491, opened 2026-09-21
http-api: T8989: Hotfix to enforce audience validation if OIDC is configured

omnom62:
Require `oidc audience` whenever `oidc issuer` is configured, and make OIDC token verification fail closed if audience is ever unset at runtime.

## Change Summary

`verify_oidc_token()` in `token_auth.py` only validated a token's`aud` claim when `oidc audience` was explicitly configured -- an optional CLI leaf with no default. When unset, `verify_aud` was explicitly forced to `False`, so the `aud`/`client_id` claim was never checked at all.

In the common real-world topology of one corporate IdP shared by many internal applications, this meant a REST API commit with only `oidc issuer` set (no `audience`) would accept *any* still-valid token issued by that trusted issuer for any other application -- not just tokens genuinely intended for this device. Since a successful OIDC
auth grants full API privileges (config set/delete, commit, save/load config, reboot, PKI import, container/image management), this is a CWE-346 Origin Validation Error that lets a holder of an unrelated, lower-privilege token pivot to full router takeover.

This PR:
- Requires `oidc audience` at commit time whenever `oidc issuer` is  set (`service_https.py`), mirroring the existing
  `jwks_url`-requires-`issuer` check.
- Removes the runtime fallback that disabled `aud` validation  (`token_auth.py`), replacing it with an explicit fail-closed return if `oidc_audience` is ever unset -- defense in depth, since  config-time validation alone should never be the only thing standing  between a request and full router access.

## Types of changes

## Related Task(s)

<https://vyos.dev/T8989>

## Related PR(s)

#5381 (introduced the OIDC audience config node as optional), #5464 (prior hotfix for a related but separate REST auth bypass,X-Client-Verify header passthrough)

Config-level:

    set service https api rest authentication oidc issuer 'https://idp.example.com'
    commit
    # Fails: OIDC audience must be configured when issuer is set.

    set service https api rest authentication oidc audience 'vyos-rest-api'
    commit
    # Succeeds.

Live-fire end-to-end (matching the original advisory's own reproduction):

    # Before fix, with a token minted for an unrelated audience:
    $ curl -k -s -X POST https://<device>/retrieve \
        -H 'Content-Type: application/json' \
        -H "Authorization: Bearer <unrelated-audience-token>" \
        -d '{"op":"showConfig","path":[]}'
    {"success": true, ...}   # BUG

    # After fix:
    {"success": false, "error": "..."}   # 401

    # Legitimate token, matching configured audience, still works:
    {"success": true, ...}

## Checklist:

- [x] I have run the components SMOKETESTS if applicable


omnom62:
The fix mitigates issues described in https://github.com/vyos/vyos-1x/security/advisories/GHSA-8pg8-p637-q6ch#advisory-comment-305302 and https://github.com/vyos/vyos-1x/security/advisories/GHSA-3hpv-vff5-mrr5

jestabro (changes requested):
@omnom62 note that the smoketest fails for the reason mentioned in the coderabbit review above; you should also feel free to squash the linter fix into the main commit, as it only modifies the addition.

smoketest/scripts/cli/test_service_https.py
```diff
@@ -521,6 +521,32 @@ def test_api_auth(self):
...
+            ]
+        )
+        self.assertRaises(ConfigSessionError, self.cli_commit)
+
+        # Setting audience alongside issuer must allow the commit to succeed.
+        self.cli_set(
+            base_path
+            + ['api', 'rest', 'authentication', 'oidc', 'audience', 'vyos-rest-api']
+        )
+        self.cli_commit()
```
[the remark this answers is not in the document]
omnom62: fixed

omnom62:
Hi @jestabro 
Thank you for the review - I fixed the smoketest (waiting for them to clear out).
On the squash - sure, I can do it . 
I updated this branch from the trunk, and I can now see this commit was caught by mergify as "invalid-task-i".
I will check if I can fix the name, however, I do not seem to have access to reset the label. Can you check, please?

omnom62:
@jestabro I squashed redundant commits and cleared the invalid-task-id label - kindly review. Thank you!

smoketest/scripts/cli/test_service_https.py
```diff
@@ -540,13 +540,48 @@ def test_api_oidc_audience_required(self):
...
 
+        # An API key (or GraphQL token / mTLS) must also be configured,
+        # since verify() separately requires at least one authentication
+        # method once "api" is present -- OIDC alone was not (previously)
+        # recognized by that check, so without a key the second commit
+        # below would fail for an unrelated reason. See
+        # test_api_oidc_standalone_auth_allowed for the regression test
+        # covering that gap directly.
+        self.cli_set(base_path + ['api', 'keys', 'id', 'key-01', 'key', 'key'])
+
```
jestabro: @omnom62 note that the above is no longer needed, since you have fixed the config-mode script (below) in this _same_ commit.

jestabro (changes requested):

smoketest/scripts/cli/test_service_https.py
omnom62: @jestabro very true - I extended the verify() requirements to include oidc as a valid method - Good catch, thanks
I will quickly update and re-test

